PR #423, opened 2020-05-16
perf(executor): improving Sentencizer by a factor of 2.5

guiferviz:
The Sententicizer is now 2.5 times faster.
Instead of replacing separators by a '\n' character and splitting later by the new line character, this improved version does that with one regex and using the `findall` method.
The complexity of the new algorithm is also linear but with a low constant time because it uses only one pass over the input string.
The regex also ignores spaces at the beginning and end of the sentences so no `strip` needed.

Adding two tests that show examples of questionable results of the Sentencizer.
If that is the expected behaviour, then it is good to keep those tests here to show that it is not considered a bug.
Those test are:
* Splitting by separators that appear in common constructions like floating point numbers.
* Splitting by the '\n' character. In this efficient version I've included '\n' among the separators to replicate the same behaviour as the previous version, but if '\n' it's not included among the separators then this new version is not going to split by new lines. If this behaviour is necessary, it can easily be modified. If it is not necessary, then it could be removed directly from the separators. Let me know and I will update the code.

## Time improvement

I've test the performance using an stress test that I've pull requested to the stress test repository:
https://github.com/jina-ai/stress-test/pull/1

Hardware details:
* Processor: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
* Memory: 16GiB SODIMM DDR4 Synchronous 2667 MHz (0,4 ns)
* OS: Linux Mint 19.3

Old code times:

    41.94MB read 3 times.
    Average per iteration: 2.876222

New code:

    41.94MB read 3 times                                                                                                       
    Average per iteration: 1.151864

Performance improvement: 2.4970
The results of both versions are exactly the same in the tested text.

### DIFF
--- a/jina/executors/crafters/nlp/split.py
+++ b/jina/executors/crafters/nlp/split.py
@@ -10,8 +10,8 @@ from .. import BaseSegmenter
 class Sentencizer(BaseSegmenter):
     """
     :class:`Sentencizer` split the text on the doc-level into sentences on the chunk-level with a rule-base strategy.
-        The text are splitted on the the punctuation characters that are listed in ``punct_chars``. The sentences that
-        are shorter than the ``min_sent_len`` or longer than the ``max_sent_len`` after stripping will be discarded.
+        The text is split by the punctuation characters listed in ``punct_chars``.
+        The sentences that are shorter than the ``min_sent_len`` or longer than the ``max_sent_len`` after stripping will be discarded.
     """
 
     def __init__(self,
--- a/jina/executors/crafters/nlp/split.py
+++ b/jina/executors/crafters/nlp/split.py
@@ -31,13 +31,14 @@ class Sentencizer(BaseSegmenter):
         self.punct_chars = punct_chars
         if not punct_chars:
             self.punct_chars = ['!', '.', '?', '։', '؟', '۔', '܀', '܁', '܂', '‼', '‽', '⁇', '⁈', '⁉', '⸮', '﹖', '﹗',
-                                '！', '．', '？', '｡', '。']
+                                '！', '．', '？', '｡', '。', '\n']
         if self.min_sent_len < self.max_sent_len:
             self.logger.warning('the min_sent_len (={}) should be smaller or equal to the max_sent_len (={})'.format(
                 self.min_sent_len, self.max_sent_len))
-        self._slit_pat = re.compile('([{0}])+([^{0}])'.format(''.join(self.punct_chars)))
+        self._slit_pat = re.compile('\s*([^{0}]+)(?<!\s)[{0}]*'.format(''.join(self.punct_chars)))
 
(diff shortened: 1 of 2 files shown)


hanxiao:
🤯 geez man! great work!

hanxiao:
@guiferviz thanks for your contribution, could you please resolve the conflict first so that the CICD could start. Btw, `self.min_sent_len < self.max_sent_len` is somehow still in your PR?

jina/executors/crafters/nlp/split.py
```diff
@@ -31,28 +31,33 @@ def __init__(self,
...
-                                '！', '．', '？', '｡', '。']
+                                '！', '．', '？', '｡', '。', '\n']
         if self.min_sent_len < self.max_sent_len:
             self.logger.warning('the min_sent_len (={}) should be smaller or equal to the max_sent_len (={})'.format(
                 self.min_sent_len, self.max_sent_len))
-        self._slit_pat = re.compile('([{0}])+([^{0}])'.format(''.join(self.punct_chars)))
+        self._slit_pat = re.compile('\s*([^{0}]+)(?<!\s)[{0}]*'.format(''.join(self.punct_chars)))
 
-    def craft(self, raw_bytes: bytes, doc_id: int, *args, **kwargs) -> List[Dict]:
+    def craft(self, raw_bytes: bytes, doc_id: int,
```
nan-wang, line 40: please reformat this line.

nan-wang:
As you have seen, `Sentencizer` is extremely simply for the time being. If the users want a customized rules, it shouldn't be very difficult to implement their own version based on regex or even use the punkt tokenizer from their favorite toolkits (e.g. `nltk`, `spaCy`, and etc.). For our side, we want to keep the core part of `jina` as lightweight as possible, and suggest the users to develop their own executors and share with the community in the `jina-hub`.

guiferviz:
I just fixed the merge conflicts. Sorry about not do in yesterday.

@hanxiao `self.min_sent_len < self.max_sent_len` is included now. When I started the branch my previous pull request had not yet been accepted, so the change was not there.

@nan-wang Ok, I understand. So I think the Sentencizer I am suggesting is simple and fast, it meets your idea of what a Sentencizer should be. As you say there are very good tokenizers out there that users can use if they want something more personalized.

guiferviz:
I'm getting the following error in one commit message.

Message:

    test(executor): Sentencizer trimming spaces

Error:

    subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]

Should I do something about this? It was in upper-case because it's the name of the class... :worried:

hanxiao:
yes, that's commit lint. please use `git commit --amend` change your commit message and then `git push --force`

jina/executors/crafters/nlp/split.py
guiferviz: Done! I have a habit of breaking all lines longer than 80 characters :grimacing:

hanxiao:
recheckcla

PR #424, opened 2020-05-16
feat(executors): Add numeric crafter - ArrayConverter

phamtrancsek12:
**Issue Link:** This commit does not refer to any issue or feature request

**Description:** In this commit, a numeric crafter is added to allow user to convert raw bytes into a numpy array, which is the input format for numeric encoder.

### DIFF
--- a/jina/executors/crafters/numeric/convert.py
+++ b/jina/executors/crafters/numeric/convert.py
@@ -1,0 +1,42 @@
+from typing import Dict, List
+import numpy as np
+from jina.executors.crafters import BaseSegmenter
+
+class ArrayConverter(BaseSegmenter):
+    """
+    :class:`ArrayConverter` convert the string of numbers into a numpy array and save to the Chunk.
+        Numbers are split on the provided delimiter, default is comma (,)
+    """
+
+    def __init__(self, delimiter=",", as_type="float32", *args, **kwargs):
+        """
+        :param delimiter: delimiter between numbers
+        :param as_type: type of number
+        """
+        super().__init__(*args, **kwargs)
+        self.delimiter = delimiter
+        self.as_type = as_type
+
+    def craft(self, raw_bytes: bytes, doc_id: int, *args, **kwargs) -> List[Dict]:
+        """
+        Split string into numbers and convert to numpy array
+
+        :param raw_bytes: the raw string in the `bytes` format
+        :param doc_id: the doc id
+        :return: a chunk dict with the numpy array
+        """
+        _string = raw_bytes.decode('utf8')
+        _string = _string.split(self.delimiter)
+        _array = np.array(_string)
+
+        try:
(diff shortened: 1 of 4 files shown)


hanxiao:
thanks for your contribution ❤️ Could you please elaborate a bit on the potential use case of this crafter?

tracy-propertyguru:
At first I was thinking about if I can index and query the stock prices data (find the stocks that have similar trends). I have tried using `IncrementalPCA` and `cosine` on S&P500 and I think it can be an example for using Jina with time-series data (Although there is other better algorithms to encode stock data, we can add them later) 

StockTrend

However, I didn't see any available crafter for this data format, so I think this crafter can be useful.

phamtrancsek12:
At first I was thinking about if I can index and query the stock prices data (find the stocks that have similar trends). I have tried using `IncrementalPCA` and `cosine` on S&P500 and I think it can be an example for using Jina with time-series data (Although there is other better algorithms to encode stock data, we can add them later) 

StockTrend

However, I didn't see any available crafter for this data format, so I think this crafter can be useful.

phamtrancsek12:
I chose to use `BaseSegmenter` because as I found in the document: `it receives value on the doc-level and returns new value on the chunk-level` (Here, it converts the `raw_bytes` of the doc into a chunk). I also noticed that `BaseSegmenter` is used to create `ImageReader` in `executors/crafters/image/io.py`, it seems like similar concept?

nan-wang:
> * From what you wrote, this is not a `Segmenter` as you didn't segment a doc into chunks. Yours should inherit from a `BaseDocCrafter`. Difference between these crafters can be found in https://docs.jina.ai/api/jina.executors.crafters.html
> * Please use single quote for string everywhere, this is the coding style of this project
> * Please add type hint for your function args

@hanxiao  `BaseDocCrafter` is not what we need in this case, because we don't save arrays at the `Document` level. The role of the `XyzReader` is similar as the `Segmenter`, i.e. read the data (e.g. file path) from the `Document` into the `Chunk` level for further processing. The differences is that `XyzReader` only returns a single `Chunk`. A better solution would be creating a new base class `Reader` for `XyzReader` to avoid confusion.  We can do this in another PR.

hanxiao:
> (Here, it converts the `raw_bytes` of the doc into a chunk). I also noticed that `BaseSegmenter` is used to create `ImageReader` in `executors/crafters/image/io.py`, it seems like similar concept?

hmm, `ImageReader` should also inherit from `BaseDocCrafter` @nan-wang . Please notice that `Segmenter` implement in this way gives you only one single chunk in the document.

hanxiao:
recheckcla

tracy-propertyguru:

phamtrancsek12:

hanxiao:
great! looks good to me! merge it and thanks for your contribution ❤️

PR #427, opened 2020-05-17
fix(pip install): Adds lz4 dependency to setup.py as current pip install jina asks for lz4 to be installed

SRITANU:
Adds `lz4` dependency to `base_dep` list in setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ base_dep = [
     'protobuf',
     'grpcio',
     'ruamel.yaml>=0.15.89',
+    'lz4'
 ]
 
 


hanxiao:
hi thanks for your contribution. Unfortunately, this is not how jina dependency works. `lz4` is already included in [`extra-requirements.txt`](https://github.com/jina-ai/jina/blob/master/extra-requirements.txt) Please refer to:

- https://docs.jina.ai/chapters/install/via-pip.html#cherry-pick-extra-dependencies
and the design pattern behind it
- https://hanxiao.io/2019/11/07/A-Better-Practice-for-Managing-extras-require-Dependencies-in-Python/

PR #452, opened 2020-05-25
feat: using black as code formatter 

ShubhamKhandelwal94:
Added black with pre-commit which will be able to check the format of code while committing on git. Haven't run  **pre-commit run --all-files** thou. Let me know if i need to run this command as this will format all existing files.


ShubhamKhandelwal94:
Create a new pull request

Issue #374, opened 2020-05-03
Using black as code formatter


maximilianwerk:
**Describe the feature**
[black](https://github.com/psf/black) is a python code formatter, which gets more and more adopted in the python community. While some of the rules might be not optimal for individual taste, it removes all further discussion on code formatting, since there are (almost) no configuration options. In my experience, this works out pretty well.

**Your proposal**
Add black as the code formatter, for example via [pre-commit hooks](https://pre-commit.com/).

PR #453, opened 2020-05-25
feat:using black as code formatter 

ShubhamKhandelwal94:
Added black with pre-commit which will be able to check the format of code while committing on git. Haven't run **_pre-commit run --all-files_** thou. Let me know if i need to run this command as this will format all existing files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,11 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    -   id: check-yaml
+    -   id: end-of-file-fixer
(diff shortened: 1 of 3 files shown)


ShubhamKhandelwal94:
Updated commit message, still getting an error on commit-lint. Not sure about this.

hanxiao:
when your merge a PR, you merge *all* commits from this PR. Therefore *all* commits messages need to comply with the commit-lint, not only the last one.

maximilianwerk:
While I absolutely support using black as the code formatter for jina, we should have a discussion about the consequences in #374 . One consequence would be that all single quotes `'` will be replaced with double quote `"`, which might not be liked too much. In case you reopen the PR once again, it would be great if you include one or two blacked files as examples. Thanks a lot.

PR #470, opened 2020-05-31
feat(encoder): improve wav2vec encoder with signal resampling

redram:
Wav2vec is trained and designed for 16 kHz sampling rate. The resampling of the input audio signal to 16 kHz should improve embedding quality.


hanxiao:
lgtm! thanks for your contribution ❤️

PR #471, opened 2020-05-31
feat(encoder): add wav2vec speech encoder

redram:
wav2vec-based speech encoder

### DIFF
--- a/jina/executors/encoders/audio/fairseq.py
+++ b/jina/executors/encoders/audio/fairseq.py
@@ -1,0 +1,75 @@
+import numpy as np
+
+from .. import BaseAudioEncoder
+from ..frameworks import BaseTorchEncoder
+from ...decorators import batching, as_ndarray
+
+
+class Wav2VecSpeechEncoder(BaseTorchEncoder, BaseAudioEncoder):
+    """
+    :class:`Wav2VecSpeechEncoder` is a speech encoder based on `wav2vec` - unsupervised pre-training for speech
+    recognition presented and implemented by Facebook: https://github.com/pytorch/fairseq/tree/master/examples/wav2vec
+    :class:`Wav2VecSpeechEncoder` uses a pre-trained model to encodes an audio signal from a `Batch x Signal Length`
+    ndarray into a `Batch x Concatenated Features` ndarray.
+    """
+
+    def __init__(self, model_path: str, *args, **kwargs):
+        """
+        Wav2vec model produces a representation for each time step at a rate of 100 Hz.
+
+        :param model_path: the path of the pre-trained model. The pre-trained model can be downloaded at
+            https://github.com/pytorch/fairseq/blob/master/examples/Wav2Vec/README.md#pre-trained-models
+        """
+        super().__init__(*args, **kwargs)
+        self.model_path = model_path
+
+    def post_init(self):
+        import torch
+        from fairseq.models.wav2vec import Wav2VecModel
+        cp = torch.load(self.model_path, map_location=torch.device('cpu'))
+        self.model = Wav2VecModel.build_model(cp['args'], task=None)
+        self.model.load_state_dict(cp['model'])
+        self.model.eval()
+        self.to_device(self.model)
+        self._sess_func = None
+        self._tensor_func = torch.tensor
+
+    @batching
+    @as_ndarray
+    def encode(self, data: np.ndarray, *args, **kwargs) -> np.ndarray:
+        """
+        Segments the audio signal of each Chunk into wav2vec frames, encodes the frames and concatenates Chunk frame
+        embeddings into a single Chunk embedding.
+
+        :param data: a `Batch x Signal Length` ndarray, where `Signal Length` is a number of samples
+        :return: a `Batch x Concatenated Features` ndarray, where `Concatinated Features` is a 512-dimensional feature
+        vector times the number of the wav2vec frames
+        """
+        assert data.shape[1] >= 465, 'the signal must have at least 465 samples'
+        embeds = []
+        with self.session():
+            for chunk_data in data:
+                signal_tensor = self.array2tensor(chunk_data.reshape(1, -1))
+                features = self.model.feature_extractor(signal_tensor)
+                embed_tensor = self.model.feature_aggregator(features)[0]
+                chunk_embed = self.tensor2array(embed_tensor).T.flatten()
+                embeds.append(chunk_embed)
+        result = np.array(embeds)
+        return result
+
+    def array2tensor(self, array):
+        tensor = self._tensor_func(array)
+        return tensor.cuda() if self.on_gpu else tensor
+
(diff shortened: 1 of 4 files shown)


jina/executors/encoders/audio/fairseq.py
```diff
@@ -0,0 +1,51 @@
...
+        super().__init__(*args, **kwargs)
+        self.model_path = model_path
+
+    def post_init(self):
+        import torch
+        from fairseq.models.wav2vec import Wav2VecModel
+        cp = torch.load(self.model_path, map_location=torch.device('cpu'))
+        self.model = Wav2VecModel.build_model(cp['args'], task=None)
+        self.model.load_state_dict(cp['model'])
+        self.model.eval()
```
nan-wang, line 29: It would be great if the codes can be run on both CPUs and GPUs. Please refer to the codes at [torchvision.py#L52](https://github.com/jina-ai/jina/blob/master/jina/executors/encoders/image/torchvision.py#L52)
```
self.to_device(self.model)
```

jina/executors/encoders/audio/fairseq.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+from .. import BaseAudioEncoder
+from ..frameworks import BaseTorchEncoder
+from ...decorators import batching, as_ndarray
+
+
+class Wav2VecSpeechEncoder(BaseTorchEncoder, BaseAudioEncoder):
+    """
+    :class:`Wav2VecSpeechEncoder` Wav2Vec-based speech encoder presented by Facebook https://github.com/pytorch/fairseq
+    that encodes an audio signal from a `Batch x Signal Length` into a `Batch * Frames x Features` ndarray.
```
nan-wang, line 10: It's better to put the model download url here.
```
https://github.com/pytorch/fairseq/blob/master/examples/wav2vec/README.md#pre-trained-models
```

jina/executors/encoders/audio/fairseq.py
```diff
@@ -0,0 +1,51 @@
...
+        embeds = []
+        for signal in data:
+            signtal_tensor = torch.from_numpy(signal).unsqueeze(0)
+            with torch.no_grad():
+                features = self.model.feature_extractor(signtal_tensor)
+                embed_tensor = self.model.feature_aggregator(features)[0]
+                embed_frames = embed_tensor.numpy().T
+                embeds.extend(embed_frames.tolist())
+        result = np.array(embeds)
+        return result
```
nan-wang, line 51: The output of encoding is expected to be `Batch x Features` so that each ``Chunk`` has one vector representation and thus the ``Indexer`` can index properly. However, in your case, the output is `(Batch x Frames) x Features`. I'd suggest to figure out one way to reduce the representations to the shape of `Batch x Features`, for example concatenating all the `Frames` for each ``Chunk``.

nan-wang: Here are some suggestions to speed it up a little bit and make it runnable on both CPUs and GPUs.

```
class Wav2VecSpeechEncoder(BaseTorchEncoder, BaseAudioEncoder):
    """
    :class:`Wav2VecSpeechEncoder` Wav2Vec-based speech encoder presented by Facebook https://github.com/pytorch/fairseq
    that encodes an audio signal from a `Batch x Signal Length` into a `Batch * Frames x Features` ndarray.
    """

    def __init__(self, model_path: str, *args, **kwargs):
        """
        Wav2Vec produces a representation for each time step at a rate of 100 Hz.

        :param model_path: the path of the pretrained model. The pretrained model can be downloaded at
            https://dl.fbaipublicfiles.com/fairseq/wav2vec/wav2vec_large.pt
        """
        self.model_path = model_path

    def post_init(self):
        import torch
        from fairseq.models.wav2vec import Wav2VecModel
        cp = torch.load(self.model_path, map_location=torch.device('cpu'))
        self.model = Wav2VecModel.build_model(cp['args'], task=None)
        self.model.load_state_dict(cp['model'])
        self.model.eval()
        self.to_device(self.model)
        self._sess_func = None
        self._tensor_func = torch.tensor

    @batching
    @as_ndarray
    def encode(self, data: np.ndarray, *args, **kwargs) -> np.ndarray:
        """
        Segments the audio signal into frames and encodes the frames.

        :param data: a `Batch x Signal Length` ndarray, where `Signal Length` is a number of samples
        :return: a `Batch * Frames x Features` ndarray, where `Features` is a 512-dimensional vector
        """
        assert data.shape[1] >= 465, 'the signal must have at least 465 samples'
        embeds = []
        with self.session():
            for signal in data:
                signtal_tensor = self.array2tensor(signal.reshape(1, -1))
                features = self.model.feature_extractor(signtal_tensor)
                embed_tensor = self.model.feature_aggregator(features)[0]
                embeds.append(self.tensor2array(embed_tensor))
        return np.concatenate(embeds, 1).T

    def array2tensor(self, array):
        tensor = self._tensor_func(array)
        return tensor.cuda() if self.on_gpu else tensor

    def tensor2array(self, tensor):
        return tensor.cpu().numpy() if self.on_gpu else tensor.numpy()

    @property
    def session(self):
        if self._sess_func is None:
            self._sess_func = self.get_session()
        return self._sess_func

    def get_session(self):
        from torch import no_grad
        return no_grad
```

hanxiao:
unittest is broken 
```
ImportError: cannot import name 'JinaTestCase' from 'tests' (/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/site-packages/tests/__init__.py)
```

redram:
> unittest is broken
I cannot reproduce the issue locally - the test runs successfully:
```
export JINA_TEST_PRETRAINED=1 && python -m unittest tests/executors/encoders/audio/test_fairseq.py
```

jina/executors/encoders/audio/fairseq.py
redram: I've changed the `encode` method to return `Batch x Concatenated Features`.

nan-wang:
recheckcla

hanxiao:
@nan-wang I don't understand why *all* unit tests failed in this PR? is something related to `__init__.py` structure in the `tests`?

redram:
The issue with the unit tests might relate to where/how `fairseq` is installed.

hanxiao:
> The issue with the unit tests might relate to where/how `fairseq` is installed.

dependencies all good. often this type of error is due to cyclic imports.

nan-wang:
> > The issue with the unit tests might relate to where/how `fairseq` is installed.
> 
> dependencies all good. often this type of error is due to cyclic imports.

This is actually due to the `fairseq` installation. I've made some trials at [https://github.com/jina-ai/jina/pull/491](https://github.com/jina-ai/jina/pull/491). Removing the `fairseq` dependency fixed the issue, while disabling the unit tests does **NOT** help.

hanxiao:
closed since #491 is merged

PR #492, opened 2020-06-03
feat(logging): use gevent as logging server

fhaase2:
Similar to the REST gateway, jina should also use gevent for the logging component. The built-in flask server is only suited for local development.


fhaase2:
reckeckcla

PR #496, opened 2020-06-04
fix(gateway): add message field to gateway

hanxiao:


hanxiao:
geez, finally. but what's wrong with `test_shards_insufficient_data`

PR #509, opened 2020-06-07
refactor(flow): refactor build inplace param

fhaase2:
While i made my way through the flow api code, i recognized that the build function has a `inplace` param, while e.g. `add` method exposes the `copy_flow` variable. I thought it might make more sense to have just one variable name. I am not sure about backwards compatibility, as this build method should be used just internally?

### DIFF
--- a/jina/flow/__init__.py
+++ b/jina/flow/__init__.py
@@ -322,7 +322,7 @@ class Flow:
 
         return op_flow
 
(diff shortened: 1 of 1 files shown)


hanxiao:
Thanks for your contribution. I see quite some active contribution from you recently? If you are interested in a full-time position, feel free to [add me on Linkedin](https://www.linkedin.com/in/hxiao87/)

PR #515, opened 2020-06-11
refactor: string formatting

fhaase2:
Right now the code base consists of a mixture of string formatting methods (format, %s, f-string...). 
I think the code base would benefit from a more or less uniform style. F-strings are typically perceived more readable in the python community and are also way faster. In this PR i have refactored already a lot (not all) formattings.


hanxiao:
wow this is a lot of work! greatly appreciated! thanks a lot! ❤️

Issue #513, opened 2020-06-10
Error message should propagate through the workflow in the query time


hanxiao:
**Describe the feature**

Right now a single error in the query time will break the whole system.

**Your proposal**

---

**Environment**

**Screenshots**

rutujasurve94:
Hi! I'd like to contribute to this issue

hanxiao:
❤️ Awesome! thanks for taking the initiative!

Let me explain the issue in more details.

Right now when an `Executor` throws an exception, the `Pod` catches and breaks from the eventloop. No message will be propagated further. The followed `Pod` in this workflow stuck there forever as they work in `recv->handle->send` way. Since there is no receive, nothing happens. The gateway receives nothing and has nothing to return to the client. So everybody stuck there forever.

There are many ways to solve this problem, the simplest way I have in mind is:
- Adding a `status` field in the Protobuf `Envelope`, when a Pod throws a RuntimeError, it changes the `status` to `error`, and propagates the message further. Also better keeps the exception message in the message.
- The followed Pods simply skip the handling part and propagate it further when they see the `error` status in the envelope.
- The gateway returns the error code and message to the client.

rutujasurve94:
Thanks a lot for the leads and details on the issue, I will begin the implementation along these steps and update here

rutujasurve94:
Hello,
I've opened PR: https://github.com/jina-ai/jina/pull/521 for this issue.
**Proposal for the feature:**
-  Changes to Protobuf ```Envelope```: this already has a ```Status``` field where status could be 'Error'. Captured the error's message as well by adding a field  ```error_message``` to this proto.
-  Built the new changes to proto with ```build-proto-with-python.sh``` script in the proto directory,  auto-modifying ```jina_pb2.py```,  ```jina_pb2_grpc.py```.
-  Added an Exception ```PodRunTimeError``` that could be generically used to raise exceptions every time the pod throws a run exception.
-  In pea: Changed the status of the message to error every time a run time exception is thrown in the loop body, propagating the message further.
-  Changed the pea's ```handle``` method: Pods simply skip the handling part and propagate it further when they see the error status in the envelope. 
- Changes to gateway: ```handle``` method: The gateway check's if the received message's envelope has Error status, in this case it returns the error code and message to the client. 

**Further evaluations:**
- [ ] Evaluated the ```msg_callback``` that catches NoExplicitMessage and passes when no message/ partial message is received.  No changes required here
[...]

PR #521, opened 2020-06-15
feat(peapod): add functionality to propagate error message through workflow

rutujasurve94:
**Adding functionality for propagating error messages through the workflow in the query time** (Issue #513)
**Problem:**
- Right now when an Executor throws an exception, the Pod catches it but then breaks from the event-loop. No message will be propagated further. 
- The followed Pod in this workflow stuck there forever as they work in recv->handle->send way. Since there is no receive, nothing will be sent. 
- The gateway also receives nothing and has nothing to return to the client. So everybody stuck there forever.

**Used the following to approach to solve the issue:**
-  Changes to Protobuf ```Envelope```: this already has a ```Status``` field where status could be 'Error'. Captured the error's message as well by adding a field  ```error_message``` to this proto.
-  Built the new changes to proto with ```build-proto-with-python.sh``` script in the proto directory,  auto-modifying ```jina_pb2.py```,  ```jina_pb2_grpc.py```.
-  Added an Exception ```PodRunTimeError``` that could be generically used to raise exceptions every time the pod throws a run exception.
-  In pea: Changed the status of the message to error every time a run time exception is thrown in the loop body, propagating the message further.
-  Changed the pea's ```handle``` method: Pods simply skip the handling part and propagate it further when they see the error status in the envelope. 
- Changes to gateway: ```handle``` method: The gateway check's if the received message's envelope has Error status, in this case it returns the error code and message to the client. 

**Further evaluations:**
- [ ] Evaluated the ```msg_callback``` that catches NoExplicitMessage and passes when no message/ partial message is received.  No changes required here
- [ ]  Evaluated the drivers: ```control.py``` and  ```jina_pb2.Request.ControlRequest.STATUS```: also don't require changes.

@hanxiao : It would be great to have your review, requesting your feedback on the implementation.


hanxiao:
really awesome work! ❤️  THANK YOU. I'm reviewing it now

jina/peapods/pea.py
```diff
@@ -319,24 +320,39 @@ def run(self):
             self.post_init()
             self.loop_body()
         except RequestLoopEnd:
-            self.logger.info('break from the event loop')
+            self._message.envelope.status = jina_pb2.Envelope.Status.ERROR
```
hanxiao, line 323: Changing `message` does not have any effect. The message won't propagate as the `send` is in the `loop_body`

jina/peapods/gateway.py
```diff
@@ -114,6 +114,14 @@ def handle(self, msg: 'jina_pb2.Message'):
             try:
                 self._request = getattr(msg.request, msg.request.WhichOneof('body'))
                 self._message = msg
+                # Check if pod had thrown a run-time exception by checking envelope's status and error message
+                if msg.envelope.status == jina_pb2.Envelope.Status.ERROR:
+                    error_message = f'pod threw run time exception {msg.envelope.error_message}'
+                    self.logger.info(error_message)
+                    # extract error message from envelope, remove envelope, send back request and error message
+                    return msg.request, msg.envelope.error_message
```
hanxiao, line 122: Note that the gateway always removes the `Envelope` as those information is often not interested to the client. Except this `error_message`, the client has to know it. So maybe `error_message` should be carried as a part of `Request`.

hanxiao:
so one way to solve this is to make all exceptions listed below inherited from something say `LogicError`

image

A `LogicError` should break the eventloop of a pea as there is no way to continue further.

All other errors, i.e. the current `except Exception as ex:` is considered as `RuntimeError`. They should not break the eventloop, and should be caught inside `handle()`.

 I will make some change on your branch to reflect that.

jina/peapods/pea.py
rutujasurve94: incorporated the change, raising pod run time exception in the last except Exception as ex clause below

jina/peapods/gateway.py
rutujasurve94: changed the Request proto to include the error message

rutujasurve94:
Thanks a lot for the review, I have made changes and updated the PR. 

1. ```Status``` and ```error_message``` are now a part of ```Request``` proto and not Envelope to enable handling in the ```gateway```. 
2.  Reverted changes to ```Logic Errors``` and raised ```PodRunTimeException``` from the except clause for ```RunTimeError``` 
3. Catching run time exception by ```pea``` inside the ```handle()``` function.

hanxiao:
since the CI is passing now, to prevent this PR to grow too big, i'm going to merge it for now. I will refactor protobuf in another PR. could you please sign the CLA first?

hanxiao:
recheckcla

rutujasurve94:

rutujasurve94:
Thank you for the feedback and commits for improving catch error logic and refactoring of excepts. I have signed the CLA as well.

PR #524, opened 2020-06-15
feat: add `if` expression to BaseDriver

hanxiao:
`if` expression is added to `BaseDriver` to allow user filter docs/chunks by a *single expression*.

Examples:

only encode MIME `text/*`
```yaml
!Encode1
requests:
  on:
    IndexRequest:
      - !EncodeDriver
        if: doc.mime_type.startswith('text')
```

One can also use it in Python:

only prune docs with even `doc_id`

```python
class PruneEvenDriver(PruneDriver):
    if_expression = 'doc.doc_id % 2 ==0'

    def __init__(self, pruned=('text',), level='doc', *args, **kwargs):
        super().__init__(pruned, level, *args, **kwargs)
```

## Limitations

- Does not work with "compositional expression":
   - Like the `lambda` function, `if` supports the **single expression** only: https://docs.python.org/3/reference/expressions.html
   - Can not filter `docs` AND `chunks` at the same time, e.g. `doc.doc_id % 2 == 0 && chunk.chunk_id % 2 == 0` does not work
- Relying on the assumption that `doc` and `chunk` are always in the context of `eval`.

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -3,7 +3,7 @@ __license__ = "Apache-2.0"
 
 import inspect
 from functools import wraps
-from typing import Callable, List
+from typing import Callable, Iterator
 
 import ruamel.yaml.constructor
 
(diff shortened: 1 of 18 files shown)


fhaase2:
Implementation looks awesome 👍  eval could be unsafe, but it could be easily extended with a validator in the future.

One thing i have noticed is that `list(self.chunks(d))` is used multiple times in `ChunkCraftDriver`, which converts the generator multiple times to a list. Would be faster to do that only once and reuse that list.

hanxiao:
yep, the `list()` part has been refactored.

Issue #522, opened 2020-06-15
callback_on_body as default output_fn behavior


hanxiao:
**Describe the feature**

**Your proposal**

---

**Environment**

**Screenshots**

policeme:
after change, it just return the request body, it will lost some informations of request, such as the status of the request. 
such as this kind of code.
```
def test_bad_flow_customized(self):
    def validate(req):
        self.assertEqual(req.status.code, jina_pb2.Status.ERROR)
        self.assertEqual(req.status.details.pod, 'r2')
        self.assertTrue(req.status.details.skipped[0].startswith('r3'))
        self.assertTrue(len(req.status.details.skipped) == 1)
        self.assertTrue(req.status.details.exception.startswith('ZeroDivisionError'))

    class DummyCrafter(BaseDocCrafter):
        def craft(self, *args, **kwargs):
            return 1 / 0

    f = (Flow().add(name='r1', yaml_path='_forward')
         .add(name='r2', yaml_path='!DummyCrafter')
         .add(name='r3', yaml_path='!BaseEncoder'))

    with f:
        f.dry_run()

    with f:
        f.index_lines(lines=['abbcs', 'efgh'], output_fn=validate)
```
so this feature is really necessary? 
or my understanding of this feature is wrong.

PR #529, opened 2020-06-16
feat(client): callback_on_body as default output_fn behavior

policeme:


nan-wang:
@hanxiao The target issue [#522](https://github.com/jina-ai/jina/issues/522) is not clear. Would you please elaborate on it?

PR #531, opened 2020-06-16
fix the bug in query with faiss

### LINKED ISSUE #530: the training function is missing in the Faiss Indexer
**Describe the bug**
the index object initialized by `faiss.index_factory()` should be trained before loading the indexed data. 

**Describe how you solve it**
- implement the `train()` function for `FaissIndexer`

---

**Environment**

**Screenshots**

nan-wang:
resolves #530

### DIFF
--- a/jina/executors/indexers/vector/faiss.py
+++ b/jina/executors/indexers/vector/faiss.py
@@ -31,19 +31,9 @@ class FaissIndexer(NumpyIndexer):
         self.index_key = index_key
         self.train_filepath = train_filepath
 
-    def post_init(self):
-        super().post_init()
-        import faiss
-        self._index = faiss.index_factory(self.num_dim, self.index_key)
-
     def get_query_handler(self):
         """Load all vectors (in numpy ndarray) into Faiss indexers """
-        if not self.is_trained:
-            _train_data = self._load_training_data(self.train_filepath)
-            if _train_data is None:
-                self.logger.warning('loading training data failed.')
-                return None
-            self.train(_train_data)
+        import faiss
         _index_data = super().get_query_handler()
         if _index_data is None:
             self.logger.warning('loading indexing data failed.')
@@ -51,6 +41,13 @@ class FaissIndexer(NumpyIndexer):
         if _index_data.ndim != 2:
             self.logger.warning('the index data should be 2D tensor, {} != 2'.format(_index_data.ndim))
             return None
+        self._index = faiss.index_factory(self.num_dim, self.index_key)
+        if not self.is_trained:
+            _train_data = self._load_training_data(self.train_filepath)
+            if _train_data is None:
+                self.logger.warning('loading training data failed.')
+                return None
+            self.train(_train_data)
         self._index.add(_index_data.astype('float32'))
         return self._index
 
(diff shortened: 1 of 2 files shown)


nan-wang:
It would be better to have one example based on `SIFT1M`.

## Reference
- [faiss demo](https://github.com/facebookresearch/faiss/blob/master/demos/demo_auto_tune.py)
- [SIFT1M](http://corpus-texmex.irisa.fr/)

jina/executors/indexers/vector/faiss.py
```diff
@@ -12,35 +12,73 @@ class FaissIndexer(NumpyIndexer):
...
+        if not self.num_dim:
+            self.num_dim = _num_dim
+        if self.num_dim != _num_dim:
+            raise ValueError('training data should have the same number of features as the index, {} != {}'.format(
+                self.num_dim, _num_dim))
+        self._index.train(data)
+        self.is_trained = True
+
+    def _load_training_data(self, train_filepath):
+        return self._load_numpy(train_filepath)
```
hanxiao, line 84: is this necessary? aren't they exactly the same?

jina/executors/indexers/vector/faiss.py
```diff
@@ -12,35 +12,73 @@ class FaissIndexer(NumpyIndexer):
...
+
+    def train(self, data: 'np.ndarray', *args, **kwargs):
+        _num_samples, _num_dim = data.shape
+        if not self.num_dim:
+            self.num_dim = _num_dim
+        if self.num_dim != _num_dim:
+            raise ValueError('training data should have the same number of features as the index, {} != {}'.format(
+                self.num_dim, _num_dim))
+        self._index.train(data)
+        self.is_trained = True
```
hanxiao, line 81: i believe there is no need to set `is_trained` manually, wrapper in the `metaclass` does that
```suggestion
```

jina/executors/indexers/vector/numpy.py
```diff
@@ -138,6 +134,20 @@ def query(self, keys: np.ndarray, top_k: int, *args, **kwargs) -> Tuple['np.ndar
...
+        result = None
+        try:
+            if self.num_dim and self.dtype:
+                with gzip.open(abspath, 'rb') as fp:
+                    result = np.frombuffer(fp.read(), dtype=self.dtype).reshape([-1, self.num_dim])
+        except EOFError:
+            self.logger.error(
+                f'{self.index_abspath} is broken/incomplete, perhaps forgot to ".close()" in the last usage?')
+        return result
+
```
hanxiao, line 150: wait, are you sure to make this refactoring just for a helper function to load training data in faiss. afaik, the faiss training data loading is just a quick temporal solution. That helper function won't last long.

In fact, you don't need to load from a gzipped numpy for the training data. A simple numpy array would be good, simply use `np.load`, it is less error-prone and more generic. Otherwise, how do you expect user to generate `train.gz` in the first place??

nan-wang: The training data loading is necessary for faiss. Therefore, I consider two scenarios.
1. the user can use indexed data for training, which is kept in `index.tgz` with the `NumpyIndexer`.
2. as you suggest, the user can simple feed a numpy array. This is the reason why we have the wrapper function `_load_training_data`.

jina/executors/indexers/vector/faiss.py
nan-wang: Check out the comments below.

jina/executors/indexers/vector/faiss.py
nan-wang: Nope. `is_trained` is not updated automatically.

PR #554, opened 2020-06-21
fix(indexer): is trained set on train call

### LINKED ISSUE #534: train wrapper does not label `is_trained`?
**Describe your problem**

According to #531 , calling `train` does not mark an object as `is_trained=True`. I thought this is automatically done with metaclass wrapper?

**What is your guess?**

I remember @nan-wang mentioned this back in Mar. but I also remember I had this tested and there is no such problem. Perhaps it does not work under multi-level inheritance or some refactoring breaks this feature?

Better write a unit test to check this.

---

**Environment**

**Screenshots**

fhaase2:
Closes #534

### DIFF
--- a/jina/executors/decorators.py
+++ b/jina/executors/decorators.py
@@ -40,7 +40,6 @@ def as_train_method(func: Callable):
                                 'training it again will override the previous training' % self.__class__.__name__)
         f = func(self, *args, **kwargs)
         self.is_trained = True
-        print(self.is_trained)
         return f
 
     return arg_wrapper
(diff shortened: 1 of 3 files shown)


fhaase2:
I do not get why CI is failing. Unittests are working fine locally. If i remove `self.is_trained =  True` from `decorators.py` unittests will fail, as they should. Also if i test the added test case seperately it works:`python -m unittest tests.test_index_train -v`.

hanxiao:
btw now you have the privilege to directly create a branch on `jina-ai/jina` not work on your own fork. Give it a try. The `cla-check` and `label-pr` actions will be done automatically and you don't need to label them by hand.

Issue #573, opened 2020-06-26
Test concurrent connections on gRPC and REST gateway


hanxiao:
**Describe your problem**

Server starts with:
- GRPC Gateway
- REST Gateway

We have not yet test the scalability here.

**What is your guess?**

First some unit test to test it.
If things turn out badly, then need to optimize on this issue.

---

**Environment**

**Screenshots**

PR #579, opened 2020-06-27
test(gateway): concurrency test for rest gateway

fhaase2:
Added a concurrency test for the REST gateway (https://github.com/jina-ai/jina/issues/573). Performance looks not too bad for me.


hanxiao:
well done! would be nice to show the avg./max/min time for a roundtrip. Could you add to it as well?

PR #582, opened 2020-06-28
docs-101-102: 101 explains jina concept, 102 explains how it works

alexcg1:
As I discussed in Slack, I've been meaning to redo our docs slightly to give users a better intro to what Jina is (101), followed by how Jina works (102)

## My update:

* Jina 101: What is neural search and Jina? (adapted from [my draft medium post](https://medium.com/@alexcg_91001/ai-powered-neural-search-search-smarter-16fb7627fa8)
* Jina 102: How does Jina Work? (the original Jina 101)

I'm not expecting this to get merged right away, since I think some feedback may be needed on accuracy of the new 101. @hanxiao I think you're the best one to do that since it's all originally based on [your post](https://hanxiao.io/2018/01/10/Build-Cross-Lingual-End-to-End-Product-Search-using-Tensorflow/)


hanxiao:
One big flaw in the new Jina 101 is the explanation about "training". In the current Jina version, we don't have good support on "training", all `train()` methods are left blank and wait for implementation. But then why Jina works reasonably in all demos that you have seen. Because they are based on pretrained deep learning model, working in a completely unsupervised way.

Around the topic of "training" there are three sub-concepts here:
- **Training**, as we usually say, train this logistic regression.
- **Training a relevance/ranking model**, as described in [my blog post](https://hanxiao.io/2018/01/10/Build-Cross-Lingual-End-to-End-Product-Search-using-Tensorflow/).
- **Pretraining**, "warmup" a deep learning model by learning pattern from a general data set.

Mapping to the current (v0.3) and near future Jina implementation:
- **Training**: interface is there but no implementation yet.
- **Training a relevance/ranking model**: no interface
- **Pretraining**: we *load and use* the pretrained model but not do *pretraining*.

In the new Jina 101, the idea of **Training** and **Training a relevance/ranking model** is over-emphasised. I'm afraid the new users when they actually start to use Jina, they feel "under-delivered" comparing to what described in 101, as none of the training paradigms are there yet.

- So will we have the training interface as you wrote in 101? **Yes**, but not in the near future.

- So why such "important" concept is not **prioritized**? Because Jina follows [the design philosophy: "training is expensive and you don't need it"](https://hanxiao.io/2019/07/29/Generic-Neural-Elastic-Search-From-bert-as-service-and-Go-Way-Beyond/#foreword-few-thoughts-on-ai-innovation-and-engineering). Important as it seems, many people don't really need "training". This tenets greatly influence the design pattern, the architecture decision and it is even the main motivation behind Jina Hub, i.e. you don't train the model, when you need one you simply pull it from the hub.

  By abstract away "training" from the system, Jina offers much simpler interface and smoother learning curve. Decoupling training from the system also allows us to focus on the actual problems in search-in-production, which otherwise will wind up into a pure research-oriented software.

All I can say is that they are still under radar, they are part of our milestones. But the current Jina works out-of-the-box without "sophisticated training".

hanxiao:
In summary, the new 101 describes how neural search system looks like. Makes totally sense to researchers or ML-oriented persons. However, it does accurately explain how Jina works today. Jina has evolved from that blog post in 2017 and ends up to a modern, production-oriented neural search system.

alexcg1:
Understood - that actually makes it even easier to explain. So we're "just" loading a pretrained model and not doing fine-tuning (yet) - that would explain Jina's speed. 'll rewrite the post to reflect that.

alexcg1:
Updated it to talk about pretrained models (and plug jina hub), and removed most stuff about actual training/fine-tuning for now. Can always add it in when we implement training.

@hanxiao could you get back to me on final check by end of your day? Thanks!

alexcg1:
Ping @hanxiao

Issue #585, opened 2020-06-28
rename `port_grpc`


hanxiao:
**Describe the feature**

`port_grpc` is used in REST Gateway as well, the name does not fit its purpose.

**Your proposal**

rename it to `port_expose`?

---

**Environment**

**Screenshots**

fhaase2:
I also recognized this. Another solution would be to add an additional parameter e.g. `port_rest`. However, for user its probably better to have only one param.

nan-wang:
When do we expect to deprecate `port_grpc`?

PR #598, opened 2020-06-29
feat: rename port grpc

fhaase2:
https://github.com/jina-ai/jina/issues/585
I added a depreciated warning in the parser


alexcg1:
Thanks for the PR @fhaase2 :smile:

hanxiao:
ok the reason of the failing is because we use `jina-devel` Docker image in some unit test (for docker support in Flow API). Your API change makes the current `jina-devel invalid, but a new `jina-devel` can only be built after PR is merged. Therefore, a deadlock.

I will break the deadlock by manually push `jina-devel` image.

btw, this is also a good indicator tells us this PR introduces a breaking change.

fhaase2:
Yes, i see, thanks @hanxiao . But why do we use the `devel` image from dockerhub in the unittest? This does not test the newest code introduced by PR... in the worst case this image is already pretty old. Should we also change it and use an image built from the PR code?

hanxiao:
That is one of the tech debts we have.

You are right, and a better solution would be: replacing `jina:devel` in the following code to `jinaai/jina:test-pip`
image

As you can see in `.github/workflows/ci.yml` L62, `jinaai/jina:test-pip` is built before the unit test.

fhaase2:
I will fix it in a different PR.

fhaase2:
> I will fix it in a different PR.

https://github.com/jina-ai/jina/pull/609

Issue #608, opened 2020-07-02
Use image built from PR in unittest


fhaase2:
**Describe your problem**

In unittests we use the `jinaai/jina:devel` image, which is pulled from dockerhub. This image does not contain the newest code changes of PR, and is therefore not suitable for the CI. We should change it to use `jinaai/jina:test-pip`

**What is your guess?**

---

**Environment**

**Screenshots**

PR #609, opened 2020-07-02
test: use image built from pr in tests

fhaase2:
Resolves https://github.com/jina-ai/jina/issues/608


nan-wang:
Great PR! Thanks for your contributions again! @fhaase2

hanxiao:
Perfect, I was going to tell you that `jina-devel` in Dockerfile blocked your previous commits this morning. But then was distracted by something else. Glad you figure that out! Nice!

JoanFM:
Regarding this, we should document, how to build a local docker image with this name so that these tests can be run in local by developers who may fear are breaking container-related functionality

PR #611, opened 2020-07-02
ci(unittest): allow test discovery from unittest

JoanFM:
Changing the CI command to execute unittests.


JoanFM:
Tests under executors.crafters.audio are failing due to the numba version in the system. librosa requires numba version==0.48 because from 0.50 and up it removes decorators which are needed. It would be important to know if there is another dependency for which we need numba library.

JoanFM:
Test executor.crafter.segmenter fails functionally

nan-wang:
> Tests under executors.crafters.audio are failing due to the numba version in the system. librosa requires numba version==0.48 because from 0.50 and up it removes decorators which are needed. It would be important to know if there is another dependency for which we need numba library.

`librosa` is only used in `executors.crafters.audio`.

nan-wang:
#612 solved in this issue.

PR #614, opened 2020-07-03
feat(executor): add on_gpu: auto

policeme:


policeme:
due to this situation, I think we can delete the on gpu in the [meta default](https://github.com/jina-ai/jina/blob/master/jina/resources/executors.metas.default.yml)

jina/executors/frameworks.py
```diff
@@ -25,11 +25,13 @@ def device(self):
             please use the environment variable `CUDA_VISIBLE_DEVICES`.
         """
         try:
+            import GPUtil
+            self.on_gpu = True if len(GPUtil.getGPUs()) > 0 else False
```
policeme, line 29: or change this line like this
```
if self.on_gpu:
     self.on_gpu = True if len(GPUtil.getGPUs()) > 0 else False
```

hanxiao: 1. `self.on_gpu = len(GPUtil.getGPUs())`
2. kinda waste to use a third party package just for detecting if GPU exist

hanxiao:
> due to this situation, I think we can delete the on gpu in the [meta default](https://github.com/jina-ai/jina/blob/master/jina/resources/executors.metas.default.yml)

no we can't

jina/executors/frameworks.py
policeme: 1. now there is no a better way to detect if GPU exist, so I use `GPUtil` to detect.
2. sorry about that, at the first commit, I tested in my GPU machine, I ignored this part. now I test in my MAC, it works
```
 def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
        """

        :param data: a 1d array of string type in size `B`
        :return: an ndarray in size `B x D`
        """
        token_ids_batch = []
        mask_ids_batch = []
        for c_idx in range(data.shape[0]):
            token_ids = self.tokenizer.encode(
                data[c_idx], pad_to_max_length=True, max_length=self.max_length)
            mask_ids = [0 if t == self.tokenizer.pad_token_id else 1 for t in token_ids]
            token_ids_batch.append(token_ids)
            mask_ids_batch.append(mask_ids)
        token_ids_batch = self.array2tensor(token_ids_batch)
        mask_ids_batch = self.array2tensor(mask_ids_batch)
```
it will use `on_gpu` before set device.

policeme:
> > due to this situation, I think we can delete the on gpu in the [meta default](https://github.com/jina-ai/jina/blob/master/jina/resources/executors.metas.default.yml)
> 
> no we can't

1. follow this comment, I keep the setting of `on_gpu` in yaml file, so the user can choose it according their needs.
2. if set `on_gpu` is `True`, it will test if GPU exist, if existed, it will use GPU, otherwise, back to CPU . 
3. if set `on_gpu` is `False`, it always use CPU.

policeme:
> > > due to this situation, I think we can delete the on gpu in the [meta default](https://github.com/jina-ai/jina/blob/master/jina/resources/executors.metas.default.yml)
> > 
> > 
> > no we can't
> 
> 1. follow this comment, I keep the setting of `on_gpu` in yaml file, so the user can choose it according their needs.
> 2. if set `on_gpu` is `True`, it will test if GPU exist, if existed, it will use GPU, otherwise, back to CPU .
> 3. if set `on_gpu` is `False`, it always use CPU.

if this, should we change the default value to `True` in [meta default](https://github.com/jina-ai/jina/blob/master/jina/resources/executors.metas.default.yml), it will auto use GPU

jina/executors/frameworks.py
```diff
@@ -14,7 +14,9 @@ class BaseFrameworkExecutor(BaseExecutor):
 
     def __init__(self, model_name: str = None, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        import GPUtil
         self.model_name = model_name
+        self.on_gpu = True if self.on_gpu and GPUtil.getGPUs() else False
```
hanxiao, line 19: again, redundant boolean logic
```suggestion
        self.on_gpu = self.on_gpu and GPUtil.getGPUs()
```

hanxiao:
im closing this PR as it has too many problems:
- not set gpu at the right place, misunderstanding on the class structure and pattern
- use third party package for detecting GPU, introduce extra dependency
- no error catching
- redundant boolean logic, poor code quality

I re-submit a new PR #617 , please refer to that.

PR #626, opened 2020-07-06
fix(encoders): fix mask prune creation

JoanFM:
**Issue link**
https://github.com/jina-ai/examples/issues/90

**Description**
Small indexing error when trying to create the prune mask if cls_pos is 'tail' as in the case of model xlnet-base-cased.

**Changes introduced**
Fix the indexing issue, add a couple of tests with the required model in the issue.
Identified another issue while running tests that some models do not implement _resize_token_embeddings_


nan-wang:
@hanxiao @JoanFM Wrapping all the models from transformers is quite error-prone because we haven’t tested all the models and the models are updated quite offen. What about including the classic models in jina core and leaving the rest in the jina hub?

JoanFM:
@nan-wang do you mean adding explicitly support only for the most common ones?

In this case, the problem was of code path coverage, but I guess that the combinations of features and paths can explode.

Issue #620, opened 2020-07-05
Potential code improvement: Incongruency with Reader crafters (ImageReader, AudioReader, ArrayReader) in the io.py of different crafter modules.


JoanFM:
**Describe your problem**
In this Issue I want to discuss a potential imrpovement of the code.

It seems to me that is a little incongruency on how these *Reader classes are built. While **AudioReader** and **ArrayReader** inherit from **BaseDocCrafter**, **ImageReader** inherits from **BaseSegmenter**.

It seems weird that they all craft chunks while setting the offset of the chunk always to 0. This could be an indicator that this needs to be a document and not a chunk.

Would it make more sense to make all of them inherit from BaseDocCrafter, always return a document in the output (no offset set in the dict)? 

Like this, a common pattern would be use to use a CompoundExecutor in the same pod where one can combine these Readers with a Segmenter. The Readers would be driven by a DocCraftDriver, while the segmenter by a SegmentDriver.

I think this way the code would be more coherent.

PR #627, opened 2020-07-06
refactor(crafters): io crafters should craft documents

JoanFM:
PR implementing proposal from #620. Linked PR in examples: https://github.com/jina-ai/examples/pull/92

**Changes introduced**
- ImageReader inherits from BaseDocCrafter and returns one document and not chunks. 
- ArrayReader and AudioReader return a document dict, so no offset provided.
- Add tests of image reading from buffer, since often pod may not have the images in a local uri.
- Use CopyFrom in DocCraftDriver to pass blob message field. Avoid error: **_assignment not allowed to field in protocol message object_**

### DIFF
--- a/jina/drivers/craft.py
+++ b/jina/drivers/craft.py
@@ -75,7 +75,13 @@ class DocCraftDriver(BaseCraftDriver):
             ret = self.exec_fn(**_args_dict)
             if ret:
                 for k, v in ret.items():
-                    setattr(d, k, v)
+                    if k == 'blob':
+                        if isinstance(v, jina_pb2.NdArray):
+                            d.blob.CopyFrom(v)
+                        else:
+                            d.blob.CopyFrom(array2pb(v))
+                    else:
+                        setattr(d, k, v)
 
 
 class SegmentDriver(BaseCraftDriver):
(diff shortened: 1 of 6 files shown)


JoanFM:
Hey @nan-wang, I did also a PR to examples in  jina-ai/examples#92 trying to adapt the cases where ImageReader is used in flower search and face-db-search

jina/drivers/craft.py
```diff
@@ -75,7 +75,13 @@ def __call__(self, *args, **kwargs):
...
             if ret:
                 for k, v in ret.items():
-                    setattr(d, k, v)
+                    if k == 'blob':
+                        if isinstance(v, jina_pb2.NdArray):
+                            d.blob.CopyFrom(v)
+                        else:
+                            d.blob.CopyFrom(array2pb(v))
+                    else:
+                        setattr(d, k, v)
```
hanxiao, line 84: I do see this part is related to https://github.com/jina-ai/jina/blob/ac6175a7d4ca90d8a49ae93a0f1ec61390ec0abf/jina/drivers/craft.py#L19 and some refactoring in the future could make it more consistent. After all, doc and chunk share the same `content` proto spec.

hanxiao: but let's leave it as a future work

PR #628, opened 2020-07-06
test(flow): improved log server tests

fhaase2:
I added tests for some endpoints. Stream endpoints are not yet covered, because its actually hard to test as they are hanging the connection


hanxiao:
👍

PR #629, opened 2020-07-07
docs(readme): small adjustments based on han's feedback

alexcg1:
Just very minor tweaks from README on `master`. For larger changes I'll roll each change into a PR to make yes/no choices easier :)


alexcg1:
Looks like the lint-checker is choking on something (which isn't my commit text)

Screenshot from 2020-07-07 15-56-31

PR #634, opened 2020-07-07
feat(flow): dryrun is running by default

hanxiao:
solves the following problems:
- `dry_run` should be triggered *from* the client, not triggered by "flow".
- `dry_run` should mimic an empty Index/Search/TrainRequest, not just Control Request. Only in this way, the driver of each peapod can be fully tested.


jina/main/parser.py
```diff
@@ -170,6 +170,8 @@ def set_flow_parser(parser=None):
                     help='type of the output')
     gp.add_argument('--output-path', type=argparse.FileType('w', encoding='utf8'),
                     help='output path of the flow')
+    gp.add_argument('--skip-dryrun', action='store_true', default=False,
```
fhaase2, line 173: Would better name it `- -skip-dry-run`

jina/flow/__init__.py
```diff
@@ -471,6 +471,9 @@ def start(self):
             self.num_pods,
             self.num_peas))
 
+        if not self.args.skip_dryrun:
+            self.dry_run()
```
fhaase2, line 475: if change below, also change `skip_dry_run`

jina/main/parser.py
fhaase2: https://helm.sh/docs/chart_template_guide/debugging/ they guys from helm also called it `- -dry-run`

hanxiao:
unfortunately dry run is not general enough now. as dryrum is implemented as a ControlRequest so it only triggers whatever driver binds to the control request. If user defines a reduce driver binds to the SearchRequest, then this reduce driver will not be triggered. One example is that the query flow part of the newly added unittest always fail. @fhaase2

fhaase2:
> unfortunately dry run is not general enough now. as dryrum is implemented as a ControlRequest so it only triggers whatever driver binds to the control request. If user defines a reduce driver binds to the SearchRequest, then this reduce driver will not be triggered. One example is that the query flow part of the newly added unittest always fail. @fhaase2

Is might be possible to extend `BaseDriver` or `BaseExecutableDriver` to always bind it to control request for dry run?

JoanFM:
> > unfortunately dry run is not general enough now. as dryrum is implemented as a ControlRequest so it only triggers whatever driver binds to the control request. If user defines a reduce driver binds to the SearchRequest, then this reduce driver will not be triggered. One example is that the query flow part of the newly added unittest always fail. @fhaase2
> 
> Is might be possible to extend `BaseDriver` or `BaseExecutableDriver` to always bind it to control request for dry run?

Regarding this,  I have been wondering if it would make sense for Jina to limit the combinations of 'Drivers', 'Executors' and 'Requests'. 

I think is cool to have this flexible design, but at some point, could the user get overwhelmed with so many options?

hanxiao:
> LGTM👍

unfortunately this changes do not pass the test for the reason mentioned above

hanxiao:
It's the problem of abstraction. 

What we (core team of Jina) are doing is still on pretty low-level. We are meta-developers, we develop something for developers. Our main goal in `jina-core` is to make it as universal as it can to serve all kinds of neural search scenarios. In the end, we do not want to expose all these low-level concepts to the users. We abstract and provide high-level interface*s* targeting to different levels of developers. Flow API is an example, dashboard is at even higher level.

Below is an old figure but very nicely illustrated the layers of abstraction we provide.

[image](https://hanxiao.io/2019/07/29/Generic-Neural-Elastic-Search-From-bert-as-service-and-Go-Way-Beyond/)



> > > unfortunately dry run is not general enough now. as dryrum is implemented as a ControlRequest so it only triggers whatever driver binds to the control request. If user defines a reduce driver binds to the SearchRequest, then this reduce driver will not be triggered. One example is that the query flow part of the newly added unittest always fail. @fhaase2
> > 
> > 
> > Is might be possible to extend `BaseDriver` or `BaseExecutableDriver` to always bind it to control request for dry run?
> 
> Regarding this, I have been wondering if it would make sense for Jina to limit the combinations of 'Drivers', 'Executors' and 'Requests'.
> 
> I think is cool to have this flexible design, but at some point, could the user get overwhelmed with so many options?

PR #635, opened 2020-07-08
feat(examples): Add changes to implement Faiss Search examples

JoanFM:
**Changes introduced**

> Added **_UnaryEncoderDriver_**, inspired by **_UnarySegmentDriver_**. Useful for testing purposes and when no encoding is needed.

> Change name of method **_search_numpy_** to **search_ndarray** to match naming of the _**index_ndarray**_ counterpart

> Remove unused **_MySegmenter_** fro HelloWorld

These changes are needed to implement changes required for https://github.com/jina-ai/examples/pull/89

### DIFF
--- a/jina/drivers/encode.py
+++ b/jina/drivers/encode.py
@@ -35,3 +35,13 @@ class EncodeDriver(BaseEncodeDriver):
                     'the first dimension must be the same' % (len(chunk_pts), embeds.shape))
             for c, emb in zip(chunk_pts, embeds):
                 c.embedding.CopyFrom(array2pb(emb))
+
+
+class UnaryEncodeDriver(EncodeDriver):
+    """The :class:`UnaryEncodeDriver` extracts the chunk-level content from documents and copies
+        the same content to the embedding
+    """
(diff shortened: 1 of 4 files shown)


jina/drivers/encode.py
```diff
@@ -35,3 +35,13 @@ def __call__(self, *args, **kwargs):
...
+
+
+class UnaryEncodeDriver(EncodeDriver):
+    """The :class:`UnaryEncodeDriver` extracts the chunk-level content from documents and copies
+        the same content to the embedding
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._method_name = None
+        self._exec_fn = lambda x: x
```
hanxiao, line 47: What does this do? Looks like a NOP or placeholder?

JoanFM: it will mean a unary encoder as an executor function to be called by the driver

nan-wang: The usage is similar as the `UnarySegmentDriver`. This is useful for testing and debugging.

PR #638, opened 2020-07-09
feat(executors): add nprobe parameter to faiss indexer

JoanFM:
**Changes introduced**

Ass nprobe parameter to FAISS Indexer. NProbe controls at search time, how many clusters are considered to search for NN in FAISS.

**Discussion**
@nan-wang 
I added the parameters in FaissIndexer constructor, but since it is a parameter that is only used at search time, seems not to be the place, because the user then would need to fix this when indexing and not being able to change it when it is serialized.

If I add nprobe to post_init, can I pass it as a constructor when loading FaissIndexer from disk?

### DIFF
--- a/jina/executors/indexers/vector/faiss.py
+++ b/jina/executors/indexers/vector/faiss.py
@@ -18,14 +18,16 @@ class FaissIndexer(BaseNumpyIndexer):
         Faiss package dependency is only required at the query time.
     """
 
-    def __init__(self, index_key: str, train_filepath: str = None, nprobe=1, *args, **kwargs):
+    def __init__(self, index_key: str, train_filepath: str = None,
+                 distance: str = 'l2', nprobe: int = 1, *args, **kwargs):
         """
         Initialize an Faiss Indexer
 
         :param index_key: index type supported by ``faiss.index_factory``
         :param train_filepath: the training data file path, e.g ``faiss.tgz`` or `faiss.npy`. The data file is expected
             to be either `.npy` file from `numpy.save()` or a `.tgz` file from `NumpyIndexer`.
-
+        :param distance: 'l2' or 'inner_product' accepted. Determines which distances to optimize by FAISS
+        :param nprobe: Number of clusters to consider at search time.
 
         .. highlight:: python
         .. code-block:: python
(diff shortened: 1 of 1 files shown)


jina/executors/indexers/vector/faiss.py
```diff
@@ -18,14 +18,15 @@ class FaissIndexer(BaseNumpyIndexer):
         Faiss package dependency is only required at the query time.
     """
 
-    def __init__(self, index_key: str, train_filepath: str = None, *args, **kwargs):
+    def __init__(self, index_key: str, train_filepath: str = None, distance = 'l2', nprobe = 1, *args, **kwargs):
```
hanxiao, line 21:
```suggestion
    def __init__(self, index_key: str, train_filepath: str = None, distance: str = 'l2', nprobe: int = 1, *args, **kwargs):
```

jina/executors/indexers/vector/faiss.py
```diff
@@ -48,19 +49,27 @@ def __init__(self, index_key: str, train_filepath: str = None, *args, **kwargs):
...
     def build_advanced_index(self, vecs: 'np.ndarray'):
         """Load all vectors (in numpy ndarray) into Faiss indexers """
-
         import faiss
-        self._index = faiss.index_factory(self.num_dim, self.index_key)
+        metric = faiss.METRIC_L2
+        if self.distance == 'inner_product':
+            metric = faiss.METRIC_INNER_PRODUCT
+        if self.distance not in ['inner_product', 'l2']:
+            self.logger.warning("Invalid distance metric for Faiss index construction. Defaulting to l2 distance")
```
hanxiao, line 62:
```suggestion
        if self.distance not in {'inner_product', 'l2'}:
            self.logger.warning('Invalid distance metric for Faiss index construction. Defaulting to l2 distance')
```

hanxiao:
You mean user may play around with `nprobe` in query time? In general, all `Executor` supports parameter-rewrite by changing `meta.name` in the YAML file, so that it does not deserialize (unpickle) the indexing-time object when loading this `Executor`. It is pretty hidden feature, and we may refactor it (say adding `override_parameters`) in the future if we find this pattern is useful.



> If I add nprobe to post_init, can I pass it as a constructor when loading FaissIndexer from disk?

JoanFM:
> You mean user may play around with `nprobe` in query time? In general, all `Executor` supports parameter-rewrite by changing `meta.name` in the YAML file, so that it does not deserialize (unpickle) the indexing-time object when loading this `Executor` (can't find => can't load). It is a pretty hidden feature, and we may refactor it (say adding `override_parameters`) in the future if we find this pattern is useful.
> 
> > If I add nprobe to post_init, can I pass it as a constructor when loading FaissIndexer from disk?

Then, having it in the constructor is fine?

hanxiao:
Yes, adding it to `__init___` is totally fine. 

It's just modifying this parameter at the query time isn't in any docs and it's quite implicit. But that's another PR.

Issue #619, opened 2020-07-05
use hello-world as a performance tracker


hanxiao:
**Describe the feature**

Tracking jina overall perf (index & search) across all releases. Use the time elapsed of hello-world as a metric.

**Your proposal**

- on every `devel` or new release, use `docker run jinaai/jina:devel hello-world` write down the elapsed time;
- keep all time in CSV/JSON and maintain them a separate repo
- plot it, x-axis version, y-axis time

---

**Environment**

**Screenshots**

hanxiao:
```bash
for j in `seq 3`; do docker run --rm -v /tmp:/tmp jinaai/jina:0.3.5 hello-world > /tmp/hw.log && perl -lne 'print "$1\t$2" if(/done in ⏱ (.*)s 🐎 (.*)\/s/)' /tmp/hw.log; done
```

hanxiao:
locally, `0.3.5`
| | time(s) | speed (docs/s) |
|---|---|---|
| download |4.3	| 866.2|
|index |31.4	| 1924.2|
|query|5.6|	28.4|

`0.3.6`
| | time(s) | speed (docs/s) |
|---|---|---|
| download |4.3|	881.2|
|index|28.0|	2160.1|
|query|3.9|	40.9|

so #618 worth it! 🥳

PR #639, opened 2020-07-09
(WIP) ci(perf): add perf tracker script to publish perf stats

JoanFM:
**Proposal Proof of Concept for #619** 

@hanxiao @nan-wang would a skeleton like this make sense to have the idea under #619 implemented?

I was thinking about having a repo perf-tracker.git or whatever name that would get pushes to master from JINA BOT and whenever it does, it recomputes a graph with the performance history.

### DIFF
--- a/perf_tracker.sh
+++ b/perf_tracker.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -ex
+
+function run_perf_tracker {
+  for j in `seq 3`;
+  do
+    docker run --rm -v /tmp:/tmp jinaai/jina:devel hello-world > /tmp/hw.log;
+    perl -lne 'print "$1\t$2" if(/done in ⏱ (.*)s 🐎 (.*)\/s/)' /tmp/hw.log
+  done
+}
+
+function git_push {
+  git config --local user.email "dev-bot@jina.ai"
+  git config --local user.name "Jina Dev Bot"
+  git add tracking_history.txt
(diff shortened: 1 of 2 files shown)


perf_tracker.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -ex
+
+function run_perf_tracker {
+  for j in `seq 3`;
```
JoanFM, line 6: it should do some work to set it in csv format.

JoanFM:
Yes, I guess it is possible, I guess you @hanxiao have permissions to allow another nightly build just for that purpose? Or maybe we can do all that in a new repo calles perf-tracker where every night we checkout jina, run the script, recompute the plot and pushes a new commit to itself?

JoanFM:
I am concerned that we have no control on the machines and under which conditions they would run. So we would have some missleading information that often can be worse than no information at all

nan-wang:
> I am concerned that we have no control on the machines and under which conditions they would run. So we would have some missleading information that often can be worse than no information at all

This is a good point. The machine provided by github might change from time to time. It is better to run it on the AWS machine.

JoanFM:
> > I am concerned that we have no control on the machines and under which conditions they would run. So we would have some missleading information that often can be worse than no information at all
> 
> This is a good point. The machine provided by github might change from time to time. It is better to run it on the AWS machine.

I close this PR because of this!

PR #641, opened 2020-07-10
perf(indexer): change sort method to argpartition, not sort all the array

JoanFM:
**Changes introduced**

When sorting chunks and obtaining top_k values, argsort sorts entire array. with argpartition, it guarantees that the **kth** element is in the sorted position, and the indexes before it are smaller and viceversa. Besides that no other ordering is guaranteed, but since it is later the job of **Rankers** to sort the documents at the end, it is not necessary.

### DIFF
--- a/jina/executors/indexers/vector/numpy.py
+++ b/jina/executors/indexers/vector/numpy.py
@@ -53,7 +53,7 @@ class NaiveIndexer(BaseNumpyIndexer):
         elif self.metric == 'cosine':
             dist = _cosine(keys, self.query_handler)
 
-        idx = dist.argsort(axis=1)[:, :top_k]
+        idx = np.argpartition(dist, kth=top_k, axis=1)[:, :top_k]
         dist = np.take_along_axis(dist, idx, axis=1)
         return self.int2ext_key[idx], dist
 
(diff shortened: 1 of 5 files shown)


hanxiao:
aha!

jina/executors/indexers/vector/numpy.py
```diff
@@ -53,7 +53,7 @@ def query(self, keys: np.ndarray, top_k: int, *args, **kwargs) -> Tuple['np.ndar
         elif self.metric == 'cosine':
             dist = _cosine(keys, self.query_handler)
 
-        idx = dist.argsort(axis=1)[:, :top_k]
+        idx = np.argpartition(dist, kth=top_k, axis=1)[:, :top_k]
```
hanxiao, line 56: so i'm guessing this is faster?

JoanFM: it should, but I have not yet empirically tested

nan-wang: image
Reference: [https://stackoverflow.com/questions/10337533/a-fast-way-to-find-the-largest-n-elements-in-an-numpy-array](https://stackoverflow.com/questions/10337533/a-fast-way-to-find-the-largest-n-elements-in-an-numpy-array)

PR #644, opened 2020-07-12
fix(ci): fix the random failing ci process

### LINKED ISSUE #607: CI pipeline timeouts randomly
**Describe your problem**

CI pipeline takes 19 minutes and times out randomly (without error on PR). This is not critical, but might get worse if we add more test cases.

Issue is already known. Might be due to resources (such as sockets, processes and threads) are not released properly during the individual tests.

One solution could be to run tests in individual processes (see e.g here: https://pypi.org/project/pytest-xdist/ or https://pypi.org/project/pytest-forked/ or some other solution such as https://pypi.org/project/nose/ which also supports that AFAIK).

**What is your guess?**

---

**Environment**

**Screenshots**

nan-wang:
resolves #607 

- fixed the brokend `extract_chunks` in the `Drivers`
- fixed the side-effects of switching unittests to discovery mode
- fixed the misusage of import in the unittests on customtorch and customkeras
- disabled the unittests on `test_gateway.py` and `test_quant.py`, which broke the CICD for some reason and we will dig into details in another PR
- disabled `test_container_volume()` and `test_container_volume()`, which occasionally leads to the issue of conflict zmq address.
- switched from `unittest` to `pytest`
- added `pytest-xdist` for running unittests in parallel
- added `pytest-timeout` for setting timeout for each test


jina/hubapi/docker.py
```diff
@@ -91,8 +91,9 @@ def pull(self):
             image = self._client.images.pull(self.args.name)
         if isinstance(image, list):
             image = image[0]
+        image_tag = image.tags[0] if len(image.tags) > 0 else ""
```
hanxiao, line 94: please do not use `len()` to check if list empty

```suggestion
        image_tag = image.tags[0] if image.tags else ''
```

tests/test_container.py
```diff
@@ -133,7 +134,9 @@ def test_flow_topo_replicas(self):
             f.dry_run()
             f.index(input_fn=random_docs(1000))
 
+    @pytest.mark.skip()
```
hanxiao, line 137: can you mark the reason of the skip for future reference? otherwise it's easy to become a tech debt and no one will take care of it anymore.

```suggestion
    @pytest.mark.skip()
```

JoanFM:
Can't zmqaddress be assigned randomly from the outside so that they hardly collide?

jina/hubapi/docker.py
nan-wang: fixed

tests/test_container.py
nan-wang: fixed.

nan-wang: @hanxiao I've created two more issues on the skipped tests. I suggest to merge this PR and make the CICD working for PRs and working on the remaining issues in another PR.

hanxiao:
seems that now we only have 1 test failed?

JoanFM:
#686

Issue #643, opened 2020-07-11
A bug about the `device` property in the class `BaseFrameworkExecutor`


Wei-weiWang:
I have found a bug in your `BaseFrameworkExecutor` class about the `device` property. Here is the link ([https://github.com/jina-ai/jina/blob/master/jina/executors/frameworks.py#L10](link)).

The attribute `_device` is assigned in the `post_init` method and used in the `device` property. In the `ImagePaddlehubEncoder` class, however, the `device` property is used without calling `post_init` in the first hand. Therefore, an exception will be raised. I suggest in the `device` property, the existence if the `_device` attribute shuold be checked so as to improve the robustness of the codes. Attached is the suggested revision.

```
class BaseFrameworkExecutor(BaseExecutor):
    """
    :class:`BaseFrameworkExecutor` is the base class for the executors using other frameworks internally, including
        `tensorflow`, `pytorch`, `onnx`, `faiss` and `paddlepaddle`.
    """
    def __init__(self, model_name: str = None, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.model_name = model_name
    
    def post_init(self):
        super().post_init()
        self._device = None
        if self.on_gpu:
            try:
                cuda_version = subprocess.check_output(['nvcc', '--version']).decode()
                self.logger.success(f'CUDA compiler version: {cuda_version}')
            except OSError:
                self.logger.warning('on_gpu=True, but you dont have CUDA compatible GPU, i will reset on_gpu=False ')
                self.on_gpu = False

    @property
    def device(self):
        """Set the device on which the executor will be running.
        ..notes:
            In the case of using GPUs, we only use the first gpu from the visible gpus. To specify which gpu to use,
            please use the environment variable `CUDA_VISIBLE_DEVICES`.
        """
        if hasattr(self, '_device') and self._device is not None:
            return self._device
        else:
            try:
                self._device = self.get_device()
                return self._device
            except Exception:
                self.logger.error(f'error when setting devices "on_gpu={self.on_gpu}"')
                raise

    @abstractmethod
    def get_device(self):
        pass

    def to_device(self, *args, **kwargs):
        """Put the model on specified device (``on_gpu``) and returns the device context"""
        raise NotImplementedError
```

nan-wang:
hi, @Wei-weiWang Would you please provide more information about your jina version? We've recently made some changes recently at [https://github.com/jina-ai/jina/pull/636](https://github.com/jina-ai/jina/pull/636).

[...]

PR #645, opened 2020-07-12
feat(executors): Applies safe access to `_device` and checks if executor should work `on_gpu` in `_posti_init_wrapper_`

JoanFM:
**Changes introduces**

This PR addresses issues raised in #643.

Applies safe access to `_device` and checks if executor should work `on_gpu` in `_posti_init_wrapper_`


hanxiao:
As `on_gpu` is universal for all executors, moving the cuda check logic to L148 is also a good idea.

https://github.com/jina-ai/jina/blob/cab33cb18bec9962ea4b65b5ed15736cd08b13bc/jina/executors/__init__.py#L135-L150

JoanFM:
> Well done! Would you please take a look at the `BaseCVPaddlehubEncoder` as well? It overwrites the `post_init`, but there is `super().post_init()` is missing. This is related to [issue-643](https://github.com/jina-ai/jina/issues/643) With current setup, every class from `BaseFrameworkExecutor` should call `super().post_init()`, which might be error-prone.

I agree that they should all call **super().post_init()** but with current changes I think we should be safe in relation to #643 since **_device** will be accessed safely. Should we still add **super().post_init()** for **BaseFrameworkExecutor** children? For me it would make sense.

nan-wang:
Aha~ I see. Then I'm work with current PR. ;)

> > Well done! Would you please take a look at the `BaseCVPaddlehubEncoder` as well? It overwrites the `post_init`, but there is `super().post_init()` is missing. This is related to [issue-643](https://github.com/jina-ai/jina/issues/643) With current setup, every class from `BaseFrameworkExecutor` should call `super().post_init()`, which might be error-prone.
> 
> I agree that they should all call **super().post_init()** but with current changes I think we should be safe in relation to #643 since **_device** will be accessed safely. Should we still add **super().post_init()** for **BaseFrameworkExecutor** children? For me it would make sense.

jina/executors/__init__.py
```diff
@@ -147,6 +157,7 @@ def _post_init_wrapper(self, _metas: Dict = None, _requests: Dict = None, fill_i
 
             _before = set(list(vars(self).keys()))
             self.post_init()
+            self._check_on_gpu()
```
hanxiao, line 160: I recommend in the following order as some user may call gpu library in `post_init` already

```suggestion
            self._check_on_gpu()
            self.post_init()
```

JoanFM:
> forget `import subprocess`

Done!

PR #647, opened 2020-07-12
feat(driver): add queryset driver

hanxiao:
seems that there is no way to **avoid** designing our own LINQ/filter/query language


jina/drivers/queryset/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    A QuerySet represents a collection of objects from your database.
+    It can have zero, one or many filters. Filters narrow down the query results
+    based on the given parameters.
+
+    In SQL terms, a QuerySet equates to a SELECT statement,
+    and a filter is a limiting clause such as WHERE or LIMIT.
+    """
+
+    def __init__(self, queryset_raw: List[str] = None, combine_policy: str = 'append', *args, **kwargs):
```
JoanFM, line 16: Should not queryset_raw be of type Union[str, List[str]] if we are considering both cases in the body?

jina/drivers/queryset/lookup.py
```diff
@@ -0,0 +1,312 @@
...
+"""
+
+import re
+
+from .dunderkey import dunder_get, dunder_partition, undunder_keys, dunder_truncate
+from .helper import *
+
+
+class QuerySet:
+    """Provides an interface to filter data and select specific fields
```
JoanFM, line 37: Should we have a file per class?

Issue #648, opened 2020-07-12
feat: proto Document as recursive structure?


hanxiao:
**Describe the feature**

I realized that there is some level of consistency between `Document` and `Chunk` in current protobuf definition. I think it's possible to unify these two interfaces and make `Document` recursive.

Example,

![image](https://user-images.githubusercontent.com/2041322/87254779-6278cd80-c485-11ea-90cc-f747f0fb9f28.png)

Doing so will enable chunking & embedding at an arbitrary level, and may improve Jina's generality. 

Some refactoring may be required at Driver level?

**Your proposal**

Something like,

```protobuf
message Document {
    // ...
    repeated Document chunks = 4;
}
``` 

---

**Environment**

**Screenshots**

nan-wang:
what about recursive chunks?

PR #652, opened 2020-07-13
feat(proto!): make document a recursive structure

hanxiao:
**Estimated Done 25th July** 
This will be part of v0.5.0

## Current test progress
- [x] hello-world can be run without any problem
- [ ] unit test is missing for all new QL drivers, will do in another PR

## Intro
This will be a super big PR with breaking changes. Simply put, it removes the protobuf spec of "Chunk". Now, a "Chunk" is defined as a "Document" that belongs to another "Document".  Hence, "Document" is now a recursive structure. One can visit chunks via `doc.chunks` and get children `List[Document]`

This is an upgrade over the previous "Document/Chunk" concept. It makes the whole thing more general, enables advance segmenting and matching (#648)

## Affects

- (Thanks to Jina's design btw) this should mostly affect drivers, not executors. 
- drivers that previously work with doc/chunk will be refactored by recursively working on the doc structure.
- all previous `chunk`-related drivers/executors are removed because the interface is now unified.
- Executors do not need to know protobuf exists. They are "document/chunk-agnostic". All structure abstraction is taken care by Drivers.

## Breakdown

- [x] Change protobuf, delete the definition of `Chunk`.
- [x] Add a `BaseRecursiveDriver` that works on single doc recursively in preorder and postoder.
- [ ] Refactoring drivers using the newly added `BaseRecursiveDriver`
   - [x] craft.py
   - [x] convert.py
   - [x] prune.py
   - [x] control.py: no need to modify
   - [x] encode.py
   - [x] index.py
   - [x] score.py
   - [x] search.py
   - [x] reduce.py
- [x] all YAML resources
- [ ] all unit tests (will cover in other PRs)
- [x] all examples, freeze the jina version this time! 

## Migration guide (draft)

- `level` is not required in all drivers anymore.

- `PruneDriver` is now `ExcludeQL`, no `level` is required, `excludes` -> `fields`, e.g.
```yaml
    - !ExcludeQL
      with:
        fields:
          - embedding
          - buffer
          - blob
          - text
```

- `DocPruneDriver` is removed, one can now use 

```yaml
    - !ExcludeQL
      with:
        fields:
          - buffer
          - chunks
```

- `ChunkPruneDriver` is removed,  one can now use the following, as one can see it is basically the same as  `DocPruneDriver` , just with different fields.

```yaml
    - !ExcludeQL
      with:
        fields:
          - embedding
          - buffer
          - blob
          - text
```

- `BaseCraftDriver` is now `CraftDriver`. `DocCraftDriver` and `ChunkCraftDriver` are removed
- `doc_id` and `chunk_id` are unified as `id` now
- `UnarySegmentDriver` is removed. Now `EncodeDriver` works directly on doc-level embedding, `extract_chunks` is now refactored as `extract_docs` works on root-level extraction (previously it works on `doc.chunks` level extraction) 
- `BaseChunkCrafter` and `BaseDocCrafter` are removed.
- `TopkFilterDriver` is removed and now is implemented with Query Language:
```yaml
!SliceMatchesQL
with:
   start: 0
   end: k
```
- `ChunkPbIndex` and `DocPbIndexer` are now removed, please use `BinaryPbIndexer`
- `ReqPruneDriver` is removed now use
```yaml
    - !ExcludeReqQL
      with:
        fields:
          - request
```
- `_merge_topk_docs` -> `_merge_all`


jina/drivers/filter.py
```diff
@@ -14,12 +14,12 @@ def __call__(self, *args, **kwargs):
...
         for d in self.req.docs:
             del d.topk_results[self.req.top_k:]
-            topk_doc_id = {md.match_doc.doc_id for md in d.topk_results}
+            topk_doc_id = {md.match.id for md in d.topk_results}
             # keep only the chunks that hit the topk docs
             for c in d.chunks:
                 # delete in reverse so that idx won't be messed up
                 for idx, mc in reversed(list(enumerate(c.topk_results))):
-                    if mc.match_chunk.doc_id not in topk_doc_id:
+                    if mc.match.id not in topk_doc_id:
```
JoanFM, line 22: match.parent_id?

tests/test_quant.py
```diff
@@ -23,8 +24,8 @@ def random_docs():
             c = d.chunks.add()
             # force sending at non-quantization
             c.embedding.CopyFrom(array2pb(np.random.random([embed_dim]), quantize=None))
-            c.chunk_id = c_id
-            c.doc_id = j
+            c.id = c_id
+            c.id = j
```
JoanFM, line 28: c.parent_id I guess it should be?

jina/drivers/filter.py
hanxiao: yep, I haven't touch driver yet. it will take some time.

JoanFM:
@hanxiao better do not activate integration-test as required until the PR #651 is merged, and integration test become a reality

jina/clients/python/request.py
```diff
@@ -82,11 +77,9 @@ def _generate(data: Union[Iterator[bytes], Iterator['jina_pb2.Document'], Iterat
             if mime_type:
                 d.mime_type = mime_type
 
-            d.doc_id = first_doc_id if not random_doc_id else random.randint(0, ctypes.c_uint(-1).value)
+            d.id = next(doc_counter)
```
JoanFM, line 80: We will not support incrementing doc id assignment? I think it can be helpful for debugging purposes.

jina/drivers/__init__.py
```diff
@@ -159,7 +159,76 @@ def __getstate__(self):
         return d
 
 
-class BaseExecutableDriver(BaseDriver):
+class BaseRecursiveDriver(BaseDriver):
```
hanxiao, line 162: @JoanFM what do you think about this? Almost all drivers in this PR are refactored based on this.

jina/clients/python/request.py
hanxiao: okay I will refactor this part back using `SimpleCounter`

jina/drivers/__init__.py
JoanFM: The intention of this class, is to define at which depth level this drivers will apply the executor `method`?

hanxiao: Some of our drivers are working with `docs`. `BaseRecursiveDriver` provides helper function so that one can apply the function on all level of `docs`, without writing recursion wrapper.

Example
```python
class KVIndexDriver(BaseIndexDriver):
    """Serialize the documents/chunks in the request to key-value JSON pairs and write it using the executor
    """

    def apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs):
        self.exec_fn(docs)




class SegmentDriver(BaseCraftDriver):
    """Segment document into chunks using the executor
    """

    def __init__(self, level_names: List[str] = None, *args, **kwargs):
        if isinstance(level_names, list) and (self._depth_end - self._depth_start + 1) != len(self.level_names):
            self.level_names = level_names
        elif level_names is None:
            pass
        else:
            raise ValueError(f'bad level names: {level_names}, the length of it should match the recursive depth + 1')

        # for adding new chunks, preorder is safer
        self.recursion_order = 'pre'
        self.counter = RandomUintCounter()

    def apply(self, doc: 'jina_pb2.Document', *args, **kwargs):
        _args_dict = pb_obj2dict(doc, self.exec.required_keys)
        ret = self.exec_fn(**_args_dict)
        if ret:
            for r in ret:
                c = doc.chunks.add()
                self.set_doc_attr(c, r, {'length', 'id', 'parent_id', 'level_depth', 'mime_type'})
                c.length = len(ret)
                c.id = next(self.counter)
                c.parent_id = doc.id
                c.level_depth = doc.level_depth + 1
                c.mime_type = doc.mime_type
            doc.length = len(ret)
        else:
            self.logger.warning(f'doc {doc.id} at level {doc.level_depth} gives no chunk')

```

JoanFM: what is an example case where we might use different than (0, 0)?

JoanFM: In this **SegmentDriver**, if I pass `depth_start` = 0 and `depth_end`=2, it will segment any **document** in **chunks** and ithse **chunks** will be segmented again?

jina/drivers/__init__.py
```diff
@@ -159,7 +159,76 @@ def __getstate__(self):
...
+        """
+        super().__init__(*args, **kwargs)
+        self._depth_start = depth_range[0]
+        self._depth_end = depth_range[1]
+        if order in {'post', 'pre'}:
+            self.recursion_order = order
+        else:
+            raise AttributeError('can only accept oder={"pre", "post"}')
+
+    def apply(self, doc: 'jina_pb2.Document', *args, **kwargs):
```
JoanFM, line 180: I think we should add the @abc.abstractmethod decorator

JoanFM: I might be wrong, not every Driver implements apply, am I right?

jina/drivers/__init__.py
hanxiao: > In this SegmentDriver, if I pass depth_start = 0 and depth_end=2, it will segment any document in chunks and ithse chunks will be segmented again?

Yes, it recursively segmented all the way to `depth_end`, that's why `recursion_order` must be set to `pre`

jina/drivers/__init__.py
hanxiao: ControlDriver doesn't need, only the Drivers that operate on `docs` need to implement `apply` or `apply_all` or both.

JoanFM: ok!

jina/drivers/__init__.py
JoanFM: And do you have a useace for 'post_order` usage?

JoanFM: > @JoanFM what do you think about this? Almost all drivers in this PR are refactored based on this.

In general, the PR looks great, it offers great flexibility avoiding the differentiation of chunks and documents.

hanxiao: postoder traverse is often used in delete and reducing. I haven't touched the `Chunk2DocRankDriver` yet, but you can imagine, `Chunk2DocRankDriver` is basically aggregating the score at `k`th level to `k-1`th level, until we reached `level_depth=0`. In this case, `post` is required

JoanFM: Ok, I see, I have a much clearer view now! Thanks!

hanxiao: i'm excited about the potentials that this PR can enable

JoanFM:
We should keep in mind that this would involve the need of some big documentation changes?

hanxiao:
> how are we gonna keep track of which id it has downstream?

you mean `if d.chunks:`?

jina/executors/crafters/numeric/io.py
```diff
@@ -43,10 +42,10 @@ def craft(self, text: str, doc_id: int, *args, **kwargs) -> Dict:
             self.logger.error(
                 f'Data type mismatch. Cannot convert input to {self.as_type}.')
 
-        return dict(doc_id=doc_id, weight=1., blob=_array)
```
JoanFM, line 46: @hanxiao I meant here for instance. How wil the indexer for instance know to which doc_id, 'now id of level 0' a particular item belongs?

hanxiao: The principle, nested structure, `chunk_id` and `doc_id` are taken care by driver. Executor does not need to worry about these things.

- In general, `doc_id` should not be changed once assigned (often at client/request_gen).
- if you are doing crafting, you always work on the same doc, so there will be no change on `doc_id`. 
- when it comes to indexer, indexer stores & retrieves all documents in **flat**, not in a nested structure. well, at least I think it is not required. I will confirm it by refactoring the hello-world.

JoanFM: Understood! Just trying to follow the main ideas. It is generally good that classes just care about the info they need to really perform their task, so good!

JoanFM:
Will solve #669

PR #665, opened 2020-07-16
refactor(pea): remove unused daemon member

JoanFM:
**Changes introduced**
Remove **daemon** from members of Pea because it is never used.
Do some more cleaning of comments and minor things as I go through the code

### DIFF
--- a/jina/flow/__init__.py
+++ b/jina/flow/__init__.py
@@ -172,10 +172,10 @@ class Flow:
 
     @classmethod
     def load_config(cls: Type['Flow'], filename: Union[str, TextIO]) -> 'Flow':
-        """Build an executor from a YAML file.
+        """Build a Flow object from a YAML file.
 
         :param filename: the file path of the YAML file or a ``TextIO`` stream to be loaded from
-        :return: an executor object
+        :return: a Flow object
         """
         yaml.register_class(Flow)
         if not filename: raise FileNotFoundError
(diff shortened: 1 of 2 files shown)


hanxiao:
`self.daemon` is a flag of `multiprocessing` and `threading` library. It is used, implicitly, as `BasePea` is inherited from either `Process` or `Thread`

https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.Process.daemon

> When a process exits, it attempts to terminate all of its daemonic child processes.


If you remove that flag, I'm not sure the process who spawns the pea can later terminate it correctly.

JoanFM:
> `self.daemon` is a flag of `multiprocessing` and `threading` library. It is used, implicitly, as `BasePea` is inherited from either `Process` or `Thread`
> 
> https://docs.python.org/3.7/library/multiprocessing.html#multiprocessing.Process.daemon
> 
> > When a process exits, it attempts to terminate all of its daemonic child processes.
> 
> If you remove that flag, I'm not sure the process who spawns the pea can later terminate it correctly.

Where is this inheritance set?

JoanFM:
> the magic happens in
> 
> ```python
> class BasePea(metaclass=PeaMeta):
> ```
> 
> Inside `PeaMeta.__call__`, it reads `args.runtime` and choose the inheritance on the fly

Can we make it more explicit when we pass the **daemon** argument to the Process or Thread?

hanxiao:
Is there a reason `daemon=False`?

JoanFM:
> Is there a reason `daemon=False`?

No, that is exactly what I mean, that having a member that is always the same seems weird, but anyway no big reason to change what is working

PR #667, opened 2020-07-17
refactor(flow): refactor Flow build

JoanFM:
**Changes introduced**

- Have **Head** and **Tail** peas separated from regular **Peas** so that logic is encapsulated, easier to reason about and to test them. Then I realised there is not much logical diference, but still feels cool to have different classes if they are different, since they may evolve in the future

- Change how **Flow** is built. While I was reading Flow code, it was hard for me to follow the code that builds it and the connections between **pods** and its **optimizations**. I think we are building and optimizing in the same pass. I think building first the Flow connecting heads and tails with the proper logic and then, in a second pass, do some **optimization** when replicas exist. I tried to simplify the process.


JoanFM:
Another good point is that this **Flow** building is **deterministic**, while the current one is not, because when running `for k in _pod_edges` the order of looping is not guaranteed. In this case, we follow the construction from the origin node of the graph (the gateway) so it can have benefits in the future to **debug** and **investigate** issues.

jina/flow/__init__.py
```diff
@@ -62,6 +62,84 @@ def arg_wrapper(self, *args, **kwargs):
...
 
 
+def _traverse_graph(op_flow: 'Flow', outgoing_map: Dict[str, List[str]], func: Callable[['Flow', str, str], None]) -> 'Flow':
+    _outgoing_idx = dict.fromkeys(outgoing_map.keys(), 0)
+    stack = deque()
+    stack.append('gateway')
+    op_flow.logger.info('Traversing dependency graph:')
+    while stack:
+        start_node_name = stack.pop()
+        print(f'start_node {start_node_name}')
```
hanxiao, line 72:
```suggestion
        # print(f'start_node {start_node_name}')
```

jina/flow/__init__.py
```diff
@@ -62,6 +62,84 @@ def arg_wrapper(self, *args, **kwargs):
...
+        end_node_idx = _outgoing_idx[start_node_name]
+        if end_node_idx < len(outgoing_map[start_node_name]):
+            # else, you are back to the gateway
+            end_node_name = outgoing_map[start_node_name][end_node_idx]
+            print(f'end_node {end_node_name}')
```
hanxiao, line 77:
```suggestion
            # print(f'end_node {end_node_name}')
```

JoanFM: Removed completely!

jina/flow/__init__.py
JoanFM: Removed completely!

hanxiao:
I don't see there is a particular need to distinguish `HeadPea` and `TailPea`. As I don't want to send a message that `HeadPea` and `TailPea` are different hence developers have to treat them with care: especially when one Pod's `TailPea` could be in fact the `HeadPea` of another Pod.

JoanFM:
> I don't see there is a particular need to distinguish `HeadPea` and `TailPea`. As I don't want to send a message that `HeadPea` and `TailPea` are different hence developers have to treat them with care: especially when one Pod's `TailPea` could be in fact the `HeadPea` of another Pod.

There is no real need, I just started doing some decoupling thinking it had some more special behavior, I can remove if wished.

PR #675, opened 2020-07-19
Help proto

JoanFM:
Help to solve conflicts.

I took your branch, pulled from master and removed one conflicting test not necessary, but it may be better for you to do it direcly @hanxiao


hanxiao:
ok, let me do this directly on my PR

PR #679, opened 2020-07-20
refactor(milvus): check connectivity status before connecting

### LINKED ISSUE #668: Improve MilvusIndexer robustness
**Describe your problem**
At the time of insertion of **MilvusIndexer** (https://github.com/jina-ai/jina/pull/651), it was not so clear how to check some information from Milvus external server using their python client, therefore an issue was opened to them. (*milvus-io/pymilvus#236**)

We should keep an eye on this issue and react to updates.

Once it is clear how to have the information requested we should make MilvusIndexer more robust.

- Check connectivity status when necessary

- Check index for a collection exist

JoanFM:
**Changes introduced**
Closes #668 
Make **MilvusIndexer** more robust.
- Check **connection** status so to avoid calling connect multiple times.
- Hesitating if we should check **index_status** before building index, expecting further input from https://github.com/milvus-io/pymilvus/issues/236

### DIFF
--- a/jina/executors/indexers/vector/milvusdb/milvusdbhandler.py
+++ b/jina/executors/indexers/vector/milvusdb/milvusdbhandler.py
@@ -76,8 +76,9 @@ class MilvusDBHandler:
         self.close()
 
     def connect(self):
-        self.logger.info(f'Setting connection to Milvus Server at {self.host}:{self.port}')
-        self.milvus_client = Milvus(self.host, self.port)
+        if self.milvus_client is None or not self.milvus_client.server_status()[0].OK():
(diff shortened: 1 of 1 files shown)


hanxiao:
ready for review?

JoanFM:
waiting for extra info from milvus issue

JoanFM:
@hanxiao, now it is ready, milvus-io/pymilvus#236. Milvus team confirmed that repeating a build_index operation has no effect, so it would be as costly as checking that an index exists. So even if it does not seem lazy from the indexer point of view. No operation is repeated and the code looks explicit

PR #681, opened 2020-07-21
feat(docs-search-box): focus on the search input when "/" is pressed

emmaadesile:
**Describe the feature**
Users will have a better experience using the documentation if they can easily access the search input without using their mouse.

**Your proposal**
I have implemented a feature that adds focus to the search input when a user presses the "/" key.

### DIFF
--- a/docs/_static/main.css
+++ b/docs/_static/main.css
@@ -366,8 +366,9 @@ a:hover {
 }
 
 .doc-searchbox input[type=text] {
+    width: 100%;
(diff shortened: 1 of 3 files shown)


JoanFM:
Hey @emmaadesile ,

Thank you for your contribution.

Do not worry about this code coverage report, apparently codecov are having some issues, and we may have not gotten the configuration right. Your change surely does not impact the code coverage.

PR #684, opened 2020-07-21
Unify "yaml_path" and "image" into one "uses"

### LINKED ISSUE #583: refactor yaml_path and image to `uses`
**Describe the feature**

the interpreter behind `yaml_path` is far too powerful, the name does not accurately represent what it actually does.

meanwhile seems no point to use extra `image` parameter

**Your proposal**

this is a breaking change in Flow API, 

if change decision is made, then change early

---

**Environment**

**Screenshots**

JoanFM:
**Changes introduced**

Resolves #583.

Unifies `yaml_path` and `image` under `uses` keyword. `reducing_yaml_path` turns into `uses_reducing` and inside a docker image, when using a `yaml_path` it becomes `uses_internal`.


JoanFM:
We need to find a proper way to distinct when `uses` is refering to a docker image.

jina/helper.py
```diff
@@ -503,6 +503,11 @@ def valid_yaml_path(path: str, to_stream: bool = False):
                                 ' or a valid file path, or a supported class name.' % path)
 
 
+def valid_docker_image(image: str):
+    # TODO: Think a better way, it is a little dangerous
```
JoanFM, line 507: @hanxiao , we need a way to better identify wether `uses` refers to a **docker image**

hanxiao: can we merge this function with `valid_yaml_path`?

JoanFM: it is not so automatic, because `valid_yaml_path`  would be called with 'uses_internal` I think

JoanFM: Actually valid_docker_image returns a boolean while valid_yaml_path actually returns a processed object; so different objects.

jina/helper.py
```diff
@@ -503,6 +503,11 @@ def valid_yaml_path(path: str, to_stream: bool = False):
...
+    return image.startswith('jinaai/') or image.startswith('jina/')
+
```
YueLiu-jina, line 509:
```
def valid_docker_image(repo_name, version="latest"):
    try:
        import docker
        import requests
    except:
        return False
    client = docker.from_env()
    try:
        client.images.get(repo_name)
        return True
    except:
        pass
    url = "https://index.docker.io/v1/repositories/" + repo_name + "/tags/" + version
    try:
        res = requests.get(url)
        return res.status_code != 404
    except:
        return False
```

JoanFM: hey @YueLiu-jina thank you very much it seems like an interesting approach, can you refactor in a single try ... except catching different kinds of exceptions?

jina/peapods/__init__.py
```diff
@@ -32,7 +33,7 @@ def Pea(args: 'argparse.Namespace' = None, allow_remote: bool = True, **kwargs):
     if args.host != __default_host__:
         from .remote import RemotePea
         return RemotePea(args)
-    elif args.image:
+    elif args.uses and valid_docker_image(args.uses):
         from .container import ContainerPea
         return ContainerPea(args)
```
YueLiu-jina, line 38:
```
elif args.uses and not args.uses.endwith(('.yml', 'yaml')):
        if valid_docker_image(args.uses):
                from .container import ContainerPea
                return ContainerPea(args)
        else:
                print(f'No docker image named {args.uses}.')
```

JoanFM: Valid yaml paths are not the only alternative to docker images for `uses`. They can also be some keywords like `_forwards`, `_route`, `_merge` or some valid python class names, like `BaseExecutor`.

YueLiu-jina: checking the validity of image online may take some time, how can we quickly tell if it's not a yaml?

jina/helper.py
JoanFM: the only problem is that the signature does not match how docker image name will be inserted to the Flow. It would come as a single string.

jina/peapods/__init__.py
JoanFM: a **try catch** over **valid_yaml_path**, but I find a little ugly! I will try to split **valid_yaml_path** and **get_valid_yaml_path**. **valid_yaml_path** would return a **boolean**, while **get_valid_yaml_path** would return a **readable** object.

YueLiu-jina: that will be great

jina/helper.py
JoanFM: @YueLiu-jina , we also can apply negative logic, and that would save us from early checking it is a valid image. I am not a fan of applying it, but in this case it can save us from trouble. Instead of checking it is a **valid_docker_image**, we can check it is not a **valid_yaml_path**.  And since `uses` should be one OR another it should work, I may also change the naming of `valid_yaml_path` to something that shows that is not only `yaml_paths`

YueLiu-jina: @JoanFM I agree with this logic. But I still feel a *valid_image_path* function is needed. If we have this function, when, say, one intends to use yaml path but make a typo, instead of treating it as a docker image and waiting docker to react, we can raise a specific error and maybe terminate the flow, and this will save lots of time. Also for the future, maybe some other formats will be used, and then this one or another won't work.

JoanFM: you have a good point there.

jina/helper.py
YueLiu-jina: code edited @JoanFM

JoanFM: good change, I suggest, remove the recursion and just reassign version and repo_name and go hit the try-catch. recursion is  sometimes a little dangerous.

JoanFM: @YueLiu-jina , maybe u can do a Pull Request to my branch with your proposed ideas, I think they are good.

JoanFM:
We can merge this then and @YueLiu-jina will create a PR to improve how we handle valid docker image names and yaml paths

JoanFM:
> We can merge this then and @YueLiu-jina will create a PR to improve how we handle valid docker image names and yaml paths

As long as examples are frozen with older jina versions

JoanFM:
@alexcg1 this can be a documentation breaking change. I have tried to change references in documentation but I am not sure I got them all right

PR #685, opened 2020-07-21
fix(flow): fix access to pod edges, consider IGNORE_GATEWAY optimization should not try to optimize gateway connections

JoanFM:
**Changes introduced**
Running in local `pytest tests/unit/test_index.py -k 'test_two_client_route_replicas'`, found an issue in the optimization code.
It solves what was not properly covered by #667

### DIFF
--- a/jina/flow/__init__.py
+++ b/jina/flow/__init__.py
@@ -62,7 +62,8 @@ def build_required(required_level: 'FlowBuildLevel'):
     return __build_level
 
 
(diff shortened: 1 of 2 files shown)


hanxiao:
hmm, i don't see why `scheduling` is required in Gateway, any example?

JoanFM:
> hmm, i don't see why `scheduling` is required in Gateway, any example?

Thanks @hanxiao, I saw what was wrong.

Issue #662, opened 2020-07-15
How to use another transformers model


iedmrc:
**Describe the feature**
Hi, According to this [docs](https://docs.jina.ai/api/jina.executors.encoders.nlp.transformer.html) encoders are limited to: 
` ‘bert-base-uncased’, ‘openai-gpt’, ‘gpt2’, ‘xlm-mlm-enfr-1024’, ‘distilbert-base-cased’, ‘roberta-base’, ‘xlm-roberta-base’, ‘flaubert-base-cased’, ‘camembert-base’, ‘ctrl’.`

How can I run another `bert/gpt2/roberta` etc.. based pretrained model? E.g. German or French or Turkish Bert & Distilbert models? I got an error if I just provide e.g. `dbmdz/distilbert-base-turkish-cased` as model_name.

nan-wang:
hi, @iedmrc Our contributors has made one example for another model from transformers, `hfl/chinese-roberta-wwm-ext`. Maybe it will be help to check it out. In general, it is quite easy to extend. If you have any trouble in extending, feel free to shoot questions.

[https://github.com/jina-ai/examples/blob/master/zh/webqa-search/pods/encoder/transformer_roberta.py](https://github.com/jina-ai/examples/blob/master/zh/webqa-search/pods/encoder/transformer_roberta.py)

moosefly123:
> hi, @iedmrc One of our contributors have made one example for another model from transformers, `hfl/chinese-roberta-wwm-ext`. Maybe it will be helpful to check it out. In general, it is quite easy to extend. If you meet any trouble, feel free to shoot questions.
> 
> https://github.com/jina-ai/examples/blob/master/zh/webqa-search/pods/encoder/transformer_roberta.py

hi~
I met the same trouble like iedmrc and I've checked out the example. 
In the example mentioned by nan-wang, the contributor uses a huggingface-style method to create a model by a shot name of "hfl/chinese-roberta-wwm-ext", which has been uploaded to the remote repository previously as a customized model. Here is my question. How can I use a local pretrained model which I just saved in local path and I can't upload it to some remote repository?
thanks.

policeme:
in your encoder yaml file, add `model_path` in  `with` like this.
```
!TransformerTorchEncoder
with:
    model_path: local_path
balabala
```
if u add like this, your model format must follow huggingface model format

iedmrc:
Hey, here is my `encode.yml`:

```
!TransformerDistilbertTurkishEncoder
[...]

Issue #553, opened 2020-06-21
🔖 Commit Message Guideline for the First Time Contributor


hanxiao:
First, thanks a lot for your initiative :heart:  The open-source community needs one like you with an open and inclusive mindset.

The first barrier to the first time contributor is writing a correct commit message. We realize many PRs have failed in our CI pipeline simply because of the bad commit messages.

**Importance of a correct commit message**

* We are using `commitlint` to check the style of your commit message as the first step of CI. If this fails, the whole pipeline will fail and skip all tests in the sequel. Very strict, yes, but necessary.
* We are generating [`CHANGELOG.md`](https://github.com/jina-ai/jina/blob/master/CHANGELOG.md) automatically on every new release. Correct commit messages help our users to understand which part is modified/improved.

**What to follow?**

* https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md#commit-style
* https://www.conventionalcommits.org/en/v1.0.0/

**Common mistakes**

* `doc(101): improved 101 document` - should be `docs(101)`
* `tests(flow): add unit test for flow exception` - should be `test(flow)`
* `DOC(101): Improved 101 Documentation` - all letters should be in lowercase
* `fix(pea): i fix this pea and this looks really awesome and everything should be working now` - too long
* `fix(pea):fix network receive of the pea` - missing space after `:`
* `hello: add hello-world` - type `hello` is not allowed

**Rule of thumb**
Keep it simple, short, don't use too many punctuations, e.g. `,.!?-_`. No need to be too specific.

**What if I commit a message in wrong style?**
You can use `git commit --amend` and then `git push force` to fix it. Note that you need to fix **all** commit messages in a PR, this is because when you merge you merge the whole commit history of your branch into master, not only the last commit.

Hope this help!

hanxiao:
hi @fhaase2 this issue is pinned on top of issue board so that everybody can see that, please do not remove it. 😃

tadejsv:
@hanxiao Please change the instruction from `git push force` to `git push --force`

PR #701, opened 2020-07-22
Support more pre-trained models for Transformer Encoder

### LINKED ISSUE #691: Support more pretrained models for Transformer Encoder
**Describe the feature**
 After this proposal users will be able to use all of the models from this list (https://huggingface.co/models) while using `BaseTransformerEncoder`. Not all of these models are intended for for information retrievial tasks, but some may be useful. Note that with the current implementation users are already having issues, see #662.

**Your proposal**
Use this [class](https://huggingface.co/transformers/model_doc/auto.html#automodelforpretraining) from transformers library.
The code is already implemented in my fork [here](https://github.com/festeh/jina/blob/feat-support-more-transformers/jina/executors/encoders/nlp/transformer.py), if you're interested, I can polish it a bit and make a PR.

Pros:
* Users will be able to load all available pre-trained models
* The number of lines of code is reduced, removed hardcoded imports

Cons:
* Hard-coding will now occur in other place. For each model one needs to decide how to get a sequence embedding from the embeddings of its tokens. I propose a simple heuristic: if a model is BERT-ish, take embedding of a first token, otherwise (XLNet, GPT-2 and other language models), take embedding of a last token. There's no guarantee that it would work for all models
* Did not test for all possible models, just considered existing corner cases (need to add a padding token for language model)

festeh:
Resolves #691

Main change:
* all models from https://huggingface.co/models are now available. It can be done via `AutoModelForPreTraining`

Other changes
* Fixed bug in `prune_mask`
* Added a heuristic for automatic creation sequence embedding from its tokens (if BERT-like, use first token, else last)
* Used `batch_encode_plus` to simplify batching
* `max_length` is set to `None` - so batches will be padded to largest sequence. I think user should determine this parameter based on the task
* Re-implemented saving logic. I got lost in fields like `raw_model_path`, `tmp_model_path`, so instead I just added `model_save_path` field. 
* Added unit tests that are using mock transformer model. The reason is that they can be run relatively (tokenizer is still downloaded) quickly as huge amount of time is saved for model initialization

Let's discuss if some of these changes are not desirable and try to come up with a better solution

### DIFF
--- a/jina/executors/encoders/helper.py
+++ b/jina/executors/encoders/helper.py
@@ -48,8 +48,8 @@ def prune_mask(mask, cls_pos='head'):
         mask_row[0, 0] = 1
         result = np.tile(mask_row, (mask.shape[0], 1))
     elif cls_pos == 'tail':
-        for num_tokens in np.sum(mask, axis=1).tolist():
-            result[:, num_tokens - 1] = 1
+        for row_index, num_tokens in enumerate(np.sum(mask, axis=1).tolist()):
+            result[row_index, num_tokens - 1] = 1
     else:
         raise NotImplementedError
     return result
--- a/jina/executors/encoders/nlp/transformer.py
+++ b/jina/executors/encoders/nlp/transformer.py
@@ -2,13 +2,32 @@ __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
 import os
-
+from typing import Optional
 import numpy as np
 
 from .. import BaseEncoder
 from ..frameworks import BaseTextTFEncoder, BaseTextTorchEncoder
 from ..helper import reduce_mean, reduce_max, reduce_min, reduce_cls
 from ...decorators import batching, as_ndarray
+from jina.logging.base import get_logger
+
+logger = get_logger("transformer_encoder")
+
+
+def auto_reduce(model_outputs, mask_2d, model_name):
+    """
+    Automatically creates a sentence embedding from its token embeddings.
+        * For BERT-like models (BERT, RoBERTa, DistillBERT, Electra ...) uses embedding of first token
+        * For XLM and XLNet models uses embedding of last token
+        * Assumes that other models are language-model like and uses embedding of last token
+    """
+    if "bert" in model_name or "electra" in model_name:
+        return reduce_cls(model_outputs, mask_2d)
+    if "xlnet" in model_name:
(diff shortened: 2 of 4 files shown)


jina/executors/encoders/helper.py
```diff
@@ -48,8 +48,8 @@ def prune_mask(mask, cls_pos='head'):
         mask_row[0, 0] = 1
         result = np.tile(mask_row, (mask.shape[0], 1))
     elif cls_pos == 'tail':
-        for num_tokens in np.sum(mask, axis=1).tolist():
-            result[:, num_tokens - 1] = 1
+        for row_index, num_tokens in enumerate(np.sum(mask, axis=1).tolist()):
```
JoanFM, line 51: nice catch!

JoanFM:
Hello @festeh,

The CI won't pass because you have some bad commit messages. please refer to:
- [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
- [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)
#553 :bookmark: Commit Message Guideline for the First Time Contributor
jina-ai/jina | Jun 21st | Added by GitHub
CONTRIBUTING.md
jina-ai/jina | Added by GitHub

festeh:
Yeah, will fix commit message soon. Meanwhile, I discovered a bug with Tensorflow model, fixing it now...

JoanFM:
The idea of using `AutoModel` and `AutoTokenizer` was also shared by #662. This PR may close this issue

JoanFM:
recheckcla

jina/executors/encoders/nlp/transformer.py
```diff
@@ -2,76 +2,92 @@
 __license__ = "Apache-2.0"
 
 import os
-
-import numpy as np
-
+from typing import Optional
```
JoanFM, line 5: numpy as np needed for type hint, lint is complaining about it

festeh: I think it's a bit hacky way, maybe explicitly use `noqa` statement if linter is an only issue? Or just remove np from type hint?

festeh: On the second thought, why not, let's return it back. Generally, there are a lot of place when `if False: import ...` are used. My comment was about this situation.

JoanFM:
LGTM! Just not sure if `auto` should have preference over `cls`. Also we can adopt more mocks to improve our testing!

PR #704, opened 2020-07-23
(W.I.P) (P.O.C): Add mode_id to support multimodal search + add logic to split messages in driver based on mode_id.

JoanFM:
**Disclaimer**

This **draft** is made to encourage **discussion** on wether this is the correct approach to support **multiple** modes in jina documents, and being able to direct output messages to different **Pods**.

This is far from being complete or ready. 

**Basics**
- Enrich **documents** with **mode_id** to refer to different document **modalities**. (Important to notice that different modalities do **NOT** imply different **mime_types**) It could refer to documents with the same type but coming from different distributions (e.g, imagine having **frontal** and **back** images)

- Right now a **Pea** processes **ONE** message at a time and manipulates it in a form that can be sent to downstream pods. This logic must change under this new requirements (Although we could implement the same functionality by just using a **QueryLanguageFilter** before entering the next **pod**. But this seems **suboptimal** at first sight). I propose a change where **Pea** returns a list of outputs messages (by default there will only be one).

- With the previous **Pea** Enrichment, create a new kind of Pea, that will be attached explicitly to the new driver **SplitRouteDriver**. (the **SplitterPea**)

- **SplitRouteDriver** is a **ControlReqDriver** inspired by **RouteDriver** that keeps a stack of the **DEALERs** that request messages for each mode_id. When processing a message, it groups **documents** by **mode_id** in different output messages and dispatches them. (It is far from ready and it must deal with challenges as to what if a mode_id has **DEALERs** waiting for it but another doesn't)

**Alternatives**
Alternatives to this implementation:
- Use of **QueryLanguage** filters before the pods that want to work on a single mode_id
- Use the **PUB-SUB** pattern, may not be as efficient as efficient in some cases but could be easier.

**To be seen**
- Implement the request call for an specific mode_id
- What should be the syntax to enable these search modes.
- How to connect these new **Peas** inside the **Flow**
- TEST, TEST, TEST
- I think I did not do properly the proto update


JoanFM:
Actually it would make more sense to have a Pea that let you concatenate 2 drivers-executors, the first would execute the actual task and the second would split the message.

In this framework having a Pub Sub pattern may make more sense and help avoiding sync issues

hanxiao:
Can you write a short "pseudo Flow code" to describe what do you want to achieve?

JoanFM:
This is more or less the logic I want to be able to achieve.

```yaml
!Flow
metas:
  prefetch: 10
pods:
  crafter:
    uses: yaml/craft.yml
  encoder_mode1:
    uses: yaml/encode_mode1.yml 
    mode_id: 1
  encoder_mode2:
    uses: yaml/encode_mode2.yml
    mode_id: 2
```

hanxiao:
lets move the discussion to Slack #contributors

PR #709, opened 2020-07-24
refactor(proto!): matches in recursive Document structure

hanxiao:
image

```
doc1 --matches-- doc3 (with score ref_id=1), doc4 (with score ref_id=1), doc7 (with score ref_id=1)
                    |
                    | --- matches -- doc6 (with score ref_id=3), doc9 (with score ref_id=3), doc10 (with score ref_id=3)
```

This design:

- improves the consistency of recursive structure introduced in #652
- allows advanced training/search paradigm working on a proximity graph/nearest-neighbor graph (which is quite common in ML). It opens a door to graph learning.

# Migration Guide:

- when looping the topk results, directly access the attribute `x` without `match.x`. i.e.
```python
for k in d.matches: 
  fn(k.match.id)
```
is now 
```python
for k in d.matches: 
  fn(k.id)
```


jina/proto/jina.proto
```diff
@@ -109,6 +101,8 @@ message Document {
 
     // the embedding array of this doc
     NdArray embedding = 19;
+
+    Relevance score = 20;
```
JoanFM, line 105: should `score` and `matches`  be `either-or`? I guess a Score would be assigned to a `retrieved` document, and the `matches` would be added to the `query document`?

tests/unit/drivers/test_querylang_drivers.py
```diff
@@ -0,0 +1,68 @@
...
+        yield d
+
+
+class DummySegmenter(BaseSegmenter):
+
+    def craft(self, text, *args, **kwargs):
+        return [{'text': 'adasd' * j} for j in range(10)]
+
+
+class MyTestCase(JinaTestCase):
```
JoanFM, line 28: Could we rename these `MyTestCase` to something more descriptive? Like QueryLangTestCase? I am trying to do it slowly when I touch some part related to them

jina/proto/jina.proto
hanxiao: setting `oneof` on `matches` and `score` will make the following pattern unimplementable.

```
doc1 --matches-- doc3 (with score ref_id=1), doc4 (with score ref_id=1), doc7 (with score ref_id=1)
                    |
                    | --- matches -- doc6 (with score ref_id=3), doc9 (with score ref_id=3), doc10 (with score ref_id=3)
```

JoanFM: Is it a pattern that can happen? or it is for the sake of full flexibility?

JoanFM: In theory, `doc1` should come from a `query set`, and `doc3, doc4 and doc7` from the `index set`. This bifurcation of `matches` to me would have a different `meaning`. How a `doc` from the `index set` has matches `with others`? Wouldn't that imply something like a kNN graph?

hanxiao: This design:
- improves the consistency of recursive structure introduced in #652 
- allows advanced training/search paradigm that requires a proximity graph/nearest-neighbor graph (which is quite common in ML). It opens the door to graph learning.

JoanFM: understood ;)

hanxiao: That being said, most of our examples can be implemented with just one level of recursion.

JoanFM: Yes, I see it more as a potential enabler for different features in the future.

JoanFM:
The access to results feel more natural now!

PR #712, opened 2020-07-24
feat(mode_id): add mode-id in proto + add query Lang driver to filter documents by mode_id if explicitly added as argument

JoanFM:
**Changes introduced**
After discussion regarding #704, we have decided to go for the simpler solution of prepending a **FilterQLDriver** to the drivers of the executors requesting a specific set of **mode_ids** to be processed. 

In addition, some restructuring of tests as a pre-work to properly add new tests has been done


hanxiao:
is `mode_id` a syntax sugar of multiple `mime_type`?

JoanFM:
Not directly; mode-id can refer to different modalities with the same mime type. It could mean data coming from different data distributions. Imagine one can leverage information from frontal and back images of objects; one may need to handle them differently. That is why I feel it is important to really have a difference betwreen mime type and mode id.

JoanFM:
I have a problem implementing the pattern in **test_flow_multimode.py**.

FilterQL has a limitation when trying to filter based on 'chunk level mode_id'. Imagine this case:

`doc0 { chunk 0 {mode_id: 'mode0'}, chunk1 {mode_id: 'mode1'}}, doc1 {chunk 0 {mode_id: 'mode0'}, chunk 1 {mode_id: 'mode1'}}
`
Ritght now if I want to extract every document from level 0 to level 1 with mode_id = mode0, I would expect it to extract doc0.chunk0 and doc1.chunk0. But currently since in level 0 it will not find any, it will totally delete the document from the docs and therefore remove its chunks.

JoanFM:
@hanxiao, since it is no clear at which **level** the **mode_ids** are looked for (which depth should be set for the driver), I take your example, and I decide **NOT** to add an explicit argument for peas. So if someone wants to **filter** on **mode_id** he/she must enforce it in the driver description. It may not be so user-friendly but I think that as you say, it can let us wait until we find common patterns.

tests/unit/drivers/test_reduce_all_driver.py
```diff
@@ -0,0 +1,63 @@
...
+                output.append([0.0, 0.0, 0.0])
+            elif "mode2" in r:
+                output.append([1.0, 1.0, 1.0])
+
+        return np.array(output)
+
+
+class ReduceAllDriverTestCase(JinaTestCase):
+    @pytest.mark.skip('Not working as expected')
+    def test_merge_chunks_with_different_mode_id(self):
```
JoanFM, line 38: @hanxiao, I was thinking about a test like this one. after reducing, I would pretend all the documents have all their chunks.

hanxiao: ok let me check

JoanFM:
@nan-wang, this PR might be interesting before tomorrow's meeting

jina/proto/jina.proto
```diff
@@ -104,6 +104,9 @@ message Document {
     NdArray embedding = 19;
 
     Relevance score = 20;
+
+    // mode_id, an identifier to the modality this document belongs to. In the scope of multi/cross modal search
+    string mode_id = 21;
```
JoanFM, line 109: For the sake of completion, should it be a repeated field? so that a document can have more than one mode and then it can be split into chunks of different modes?

JoanFM: In addition to this, in content should we consider another field option that is any repeated combination of content? or should we consider that if we use multimodality, one should pass content that is his/her responsibility to split (he/she should provide a crafter able to split into different modes if required)?

JoanFM: Any thoughts on this @hanxiao, @nan-wang

hanxiao: let me switch to this branch

tests/unit/drivers/test_reduce_all_driver.py
hanxiao: what does this do? this part feels strange, given your doc text, it basically encodes everything into `[0.0, 0.0, 0.0]`
```
    def encode(self, data: str, *args, **kwargs) -> 'np.ndarray':
        output = []
        for r in data:
            if "mode1" in r:
                output.append([0.0, 0.0, 0.0])
            elif "mode2" in r:
                output.append([1.0, 1.0, 1.0])

        return np.array(output)
```

hanxiao: ok first observation is that doc_pointer mixes docs from multiple `depth_level` together, which is always dangerous. Let me check how this affects the result

image

hanxiao: and in `ReduceAllDriver`, this "force a shallow copy" does not guarantee it copies from the docs in the current request
```
class ReduceAllDriver(ReduceDriver):
    """:class:`ReduceAllDriver` merges chunks and matches from all requests """

    def _apply(self, doc: 'jina_pb2.Document', *args, **kwargs):
        if doc.id not in self.doc_pointers:
            self.doc_pointers[doc.id] = copy(doc)  # force a shallow copy
```

hanxiao: ok let me fix `ReduceAllDriver` on this branch

jina/proto/jina.proto
hanxiao: if it is called `something_id` then better keep it as `int`.

maybe `repeated string modalities`?

in principal it can be also part of `tags`, but given this is one of the highlights of Jina, we promote this field as an independent field.

> For the sake of completion, should it be a repeated field? so that a document can have more than one mode and then it can be split into chunks of different modes?

JoanFM: I would agree with ` repeated string modalities`. I think a string can have more meaningful names that can map business or function value. Another point is whether we need to keep multiple content types in the same document?.

hanxiao:
this commit should fix the reduce driver. but let me work on the docstring of `ReduceAllDriver` to clarify what are the expectations here.

jina/proto/jina.proto
hanxiao: do you mean something like this?
```protobuf
message MixedContent {
  repeated Content values = 1;
}

message Content {
  oneof content {
    bytes buffer = 1;
    
    NdArray blob = 2;
    
    string text = 3;
    
    MixedContent mix = 4;
  }
}
```

and then 

```protobuf
message Document {
   //...
   Content content = 1
   //...
}
```

JoanFM: yes, something like this. I would expect that if I reconstruct the content of my chunks, I can exactly map it into the content of the parent document. Does it make sense? Or is it adding another layer of complexity?

hanxiao: let's keep the content "flat" as it is and uses nested "Chunks" to improve the expressivity of a Document.

JoanFM: Ok, but then if someone wants to build a `crafter` to split them, it must have the content in a way that he/she can split, right? But I agree, let's keep it for now `flat` and if we see that is needed we can always add later

hanxiao: Imagine when a doc first enters Jina Flow, it is a quite reasonable and general assumption that this doc is always "single-content" as `buffer.

It then depends on the `Crafter`/`Segmenter` to interpret the buffer and translate it into more readable content. The multi-modality interpretation happens inside the Flow.

If we assume a doc enters Jina with "multi-content", then this means there is a powerful client does what our `Crafter` supposed to do: it reads from buffer and parses it and fills in the "multi-content".

We can not make such assumption on the client side, because:
- it significantly increases the maintaining cost of clients (especially when it comes to multi-language)
- it is easier to scale and maintain when such crafting job is done inside the flow.

hanxiao: now i start to think we should use `string modality` rather than `repeated string`

> let's keep the content "flat" as it is and uses nested "Chunks" to improve the expressivity of a Document.

hanxiao: given that a recursive document can give us *infinite* expressivity, we should really keep the fields inside `Document` simple and flat

tests/unit/drivers/test_reduce_all_driver.py
JoanFM: is it now working this test?

JoanFM: we need to add more assertions

JoanFM: added!

jina/proto/jina.proto
```diff
@@ -249,6 +252,8 @@ message Request {
         Command command = 1; // the control command
 
         map<string, string> args = 2; // the pea arguments, useful in STATUS
+
+        uint32 modality = 3; // modality of the requested, used in ROUTER-DEALER pattern, tells the router which modality wants to get documents for.
```
JoanFM, line 256: We can remove this, it is not necessary anymore

Issue #711, opened 2020-07-24
Codecoverage does not seem to get accurate delta descriptions on the PR


JoanFM:
**Describe your problem**
Codecoverage does not seem to make accurate PR comments on the specific impact of a PR on code coverage. 

I have the suspicion that it mixes the impact of the `previous` PR merged into master and the one being analyzed, and I guess it is because that no report is done with `tests` only on `master` branch.

https://docs.codecov.io/docs/comparing-commits

hanxiao:
my guess is codecov should also be run in `cd.yml`

JoanFM:
Yes, do we have a test of master after every merge is done?

hanxiao:
not yet. you can add it

JoanFM:
It seems to be working since #713

PR #713, opened 2020-07-25
feat(ci): add tests and coverage when push on master

JoanFM:
**Changes introduced**
Runs tests and coverage reports when merge on master happens, intentds to solve #711 

I would like to know if I can abstract these [`unit-test, integration-test. codecov`] steps into its own single `.yml `file to be called from any workflow.


hanxiao:
fyi, YAML support anchor reference with `&` and `<<`, but it must be in the same file

http://blogs.perl.org/users/tinita/2019/05/reusing-data-with-yaml-anchors-aliases-and-merge-keys.html

PR #714, opened 2020-07-25
feat(drivers): filterQL cannot handle doc in place

JoanFM:
**Changes introduced**
- Make `apply` and `apply_all` methods 'private' for drivers.
- `FilterQL` currently pretends to handle document in place by using `del doc` operation. This just sets the garbage collector to erase that variable if possible at some point, but it is no guarantee that the doc will not be in the list of docs any longer. Here I have an **specialized** way of traversing documents for `FIlterQL`. 

**Disclaimer**
I am not very happy with the layout of this PR. Maybe there is a better way that can avoid specializing `_preorder_apply` and `_postorder_apply`, plus there must be a more efficient way to tell python that I want to remove a specific set of indexes from an array.


hanxiao:
another way is to use `del` in `apply_all`, and that should work let me try.

> FilterQL currently pretends to handle document in place by using del doc operation. This just sets the garbage collector to erase that variable if possible at some point, but it is no guarantee that the doc will not be in the list of docs any longer. Here I have an specialized way of traversing documents for FIlterQL.

JoanFM:
> another way is to use `del` in `apply_all`, and that should work let me try.
> 
> > FilterQL currently pretends to handle document in place by using del doc operation. This just sets the garbage collector to erase that variable if possible at some point, but it is no guarantee that the doc will not be in the list of docs any longer. Here I have an specialized way of traversing documents for FIlterQL.

Feel free to take this branch, so that u can get the tests changes and so ...

Issue #692, opened 2020-07-22
Continuously improve testing and test coverage


JoanFM:
**Description**
We have to **continuously** push for **quality**, thus we always need to improve our **testing** and **test coverage**. Any improvement of testing is a great **contribution** to **jina**.

If you have no idea how to get started on this, here there are some ideas that could be good to have.

- [ ] **Increase coverage**
Checking the reports on `https://codecov.io/gh/jina-ai/jina` one can see what parts of the code are less tested and we can work together to increase the coverage

- [ ] **Increase the `unitaryism` (I am sure this word does not exist ;)) of unittests.**
Some tests (see, for instance,  test_mime.py, test_segmenter.py) heavily rely on building a Flow to test a functionality. Move those tests under `flow` tests and have the same functionality tested as a simple unittest.
Having a more `unitary` test helps to have better understanding on how pieces of code and classes have to be used.
It also helps to identify potential code quality improvements (It can detect too much coupling between classes and modules) (It should be easy without mocking too many classes to implement these tests) 

- [ ] **Better pytest standards** 
We are trying to improve in our usage of `pytest` features (fixtures, parametrization, etc...) so it can be good to bring old-style tests to this new wave.

lusloher:
Where do we stand on this item?

MiaAltieri:
I'd love to contribute and help to convert either/both test_mime.py or test_segmenter.py into unit tests underneath the flow tests, since their codebase relies heavily on flows. 

I saw that you asked to make sure they have the same functionality tested but as a simple unit tests. They seem like short simple tests already, what can I do to help further simply/unitize them? Is there something you would specifically like to see?

winstonww:
Hi Team,

I'd love to contribute to this issue as well. If I read `https://codecov.io/gh/jina-ai/jina` correctly, the file `jina/peapods/zmq/__init__.py` is currently less tested than the other components. For completeness, I am thinking of adding a unit test to test `Zmqlet`'s `recv_message` method. I have published draft PR #2361  to show this test. If there are any suggestions, please feel free to let me know. Thank you!

winstonww:
> Hi Team,
> 
[...]

PR #719, opened 2020-07-26
feat(proto): change tags from map to struct

hanxiao:


tests/unit/test_protobuf.py
```diff
@@ -0,0 +1,31 @@
...
+
+class ProtobufTestCase(JinaTestCase):
+
+    def test_tags(self):
+        d = Document()
+        d.tags['int'] = 1  # will convert to float!!!
+        d.tags['str'] = 'blah'
+        d.tags['float'] = 0.1234
+        d.tags['bool'] = True
+        d.tags['nested'] = {'bool': True}
```
bwanglzu, line 16: One suggestion is we can setup a set of test fixtures in tests and reuse these fixtures in other tests.

Reference:
> One great use for fixtures is to set up structures or variables the same way for all tests. This is to make sure that tests can run individually as well as a set and in any order. [link](https://pythontesting.net/framework/unittest/unittest-fixtures/)

Might be related to #692

tests/unit/test_protobuf.py
```diff
@@ -0,0 +1,31 @@
...
+from jina.proto.jina_pb2 import Document
+from tests import JinaTestCase
+from google.protobuf.json_format import MessageToJson, Parse
+
+
+class ProtobufTestCase(JinaTestCase):
+
+    def test_tags(self):
+        d = Document()
+        d.tags['int'] = 1  # will convert to float!!!
```
hanxiao, line 12: @JoanFM don't understand why but apparently this `Struct.Value` does not support `int`
 https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value

JoanFM:
@hanxiao, it is strange, I will look into it, but anyway it would be an improvement with respect with current tags right?

JoanFM:
Found this stackoverflow conversation discussing the reason behind no ints. https://stackoverflow.com/questions/51818125/how-to-use-ints-in-a-protobuf-struct

hanxiao:
- we can later make our own `Struct` by including `int32/64`.
- Note, the original `Struct` is designed as JSON compatible. In JS there is only `Number` no float or int. So when it comes to REST API, we may still have to actively parse float into int, e.g. `top_k`.

For now, we can merge it. Later we see how the current `Stuct` works for the doc enrichment and query lang and then decide how to refactor it.

PR #725, opened 2020-07-27
fix(encode): extract should work at chunk level

JoanFM:
**Changes introduced**
Right now, **extract_docs** is not looping on document **chunks**.  I have identified something strange working on #712 . Encoders are receiving the complete document content instead of encoding one chunk at a time.

**Disclaimer**
Maybe I am missing something on how **encoders** should receive now information since the big refactor from #652. 

**TODO**
Add encoder driver test

### DIFF
--- a/jina/drivers/encode.py
+++ b/jina/drivers/encode.py
@@ -22,11 +22,8 @@ class EncodeDriver(BaseEncodeDriver):
     """
 
     def _apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs):
-        contents, chunk_pts, no_chunk_docs, bad_chunk_ids = extract_docs(docs,
-                                                                         embedding=False)
-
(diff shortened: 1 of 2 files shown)


hanxiao:
the default `depth_range` is `[0,0]` so all drivers only work on "previous doc-level". Setting the `depth_range` to `[0,1]` can often solve the problem

JoanFM:
> the default `depth_range` is `[0,0]` so all drivers only work on "previous doc-level". Setting the `depth_range` to `[0,1]` can often solve the problem

Should it be set by default for `EncodeDriver` maybe?

hanxiao:
yes. we can. We can also set default `depth_range` at `BaseRecursiveDriver` to `[0,1]`. For now, I choose to write the `depth_range` explicitly in YAML and see if I can find some patterns.

PR #729, opened 2020-07-28
feat(rankers): add tests for Chunk2DocRankerDriver + fix bug in the column indexes 

JoanFM:
**Changes introduced**

- Fix a bug in **Chunk2DocRankerDriver**. Executor sets these values `self.col_doc_id = 0; self.col_chunk_id = 1`. But the driver fills the match_idx: `match_idx.append((match.id, match.parent_id, c.id, match.score.value))`, while in RankerTestCase it is fileld as expected: ` {'parent_id': 1, 'id': 10, 'score': 0.4, 'length': 200}`

- Added tests for **Chunk2DocRankerDriver**


hanxiao:
Hi Joan, I will work on this branch this afternoon.

JoanFM:
> Hi Joan, I will work on this branch this afternoon.

Okey, I was also thinking about looking at how to refactor the Ranker Executor to make it seamlessly work for **Chunk2DocRanker** and for **Doc2DocRanker** (do not know if it is the proper naming)

hanxiao:
What is `Doc2DocRanker`?

JoanFM:
> What is `Doc2DocRanker`?

It is not yet a name, I meant a Ranker that re-scores `matched documents` based on document level information.
Currently Rankers work as some kind of reduction of `chunk` level scores and info into the `doc-level`.

JoanFM:
btw, it may be better, to merge this since it only adds tests and a small fix. Like this, when you work this afternoon on the issue, u will have less noise.

hanxiao:
I did a quick job and refactor it a bit. Basically the same thing. I try to avoid looping on chunks/drivers inside `__apply__` or `__apply_all__` as the recursive looping should be taken care by `BaseRecursiveDriver.__call__`. Same thing I did in https://github.com/jina-ai/jina/pull/714 to refactor `KVSearchDriver`.

Before that, if you set `depth_range=[0,0]` to `Chunk2DocRankDriver`, it will still access info at level 1 due to `for c in doc.chunks`. This commit fixes the inconsistent semantic on the depth boundary.

hanxiao:
I see. 

Yep, I don't like this name `Chunk2DocRankDriver`. But I don't have a better idea now. Regardless the names,

- `Chunk2DocRank` works as a reduction to aggregate `k`th level scores into `k-1`th level
- `Doc2DocRank` works as a mutation on the `k`the level scores and re-score them.

Is this what you pictured?

> > What is `Doc2DocRanker`?
> 
> It is not yet a name, I meant a Ranker that re-scores `matched documents` based on document level information.
> Currently Rankers work as some kind of reduction of `chunk` level scores and info into the `doc-level`.

JoanFM:
> I see.
> 
> Yep, I don't like this name `Chunk2DocRankDriver`. But I don't have a better idea now. Regardless the names,
> 
> * `Chunk2DocRank` works as a reduction to aggregate `k`th level scores into `k-1`th level
> * `Doc2DocRank` works as a mutation on the `k`the level scores and re-score them.
> 
> Is this what you pictured?
> 
> > > What is `Doc2DocRanker`?
> > 
> > 
> > It is not yet a name, I meant a Ranker that re-scores `matched documents` based on document level information.
> > Currently Rankers work as some kind of reduction of `chunk` level scores and info into the `doc-level`.

It is exactly what I see. But I do not know if it can be done seamlessly with the same Rankers.

PR #734, opened 2020-07-29
test(drivers): add encode driver test

JoanFM:
**Changes introduced**
Add unittests for **EncodeDriver**


hanxiao:
btw, the reason I used a very compact variable naming style in `/drivers` is that originally I expect this part to be rewritten in Cython.

PR #738, opened 2020-07-29
test(drivers): add test for segment driver

JoanFM:
**Changes introduced**
Add unittests fro `SegmentDriver`

### DIFF
--- a/jina/drivers/craft.py
+++ b/jina/drivers/craft.py
@@ -69,6 +69,5 @@ class SegmentDriver(CraftDriver):
                 c.parent_id = doc.id
                 c.level_depth = doc.level_depth + 1
                 c.mime_type = doc.mime_type
-            doc.length = len(ret)
         else:
             self.logger.warning(f'doc {doc.id} at level {doc.level_depth} gives no chunk')
--- a/jina/clients/python/request.py
+++ b/jina/clients/python/request.py
@@ -80,6 +80,7 @@ def _generate(data: Union[Iterator[bytes], Iterator['jina_pb2.Document'], Iterat
 
             d.id = next(doc_counter)
             d.weight = 1.0
+            d.length = batch_size
         yield req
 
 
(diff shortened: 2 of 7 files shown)


jina/drivers/craft.py
```diff
@@ -64,7 +64,7 @@ def _apply(self, doc: 'jina_pb2.Document', *args, **kwargs):
             for r in ret:
                 c = doc.chunks.add()
                 self.set_doc_attr(c, r, self._protected_fields)
-                c.length = len(ret)
+                c.length = 0
```
hanxiao, line 67: why is this zero? `len(ret)` is correct. the idea is to keep track of how many siblings (i.e. other `Document` belongs to the same parent at the same level) this chunk has, which can later be used for normalization in the ranking.

JoanFM: It feels weird, to set length to 0, when in documentation it says that refers to the number of chunks. Besides this, this change  not breaking any tests means that we are missing some tests.

hanxiao: let's revert it back on this line but instead change the proto docstring to `total number of siblings of this document`. In the next PR, we can think about if there is a better name instead of `length`

```suggestion
                c.length = len(ret)
```

JoanFM: ok, let me revert it back together with the test

JoanFM: then the assigning of `doc.length = len(ret)` must be wrong?

hanxiao: hmm...let me check how `length` is used now

JoanFM: I think is not used anywhere

hanxiao: `length` is mostly used in `executors/rankers/tfidf.py` (`TfIdfRanker`) as normalizer

JoanFM: ok, I see. But one of the 2 usages that is assigned in this PR must be wrong right? They are contradictory in the meaning of `length`. Should it mean number of `children` or number of `siblings`?

JoanFM: From the description of `tf-idf` it should be `number of siblings` I think. So we should remove the assignment `doc.length = len(ret)`

PR #742, opened 2020-07-30
refactor(indexer): changes detected by using mypy static type checker

JoanFM:
**Changes introduced**
Some changes suggested by the usage of `mypy` as static type checker for jina.

Usage: 

```bash
pip install mypy
mypy --ignore-missing-imports jina
```
This PR solves:

```bash
jina/executors/indexers/vector/ngt.py:21: error: Incompatible default for argument "epsilon" (default has type "float", argument has type "int")
jina/executors/indexers/vector/ngt.py:47: error: Signature of "query" incompatible with supertype "BaseIndexer"
jina/executors/indexers/vector/ngt.py:58: error: "append" of "list" does not return a value
```

### DIFF
--- a/jina/executors/indexers/vector/ngt.py
+++ b/jina/executors/indexers/vector/ngt.py
@@ -18,7 +18,7 @@ class NGTIndexer(BaseNumpyIndexer):
         Quick Install : pip install ngt
     """
 
-    def __init__(self, num_threads: int = 2, metric: str = 'L2', epsilon: int = 0.1, *args, **kwargs):
+    def __init__(self, num_threads: int = 2, metric: str = 'L2', epsilon: float = 0.1, *args, **kwargs):
         """
         Initialize an NGT Indexer
         :param num_threads: Number of threads to build index
@@ -53,9 +53,7 @@ class NGTIndexer(BaseNumpyIndexer):
         idx = []
         for key in keys:
             results = index.search(key, size=top_k, epsilon=self.epsilon)
-            index_k = []
-            distance_k = []
-            [(index_k.append(result[0]), distance_k.append(result[1])) for result in results]
+            index_k, distance_k = zip(*results)
             idx.append(index_k)
             dist.append(distance_k)
 
--- a/jina/executors/indexers/vector/ngt.py
+++ b/jina/executors/indexers/vector/ngt.py
@@ -53,8 +53,9 @@ class NGTIndexer(BaseNumpyIndexer):
         idx = []
         for key in keys:
             results = index.search(key, size=top_k, epsilon=self.epsilon)
-            index_k, distance_k = zip(*results)
-            idx.append(index_k)
-            dist.append(distance_k)
+            if results:
+                index_k, distance_k = zip(*results)
+                idx.append(index_k)
+                dist.append(distance_k)
 
         return self.int2ext_key[np.array(idx)], np.array(dist)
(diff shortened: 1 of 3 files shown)


jina/executors/indexers/vector/ngt.py
```diff
@@ -55,7 +55,9 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
             results = index.search(key, size=top_k, epsilon=self.epsilon)
             index_k = []
             distance_k = []
-            [(index_k.append(result[0]), distance_k.append(result[1])) for result in results]
+            for result in results:
+                index_k.append(result[0])
```
bwanglzu, line 59: guess you first need to do a len(result) > 2 check, might cause `IndexError`

JoanFM: I think, it would mean that the search launched an exception. Wouldn't it be too overkilling to check for every result the length? I agree though that we should have a clearer policy into sanity checks

JoanFM: @nan-wang what is your thought on this?

hanxiao: I didn't notice this before, bad coding style 😢 
> [(index_k.append(result[0]), distance_k.append(result[1])) for result in results]
plus, this can be as simple as
```python
index_k, distance_k = zip(*results)
```

JoanFM: Changed to @hanxiao suggestion!

jina/executors/indexers/vector/ngt.py
```diff
@@ -53,9 +53,7 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
         idx = []
         for key in keys:
             results = index.search(key, size=top_k, epsilon=self.epsilon)
-            index_k = []
-            distance_k = []
-            [(index_k.append(result[0]), distance_k.append(result[1])) for result in results]
+            index_k, distance_k = zip(*results)
```
nan-wang, line 56: It's better to verify `results` is empty
```suggestion
            if results:
                index_k, distance_k = zip(*results)
                idx.append(index_k)
                dist.append(distance_k)
```

## Reference:
Index.search(): https://github.com/yahoojapan/NGT/blob/master/python/ngt/base.py#L363

JoanFM:
> Minor changes

Done!

jina/executors/indexers/vector/ngt.py
hanxiao: wait this is wrong

jina/executors/indexers/vector/ngt.py
```diff
@@ -53,9 +53,10 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
...
             results = index.search(key, size=top_k, epsilon=self.epsilon)
-            index_k = []
-            distance_k = []
-            [(index_k.append(result[0]), distance_k.append(result[1])) for result in results]
+            if results:
+                index_k, distance_k = zip(*results)
+                idx.append(index_k)
+                dist.append(distance_k)
             idx.append(index_k)
             dist.append(distance_k)
```
hanxiao, line 61: append twice

JoanFM: Done, thanks!

Issue #693, opened 2020-07-22
QueryLangDriver needs to pick up values from payload


hanxiao:
**Describe the feature**

#652 introduces a new set of driver in `drivers/querylang`. These drivers should pick up configs from two places:
1. its own configs from YAML,
2. the payload carried in `Request`, e.g. the value of `top_k`, `filter_by`
3. when both settings are set, QueryLangDriver also needs to solve the conflict.

Right now (2) and (3) are missing. All drivers in `drivers/querylang` do not respect values in the payload, making `top_k` completely useless.

**Your proposal**

- Refactor `QuerySetDriver` and make all Drivers inherited from it can easily pick up values from payload in `Request`. Once for all. We don't want to write read-payload logic for each driver.
- Add a conflict-solving-logic that determines who overrides who.

---

**Environment**

**Screenshots**

PR #751, opened 2020-07-31
feat: driver reads queryset from the request

hanxiao:
Solving #693 . QueryLang can be defined either fixed (in the driver), or dynamically (from the request).

Think about the query time, before this PR, QueryLang is fixed and there is no way to change the filter, slicing's parameters in query time. If the user what to change top_k value in the query time then there is no way to change it.

This PR grants whoever inherited from `QueryLangDriver` the ability to access the field inside the message. The access is simply a Pythonic "get attribute", i.e. `self.key`, hiding the complexity to QL/driver developer.


jina/proto/jina.proto
```diff
@@ -255,6 +255,8 @@ message Request {
     }
 
     Status status = 6; // status info
+
+    map<string, google.protobuf.Struct> queryset = 7; // the query set
```
JoanFM, line 259: what is the idea behind it?

JoanFM:
Is the point of this PR, to avoid defining complex query language logic in the `yaml` files and get them encoded in the request? so that we have flexibility at `Document` level?

hanxiao:
it's about solving #693

JoanFM:
> it's about solving #693 . `QueryLang` can be defined either fixed (in the driver), or dynamically (from the request).
> 
> Think about the query time, before this PR, `QueryLang` is fixed and there is no way to change the filter, slicing's parameters in query time. If the user what to change `top_k` value in the query time then there is no way to change it.
> 
> This PR grants whoever inherited from `QueryLangDriver` the ability to access the field inside the message. The access is simply a Pythonic "get attribute", i.e. `self.key`, hiding the complexity to QL/driver developer.

Yes, what I thought, I think it is a great feature!

Issue #753, opened 2020-07-31
Update extra_requirements: Error when running `transformers` test locally


JoanFM:
**Describe your problem**
When running locally: 

`pytest tests/unit/executors/encoders/nlp/test_mock_transformer.py`

I tried this `test` after observing this same behavior running `urbandict` example.

I receive `FAILED` results:

```bash
FAILED tests/unit/executors/encoders/nlp/test_mock_transformer.py::TransformerEncoderWithMockedModelTestCase::test_encodes_bert_like - ValueError: expected sequence of length 6 at dim 1 (got 5)
FAILED tests/unit/executors/encoders/nlp/test_mock_transformer.py::TransformerEncoderWithMockedModelTestCase::test_encodes_lm_like - ValueError: expected sequence of length 4 at dim 1 (got 3)
FAILED tests/unit/executors/encoders/nlp/test_mock_transformer.py::TransformerEncoderWithMockedModelTestCase::test_loads_tf_encoder - ValueError: Can't convert non-rectangular Python sequence to Tensor.
```

**Possible reason**
Is there any guarantee that batch_encode_plus will return input_ids  of the same size for a given batch? Even when there are less tokens than `max_length`?

**Environment**

```bash
platform linux -- Python 3.7.5, pytest-5.4.2, py-1.8.1, pluggy-0.13.1 -- /usr/bin/python3.7

Name: transformers
Version: 2.9.0
Summary: State-of-the-art Natural Language Processing for TensorFlow 2.0 and PyTorch
Home-page: https://github.com/huggingface/transformers
Author: Thomas Wolf, Lysandre Debut, Victor Sanh, Julien Chaumond, Sam Shleifer, Google AI Language Team Authors, Open AI team Authors, Facebook AI Authors, Carnegie Mellon University Authors
Author-email: thomas@huggingface.co
License: Apache
Location: LOCATION
Requires: regex, filelock, tokenizers, sentencepiece, tqdm, numpy, sacremoses, requests
Required-by: flair

Name: torch
Version: 1.5.0
Summary: Tensors and Dynamic neural networks in Python with strong GPU acceleration
Home-page: https://pytorch.org/
Author: PyTorch Team
Author-email: packages@pytorch.org
License: BSD-3
Location: LOCATION
Requires: numpy, future
Required-by: torchvision, laserembeddings, flair

Name: tensorflow
Version: 2.0.0
Summary: TensorFlow is an open source machine learning framework for everyone.
[...]

PR #756, opened 2020-08-02
fix(transformers): Force padding to max length for old and new transf…

JoanFM:
**Changes introduced**
Changes to solve #753 without need to update requirements. Force in different versions the setting of pad to `max_length`


JoanFM:
Not sure if better to keep this compatibility or just pin 3.0.2 in requirements.

On one hand we keep more compatibility but on the other hand it makes it harder to change the behavior in the future at the risk of keeping incoherent behavior depending on the transformers version

nan-wang:
> the other hand it makes it harder to chan

I'd prefer to merge this to further improve the compatibility. The tenet of keeping a model in jina core is to support the general models and to make it easy-to-use. As for the downside, we can safely add version info in the requirements.

PR #759, opened 2020-08-03
refactor(images): refactor image crafters, will add location

JoanFM:
**Changes introduced**

- `ImageCrafters` are not compliant with the current split between crafter and segmenter where their `craft` API is different. 
Split this logic and remove the ImageCrafter class which is turned into a series of `helper` methods.
- Add `location` information about the position of a `cropped` image in its original image. 
- Fix some mixtures of `height` and `weight` dimensions.

**TO-DO**

- [x]  Add location information from the chunks
- [x]  Extensive unit-testing


jina/executors/crafters/image/segmenter.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def _location_generator_without_padding(self, img: 'np.ndarray') -> Generator[Tuple[int, int], None, None]:
+        height, width, _ = img.shape
+        top, left = (0, 0)
+        if isinstance(self.target_size, Tuple):
+            w_size, h_size = self.target_size
+        else:
+            w_size = h_size = self.target_size
+
+        while top + h_size <= height:
```
JoanFM, line 150: I wonder if there is a more `pythonic` way to construct these `generators`

hanxiao: using Python here will be very slow. numpy can do this directly, here is a code snippet of my previous implementation on sliding windows

```python
    def _get_all_sliding_window(self, img: 'np.ndarray'):
        extend_height = self.window_size - (img.shape[0]) % self.stride_height
        extend_wide = self.window_size - (img.shape[1]) % self.stride_wide

        input = np.pad(img, ((0, extend_height),
                             (0, extend_wide),
                             (0, 0)),
                       mode='constant', constant_values=0)
        expanded_input = np.lib.stride_tricks.as_strided(
            input,
            shape=(
                1 + int((input.shape[0] - self.window_size) / self.stride_height),
                1 + int((input.shape[1] - self.window_size) / self.stride_wide),
                self.window_size,
                self.window_size,
                3
            ),
            strides=(
                input.strides[0] * self.stride_height,
                input.strides[1] * self.stride_wide,
                input.strides[0],
                input.strides[1],
                1
            ),
            writeable=False
        )
        center_point_list = [
            [self.window_size / 2 + x * self.stride_wide, self.window_size / 2 + y * self.stride_height]
            for x in range(expanded_input.shape[0])
            for y in range(expanded_input.shape[1])]

        expanded_input = expanded_input.reshape((-1, self.window_size, self.window_size, 3))
        return [np.array(Image.fromarray(img)) for img in expanded_input], center_point_list
```

JoanFM: ok, thanks, will change! 

Should we use `center_points` as the `location` or the coordinates of the `(top, left)` coordinates?

hanxiao: let's keep `top, left`. `center_points` can be derived from it anyway

JoanFM: Done!

PR #764, opened 2020-08-04
refactor: Turn BaseFrameworkExecutor into a mixin BaseDeviceHandler

JoanFM:
**Changes introduced**
- Change `BaseFrameworkExecutor` to `BaseDeviceHandler` and have all its direct subclasses be usable as `mixins` (no members are added, they just add functionality to handle `devices` to the classes they are mixed with). This solves the diamond problem in some classes.

- Make sure to call `super().post_init__` when needed, if we add some `post_init()` to a base class we expect to be called from all its children.

- Rename and refactor `test` classes under `Encoding`

- Make sure TF based image encoders set `channel_axis=1` by default. 

- Fix some issue in `CustomTFKeras`, having `channel_axis` defined twice was giving problem at reconstruction time in test.

- Fix pickle issue in `test_customtorchvision`.

**Discussion**
-  Having `BaseFrameworkExecutor` inherit from `BaseExecutor`, all the `child` classes are inheriting twice `BaseExecutor`.

- `BaseExecutor` should not have information about `gpu` or not. It is really specific to `framework executor` but I am not sure if removing `on_gpu` from `BaseExecutor` can break compatibility. I would like to `remove` from it, `on_gpu` and `_check_on_gpu`.


hanxiao:
Imagine, FaissIndexer, it is from `BaseExecutor` and uses `on_gpu`

JoanFM:
> FaissIndexer, it is from `BaseExecutor` and uses `on_gpu`
> 
> > BaseExecutor should not have information about gpu or not. It is really specific to framework executor but I am not sure if removing on_gpu from BaseExecutor can break compatibility. I would like to remove from it, on_gpu and _check_on_gpu.

It inherits from `BaseFaissExecutor` which comes from `BaseFrameworkExecutor`

hanxiao:
ok then I would suggest to remove `BaseFrameworkExecutor` completely and move all of its features into `BaseExecutor`.

As an AI framework, it is not crazy to have `on_gpu` and `device` in `BaseExecutor`. After all, we brand ourselves as providing "first-class support on AI models". Making `on_gpu` and `device` available at the very beginning is a good way to carry out the words.

JoanFM:
> ok then I would suggest to remove `BaseFrameworkExecutor` completely and move all of its features into `BaseExecutor`.
> 
> As an AI framework, it is not crazy to have `on_gpu` and `device` in `BaseExecutor`. After all, we brand ourselves as providing "first-class support on AI models". Making `on_gpu` and `device` available at the very beginning is a good way to carry out the words.

The problem that I see:

Having `BaseFrameworkExecutor` independent from `BaseExecutor` allows us to have this curent usage for instance: 

```python
class BaseTorchEncoder(BaseTorchExecutor, BaseEncoder):
    pass
```

If the logic coming from the `framework` is set at the `BaseExecutor` level we cannot recicle the `Framework` logic for different `executors`. (Important to notice that current solution in `master` suffers from the `diamond` problem)

I would not be able to do 

```python
class BaseTorchSegmenter(BaseTorchExecutor, BaseSegmenter)
pass 
```
where I can `mix-in` the handling of the `devices` and `gpu` info and the interface of the `segmenters`.

hanxiao:
by removing inheritance `BaseExecutor` from `BaseFrameworkExecutor`, you lose ALL functionalities that `BaseExecutor` and `ExecutorType` implemented, in particular YAML config serialize/deserialize, etc. The resulting `BaseFrameworkExecutor` is not usable without "mixin" with `BaseExecutor`

JoanFM:
Hey @hanxiao.

These are the alternatives that I see we have here and the benefits and problems that I see (sorry for the formatting, not my strength). In yellow is a supposed inheritance we may add.

Current-inheritance

PR-structure

merge-structure

hanxiao:
👍 very clear, I see your point.  in fact, the only thing I'm against is calling it `BaseFrameworkExecutor`. As you draw `BaseFramework` is good, or even `BaseDevice` sounds more correct. 
```
BaseDevice
|
|--- TFDevice
|--- TorchDevice
|--- OtherBlahBlahDevice

```

JoanFM:
>  very clear, I see your point. in fact, the only thing I'm against is calling it `BaseFrameworkExecutor`. As you draw `BaseFramework` is good, or even `BaseDevice` sounds more correct.
> 
> ```
> BaseDevice
> |
> |--- TFDevice
> |--- TorchDevice
> |--- OtherBlahBlahDevice
> ```
> 
> Then one can do
> 
> ```python
> class AwesomeTFCrafter(BaseCrafter, TFDeivce)
> ```

Yes, that is the idea, let's see how it looks.

jina/executors/encoders/frameworks.py
```diff
@@ -54,16 +58,23 @@ def _append_outputs(input_fn, outputs_name_to_append, output_fn):
...
+    def run_on_gpu(self):
+        return self.on_gpu
+
 
+class BaseTorchEncoder(BaseEncoder, BaseTorchDeviceHandler):
 
-class BaseTorchEncoder(BaseTorchExecutor, BaseEncoder):
-    pass
+    @property
+    def run_on_gpu(self):
```
JoanFM, line 70: This pretends to add as a `bridge` from executors to DeviceHandler, to avoid removing `on_gpu` from `Executors`. I think it is something that is already used, and can be part of the reconstruction of `executors`.

nan-wang:
Good initiatives! The diamond issue is a potential trouble-maker. Splitting the framework logic into the `BaseExecutor` and `BaseFramework` is a good idea. I love it!

jina/executors/encoders/nlp/transformer.py
```diff
@@ -180,10 +181,21 @@ def get_tensor_func(self):
         raise NotImplementedError
 
 
-class TransformerTFEncoder(BaseTransformerEncoder, BaseTextTFEncoder):
+class TransformerTFEncoder(BaseTFDeviceHandler, BaseTransformerEncoder):
```
JoanFM, line 184: Avoiding `diamond` problem

hanxiao:
im going to switch to this branch and do the review locally

hanxiao:
# WIP

I'm working on this PR directly to solve the following problems

- `BaseDeviceHandler` -> `BaseDevice`, remove `Base` and `Handler` in or inheritance
- `run_on_gpu` is meaningless as no code ever sets its value. meanwhile `on_gpu` is not refactored
- `executors/frameworks.py` -> `executors/devices.py`
- ...

jina/executors/devices.py
```diff
@@ -22,11 +22,6 @@ def device(self):
             please use the environment variable `CUDA_VISIBLE_DEVICES`.
         """
 
-    @property
-    @abstractmethod
-    def is_gpu(self) -> bool:
```
JoanFM: I think keeping this is a safer way to have this `mixin`. You force anyone who uses it to define it. By just using `self.on_gpu` there is a risk that you are mixing it with a class that does not provide this method. So it affects its independence.

hanxiao: `on_gpu` is a meta tag available to all executors. it indicates whether user wants to put it on gpu. That's it. no guarantee it will be put on to gpu.

hanxiao: oh btw, all docstring inside `executors.metas` are *not* being picked up to our docs!

tests/unit/executors/encoders/image/test_customtorchvision.py
```diff
@@ -1,6 +1,6 @@
 import tempfile
 
-from jina.executors.encoders.image.customtorchvision import CustomImageTorchEncoder
+from jina.executors.encoders.image.torchvision import CustomImageTorchEncoder
```
JoanFM, line 3: has it been moved?

PR #757, opened 2020-08-03
[W.I.P] (Do not merge) fix(drivers): loop on chunks matches not matches matches

JoanFM:
**Changes introduced**
It is just a question, is it correct to have `matches-matches`, or we want the `chunk-matches`?

Issue #702, opened 2020-07-22
🚨 BREAKING CHANGES 0.4.0 -> 0.4.1 & Migration Guide


hanxiao:
# Background

#652, #684, #700, #709  introduce breaking changes and new features. These features will be released in `v0.4.1`. 
Those changes are unfortunately not back-compatible. Let's write down the migration guide here.

# What has changed?

In short, the protobuf definition of `Document` and `Chunk` are unified to have the same representation. In this new representation, `Document` has a recursive structure. The previous `Chunk` is just a nested `Document` at one level deeper. The next figure illustrates this idea:

![image](https://user-images.githubusercontent.com/2041322/88682088-904f4a80-d0f2-11ea-9c99-1885a3e79d12.png)

This new proto enables cleaner driver design, yields more consistent low-level APIs, and provides great extensibility on future features, e.g. proximity graph/graph-based learning.

# Affected Versions

Since `v0.4.1`, expected on 31.07.2020.

The last version unaffected by those changes is `v0.4.0`, i.e. `pip install jina==0.4.0`

# What is affected?

Mostly drivers, some time executors. The affection is mostly due to
- As we don't distinguish chunk and doc anymore, special logic designed for `Chunk` is removed.
- As `Document` is now a recursive structure, all drivers have to work correctly by traversing the recursive structure.

# Migration Guide

- `level` is not required in all drivers anymore.

- `depth_range` controls the driver `_apply` level, the default is `depth_range=[0,0]`, i.e. working on doc level only. To mimic the previous chunk-level processing, use `depth_range=[1,1]`. `depth_range=[0,1]` means working on both doc and chunk level

- `PruneDriver` is now `ExcludeQL`, no `level` is required, `excludes` -> `fields`, e.g.
```yaml
    - !ExcludeQL
      with:
        fields:
          - embedding
          - buffer
          - blob
          - text
```

- `DocPruneDriver` is removed, one can now use 

```yaml
    - !ExcludeQL
      with:
        fields:
          - buffer
          - chunks
```

- `ChunkPruneDriver` is removed,  one can now use the following, as one can see it is basically the same as  `DocPruneDriver` , just with different fields.

```yaml
    - !ExcludeQL
      with:
        fields:
          - embedding
          - buffer
          - blob
          - text
```

- `BaseCraftDriver` is now `CraftDriver`. `DocCraftDriver` and `ChunkCraftDriver` are removed
- `doc_id` and `chunk_id` are unified as `id` now
[...]

PR #765, opened 2020-08-04
fix Chunk2DocRankDriver 763

### LINKED ISSUE #763: MinRanker is not using the proper driver
**Describe the bug**
`MinRanker` should use `Chunk2DocRankDriver` instead of `BaseRankDriver`. The `BaseRankDriver` does not merge the chunk level information to the doc level

**Describe how you solve it**

- - -

**Environment**

**Screenshots**

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-145) by [Unito](https://www.unito.io/learn-more)

nan-wang:
resolves #763


jina/drivers/__init__.py
```diff
@@ -258,7 +258,7 @@ def pre_traverse(_docs, traverse_on, context_doc=None):
             _traverse(docs, 'chunks')
         if 'matches' in self.traverse_fields:
             for d in docs:
-                _traverse(d.matches, 'matches')
+                _traverse(d, 'matches')
```
hanxiao, line 261: no this is wrong, please revert it back.

hanxiao: _traverse do not work on single doc

nan-wang: yes, u r right.

hanxiao:
is this ready for review? @nan-wang

jina/resources/executors.requests.Chunk2DocRanker.yml
```diff
@@ -0,0 +1,10 @@
+on:
+  ControlRequest:
+    - !ControlReqDriver {}
+  SearchRequest:
+    - !Chunk2DocRankDriver
+      with:
+        depth_range: [0, 2]  # prune root and 1st level
+        start: 0
+        end: 50
```
hanxiao, line 9: what are these start and end is about?

hanxiao, line 9:
```suggestion

```

hanxiao: and is there any reason why you want to promote a specialized executor to `resources` level?

JoanFM:
Enables https://github.com/jina-ai/examples/pull/132

nan-wang:
> is this ready for review? @nan-wang 

not yet. This is for fixing failure of retrieving docs in examples.

tests/unit/drivers/test_rank_driver.py
```diff
@@ -0,0 +1,100 @@
...
+class Chunk2DocRankDriverTestCase(JinaTestCase):
+    def test_chunk2doc_driver_mock_indexer(self):
+        doc = create_document_to_rank()
+        print(doc)
+        driver = Chunk2DocRankDriver()
+        executor = MockRanker()
+        args = set_pea_parser().parse_args(['--runtime', 'process'])
+        pea = MockPea(args)
+
+        driver.attach(executor=executor, pea=pea)
```
JoanFM, line 91: You can safely set pea to None I think and avoid having a MockPea.

nan-wang: Nope, the driver needs to check the `self.pea.message.envelope.status.code` in order to decide on which level to skip the execution
https://github.com/jina-ai/jina/blob/master/jina/drivers/__init__.py#L293

tests/unit/drivers/test_rank_driver.py
```diff
@@ -0,0 +1,100 @@
...
+
+class MockPea(BasePea):
+    def __init__(self, args):
+        super().__init__(args)
+        self._message = jina_pb2.Message()
+        self._message.envelope.status.code = jina_pb2.Status.SUCCESS
+
+
+class Chunk2DocRankDriverTestCase(JinaTestCase):
+    def test_chunk2doc_driver_mock_indexer(self):
```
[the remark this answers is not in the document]
nan-wang: fair enough. I'll change it.

tests/unit/drivers/test_rank_driver.py
```diff
@@ -0,0 +1,100 @@
+from tests import JinaTestCase
```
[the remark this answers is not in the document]
nan-wang: hmm, I've just found it. `test_chunk2doc_rank_drivers.py`.

jina/drivers/rank.py
```diff
@@ -28,11 +29,12 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.recursion_order = 'post'
 
-    def _apply_all(self, docs: 'jina_pb2.Document', context_doc: 'jina_pb2.Document', *args, **kwargs):
+    def _apply_all(self, docs: List['jina_pb2.Document'], context_doc: 'jina_pb2.Document', travers_on: str, *args, **kwargs):
```
hanxiao, line 32: if you don't need `traverse_on`, no need to add it. That's how *args and **kwargs for

jina/drivers/rank.py
```diff
@@ -28,11 +29,12 @@ def __init__(self, *args, **kwargs):
...
         """
 
         :param docs: the chunks of the ``context_doc``, they are at depth_level ``k``
         :param context_doc: the owner of ``docs``, it is at depth_level ``k-1``
+        :param traverse_on: "matches" or "chunks"
```
hanxiao, line 37: same here, if you don't need in the function, no need to add it

jina/drivers/__init__.py
```diff
@@ -254,6 +255,7 @@ def pre_traverse(_docs, traverse_on, context_doc=None):
         else:
             raise ValueError(f'{self.recursion_order}')
 
+        # context_doc is missing, and this leads to `_apply_all` is skipped
```
hanxiao, line 258: what does this mean? can you remove this comment if nothing wrong with the logic?

also, always refer to #702 illustration about the doc structure

jina/drivers/rank.py
nan-wang: Yes, u r right. The codes need clean up.

jina/drivers/__init__.py
nan-wang: Some debug messages left over. As above, it has been cleaned up.

tests/test_helper.py
```diff
@@ -0,0 +1,23 @@
+from tests import JinaTestCase
+from jina.helper import cached_property
+
+
+class MyTestCase(JinaTestCase):
```
JoanFM, line 5: can you put a more descriptive name to the `TestCase` class? And shouldn't it be under `tests/unit`?

tests/test_helper.py
```diff
@@ -0,0 +1,23 @@
+from tests import JinaTestCase
```
JoanFM, line 1: I think these tests are a great idea, we should also add a `test` for `batching` decorator

jina/helper.py
```diff
@@ -647,6 +647,6 @@ def __init__(self, func):
     def __get__(self, obj, cls):
         if obj is None:
             return self
-
-        value = obj.__dict__[f'CACHED_{self.func.__name__}'] = self.func(obj)
```
JoanFM, line 651: In `jina/executors/__init__.py` there is some logic related to this that needs to be reviewed maybe?

```python
    def __getstate__(self):
        d = dict(self.__dict__)
        del d['logger']
        for k in self._post_init_vars:
            del d[k]
        cached = [k for k in d.keys() if k.startswith('CACHED_')]
        for k in cached:
            del d[k]
        return d
```

JoanFM: or was it the source of the problem? that we should say `if not  k.startswith('CACHED_')`?

nan-wang: Good catch! This part of the codes is for dumping `executors` into pickles. We don't want to save the cached variables, e.g. the models. However, using the `CACHED_` prefix leads to the failure of retrieving the variable from the `self.__dict__`.

hanxiao: your changes will fail. because you save cached which are not pickable. please revert it back

hanxiao:
@nan-wang I'm confused, what is this PR really about? As far as I can tell, it adds some unit tests. problem reported in #763 is either not-an-issue or this hasn't resolved in this PR yet. And now it tries to resolve `cached_property` without supported unit tests.

## Suggestions

- change the name of this PR, label  #763 with correct status; 
- resolve cached_property separately in another PR with supported unit test.

jina/drivers/__init__.py
```diff
@@ -260,6 +260,10 @@ def pre_traverse(_docs, traverse_on, context_doc=None):
         if 'matches' in self.traverse_fields:
             for d in docs:
                 _traverse(d.matches, 'matches')
+        if 'chunk_matches' in self.traverse_fields:
```
JoanFM, line 263: I think this effect is already achieved by adding chunks, matches. Given that is if, if, first will iterate on chunks and later on matches.

hanxiao: @JoanFM is right. no need of this logic . simply set traverse_on to [chunks, matches] does this job.

jina/helper.py
JoanFM: This has been handled by another PR

jina/drivers/__init__.py
hanxiao: https://github.com/jina-ai/jina/pull/757#issuecomment-667923756

plz see my previous comment.

btw, this PR is really sidetracked and takes too long time. i strongly suggested to close this PR with your added unittest only, if there are fixes on Transformers then started another PR.

or at least change the title of this PR or the issue description #762

nan-wang:
rephrased in #799

PR #769, opened 2020-08-06
fix(crafters): fix crop dimension

JoanFM:
**Changes introduced**
- Spotted an issue when calling cropping, we were not respecting that `_crop_image` expects dimensions in format (height, width). Issue was not spotted because tests were done with `square` shaped images.

- Add some test with different dimensions

**WARNING - QUESTIONS**
- Qualitatively I have checked the crop, but for `precision` issues that seem to happen when saving by `PIL` cannot test the perfect reconstruction of the `crop`. Any idea how to fix this?

### DIFF
--- a/jina/executors/crafters/image/crop.py
+++ b/jina/executors/crafters/image/crop.py
@@ -39,7 +39,7 @@ class ImageCropper(BaseCrafter):
         :returns: a chunk dict with the cropped image
         """
         raw_img = _load_image(blob, self.channel_axis)
-        _img, top, left = _crop_image(raw_img, target_size=(self.width, self.height), top=self.top, left=self.left)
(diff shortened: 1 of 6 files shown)


nan-wang:
As for the reconstruction issue, is it related to the scale and crop operation?

JoanFM:
Not directly

nan-wang:
The PR is OK. Would you elaborate more on the precision?

JoanFM:
Yes, the PR is ready. I wanted to add tests with real image to validate the cropping. But when saving to disk and comparing although they are the same, some precision errors make the assert Fail.

JoanFM:
@nan-wang  I think this can be merged, not related to the precision issue, if I do not find the reason I might create an issue and ask for help from the community

PR #770, opened 2020-08-06
feat: object detection segmenter

JoanFM:
**Changes introduced**
- Add `torchvision` detection models pretrained on `COCO dataset` for `object detection`. They will be part of `image` segmenters. They can be viewed as an intelligent AI-powered way of `cropping` the images to get `chunks` with meaningful `semantic` content.

- Try to unify the `axis` moving logic to move `channel` axis to the required position

- Add extensive tests with mocking model and with real images. Images have been checked visually but due to some precision issues cannot compare exactly the content of the images. (may be worth investigating or adding an issue for the community to help)?

**Left out TODO**
It may be overkilling at this point, adding the possibility to change the `backbone` (the CNN component that is responsible for the `classification part` of object detection.


jina/executors/crafters/frameworks.py
```diff
@@ -0,0 +1,17 @@
...
+        super().__init__(*args, **kwargs)
+        self.model_name = model_name
+
+    def post_init(self):
+        super().post_init()
+        self._device = None
+
+    @property
+    def run_on_gpu(self):
+        return self.on_gpu
```
hanxiao, line 17:
```suggestion
```

I don't think these are necessary and being used anywhere

jina/executors/crafters/frameworks.py
```diff
@@ -0,0 +1,17 @@
+from jina.executors.crafters import BaseSegmenter
+from jina.executors.devices import TorchDevice
+
+
+class BaseTorchSegmenter(TorchDevice, BaseSegmenter):
+
+    def __init__(self, model_name: str = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.model_name = model_name
```
hanxiao, line 9: I also recommend completely remove this class, as the only thing we inherit is `model_name`, which is not a torch/segment related pattern and there is no clue that this pattern is general.

```suggestion
class BaseTorchSegmenter(TorchDevice, BaseSegmenter):

    def __init__(self, model_name: str = None, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.model_name = model_name
```

JoanFM:
> no need on `BaseTorchSegmenter`

Done!

PR #782, opened 2020-08-07
test: separate test for sliceql

JoanFM:
**Changes introduced**
Example on how to `test` a class without the use of `Flows`. 
Also introduce an example using a really complex `recursive document structure`


tests/unit/drivers/querylang/test_querylang_drivers.py
```diff
@@ -55,48 +53,6 @@ def craft(self, text, *args, **kwargs):
 
 class QueryLangTestCase(JinaTestCase):
 
-    def test_segment_driver(self):
```
JoanFM, line 58: there is already a dedicated test for `SegmentDriver`

PR #783, opened 2020-08-07
fix: add missing section

alexcg1:
Vim somehow undid my section addition. I've re-added it here


alexcg1:
@all-contributors please add @alexcg1 for doc

PR #788, opened 2020-08-10
fix: test object detection check real image content

JoanFM:
**Changes introduced**
Found the way to check the `actual content` of the resulting crop image to have a better testing

@nan-wang it is related to the "precision problem" I was facing.


alexcg1:
@all-contributors please add @JoanFM  for code

PR #812, opened 2020-08-11
docs: add docs about common jina practices

JoanFM:
@alexcg1 @hanxiao 

We have a task to add some documentation about `jina best practices` or `patterns`. Here I showcase the schema I want to introduce and the place. Do you think is good way? Is it the right place in the docs to place it? Are we missing some common pattern that might deserve a little explanation for a user to grasp?

It would be good to review the content @hanxiao, @nan-wang, it may not be so accurate?

### DIFF
--- a/docs/chapters/flow/index.md
+++ b/docs/chapters/flow/index.md
@@ -302,3 +302,17 @@ With Jina Dashboard, you can interactively drag-n-drop Pod, set its attribute an
 
 
 More information on the dashboard can be found [here](https://github.com/jina-ai/dashboard).
+
+# Common design patterns
+
+jina is a really flexible AI-powered neural search framework. It is designed to enable any pattern that can be framed as  
+a neural search problem.
+However, there are basic common patterns that show up when developing search solutions with jina and here is a recopilation of some of them.
+
+- CompoundIndexer (Vector + KV Indexers):
+
+- Indexers at different depth levels:
+
+- NLP Document chunks (Segment + encode):
+
+- Reference Indexer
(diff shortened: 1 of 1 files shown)


alexcg1:
I like the concept and it looks good - kind of like Jina 101 but much deeper. And I dig that you use examples

alexcg1:
@hanxiao Should we do Jina 10X? 20X? (e.g. 102, 201, 202) - I think this is a good concept to treat knowledge as a progression and something users can see is a step by step process. This would be a good match for a "Jina curriculum"

I've got lots of thoughts in this area ;)

JoanFM:
I think it is important also here to decide the good section:

As I see it, it may have impact as how `jina` brands itself. 

If we put a title such as `jina Design Pattern`, we are branding `jina` almost like a `pseudo programming language for search`. And if we go for this, it may make sense to have a dedicated section.

alexcg1:
@JoanFM I like the title `Jina Design Pattern` since `design pattern` is a phrase that comes up in our presentations a lot.

lusloher:
Hi team @JoanFM such good work, exactly what we need, not just a new framework of how to neural search, but also what dominant patterns are to use Jina :) 
I was looking through it and love the build up, I would just suggest a small change in order, that for each pattern, before closely describing it we give an important exaple (e.g. long-text search) right away, so this way people can read instructions with the set example in mind :)

theUnkownName:
Really cool! I like a lot this style of explaining the concept followed by an example, we should try to make more of this.

PR #817, opened 2020-08-13
fix(drivers): fix the default value for depth_level 814

### LINKED ISSUE #814: level_depth=(0, 0) is invalid
**Describe the bug**
The default value of `BaseRecursiveDriver ` is `level_depth=(0, 0)`, which is an invalid option. The traverse range is **right exclusive**.

**Describe how you solve it**
change it to `(0, 1)`

- - -

**Environment**

**Screenshots**

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-163) by [Unito](https://www.unito.io/learn-more)

nan-wang:
resolved #814

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -170,11 +170,11 @@ class BaseDriver(metaclass=DriverType):
 
 class BaseRecursiveDriver(BaseDriver):
 
-    def __init__(self, depth_range: Tuple[int] = (0, 0), apply_order: str = 'post',
+    def __init__(self, depth_range: Tuple[int] = (0, 1), apply_order: str = 'post',
                  traverse_on: Tuple[str] = ('chunks',), *args, **kwargs):
         """
 
-        :param depth_range: right-exclusive range of the recursion depth, (0,0) for root-level only
+        :param depth_range: right-exclusive range of the recursion depth, (0, 1) for root-level only
         :param apply_order: the traverse and apply order. if 'post' then first traverse then call apply, if 'pre' then first apply then traverse
         :param args:
         :param kwargs:
@@ -226,8 +226,10 @@ class BaseRecursiveDriver(BaseDriver):
             """
             if _docs:
                 for d in _docs:
+                    # check if apply to next level
                     if d.level_depth < self._depth_end:
                         post_traverse(getattr(d, traverse_on), traverse_on, d)
+                    # check if apply to the current level
                     if self.is_apply and (d.level_depth >= self._depth_start):
                         self._apply(d, context_doc, traverse_on, *args, **kwargs)
 
@@ -242,9 +244,11 @@ class BaseRecursiveDriver(BaseDriver):
                     self._apply_all(_docs, context_doc, traverse_on, *args, **kwargs)
 
                 for d in _docs:
+                    # check if apply on the current level
                     if self.is_apply and d.level_depth >= self._depth_start:
                         self._apply(d, context_doc, traverse_on, *args, **kwargs)
-                    if d.level_depth < self._depth_end:
+                    # check if apply to the next level
+                    if (d.level_depth + 1) < self._depth_end:
                         pre_traverse(getattr(d, traverse_on), traverse_on, d)
 
         if self.recursion_order == 'post':
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -245,10 +245,10 @@ class BaseRecursiveDriver(BaseDriver):
 
                 for d in _docs:
                     # check if apply on the current level
-                    if self.is_apply and d.level_depth >= self._depth_start:
+                    if self.is_apply and self._depth_start <= d.level_depth < self._depth_end:
                         self._apply(d, context_doc, traverse_on, *args, **kwargs)
                     # check if apply to the next level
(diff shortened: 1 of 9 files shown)


jina/drivers/__init__.py
```diff
@@ -242,9 +244,11 @@ def pre_traverse(_docs, traverse_on, context_doc=None):
                     self._apply_all(_docs, context_doc, traverse_on, *args, **kwargs)
 
                 for d in _docs:
+                    # check if apply on the current level
                     if self.is_apply and d.level_depth >= self._depth_start:
                         self._apply(d, context_doc, traverse_on, *args, **kwargs)
-                    if d.level_depth < self._depth_end:
+                    # check if apply to the next level
+                    if (d.level_depth + 1) < self._depth_end:
```
JoanFM, line 251: what if we are traversing on matches? next level is not ensured right?

hanxiao: > `(d.level_depth + 1)`

+1, -1, is a pretty bad way to handle boundary. Can you explain why do you need this change?

hanxiao: if you look at the changes of this PR, it seems that you `(d.level_depth + 1)` but also changes all right value `+1`, i.e. from `[1, 1]` to `[1,2]`. **Note, these two changes cancel each other out**, it means there must be something else causing your bug.

nan-wang: the following test will fail because the `_apply` will run for `self._depth_end` in the case of `pre_traverse`

```
FAILED tests/unit/drivers/test_reduce_all_driver.py::ReduceAllDriverTestCase::test_merge_chunks_with_different_modality
FAILED tests/unit/drivers/querylang/test_querylang_drivers.py::QueryLangTestCase::test_filter_ql_modality
FAILED tests/unit/flow/test_flow.py::FlowTestCase::test_flow_with_modalitys_simple
```

nan-wang: > what if we are traversing on matches? next level is not ensured right?

Nope, the following codes tested.
```
In [7]: d = jina_pb2.Document()
In [8]: getattr(d, 'matches')
Out[8]: []
In [9]: m = d.matches.add()
In [10]: getattr(m, 'matches')
Out[10]: []
```

hanxiao: okay noted, then let's keep it in this way. `pre_traverse` and `post_traverse` work differently that's fine.

JoanFM: Well, this is expected, but what if `doc` actually has `matches` and we are traversing on them for this level? We are skipping under the assumption that next iteration will be in a different level right?

JoanFM: > > what if we are traversing on matches? next level is not ensured right?
> 
> Nope, the following codes tested.
> 
> ```
> In [7]: d = jina_pb2.Document()
> In [8]: getattr(d, 'matches')
> Out[8]: []
> In [9]: m = d.matches.add()
> In [10]: getattr(m, 'matches')
> Out[10]: []
> ```

'd' and `m` have no matches when the getattr is called

nan-wang:
> if you look at the changes of this PR, it seems that you (d.level_depth + 1) but also you give all right values +1, i.e. from [1, 1] to [1,2]. Note, these two changes cancel each other out! This means there must be something else causing your bug.

First of all, `(d.level_depth + 1)` is only for the `_pre_traverse`. The `_pre_traverse` used to work well with `(1, 1)`. After changing the default value, it stops working due to the boundary issue.

hanxiao:
what are all these changes for? if they work fine in unit test previously, why do we have to change them?

image

nan-wang:
> what are all these changes for? if they work fine in unit test previously, why do we have to change them?
> 
> image

In https://github.com/jina-ai/jina/blob/master/jina/drivers/__init__.py#L177, we states that `depth_range` is right exclusive. This is only true when we set `apply_order='post'`. For `apply_order='pre'`, it is right inclusive.

hanxiao:
okay getting clear after a call, please proceed @nan-wang

jina/drivers/__init__.py
nan-wang: > > `(d.level_depth + 1)`
> 
> +1, -1, is a pretty bad way to handle boundary. Can you explain why do you need this change?

Good point. I've made another try.

nan-wang: > Well, this is expected, but what if `doc` actually has `matches` and we are traversing on them for this level? We are skipping under the assumption that next iteration will be in a different level right?

@JoanFM I'm not sure I've got your point. Note that all the `matches` of the `doc` should have the same `level_depth` as the `doc`. Is it related to your concern?

```
doc: level_depth=k
  |-- matches: level_depth=k
  |        |-- matches: level_depth=k
  |        |-- matches: level_depth=k
  |                |-- matches: level_depth=k
  |                |-- matches: level_depth=k
  |
  |-- matches: level_depth=k
```

JoanFM: that's why I do not see while we have to avoid doing `pre_traverse` checking a level that may not apply. But anyway I am sure you had a proper analysis of the issue.

hanxiao:
all good well done!

PR #830, opened 2020-08-14
refactor(driver): QueryLangDriver is now QuerysetReader

hanxiao:
As an example, `VectorSearchDriver` is now a mixin of `QuerySetReader` and `BaseSearchDriver`


jina/drivers/__init__.py
```diff
@@ -54,6 +52,46 @@ def arg_wrapper(self, *args, **kwargs):
     return arg_wrapper
 
 
+class QuerysetReader:
```
JoanFM, line 55: should we call it `QuerySetReader` with capital `S`?

PR #839, opened 2020-08-17
fix: fit transform should not fit at encode time plus refactor

JoanFM:
**Changes introduced**
- All numeric encoders fit in the structure of FitTransformEncoders.
- Fit should not be used when encoding.
- When encoding, model should be loaded from file

**TODO**
- [x] Add and fix tests


JoanFM:
@JamesTang-jinaai, please review since changes in code you added

PR #840, opened 2020-08-17
chore(readme): add new example

alexcg1:
Added my [new example](https://github.com/jina-ai/examples/tree/master/my-first-jina-app)


JoanFM:
We have to keep into account to revisit this docs when we revisit the examples after depth level rebuild.

Southpark has evolved so that we do not show the ugly `SEP` everywhere for instance

alexcg1:
Nice to hear the <SEP>'s are gone!

PR #851, opened 2020-08-19
feat(client): add level depth param to _generate

JoanFM:
**Design changes proposal**
Digging into the concept of `level_depth` has lead me to some thoughts about our current design related to `recursivity`.

- I think `level_depth` concept should only be used for `chunks` and not `matches`. Why? I think a `match.level_depth` should be forced by the `level_depth` of the underlying `document` it is referring to. (At the end a `match` is just another "view" of a `document`).

- The main benefit of this approach, is that `level_depth` in the output can make `conceptual` sense to the user. 
See for instance the test `tests/integration/level_depth/test_search_different_depths.py` introduced in this PR. It would be inconsistent to see a match representing the `doc` with `id = 1` to have a different depth level than the depth level of the actual `indexed doc with id = 1`. Having inconsistencies it is dangerous since it would make it hard to evolve and it is really error prone.

- I think that although it may make sense in the future to `recursively` iterate on `matches`, we currently do not have a case or examples using this (and I can't see any in the near future). It is currently adding noise and makes the code more complicated and less mantainable. I do suggest in this PR to remove `matches` recursivity and simply apply the driver to `matches` when traverse_on `matches` is added.

**Changes introduced**

- Added the option to add `level_depth` at _generate in client. Would it make sense to just allow this when `search` mode?

- Extensive testing of `_generate` 

- Change design on depth level traversal

- Need to introduce new Ranking Driver? 

**TODO**
- Adapt `urbandict` example (to do in another PR)


jina/drivers/rank.py
```diff
@@ -46,11 +46,10 @@ def _apply_all(self, docs: 'jina_pb2.Document', context_doc: 'jina_pb2.Document'
         match_idx = []
         query_chunk_meta = {}
         match_chunk_meta = {}
-        for c in docs:
-            for match in c.matches:
-                match_idx.append((match.parent_id, match.id, c.id, match.score.value))
-                query_chunk_meta[c.id] = pb_obj2dict(c, self.exec.required_keys)
-                match_chunk_meta[match.id] = pb_obj2dict(match, self.exec.required_keys)
+        for match in docs:
```
JoanFM, line 49: I had to make this change @nan-wang @hanxiao, 

in the case of the added example:

When we are traversing on `matches`, all these `docs` will be matches from some `previous` indexer. Therefore they are the matches that we need to sort.

I am sure I am breaking some case here, but it seems to me that this has to be reviewed to cover all cases, or we are maybe needing a different Driver and not every case can be handled by this one.

jina/drivers/__init__.py
```diff
@@ -319,7 +308,7 @@ def pre_traverse(_docs, traverse_on, context_doc=None):
             _traverse(docs, 'chunks')
         if 'matches' in self.traverse_fields:
             for d in docs:
-                _traverse(d.matches, 'matches')
+                _traverse(d.matches, 'matches', context_doc=d)
```
JoanFM, line 311: when ranking, `d` should be the context_doc?

jina/drivers/__init__.py
```diff
@@ -242,10 +231,10 @@ def __init__(self, depth_range: Tuple[int] = (0, 1), apply_order: str = 'post',
         if isinstance(traverse_on, str):
             traverse_on = (traverse_on,)
         self.traverse_fields = set(traverse_on)
-        if apply_order in {'post', 'pre'}:
```
JoanFM, line 245: right now `pre` is not used anywhere and it just adds noise to the code, it is easy to add when required

tests/integration/level_depth/test_search_different_depths.py
```diff
@@ -0,0 +1,56 @@
+import os
+from jina.flow import Flow
+from tests import JinaTestCase
+
+cur_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+class SearchDifferentLevelDepthTestCase(JinaTestCase):
```
JoanFM, line 8: this is the `test` that expresses the requirement of this `feature`

jina/drivers/rank.py
JoanFM: Done!

nan-wang (changes requested):
For adding the new inputs, you would also need to adapt the `set_client_cli_parser` accordingly.

tests/unit/drivers/querylang/test_querylang_drivers.py
```diff
@@ -12,16 +11,10 @@ def random_docs(num_docs):
         d.uri = 'doc://'
         for m in range(10):
             dm = d.matches.add()
-            dm.text = 'match to hello world'
+            dm.text = 'match to hllo world'
```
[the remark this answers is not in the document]
JoanFM: I think this is on purpose so that is filtered out

nan-wang:
With this PR, we introduce a new argument `level_depth` to the `search` api in the `Flow`. The new argument is used to specify on the `level_depth` of the query document.

As for the retrieved `Document` in the `matches` field, we don't require they have the same of `level_depth` as the query `Document`.

hanxiao:
finally, i'm reviewing this PR

hanxiao:
> * I think `level_depth` concept should only be used for `chunks` and not `matches`. Why? I think a `match.level_depth` should be forced by the `level_depth` of the underlying `document` it is referring to. (At the end a `match` is just another "view" of a `document`).

make sense, I will

hanxiao:
well, i have gone through this PR. @JoanFM made a good point. To that, I will use `depth` to represent `chunks` and use `degree` to represent `matches`. This corresponds to the `depth` and `width` of the following illustration:

image

but i would make some changes on the following points

- keep recursive on matches
- keep `pre` recursion in general, it is used in segmenter, please read here https://hanxiao.io/2020/08/28/What-s-New-in-Jina-v0-5/#recursive-drivers-work-as-convolutiondeconvolution @nan-wang @JoanFM 

I will work on this branch in the next hours.

hanxiao (changes requested):
but i would make some changes on the following points

- keep recursive on matches
- keep `pre` recursion in general, it is used in segmenter, please read here https://hanxiao.io/2020/08/28/What-s-New-in-Jina-v0-5/#recursive-drivers-work-as-convolutiondeconvolution @nan-wang @JoanFM

hanxiao:
based on the last comment, I'm reverting `_traverse_apply` function.

jina/drivers/rank.py
```diff
@@ -54,15 +71,76 @@ def _apply_all(self, docs: 'jina_pb2.Document', context_doc: 'jina_pb2.Document'
...
+        matches: indexed full documents (documents of level depth 0).
+    `
+    Using this Driver before querying a Binary Index with full binary document data can be very useful to implement a search system.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.recursion_order = 'post'
+
+    def _apply_all(self, context_doc: 'jina_pb2.Document', *args, **kwargs):
```
hanxiao, line 113: this has an inconsistent signature as `_apply_all` of the parent class

JoanFM: I did that on purpose because although not best practice if used with named parameters it would work. Since the matched documents would come from the context doc, I thought having a single point of entry of info would make the code less error prone and more self explainable.

hanxiao (approved):

PR #859, opened 2020-08-20
fix: add ignore pattern to import class

hanxiao:


tests/integration/test_helloworld.py
```diff
@@ -105,7 +105,7 @@ def test_download_proxy(self):
         # head check
         req = urllib.request.Request(args.index_data_url, method="HEAD")
         response = urllib.request.urlopen(req, timeout=5)
-        self.assertEqual(response.status, 200)
+        assert response.status == 200
 
 
 if __name__ == '__main__':
```
JoanFM, line 111: we can remove this here

tests/unit/drivers/querylang/test_querylang_reader.py
```diff
@@ -65,7 +65,7 @@ def validate2(req):
 
     def test_querlang_driver(self):
         qld2 = dummyDriver(arg1='world')
-        self.assertEqual(qld2.arg1, 'world')
+        assert qld2.arg1 == 'world'
 
 
 if __name__ == '__main__':
```
JoanFM, line 71: and here

hanxiao: okay seems that my regex missed that

hanxiao:
horray!

PR #865, opened 2020-08-22
feat(flow): add visualize method to Flow class

kaushikb11:
## Description

The motivation for this PR is to able to visualize the Flow Object using the Flow API similar to the Graph being created in the Jina Dashboard.

The Graph would be a static representation of the underlying flow graph with Pods as nodes and the edges as the dependencies. I believe this could be helpful for understanding pods dependencies and possibly debugging & this could be further extended to custom Node attributes(shape, color) depending on the Executor it's inheriting from as well as the Pod status.

### Example

Let's consider the Flower Search Example from the Jina [Examples](https://github.com/jina-ai/examples/tree/master/flower-search) Repo:

#### Index Flow YAML Spec

flow-viz

#### Index Flow Graphviz 

flow-index-v1 gv

This Graphviz object will be rendered inline if called from an IPython notebook, otherwise, it will be rendered in a new window.  If a `filename` is provided, the object will not be rendered and instead saved to the location specified.

Ref: I had discovered this feature while evaluating [Prefect](https://docs.prefect.io/core/advanced_tutorials/visualization.html) Workflow Engine a few months back.

### DIFF
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -6,7 +6,7 @@
 #
 # AVAILABLE TAGS
 #
-# nlp, cv, audio, numeric, index, http, sse, framework, perf, network
+# nlp, cv, audio, numeric, index, http, sse, framework, perf, network, viz
 #
 # REMARKS ON TAGS
 # 1. Try to reuse the existing tags if possible.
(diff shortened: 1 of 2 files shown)


kaushikb11:
@hanxiao Thanks for sharing this! Didn't know about Mermaid, the interface seems really cool and helpful for building Flows/Diagrams for multiple use cases.

hanxiao:
@kaushikb11 We finally decided to add it back. Now it is there in the master
image

kaushikb11:
Great work @theUnkownName on the PR! 🥳

PR #870, opened 2020-08-24
fix: go back to old sorting

JoanFM:
**Changes introduced**
Until a new way of Ranking is introduced, for some cases as `faiss indexer` where a flat document structure is indexed and queried, we need to fully go back to sorting the elements by distance. 

Currently we cannot assume that a Ranker will do the sorting afterwards.

Linked to https://github.com/jina-ai/examples/pull/166


nan-wang (approved):
LGTM👍

Issue #873, opened 2020-08-25
Problem when pickling indexer


JoanFM:
**Describe the bug**
Related to #867

When indexing data twice, the `indexer` does not properly close the index files

**Steps to reproduce**

Please run this using `jina version 0.4.1`
[867.tar.gz](https://github.com/jina-ai/jina/files/5124878/867.tar.gz)

Download and untar the file here,

Run twice `python app.py -t index`

and then when you do 
`python app.py -t query`

An error occurs because `index` vectors where not properly flushed and closed

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-194) by [Unito](https://www.unito.io/learn-more)

JoanFM:
It seems that the problem is that the only `peas` that properly `terminate` their executors are `HeadPea` and `TailPea` leaving the index in an inconsistent state

JoanFM:
At the second saving time, the indexer does not manage to pickle itself:

Here I print the differences of attributes I see the first time I index compared to the second. The first thing that strikes is the appearance of `int2ext_key` that may be two big to pickle?

```
[...]

Issue #867, opened 2020-08-24
Stress Test in CICD with big datasets


jmferrete:
**Describe your problem**
We are building a semantic text search system, using NumpyIndexer and cosine distance metric. We have been testing Jina with smaller sets for the moment, running in single instances, but our final goal is to run Jira in a distributed system over datasets with billions of documents.

As a first step, and to validate our configuration when querying smaller datasets, we have indexed around **30 million** text documents (simple sentences), using the following settings:

* Prefetch: 200
* Batch size: 1000
* Shards: 10
* Parallel: 10

(29889733 docs :white_check_mark: done in :stopwatch: 6595.7s :racehorse: 4531.8/s)
So the indexing part is fine.

But we are having problems while performing queries over this data. Any query causes an out of memory error at OS level.

We tried running queries on differents environments:

* Machines with up to 240 GB of RAM.
* Decreasing settings down to:
* Prefetch: 10
* Batch size: 50
* Shards: 1
* Parallel: 1

But we are still having the same memory problems.

We would like to get some feedback about the problems we are having.

Dataset details:
Regarding the input data (around 150GB of plain text files), it is spread into multiple files in the `./source/` folder. Each input file has the following structure:
file1

```
{"text": "my sample text", "vector": [ 0.1234567, -0.2345678, 0.3456789...]}
{"text": "my other sample text", "vector": [ 0.4567890, 0.5678901, -0.6789012...]}
...
```

file2

```
{"text": "my sample text", "vector": [ 0.1234567, -0.2345678, 0.3456789...]}
{"text": "my other sample text", "vector": [ 0.4567890, 0.5678901, -0.6789012...]}
...
```

Basically, multiline json files, where each line is a json document. And each document is formed by two fields:

* `text`: The text string (to be added as meta-information during indexing).
* `vector`: A pre-generated [512,] `float32` array generated from the `text` field by using the universal sentence encoder ("https://tfhub.dev/google/universal-sentence-encoder/4")

Once the data is indexed, the resulting compressed dataset is around 56GB.

[Here you can find the setup files tar.gz](https://com-ravenpack-nchan.s3.amazonaws.com/jina-ai-config.tar.gz)

**What is your guess?**
We guess there must be something wrong in our configuration, as we didn’t expect to run into memory problems, given the dataset size and the instances we have used for testing.

**Environment**
We have tried with differents environments, but the common one is this:

```bash
(jina) [ec2-user@ip-10-200-7-102 ~]$ jina --version-full
[...]

PR #874, opened 2020-08-26
fix: add key mappings to post init to avoid serialize

JoanFM:
Related to #873 and #867


hanxiao (approved):

JoanFM:
I am still not sure it solves the issues

PR #901, opened 2020-09-03
refactor: have flow and pods inherit from exit stack

JoanFM:
**Changes introduced**
I think it does not have functional impact. But it feels like is a more correct way to use `ExitStack` in our context.

It does not fix any issue, but it feels like a more correct way to handle the stack of context Managers.

The idea is in ExitStack docstring it says:

```
    """Context manager for dynamic management of a stack of exit callbacks.

    For example:
        with ExitStack() as stack:
            files = [stack.enter_context(open(fname)) for fname in filenames]
            # All opened files will automatically be closed at the end of
            # the with statement, even if attempts to open files later
            # in the list raise an exception.
    """
```

So the idea is that when we use Flow or a Pod as a context manager (ex: `with Flow().add(...) as flow:` ) we are explicitly using the ExitStack in the way it is supposed to. It changes from `has` an ExitStack to `is` an ExitStack logic


jina/peapods/pod.py
```diff
@@ -56,8 +58,7 @@ def close_if_idle(self):
         """Check every second if the pod is in idle, if yes, then close the pod"""
         while True:
             if self.is_idle:
-                self.close()
-                break  # only run once
+                raise PodIdleException('Pod has been found idle, needs to exit its context')
```
JoanFM, line 61: I am not so sure about this, since it runs in a different Thread

hanxiao (approved):

PR #905, opened 2020-09-03
fix: move pea closing from run loop to close as context manager

JoanFM:
**Changes introduced**
Related to #873 and #867 

Right now the closing of the `Pea` is handled inside the `run` loop inside `finally`.

But when Flow is done sending messages, it tries to close its Pods as context manager, and each Pod tries to close its Peas as a context manager.

This can happen while Pea is tearing down itself and closing its files.

Therefore, I propose to move the logic of closing the Pea context and its files and binaries inside the close and __exit__ logic in a  more robust way.

**TODO** (QUESTIONS AND DOUBTS)

- [ ] When the Flow calls exit, and before closing the Pea, we need to ensure that it has handled all the messages or just the last one? Response: I think this is not an issue since Flow waits to have response to start closing everything

### DIFF
--- a/jina/peapods/pea.py
+++ b/jina/peapods/pea.py
@@ -364,12 +364,14 @@ class BasePea(metaclass=PeaMeta):
 
     def close(self) -> None:
         """Gracefully close this pea and release all resources """
-        if self.is_ready.is_set() and hasattr(self, 'ctrl_addr'):
-            send_ctrl_message(self.ctrl_addr, jina_pb2.Request.ControlRequest.TERMINATE,
-                              timeout=self.args.timeout_ctrl)
+
         self.teardown()
-        self.unset_ready()
         self.is_shutdown.set()
+        if self.is_ready.is_set():
+            self.unset_ready()
+            if hasattr(self, 'ctrl_addr'):
+                send_ctrl_message(self.ctrl_addr, jina_pb2.Request.ControlRequest.TERMINATE,
+                                  timeout=self.args.timeout_ctrl)
 
     @property
     def status(self):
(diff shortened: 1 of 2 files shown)


jina/peapods/pea.py
```diff
@@ -365,8 +365,11 @@ def post_init(self):
     def close(self) -> None:
         """Gracefully close this pea and release all resources """
         if self.is_ready.is_set() and hasattr(self, 'ctrl_addr'):
-            return send_ctrl_message(self.ctrl_addr, jina_pb2.Request.ControlRequest.TERMINATE,
-                                     timeout=self.args.timeout_ctrl)
+            send_ctrl_message(self.ctrl_addr, jina_pb2.Request.ControlRequest.TERMINATE,
```
JoanFM, line 368: does it make sense sending it before or after tearing down? maybe the order should be reversed

rutujasurve94: Yes, I think it makes sense to send TERMINATE request after teardown is complete

hanxiao:
my general recommendation is to have the following understanding before touching the eventloop of the Pea:
- How zmq socket works
- How zmq socket and context lifecycle works
- How multiprocessing/threading works
- Is zmq socket and context thread-safe/cross-multiprocess?
- How TERMINATE signal works
- How ContainerPea and RemotePea depends on BasePea
- What is the lifecycle of a Pea

jina/peapods/pea.py
hanxiao: better not change it until you fully understand the current eventloop with the knowledge I pointed here https://github.com/jina-ai/jina/pull/905#issuecomment-686640530

PR #907, opened 2020-09-04
fix(peapod): fix eventloop and shutdown logic

hanxiao:


jina/peapods/zmq.py
```diff
@@ -313,6 +313,8 @@ def register_pollin(self):
     def close(self):
         """Close all sockets and shutdown the ZMQ context associated to this `Zmqlet`. """
         if not self.is_closed:
+            # wait until the close signal is received
+            time.sleep(.01)
```
JoanFM, line 317: isn't there a better way to ensure this?

hanxiao: nah, not for now.

PR #909, opened 2020-09-04
feat(driver): add VectorFillDriver

hanxiao:


jina/drivers/search.py
```diff
@@ -49,6 +51,19 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs):
             del docs[j]
 
 
+class VectorFillDriver(QuerySetReader, BaseSearchDriver):
+    """ Fill in the embedding by their id
+    """
+
+    def __init__(self, executor: str = None, method: str = 'query_by_id', *args, **kwargs):
```
JoanFM, line 58: would it make sense to change `query_by_id` name to `get_embedding_by_id`? Since `query` returns such different things it can be less confusing

PR #913, opened 2020-09-06
feat(driver): add fill_embedding in search driver

hanxiao:


tests/unit/test_adjacency.py
```diff
@@ -31,6 +31,19 @@ def test_high_order_matches():
     shutil.rmtree('test-index-file', ignore_errors=False, onerror=None)
 
 
+def test_high_order_matches_integrated():
```
JoanFM, line 34: Can this be moved to integration test?

PR #914, opened 2020-09-07
Feat device map 896

xiongma:
1. add device map at Pod level


xiongma:
there are some conflicts, I will pull the master and push again.

PR #918, opened 2020-09-07
test: add tests with executor that sleeps on save

JoanFM:
**Changes introduced**
Simple test that spins a Pod with an Executor that `sleeps` when pickling. 

Now it passes, I want to ensure that it does not pass previous to #907 

Confirmed! This test does not work on previous versions!

### DIFF
--- a/tests/integration/issues/867/test_close_executor.py
+++ b/tests/integration/issues/867/test_close_executor.py
@@ -1,0 +1,48 @@
+import time
+import os
+from jina.executors import BaseExecutor
+from jina.flow import Flow
+
+cur_dir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


hanxiao (approved):

tests/integration/issues/867/test_close_executor.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __getstate__(self):
+        d = super().__getstate__()
+        time.sleep(2)
+        d['test'] = 10
+        return d
+
+
+def test_close_and_load_executor():
+    with Flow().add(uses='!SlowSaveExecutor').build() as f:
```
hanxiao, line 39:
```suggestion
    with Flow().add(uses='SlowSaveExecutor').build() as f:
```
dont think `!` is necessary if none args is followed

hanxiao:
resolves #867

Issue #891, opened 2020-09-02
Enable post-retrieval scoring of document-match pairs


maximilianwerk:
**Describe the feature**

Currently jina only supports ranking matches based on the scores, that the retrieval step provides. Adding the possibility to add more query <> match metrics in order to fine-tune the ranking is needed. Possible applications are a simple edit distance or complex deep learning scoring techniques as BERT.

Furthermore, it might be necessary, to add a `WeightedRanker`, which takes multiple scores encoded in the `score.operands` field and combines them to one top-level score per match.

**Proposal**

* Add a driver for unpacking the document and the matches (named `ContentMatchDriver`).
* Add a executor interface for scoring the document with the matches (named `ContentMatcher`). This should be configurable to either overwrite the match score or add a score in the `score.operands` field.
* Add a concrete implementation of the `ContentMatcher` in the form of a simple Levenshtein distance in the hub (named `LevenshteinMatcher`).
* Add a concrete implementation of the `ContentMatcher` in the form of a BERT scoring in the hub (named `BertMatcher´).

Adding a `WeightedRanker` as a consecutive step might be necessary. This could be a simple linear-combination of the existing scores or something like lambda-mart in the long run. Anyhow, I would rather add this as a consecutive task, to not overload this issue.

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-204) by [Unito](https://www.unito.io/learn-more)

JoanFM:
Important to keep in mind, that Rankers should be chainable to allow different phases of ranking

PR #919, opened 2020-09-07
feat: added Matches2DocRankDriver

maximilianwerk:
This is a suggested implementation for #891. Missing:

- [x]  tests for the `Matches2DocRankDriver`
- [x]  concrete demo-implementations for the `Match2DocRanker`

Open questions: 

- Rename `Match2DocRanker` to `Match2DocScorer`, since effectively only scores are provided?
- Fix allowed adjacency depth to 0, meaning only the direct matches of a document are re-scored?
- Any obvious design error when thinking about the neural search use case?

Thanks a lot for your feedback already!

### DIFF
--- a/jina/drivers/rank.py
+++ b/jina/drivers/rank.py
@@ -187,12 +187,11 @@ class Matches2DocRankDriver(BaseRankDriver):
 
     def _sort_matches_in_place(self, context_doc, match_scores):
         sorted_scores = self._sort(match_scores)
-
+        old_matches = {match.id: match for match in context_doc.matches}
         context_doc.ClearField('matches')
         for match_id, score in sorted_scores:
             new_match = context_doc.matches.add()
-            new_match.id = int(match_id)
-            new_match.score.ref_id = context_doc.id
+            new_match.CopyFrom(old_matches[match_id])
             new_match.score.value = score
             new_match.score.op_name = exec.__class__.__name__
 
--- a/jina/drivers/rank.py
+++ b/jina/drivers/rank.py
@@ -148,3 +148,55 @@ class CollectMatches2DocRankDriver(BaseRankDriver):
                 r.score.ref_id = context_doc.id  # label the score is computed against doc
                 r.score.value = score
                 r.score.op_name = exec.__class__.__name__
+
+
+class Matches2DocRankDriver(BaseRankDriver):
+    """ This driver is intended to only resort the given matches on the 0 level granularity for a document.
+    It gets the scores from a Ranking Executor, which does only change the scores of matches.
+    Afterwards, the Matches2DocRankDriver resorts all matches for a document.
+    Input-Output ::
+        Input:
+        document: {granularity: 0, adjacency: k}
+            |- matches: {granularity: 0, adjacency: k-1}
+        Output:
+        document: {granularity: 0, adjacency: k}
+            |- matches: {granularity: 0, adjacency: k-1} (Sorted according to scores from Ranker Executor)
+    """
+
+    def __init__(self, reverse=False, *args, **kwargs):
+        super().__init__(recur_on='matches', *args, **kwargs)
+        self.recursion_order = 'post'
+        self.reverse = reverse
+
+    def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2.Document', *args, **kwargs):
+        """ Call executer for score and sort afterwards here. """
+
+        # if at the top-level already, no need to aggregate further
+        if context_doc is None:
+            return
+
+        query_meta = pb_obj2dict(context_doc, self.exec.required_keys)
+        old_match_scores = {match.id: match.score.value for match in docs}
+        match_meta = {pb_obj2dict(match, self.exec.required_keys) for match in docs}
+
+        # if there are no matches, no need to sort them
+        if not old_match_scores:
+            return
+
+        new_match_scores = self.exec.score(query_meta, old_match_scores, match_meta)
+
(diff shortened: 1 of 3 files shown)


jina/drivers/rank.py
```diff
@@ -148,3 +148,55 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2
                 r.score.ref_id = context_doc.id  # label the score is computed against doc
                 r.score.value = score
                 r.score.op_name = exec.__class__.__name__
+
+
+class Matches2DocRankDriver(BaseRankDriver):
```
JoanFM, line 153: I like this naming, but for now to highlight the difference of `Chunk2DocRankerDriver` I may name it `Doc2DocRankerDriver`? In a future PR I would totally refactor names, what do you think?

maximilianwerk: But this ranker explicitly only goes from matches to doc. So why make the naming more general then it is?

jina/drivers/rank.py
```diff
@@ -148,3 +148,55 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2
...
+    """ This driver is intended to only resort the given matches on the 0 level granularity for a document.
+    It gets the scores from a Ranking Executor, which does only change the scores of matches.
+    Afterwards, the Matches2DocRankDriver resorts all matches for a document.
+    Input-Output ::
+        Input:
+        document: {granularity: 0, adjacency: k}
+            |- matches: {granularity: 0, adjacency: k-1}
```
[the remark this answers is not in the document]
maximilianwerk: :+1:

JoanFM:
It would be cool to have a test showing the capacity?

jina/drivers/rank.py
```diff
@@ -148,3 +148,53 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2
...
+            return
+
+        new_match_scores = self.exec.score(query_meta, old_match_scores, match_meta)
+        self._sort_matches_in_place(context_doc, new_match_scores)
+
+    def _sort_matches_in_place(self, context_doc, match_scores):
+        sorted_scores = self._sort(match_scores)
+
+        context_doc.ClearField('matches')
+        for match_id, score in sorted_scores:
```
JoanFM, line 192: I think we may need to take care of this, we should keep the content that `matches` carry with them? They may actually carry payload with them that need to be kept

JoanFM: Not sure how to go about this though

tests/unit/drivers/test_matches2doc_rank_drivers.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+
+from jina.drivers.rank import Matches2DocRankDriver
+from jina.executors.rankers import Chunk2DocRanker
+
+from jina.proto import jina_pb2
+
+
+class MockAbsoluteLengthRanker(Chunk2DocRanker):
```
JoanFM, line 9: Do we want it to use Chunk2DocRanker or simple BaseRanker?

jina/drivers/rank.py
maximilianwerk: :+1:

tests/unit/drivers/test_matches2doc_rank_drivers.py
maximilianwerk: Apparently, we want to use `Match2DocRanker`

jina/drivers/rank.py
JoanFM: good solution!

tests/unit/drivers/test_matches2doc_rank_drivers.py
JoanFM: yes, but interface will be exactly as BaseRanker right? u can add an extra inheritance layer if you want, but I do believe that Chunk2DocRanker (although it could somehow work) is missleading.

JoanFM:
LGTM!

tests/unit/drivers/test_matches2doc_rank_drivers.py
nan-wang: Yes, `BaseRanker` is enough

nan-wang (changes requested):

nan-wang (dismissed):
LGTM👍

tests/unit/drivers/test_matches2doc_rank_drivers.py
maximilianwerk: But the interface tells you what exactly are the to-expected arguments. When I want to implement a Ranker in the hub, that works well with the `Matches2DocRankDriver`, it makes the life much easier, if there is an explicit interface, that I have to implement, instead of guessing the right function arguments.

JoanFM: @maximilianwerk do u mean is better to add `Match2DocRanker`? For me it is even better than `BaseRanker`, I am just saying that `Chunk2DocRanker` would be missleading.

JoanFM: And that's what you put now so good for me

maximilianwerk: I fixed it to `Match2DocRanker` and made a proper documentation of how to use it.

maximilianwerk: please :+1: once again

hanxiao (approved):

PR #920, opened 2020-09-07
feat(cli): add daemon as argument of pea

hanxiao:


jina/docker/hubio.py
```diff
@@ -208,8 +208,12 @@ def build(self) -> Dict:
                 if self.args.test_uses:
                     try:
                         from jina.flow import Flow
-                        with Flow().add(uses=image.tags[0]):
-                            pass
+                        if self.args.daemon:
```
JoanFM, line 211: `with Flow().add(uses=image.tags[0], daemon=self.args.daemon):` should suffice?

hanxiao: oh, looks like our parser is able to recognize `daemon=False` as not pass anything to the docker, good

PR #922, opened 2020-09-08
fix(api): change uses to yaml_path

alasdairtran:
A follow-up on commit a53185d5bb20d84b26dd5fd6b01e61f8b9b33c6d which forgets to change `uses` to `yaml_path` in one place.

### DIFF
--- a/jina/main/api.py
+++ b/jina/main/api.py
@@ -80,7 +80,7 @@ def flow(args):
     """Start a Flow from a YAML file or a docker image"""
     from ..flow import Flow
     if args.yaml_path:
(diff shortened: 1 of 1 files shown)


JoanFM:
Hey @alasdairtran ,

Thank you very much for your contribution. I was not aware we missed so many points when migrating! I wonder if you have had any error when using `jina` because of this?

JoanFM (approved):

hanxiao:
btw, @alasdairtran thanks for reminding, and I decide to go for `uses` in Flow API as well, see #925

Issue #885, opened 2020-08-30
🚨 BREAKING CHANGES 0.5.0 -> 0.6.0


hanxiao:
This issue is created to keep track breaking changes between `0.5.0` and `0.6.0` (scheduled release on Oct. 1, 2020)

## Breaking Changes & Migration Guide
| PR | Changes | Rolled Out? | Override |
|----|----| --- | --- |
| #889  | Drivers  <ul><li>`level_depth` -> `granularity` </li><li> add a new field `adjacency` to represent previous `level_depth` for `matches`. Note that `granularity` and `adjacency` are _orthogonal_, represent the depth and width of a recursive doc structure. </li><li> `depth_range` -> `recurring_range`, covering both `granularity` and `adjacency`, depending on what you want to traverse</li></ul> | 0.5.1 ✅ | |
| #893 | Drivers <ul><li> `traverse_on` -> `recur_on`</li><li> `recurring_range -> recur_range`</li></ul>  | 0.5.2 ✅ |  #889|
| #925 | Flow CLI `--yaml-path` -> `--uses` | 0.5.3 ✅ | |
| #925 | Pod CLI <ul><li>`--uses-reducing` -> `--uses-after`</li><li>  add `--uses-before`</li></ul> | 0.5.3 ✅ | |
| #923 | Driver args <ul><li>`recur_on` is removed</li><li>`recur_range` is now decoupled into two orthogonal directions: `granularity_range` and `adjacency_range`</li><li>`recur_on='chunks'; recur_range=[0, k]` -> `granularity_range=[0, k], adjacency_range=[0,0]`</li><li>`recur_on='matches'; recur_range=[0, k]` -> `granularity_range=[0, 0], adjacency_range=[0,k]`</li><li>one can also fix one direction and traverse on the other, e.g. `granularity_range=[m, m], adjacency_range=[0,k]` will traverse on up to `k`-order matches at granularity `m`</li></ul> | 0.5.4 ✅ | #893 |
| #933 | Driver args <ul><li>`recur_on` is added back</li><li>`granularity_range` and `adjacency_range` are now right-inclusive and have the same behavior. They are now truly orthogonal.</li><li>When traversing on `chunks` the starting `match` depth `x` is `adjacency_range=(x, _)`</li><li>When traversing on `matches` the starting `chunk` depth `x` is `granulartiy_range=(x, _)` | 0.5.5 ✅| #923 |

[...]

PR #923, opened 2020-09-08
[P.O.C] feat: add adjacency range

JoanFM:
**Changes proposed**
- Generalize better recursion by providing two ranges (one for depth and the other for adjacency). Also remove recur_on since information would be redundant

- In SearchRequest (assume every document has as much as a chunk), when looping on adjacency make sure that you work on the depth level of the starting depth range.

**TODO**
Adapt drivers interfaces, tests, etc ... to work witht his


maximilianwerk (commented):
I like, the it makes the recursion much more explicit. Anyhow, I am not sure, what graph model we will have behind or result structure. If we have a tree-like structure (e.g. query->chunks and query -> matches and matches->matches-chunks are the edges), we need to take care of the order of walking depending on post/pre `apply-oder`. On the other hand, that would not make much sense in the image retrieval scenario, where we definitely want query-chunk -> match-chunk edges. So perhaps keeping the order of depth/adjacency traversal configurable is the most viable solution for now.

jina/drivers/__init__.py
```diff
@@ -215,21 +219,21 @@ def __getstate__(self) -> Dict[str, Any]:
 
 class BaseRecursiveDriver(BaseDriver):
 
-    def __init__(self, recur_range: Tuple[int] = (0, 1), apply_order: str = 'post',
-                 recur_on: Tuple[str] = ('chunks',), *args, **kwargs):
+    def __init__(self, recur_depth_range: Tuple[int, int] = (0, 1), recur_adjacency_range: Tuple[int, int] = (0, 0),
```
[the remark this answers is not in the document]
JoanFM: This I put to keep exactly the default `chunks` recur_on (I translate `recur_on` = 'chunks'' to `recur_depth_range: Tuple[int, int] = (0, 1)`. Not that I have a strong opinion on that, but to keep the changes at minimmum

jina/drivers/__init__.py
```diff
@@ -303,14 +307,26 @@ def pre_traverse(_docs, recur_on, context_doc=None):
         else:
             raise ValueError(f'{self.recursion_order}')
 
-        if 'chunks' in self.traverse_fields:
+        if (self._depth_start < self._depth_end) or \
```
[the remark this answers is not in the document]
JoanFM: There is currently no executor that receives a complete `document`. The `driver` right now cares about protobuf making `executors` blind to the `protobuf` structure, so this has to be taken care at driver level (At least under current design)

jina/drivers/__init__.py
```diff
@@ -262,7 +266,7 @@ def __call__(self, *args, **kwargs):
     def _traverse_apply(self, docs, *args, **kwargs):
         """often useful when you delete a recursive structure """
 
-        def post_traverse(_docs, recur_on, context_doc=None):
+        def post_traverse(_docs, recur_on, context_doc=None, depth_name='granularity', recur_start=0, recur_end=1):
```
maximilianwerk, line 269: Please update the docstring. Please put function arguments on new lines (line length = 115 is not `PEP8` conformant). I know we don't enforce `PEP8` too much until now, but I'd love to start a bit. Or, at some point in time, we just go to black.

Apart from the: Great refactoring of this function!

jina/drivers/__init__.py
```diff
@@ -303,14 +307,20 @@ def pre_traverse(_docs, recur_on, context_doc=None):
         else:
             raise ValueError(f'{self.recursion_order}')
 
-        if 'chunks' in self.traverse_fields:
-            depth_name = 'granularity'
-            _traverse(docs, 'chunks')
+        if (self._depth_start < self._depth_end) or \
+                (self._depth_start == self._depth_end and self._depth_start > 0):
```
maximilianwerk, line 311: Currently we are catching the `_depth_start=0` case on `apply` and `apply_all` level (at least for the ranker). I like, preventing it beforehand. Would it be a good idea, to remove the `if context_doc is None: return` logic from the driver or is this it better to catch twice?

maximilianwerk (commented):
Apart from small comments: Looks good to me.

jina/drivers/__init__.py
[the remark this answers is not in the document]
maximilianwerk: Seeing how often you did `recur_depth_range: [0, 0]`, I still vote for `(0,0)` as the default. ;) Anyhow, since this PR is for getting us running for the Hackaton, keep it with `(0,1)` and we might refactor this (or not) in the big change to come when we restructure the rankers.

jina/drivers/__init__.py
JoanFM: I will do that for now, but yes I would love to have u help me enforce some style, I am not so experienced when it comes to Python standards.

jina/drivers/__init__.py
JoanFM: I honestly was not thinking about this when doing this, here I was mostly focused on translating really explicitly the old content. I would like to remove but I would prefer to have it in another PR if possible

jina/drivers/__init__.py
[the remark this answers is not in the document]
maximilianwerk: I still believe, that the `pre` iteration often requires to first propagate along the adjacency level and only afterwards along the chunk level. Anyhow, I am OK with waiting for the first use case to appear before making it flexible.

maximilianwerk:
Don't forget to add the changes to the breaking changes issue.

jina/drivers/__init__.py
```diff
@@ -215,21 +219,21 @@ def __getstate__(self) -> Dict[str, Any]:
 
 class BaseRecursiveDriver(BaseDriver):
 
-    def __init__(self, recur_range: Tuple[int] = (0, 1), apply_order: str = 'post',
-                 recur_on: Tuple[str] = ('chunks',), *args, **kwargs):
+    def __init__(self, recur_depth_range: Tuple[int, int] = (0, 1), recur_adjacency_range: Tuple[int, int] = (0, 0),
+                 apply_order: str = 'post', *args, **kwargs):
```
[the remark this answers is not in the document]
nan-wang: The only thing I'm not into this style is that we follow this rule even for the functions with a few arguments. Another option would be that we only do this for the multi-line argument list.

jina/drivers/__init__.py
```diff
@@ -215,21 +215,25 @@ def __getstate__(self) -> Dict[str, Any]:
...
         super().__init__(*args, **kwargs)
-        self._depth_start = recur_range[0]
-        self._depth_end = recur_range[1]
-        if isinstance(recur_on, str):
-            recur_on = (recur_on,)
-        self.traverse_fields = set(recur_on)
+        self._depth_start = recur_depth_range[0]
+        self._depth_end = recur_depth_range[1]
+        self._adjacency_start = recur_adjacency_range[0]
+        self._adjacency_end = recur_adjacency_range[1]
```
nan-wang, line 236:
```suggestion
        self._depth_start, self._depth_end = recur_depth_range
        self._adjacency_start, self._adjacency_end = recur_adjacency_range
```

jina/drivers/__init__.py
nan-wang: I agree with Joan, the changes on the logic need more test. Let's make another PR for that.

jina/drivers/__init__.py
```diff
@@ -215,21 +215,25 @@ def __getstate__(self) -> Dict[str, Any]:
 
 class BaseRecursiveDriver(BaseDriver):
 
-    def __init__(self, recur_range: Tuple[int] = (0, 1), apply_order: str = 'post',
-                 recur_on: Tuple[str] = ('chunks',), *args, **kwargs):
+    def __init__(self,
+                 recur_depth_range: Tuple[int, int] = (0, 1),
+                 recur_adjacency_range: Tuple[int, int] = (0, 0),
```
nan-wang, line 220:
```suggestion
                 recur_adjacency_range: Tuple[int, int] = (0, 1),
```

nan-wang (changes requested):

nan-wang:
generally, LGTM

jina/drivers/__init__.py
JoanFM: My concern with this is that it is like setting `recur_on: (chunks, matches)` by default.

JoanFM: I think adding this a lot of tests will fail

jina/drivers/__init__.py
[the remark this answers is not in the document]
maximilianwerk: The rule of thumb: if the line with a function and all parameters exceeds ~90 characters, put the parameters on a new line. If this new line still exceeds the 90 characters, put each parameter on a separate line. This is also how black does it.

hanxiao (changes requested):
good to go after renaming,

recur_depth_range -> depth_range
recur_adjacency_range -> adjacency_range

hanxiao:
breaking changes are recorded in #885

hanxiao (approved):

PR #928, opened 2020-09-08
fix: typo in top k

JoanFM:


JoanFM:
@deepankarm may it be the source of your problem?

hanxiao:
this is not a typo

hanxiao:
https://docs.jina.ai/api/jina.drivers.html?highlight=querysetreader#jina.drivers.QuerySetReader

hanxiao:
you use `_` then you lose reading from queryset in the payload, this is how `QuerySetReader` is implemented

PR #931, opened 2020-09-09
[Discussion enabler] fix: remove vecs after bulding advanced index

JoanFM:
Would this help mitigate the memory problem that is needed for more advanced indexers? 

I know Python is not the best programming language to handle this things, but when loading an indexer such as FaissIndexer or AnnoyIndexer, are we having at the same time the `np.array of the complete index in memory` together with the complex index object created by the library in use?

### DIFF
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -90,7 +90,9 @@ class BaseNumpyIndexer(BaseVectorIndexer):
         """
         vecs = self.raw_ndarray
         if vecs is not None:
-            return self.build_advanced_index(vecs)
+            index = self.build_advanced_index(vecs)
+            del vecs
+            return index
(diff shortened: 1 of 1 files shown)


hanxiao:
no. 

`del` does not anything in Python

hanxiao:
also, `del vecs` basically disable the usage of `query_by_id`.

JoanFM:
> also, `del vecs` basically disable the usage of `query_by_id`.

I know, I know, I have been giving quite some thought. But I think that to keep this we are making it impossible for `garbage collector` to even consider removing `raw_ndarray`. So in case of AnnoyIndexer, we would keep the `ANN index` as handler and the `raw_data` with no option of the "only needed by `query_by_id`  `raw_ndarray`".

PR #932, opened 2020-09-10
feat(indexer): add shelfpbindexer for data persistance

hanxiao:


tests/integration/issues/github_929/test_shelfindexer.py
```diff
@@ -0,0 +1,38 @@
...
+    with f:
+        f.index(docs)
+
+    m2 = used_memory()
+    d = jina_pb2.Document()
+    d.id = 9523
+
+    def validate(req):
+        m4 = used_memory()
+        print(f'before: {m1}, after index: {m2}, after loading: {m3} after searching {m4}')
```
JoanFM, line 32: Shouldn't there be an assert here? What exactly is being tested?

PR #933, opened 2020-09-10
fix: traversing adjacency graph for multiple query chunks

maximilianwerk:
This is needed for making the lyrics demo work again. Be aware, that this will break tests and is meant as a proposal. The deleting of non-found chunks might be crucial for some application and switching it on/off might be a parameter.


jina/drivers/__init__.py
```diff
@@ -329,10 +329,14 @@ def pre_traverse(_docs, recur_on, context_doc=None,
             for d in docs:
                 # Move to starting depth range
                 # assume that these search docs have a maximum of one chunk per document (common pattern in search)
-                working_doc = d
-                while working_doc.granularity < self._granularity_start:
-                    working_doc = working_doc.chunks[0]
-                _traverse(working_doc.matches, 'matches', None, 'adjacency', self._adjacency_start, self._adjacency_end)
+                working_docs = [d]
+                while working_docs[0].granularity < self._granularity_start:
```
JoanFM, line 333: Smart move

jina/drivers/search.py
```diff
@@ -46,9 +46,9 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs):
             else:
                 miss_idx.append(idx)
 
-        # delete non-existed matches in reverse
-        for j in reversed(miss_idx):
-            del docs[j]
+        # # delete non-existed matches in reverse
```
JoanFM, line 49: Can u explain here how we identified the need to avoid this deletion, it was quite a case, not obvious at all

maximilianwerk: We had a query which was split into sentences (e.g. documents with granularity 1). For some reason they completely disappeared from the data stream. After going backwards, we found out, that something around the `KVSearchDriver` caused. Until ultimately we saw that driver deleting the chunks, since they have no entry in the `BinaryPbIndexer` (they are part of the query). And the driver things, whatever has no value in my database, is not worth returning.

jina/drivers/__init__.py
JoanFM: If we go for this, please remember to adapt the tests I added to have more chunks per query doc.

jina/drivers/search.py
hanxiao: this logic has to be reserved. this handles the miss docs in sharding.

JoanFM: so @maximilianwerk we need to adapt somehow the logic of traversing to adapt then to this without removing this, maybe we need to recover the `recur_on` parameter?

JoanFM: @hanxiao it disappeared because the way we provide now `granularity_range` and `adjacency_range` is limiting. It came from an unwanted loop on `chunks` before `matches`. So we will need to tackle this.

nan-wang: @maximilianwerk It would be great to add a unit test so that we ensure everyone is on the same page.

maximilianwerk: Yes adding tests was anyhow on my agenda.

JoanFM: it is mostly caused by the limitation on how we traverse. In one case we are traversing on chunks before traversing in matches because we `set granularity_range: [1, 1] and adjacency_range: [0 , 1]` to go for the KVSearchDrivrer. So it is a "bug more or less" it first does an iteration on chunks and does not find anything because is a query chunk, therefore it deletes the chunk from the query. So later when we traverse on matches it has no chunk to go down to

hanxiao:
make sure to fix the commit lint error so that CI keeps tracking the health status of this PR

jina/drivers/__init__.py
```diff
@@ -319,20 +323,34 @@ def pre_traverse(_docs, recur_on, context_doc=None,
...
+
+        if 'matches' in self._recur_on:
+            working_docs = docs
+            while working_docs[0].granularity < self._granularity_start:
+                _temp = []
+                for working_doc in working_docs:
+                    _temp.extend(working_doc.chunks)
+                working_docs = _temp
+
+            if 'chunks' in self._recur_on and self._granularity_start == 0 and self._adjacency_start == 0:
```
JoanFM, line 346: Cant this option be handled before filling the working docs?

JoanFM (commented):
I would add some integration test to show that the issue found for multires search is enabled

jina/drivers/__init__.py
maximilianwerk: Great idea. :+1:

maximilianwerk:
I'll finish the tests for showcasing the traversal power tomorrow.

JoanFM (commented):
Just being paranoic, can we have a couple of theses tests run with Rankers? I feel they may have some differences with the others that may deserve testing. And these tests can work as good reference and "documentation" (note the quotes)

jina/resources/executors.requests.CompoundIndexer.yml
```diff
@@ -11,7 +11,7 @@ on:
       with:
         executor: BaseKVIndexer
         granularity_range: [0, 0]
-        adjacency_range: [0, 1]
+        adjacency_range: [1, 1]
```
maximilianwerk, line 14: This is a change which I am not utterly sure of. But it somehow makes sense: once you retrieved the matches, you only collect the metadata for the matches and not of the query itself. BTW: Getting the metadata for the query itself broke the hello-world

jina/resources/helloworld.flow.query.yml
```diff
@@ -6,11 +6,9 @@ with:
 pods:
   encode:
     uses: $RESOURCE_DIR/helloworld.encoder.yml
-    parallel: $PARALLEL
```
maximilianwerk, line 9: Having `parallel` does not make sense, since the encoder saves the used matrix and having two in parallel means, they overwrite each others matrix. It works during querying, but half of the indexed documents will never retrieved, since they are encoded with a thrown away matrix.

jina/resources/helloworld.flow.query.yml
```diff
@@ -6,11 +6,9 @@ with:
...
   index:
     uses: $RESOURCE_DIR/helloworld.indexer.yml
     shards: $SHARDS
     separated_workspace: true
     polling: all
-    uses_after: $RESOURCE_DIR/helloworld.reduce.yml
```
maximilianwerk, line 15: I guess that was an artifact from old!?

JoanFM: reduce is needed if used in parallel right? but it may use `_reduce` anyhow? why was it needed to be removed?

jina/resources/helloworld.flow.query.yml
JoanFM: oh! good catch!

jina/resources/helloworld.flow.query.yml
maximilianwerk: It just does not provide any value and was not working before. So I thought removing complexity from a `hello-world` is a good thing.

JoanFM: All in for reducing complexity, and since u found the problem with parallelism in encoder, it makes it more obvious, reducer only comes to play when parallelism is up

hanxiao: this is required, do not remove

jina/clients/python/io.py
```diff
@@ -88,5 +88,5 @@ def input_numpy(array: 'np.ndarray', axis: int = 0, size: int = None, shuffle: b
     for r in array:
         yield r
         d += 1
-        if size is not None and d > size:
+        if size is not None and d >= size:
```
maximilianwerk, line 91: Otherwise 129 results where returned, when size is 128.

jina/resources/helloworld.reduce.yml
```diff
@@ -20,6 +20,6 @@ requests:
           start: 0
           end: 50
           granularity_range: [0, 0]
-          adjacency_range: [0, 1]
+          adjacency_range: [0, 0]
```
maximilianwerk, line 23: New ranges are so much better than the old ones. This slice says: get me the top 50 in the first dimension e.g. for the first 50 queries submitted. I like it.

hanxiao (approved):

hanxiao (dismissed):
please append the breaking changes introduced by this PR to here https://github.com/jina-ai/jina/issues/885

tests/unit/drivers/test_recursive_traversal.py
```diff
@@ -0,0 +1,200 @@
...
+    Thus, the remaining document structure should still be intact.
+    In practice this combination should not be chosen, anyhow it nicely demonstrates the behaviour.
+    """
+    docs = build_docs()
+    driver = SliceQL(
+        start=0,
+        end=1,
+        adjacency_range=(0, 1),
+        granularity_range=(0, 0),
+        recur_on=["chunks"]
```
nan-wang, line 95: I've written the following test, which failed unexpectedly.
```
def test_granularity_matches_with_chunks_0_1():
    docs = build_docs()
    driver = SliceQL(
        start=0,
        end=1,
        adjacency_range=(0, 0),  # no apply along the adjacency
        granularity_range=(0, 1),  # only apply to the documents of the granularity=0
        recur_on=["chunks"]
    )
    print(f'len(docs): {len(docs)}')
    driver._traverse_apply(docs)
    assert len(docs) == DOCUMENTS_PER_LEVEL  # This failed unexpectedly. Intuitively, I'm expecting this only apply to the chunks of the documents. The documents have granularity=0. Therefore, the two documents having granularity=0 should stay intact and only their chunks are sliced.
    assert len(docs[0].chunks) == 1  # This is expected. The chunks of the documents is sliced to top 1.
```

tests/unit/drivers/test_recursive_traversal.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    This combination of `range` and `recur_on` will only `_apply_all` to the root node.
+    Thus, the remaining document structure should still be intact.
+    In practice this combination should not be chosen, anyhow it nicely demonstrates the behaviour.
+    """
+    docs = build_docs()
+    driver = SliceQL(
+        start=0,
+        end=1,
+        adjacency_range=(0, 0),
```
nan-wang, line 117: Shall we at least mark the adjaceny_range should start from `1`? `adjacency_range=0` means apply should work on `chunks`.

jina/resources/helloworld.reduce.yml
nan-wang, line 23: This is something related to the unit test I proposed. Setting both ranges to `[0, 0]` means applying to neither granularity nor adjacency. The document should stay intact. But the following test failed.

```
def test_both_with_0_0():
    docs = build_docs()
    driver = SliceQL(
        start=0,
        end=1,
        adjacency_range=(0, 0),
        granularity_range=(0, 0),
        recur_on=["chunks", 'matches']
    )
    print(f'len(docs): {len(docs)}')
    driver._traverse_apply(docs)
    assert len(docs) == DOCUMENTS_PER_LEVEL  # this failed unexpectedly
    assert len(docs[0].chunks) == DOCUMENTS_PER_LEVEL
    assert len(docs[0].chunks[0].matches) == DOCUMENTS_PER_LEVEL
    assert len(docs[0].matches) == DOCUMENTS_PER_LEVEL
    assert len(docs[0].matches[0].chunks) == DOCUMENTS_PER_LEVEL
    assert len(docs[0].matches[0].matches) == DOCUMENTS_PER_LEVEL
    assert len(docs[0].matches[0].matches[0].chunks) == DOCUMENTS_PER_LEVEL
```

nan-wang (changes requested):
Some behaviors are unexpected. At least they are controversial against our definition of right-exclusive ranges

```
        :param granularity_range: right-exclusive range of the recursion depth, (0, 1) for root-level only
        :param adjacency_range: right-exclusive range of the recursion adjacency, (0, 1) for single matches
```

nan-wang:
There are two things to change. @maximilianwerk 

- adapt the docstring in `BaseRecursiveDriver`. The ranges are right-inclusive.
- require `adjacency_range` to start with `1` or at least remark this in the docstring. Otherwise, this leads to confusions and is error-prone. When setting `adjacency: [0, 0], recur_on=['matches', ]`, `apply_all` actually will work on the `chunks`. This means it has the same effect as `granularity: [0, 0], recur_on=['chunks', ]`. Specifically, the following two tests lead to the same results.

```
def test_matches_with_0_0():
    docs = build_docs()
    driver = SliceQL(
        start=0,
        end=1,
        adjacency_range=(0, 0),
        granularity_range=(0, 0),
        recur_on=['matches', ]
    )
    driver._traverse_apply(docs)
    assert len(docs) == 1

def test_chunks_with_0_0():
    docs = build_docs()
    driver = SliceQL(
        start=0,
        end=1,
        adjacency_range=(0, 0),
        granularity_range=(0, 0),
        recur_on=['chunks', ]
    )
    driver._traverse_apply(docs)
    assert len(docs) == 1
```

jina/drivers/__init__.py
```diff
@@ -216,20 +216,24 @@ def __getstate__(self) -> Dict[str, Any]:
 class BaseRecursiveDriver(BaseDriver):
 
     def __init__(self,
+                 recur_on: Union[str, Tuple[str, str]] = ('chunks', 'matches'),
```
nan-wang, line 219: I'd prefer to rename chunks to documents because the apply() function take effects on the top level documents as well. When granularity=0, the documents are **NOT** chunks.

nan-wang (changes requested):
Minor naming issue.

tests/unit/flow/test_flow_multimode.py
```diff
@@ -45,6 +46,13 @@ def encode(self, data: str, *args, **kwargs) -> 'np.ndarray':
         return np.array(output)
 
 
+@pytest.fixture(autouse=True)
+def run_around_tests():
+    yield
+    rm_files(['vec1.gz', 'vec2.gz', 'chunk1.gz', 'chunk2.gz',
+              'vecidx1.bin', 'vecidx2.bin', 'kvidx1.bin', 'kvidx2.bin'])
+
+
```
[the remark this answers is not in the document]
maximilianwerk: We should get rid of the `rm_files` method, but rather you pytest temporary folders. The `rm_files` wiped my whole home folder, since I used it a bit wrongly... Don't even know, what I did wrong, since it was wiped... :D

jina/drivers/__init__.py
maximilianwerk: But it is still run in the chunk direction. And matches are documents as well. From a data structure chunks and matches are equal (just different name). Thus having the same behavior in the recursion for both makes the idea of "they are orthogonal" even more true. And the documents at depth=0 and adjacency=0 are iterable via the chunk or the matches direction. This also explains the behavior you mentioned above: When `granularity_range = (0,0)` and `adjacency_range = (0,0)` the following think are equivalent: `recur_on=["matches"]`, `recur_on=["chunks"]` and `recur_on=["chunks", "matches]`. I made the last case explicitly equivalent, since a double `_apply_all` for a executor like `Reverse` would mean they are twice reversed and thus the same again. Also it somehow makes sense, that `apply` and `apply_all` is only called once for each document.

maximilianwerk:
Good catch with the right-inclusivity. :+1:

maximilianwerk:
I think what is needed now is a good documentation and visual explanation of how to set the ranges correctly. I'll look into that after the Hackaton.

maximilianwerk:
> please append the breaking changes introduced by this PR to here #885

Done.

nan-wang (approved):

PR #935, opened 2020-09-10
fix: Fix deadlock when __sse_queue__ QueueHandler receives logs

JoanFM:
Linked to #936


tests/unit/flow/test_flow_with_sse.py
```diff
@@ -0,0 +1,9 @@
+from jina.flow import Flow
+
+
+def test_flow_with_sse():
+    f = Flow(logserver=True).\
+        add(uses='BaseExecutor', parallel=1).add(uses='BaseExecutor', parallel=2).build()
+    with f:
```
JoanFM, line 7: this test is blocked forever without the changes

hanxiao:
note that github action may restrict the access to `0.0.0.0:5000`, you may want to switch another port, and/or use `127.0.0.1/localhost`

JoanFM:
> note that github action may restrict the access to `0.0.0.0:5000`, you may want to switch another port, and/or use `127.0.0.1/localhost`

Ritght now the problem in local is that by using `pytest-repeat` when repeating this test more than once there is still a `deadlock`,

hanxiao:
i think it's Flask problem, terminating Flask programmatically is not straightforward

tests/unit/flow/test_flow_with_sse.py
JoanFM: Still does not solve, when using `pokedex-with-bit` with logserver there is still a deadlock

extra-requirements.txt
```diff
@@ -56,5 +56,6 @@ pytest-subtests:            test
 pytest-xdist==1.34.0:       test
 pytest-timeout:             test
 pytest-cov:                 test
+pytest-repeat               test
```
hanxiao, line 59:
```suggestion
pytest-repeat:               test
```

hanxiao:
btw in `flow/__init__.py` 474, `daemon` should be set to false to ensure the eventloop is ended without interruption
```
self._sse_logger = threading.Thread(name='sentinel-sse-logger',
                                                target=start_sse_logger, daemon=True,
                                                args=(self.args.logserver_config,
                                                      self.yaml_spec))
```

JoanFM:
> btw in `flow/__init__.py` 474, `daemon` should be set to false to ensure the eventloop is ended without interruption
> 
> ```python
> self._sse_logger = threading.Thread(name='sentinel-sse-logger',
>                                                 target=start_sse_logger, daemon=True,
>                                                 args=(self.args.logserver_config,
>                                                       self.yaml_spec))
> ```

Still same error when using pokedex

jina/peapods/pea.py
```diff
@@ -367,7 +370,7 @@ def post_init(self):
         """
         pass
 
-    def close(self) -> None:
+    def send_terminate_signal(self) -> None:
```
hanxiao, line 373: please keep `close()`, as it is the convention `of context_manager`. It also allows people to manually close a CM without using `with`.

Suggest:

```python
def __exit__(self, exc_type, exc_val, exc_tb) -> None:
       self.close()

def close(self) -> None:
        self.send_terminate_signal()
        self.is_shutdown.wait()
             clear_queue()
             self.join()
```

JoanFM: Done!

hanxiao:
looks like CI is taking forever, something is not closed correctly

JoanFM:
> note that github action may restrict the access to `0.0.0.0:5000`, you may want to switch another port, and/or use `127.0.0.1/localhost`

may the timeout in CI be related to this?

JoanFM:
It is `containerPeas` we are not properly closing now

hanxiao (approved):

deepankarm (approved):

PR #937, opened 2020-09-10
Bookkeeping for jina hub push flows

deepankarm:
- Enabled `jina hub build --push` & `jina hub push` flows to update summaries to Mongodb Atlas.
- Currently reads the database credentials from `os.environ`
- Uses the following env variables 
  - db_username
  - db_password
  - db_hostname
  - db_name
  - db_collection

Doing `jina hub build` - builds the docker & writes the summary json to a local file.

Doing `jina hub push` - pushed the docker to docker hub. Tries reading the json from local disk. If unable, builds its summary (without build data). writes the summary to mongo collection.

Doing `jina hub build --push` - builds the docker, pushes it to docker hub, writes the summary from build to mongo collection.

Pending - 
- Enable hub-builder to push the image too
- Enable CD in jina-hub via updated hub-builder
- Unit tests for DB related modules

### DIFF
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -58,4 +58,4 @@ pytest-timeout:             test
 pytest-cov:                 test
 flaky:                      test
 mock:                       test
-pymongo[srv]:               hub
+pymongo[srv]:               hub, devel
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -58,3 +58,4 @@ pytest-timeout:             test
 pytest-cov:                 test
 flaky:                      test
 mock:                       test
+pymongo[srv]:               hub
--- a/jina/docker/checker.py
+++ b/jina/docker/checker.py
@@ -5,6 +5,7 @@ import os
 import re
 import unicodedata
 
+import tempfile
 from pkg_resources import resource_stream
 
 from ..helper import yaml
@@ -74,6 +75,9 @@ def get_exist_path(directory, s):
     r = os.path.join(directory, s)
     if os.path.exists(r):
         return r
+
+def get_summary_path(image_name: str):
+    return os.path.join(tempfile.gettempdir(), image_name.replace('/', '_'), 'summary.json')
 
 
 def is_error_message(s):
--- a/jina/docker/checker.py
+++ b/jina/docker/checker.py
@@ -81,9 +81,8 @@ def is_error_message(s):
 
 def db_env_variables_set():
     """ Checks if any of the db env variables are not set """
-    none_exists = None in [
-        os.environ.get('db_username', None), os.environ.get('db_password', None),
-        os.environ.get('db_hostname', None), os.environ.get('db_name', None),
-        os.environ.get('db_collection', None)
-    ]
-    return not none_exists
+    keys = ['JINA_DB_HOSTNAME', 'JINA_DB_USERNAME', 'JINA_DB_PASSWORD', 'JINA_DB_NAME', 'JINA_DB_COLLECTION']
+    for k in keys:
+        if k not in os.environ:
+            return False
+    return True
--- a/jina/docker/checker.py
+++ b/jina/docker/checker.py
@@ -77,7 +77,7 @@ def get_exist_path(directory, s):
         return r
 
 def get_summary_path(image_name: str):
-    return os.path.join(tempfile.gettempdir(), image_name.replace('/', '_'), 'summary.json')
+    return os.path.join(tempfile.gettempdir(), image_name.replace('/', '_') + '_summary.json')
 
 
 def is_error_message(s):
--- a/jina/docker/checker.py
+++ b/jina/docker/checker.py
@@ -77,3 +77,13 @@ def get_exist_path(directory, s):
 
 def is_error_message(s):
     return re.search(excepts_regex, s, re.IGNORECASE | re.UNICODE) is not None
+
+
+def db_env_variables_set():
+    """ Checks if any of the db env variables are not set """
+    none_exists = None in [
+        os.environ.get('db_username', None), os.environ.get('db_password', None),
+        os.environ.get('db_hostname', None), os.environ.get('db_name', None),
+        os.environ.get('db_collection', None)
+    ]
+    return not none_exists
--- a/jina/docker/checker.py
+++ b/jina/docker/checker.py
@@ -5,8 +5,9 @@ import os
 import re
 import unicodedata
 
-from ..helper import yaml
 from pkg_resources import resource_stream
+
+from ..helper import yaml
 
 image_tag_regex = r'^hub.[a-zA-Z_$][a-zA-Z_\s\-\.$0-9]*$'
 required = {'name', 'description'}
@@ -79,10 +80,7 @@ def is_error_message(s):
     return re.search(excepts_regex, s, re.IGNORECASE | re.UNICODE) is not None
 
 
-def db_env_variables_set():
+def is_db_envs_set():
     """ Checks if any of the db env variables are not set """
     keys = ['JINA_DB_HOSTNAME', 'JINA_DB_USERNAME', 'JINA_DB_PASSWORD', 'JINA_DB_NAME', 'JINA_DB_COLLECTION']
-    for k in keys:
-        if k not in os.environ:
-            return False
-    return True
+    return all(k in os.environ for k in keys)
--- a/jina/docker/database.py
+++ b/jina/docker/database.py
@@ -9,7 +9,7 @@ class MongoDBHandler:
     """
     Mongodb Handler to connect to the database & insert documents in the collection
     """
-    def __init__(self, hostname: str, username: int, password: str,
+    def __init__(self, hostname: str, username: str, password: str,
                  database_name: str, collection_name: str):
         self.logger = get_logger(self.__class__.__name__)
         self.hostname = hostname
--- a/jina/docker/database.py
+++ b/jina/docker/database.py
@@ -1,0 +1,56 @@
+
+import pymongo
+
+from ..excepts import MongoDBException
+from ..logging import get_logger
+
+
+class MongoDBHandler():
+    """
+    Mongodb Handler to connect to the database & insert documents in the collection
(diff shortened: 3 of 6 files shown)


jina/docker/hubio.py
```diff
@@ -73,7 +74,46 @@ def new(self):
             self.logger.info('nothing is created, bye!')
 
     def push(self, name: str = None, readme_path: str = None):
-        """A wrapper of docker push """
+        """ A wrapper of docker push """
+        is_build_success, is_push_success = True, False
+        try:
+            # self._push(name, readme_path)
```
rutujasurve94, line 80: Do we intend on calling `_push` from here? It could return `True` on successful push and could be used for setting `is_push_success`
It could be renamed differently for more clarity.

Could you document what the args `name` and `readme_path` refer to?

deepankarm: Yes. That was an unintended comment. Fixed the pytest failure too.

jina/docker/hubio.py
```diff
@@ -73,7 +74,46 @@ def new(self):
...
+            
+            result = {
+                'name': name,
+                'path': getattr(self.args, 'path', ''),
+                'details': _details,
+                'last_build_time': get_now_timestamp(),
+                'build_duration': '',
+                'is_build_success': is_build_success,
+                'is_push_success': is_push_success,
+                'build_logs': [],
```
rutujasurve94, line 108: In case build logs are not found in the temporary location - does it mean that the build was never triggered or it was an unsuccessful build? 
In the first case, we'd have to trigger the build right?
Also, `build_duration` is an empty field. In this case `is_build_success` and `is_push_success` can never be true (it makes sense to set it false here?)

jina/docker/hubio.py
```diff
@@ -73,7 +74,46 @@ def new(self):
...
-        """A wrapper of docker push """
+        """ A wrapper of docker push """
+        is_build_success, is_push_success = True, False
+        try:
+            # self._push(name, readme_path)
+            is_push_success = True
+        except Exception:
+            self.logger.error(f'can not push to the registry')
+        
+        file_path = f'/tmp/{self.args.name.replace("/", "_")}.json'
```
rutujasurve94, line 85: Could we add a documentation comment for better readability that this temporary file path refers to the `build log` location?

In case this path doesn't exist, do we have a build re-try and push re-try attempt logic?

deepankarm: For the cases, where this file doesn't exist, that would mean, manual deletion of the file. We don't want to rebuild during a `push` flow. Hence, fallback logic is to write whatever information we can have during a `push` only flow. I'll add a comment which says reading from tmp build log file.
Please suggest, if you feel we can have a better file to write to, rather than `/tmp/{name}.json`

jina/docker/hubio.py
deepankarm: In `push` logic, we have a `_check_docker_image` method, which checks if the image was built locally or not. 

For the case where the tmp file doesn't exist 

- The user might not have built the image. During `_push`, we'll check if it is a valid image in the local machine. If `_push` raises an exception, we'll never write to the db

- Even though the build was successful, the user might have deleted the file manually or for some reason, the file didn't get created. In this case, due to lack of build information, we'll write `build_duration` as empty, `is_build_success` is assumed to be True, `is_push_success` depends on `_push` call.

jina/docker/database.py
```diff
@@ -0,0 +1,56 @@
...
+    @property
+    def database(self):
+        return self.client[self.database_name]
+    
+    @property
+    def collection(self):
+        return self.database[self.collection_name]
+    
+    def insert(self, document: str):
+        result = self.collection.insert_one(document)
```
rutujasurve94, line 48: Could you do also add a primary key extracting fields from the document as:
`id = jina-<version>-<executor-family>-<executor-type>-<timestamp-now>`
 `self.collection.insert_one( _id: id, document)`

This would help querying the documents from the collection better in the future.

extra-requirements.txt
```diff
@@ -58,3 +58,4 @@ pytest-timeout:             test
 pytest-cov:                 test
 flaky:                      test
 mock:                       test
+pymongo[srv]:               hub
```
hanxiao, line 61:
```suggestion
pymongo[srv]:               hub, devel
```

jina/docker/checker.py
```diff
@@ -77,3 +77,13 @@ def get_exist_path(directory, s):
...
     return re.search(excepts_regex, s, re.IGNORECASE | re.UNICODE) is not None
+
+
+def db_env_variables_set():
+    """ Checks if any of the db env variables are not set """
+    none_exists = None in [
+        os.environ.get('db_username', None), os.environ.get('db_password', None),
+        os.environ.get('db_hostname', None), os.environ.get('db_name', None),
+        os.environ.get('db_collection', None)
+    ]
```
hanxiao, line 88:
```suggestion
     keys = ['db_username', 'db_password', ...]
     for k in keys:
        if k not in os.environ:
           return False
```

also I suggest `db_username`, `db_password` to have all-caps letter and with `JINA_` prefix

jina/docker/database.py
```diff
@@ -0,0 +1,56 @@
+
+import pymongo
+
+from ..excepts import MongoDBException
+from ..logging import get_logger
+
+
+class MongoDBHandler():
```
hanxiao, line 8:
```suggestion
class MongoDBHandler:
```

jina/docker/database.py
```diff
@@ -0,0 +1,56 @@
...
+    
+    def connect(self):
+        try:
+            self.client = pymongo.MongoClient(self.connection_string)
+            self.client.admin.command('ismaster')
+            self.logger.info('Successfully connected to the database')
+        except pymongo.errors.ConnectionFailure:
+            raise MongoDBException("Database server is not available")
+        except pymongo.errors.ConfigurationError:
+            raise MongoDBException("Credentials passed are not correct!")
```
hanxiao, line 34:
```suggestion
        except pymongo.errors.ConnectionFailure:
            raise MongoDBException('Database server is not available')
        except pymongo.errors.ConfigurationError:
            raise MongoDBException('Credentials passed are not correct!')
```

extra-requirements.txt
deepankarm: Added devel.
What is use of `devel`?

jina/docker/checker.py
deepankarm: Changed this.

jina/docker/database.py
deepankarm: Fixed!

jina/docker/database.py
deepankarm: Fixed!

jina/docker/database.py
```diff
@@ -0,0 +1,56 @@
+
+import pymongo
+
+from ..excepts import MongoDBException
+from ..logging import get_logger
+
+
+class MongoDBHandler:
```
hanxiao, line 8: this could be a useful `KVIndexer` in `jina-hub` btw

hanxiao:
im working on this branch to fix some logic

jina/docker/database.py
```diff
@@ -9,7 +9,7 @@ class MongoDBHandler:
     """
     Mongodb Handler to connect to the database & insert documents in the collection
     """
-    def __init__(self, hostname: str, username: int, password: str,
+    def __init__(self, hostname: str, username: str, password: str,
```
deepankarm, line 12: This was silly from me 👎

jina/docker/hubio.py
hanxiao: From L91 to 107, I think we don't need it. This surpasses our (implicit) validation in the build process, and allows user to upload meta info of an **arbitrary image**, which is not preferred.

jina/docker/hubio.py
```diff
@@ -66,14 +67,52 @@ def new(self):
...
+        try:
+            self._push_docker_hub(name, readme_path)
+            is_push_success = True
+        except:
+            self.logger.error(f'can not push to the registry')
+
+        file_path = get_summary_path(name)
+        if os.path.isfile(file_path):
+            with open(file_path) as f:
+                result = json.load(f)
```
hanxiao, line 89: Given our discussion in Slack, the order here has to be changed.

1. check if we have "certificate" aka that temp file, if yes, continue, otherwise, log.error and return
2. push to docker
3. write to mongodb

hanxiao:
@deepankarm you can continue on this. I'm not working on this in short term

jina/docker/hubio.py
```diff
@@ -225,7 +264,7 @@ def build(self) -> Dict:
                         self.logger.error(f'can not use it in the Flow')
                         is_build_success = False
 
-                if self.args.push:
+                if is_build_success and self.args.push:
                     try:
                         self.push(image.tags[0], self.readme_path)
```
deepankarm, line 269: For the `build --push` case, `self.push()` function will search for temp file to be present in the disk and hence fail. We cannot write to the file before push, as we keep push info in summary too. 
We need to call `self._push_docker_hub()` here & do the file write + db write in the same function.

jina/docker/checker.py
```diff
@@ -74,6 +76,15 @@ def get_exist_path(directory, s):
     if os.path.exists(r):
         return r
 
+def get_summary_path(image_name: str):
+    return os.path.join(tempfile.gettempdir(), image_name.replace('/', '_'), 'summary.json')
```
deepankarm, line 80: This will fail to create the directory `image_name.replace('/', '_')` and hence cannot find the file to write. Fixing this (avoid joining part 2 & 3)

hanxiao: oh, right, you can also do `f"image_name.replace('/', '_')-summary.json"`

deepankarm: This is done.

jina/docker/hubio.py
deepankarm: This is done & tested.

hanxiao (approved):

PR #938, opened 2020-09-11
feat: add capacity to lookup and use querylang with tags in protobuf …

JoanFM:
**Changes introduced**
Enable QueryLangDrivers to work for keys and values from `tags` in protobuf `Document`

### DIFF
--- a/jina/drivers/querylang/queryset/dunderkey.py
+++ b/jina/drivers/querylang/queryset/dunderkey.py
@@ -32,7 +32,9 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 ## This module deals with code regarding handling the double
 ## underscore separated keys
-from typing import Tuple, Union
+from typing import Tuple
+from google.protobuf.struct_pb2 import Struct
+from google.protobuf import json_format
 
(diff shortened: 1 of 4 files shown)


jina/drivers/querylang/queryset/dunderkey.py
```diff
@@ -118,6 +124,8 @@ def dunder_get(_dict, key):
 
     if isinstance(parts[0], int):
         result = guard_iter(_dict)[parts[0]]
+    elif isinstance(_dict, dict):
+        result = _dict[parts[0]]
     else:
         result = getattr(_dict, parts[0])
```
hanxiao, line 130: whats wrong with accessing Pb via `getattr`? why converting to `dict` is necessary?

hanxiao: aha, okay, this is only for `Stuct`

JoanFM: yes Struct has a weird representation:

https://github.com/protocolbuffers/protobuf/pull/6451

It may be still possible (did not manage myself) to access it but with real weird syntax, (like adding __fields ...)

hanxiao (approved):

PR #941, opened 2020-09-11
fix: set request top_k, needed to be read by jinabox

### LINKED ISSUE #942: Not able to interpet top_k introduced by jinabox (And I guess also not by python client)
**Describe the bug**
Currently `jina` is not able to pass and interpret the `top_k` parameter sent by `jinabox`. 

PR #941 fixes the issue 

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-229) by [Unito](https://www.unito.io/learn-more)

JoanFM:
Not working because SearchRequest does not have `top_k`, so how do we interpeet the message from `jinabox`?

Fixes #942

### DIFF
--- a/jina/clients/python/request.py
+++ b/jina/clients/python/request.py
@@ -58,7 +58,7 @@ def _add_document(request: 'jina_pb2.Request', content: Union['jina_pb2.Document
 
 def _generate(data: Union[Iterator['jina_pb2.Document'], Iterator[bytes], Iterator['np.ndarray'], Iterator[str], 'np.ndarray'],
               batch_size: int = 0, first_doc_id: int = 0, first_request_id: int = 0,
-              random_doc_id: bool = False, mode: ClientMode = ClientMode.INDEX,
+              random_doc_id: bool = False, mode: ClientMode = ClientMode.INDEX, top_k: int = 50,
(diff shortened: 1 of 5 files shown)


jina/clients/python/request.py
```diff
@@ -87,6 +87,16 @@ def _generate(data: Union[Iterator['jina_pb2.Document'], Iterator[bytes], Iterat
                 queryset = [queryset]
             req.queryset.extend(queryset)
 
+        if top_k and mode == ClientMode.SEARCH:
+            if top_k <= 0:
+                raise ValueError(f'"top_k: {top_k}" is not a valid number')
```
maximilianwerk, line 92: I am not sure about the context of this exception, but I read it as: The client can define a value, which will make jina throw an exception. Is this exception catches on a per-request basis? If so, everything is fine. If not, a client can crash jina with a bad request. That would be not good.

maximilianwerk:
Just tested on the freshly refactored southpark example and it worked. :+1:

maximilianwerk:
LGTM

maximilianwerk (approved):
LGTM

jina/clients/python/request.py
JoanFM: This works on the client, so jina would, so not necessarily. This exception is taken from an old version where this functionality was present.

PR #944, opened 2020-09-13
fix: add recur_on param back

JoanFM:
Hey @maximilianwerk  may this idea solve the issue?

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -323,6 +323,9 @@ class BaseRecursiveDriver(BaseDriver):
         else:
             raise ValueError(f'{self.recursion_order}')
 
+        if not docs:
+            return
+
         if 'chunks' in self._recur_on:
             working_docs = docs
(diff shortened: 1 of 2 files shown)


jina/drivers/__init__.py
```diff
@@ -336,18 +336,13 @@ def pre_traverse(_docs, recur_on, context_doc=None,
             _traverse(working_docs, 'chunks', None, 'granularity', self._granularity_start, self._granularity_end)
 
         if 'matches' in self._recur_on:
-
-            for d in docs:
```
JoanFM, line 329: I think this change may be dangerous, the recursion was never equal.

maximilianwerk: Hmm but what it really changes is: instead of going deep on every top-level document one-by-one, go deep by all together. In theory there should be no difference. During fixing of the unit-tests I'll see how it hurts.

JoanFM: I just wanted to raise attention

PR #950, opened 2020-09-15
feat(indexer): add memmap numpy indexer

hanxiao:
## Content

This PR makes the first tackle on the memory footprint problem: it improves the current `NumpyIndexer` while keeping the interface the same as previous. Specifically:

- **By default `compress_level=0` means no `gzip` compression.** In practice, we found that the compression isn't worthy: by setting it to `1` it decreases the file size by 10% but increasing the IO overhead by 17x. Note that this overhead sounds scary but is not really observable as Jina's Pods work in parallel & async.
- **By default use `np.memmap` for reading the vector from HDD/SDD.** It gives much less memory footprint and only load data on-demand. It also improves the speed when combining with `@batching`. 
- **By default set `batching=512`** to distance computing functions. This is the empirical size we found good in experiment.

Other changes include `BaseNumpyIndexer` refactoring, `batching` decors, helper functions improvement etc.

## Follow-up

- As the next step, `build_advanced_index` in `BaseNumpyIndexer` has to be refactored as the distance computing functions in `NumpyIndexer` to allow batching and make it "memmap-friendly". This will be then a new PR.
- Advanced indexers that inherit from `BaseNumpyIndexer` should be refactored accordingly to enjoy this benefit.


jina/executors/indexers/vector.py
```diff
@@ -145,12 +144,19 @@ def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> '
...
+            else:
+                self.logger.error(
+                    f'the size of the keys and vectors are inconsistent ({r.shape[0]} != {self._size}), '
+                    f'did you write to this index twice? or did you forget to save indexer?')
 
     @cached_property
     def ext2int_id(self) -> Dict:
         """Convert external ids (random index) to internal ids (0,1,2,3,4,...) """
-        return {k: idx for idx, k in enumerate(self.int2ext_id)}
+        if self.int2ext_id is not None:
```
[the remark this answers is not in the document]
hanxiao: then return `None`, in python no return = `return None`. btw, this logic is the same as the current master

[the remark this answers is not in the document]
JoanFM: then return type `Option[Dict]`?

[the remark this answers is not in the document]
hanxiao: fixed

jina/executors/indexers/vector.py
```diff
@@ -123,78 +130,92 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
...
+
+
+def _cosine(A_norm_ext, B_norm_ext):
+    return A_norm_ext.dot(B_norm_ext).clip(min=0) / 2
 
 
 class NumpyIndexer(BaseNumpyIndexer):
     """An exhaustive vector indexers implemented with numpy and scipy. """
 
+    batch_size = 512
```
JoanFM, line 214: what if we put it as parameter and have it in a `cached_property` so that is not dumped?

hanxiao: `batch_size` is part of `metas` and for `metas`, the priority is 

metas in YAML > class attribute > default_jina_config

hanxiao: btw our docs on `metas` has been broken for quite a while

jina/executors/indexers/vector.py
```diff
@@ -123,78 +130,92 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
...
-        if vecs is None:
-            return None
+        if not (path.exists(self.index_abspath) or self.num_dim or self.dtype):
+            return
 
-        if self.key_bytes and self.key_dtype:
-            self.int2ext_key = np.frombuffer(self.key_bytes, dtype=self.key_dtype)
+        if self.compress_level > 0:
+            return self._load_gzip(self.index_abspath)
+        elif self.size is not None and os.stat(self.index_abspath).st_size:
```
JoanFM, line 148: an extra else: raising an Exception would help to tackle the problem if ever found. Maybe u run it without the proper previous `index` flow or something?

hanxiao: will just return `None`, `get_query_handler` will handle `None`

Issue #952, opened 2020-09-17
bug in traversing on chunks


nan-wang:
**Describe the bug**
In line https://github.com/jina-ai/jina/blob/master/jina/drivers/**init**.py#L329,
when passing a document with no match at one granularity. `_temp.extend(working_doc.matches)` will leads to an empty `working_docs` and accessing `working_docs[0]` results an error. 

```
       if 'chunks' in self._recur_on:
            working_docs = docs
            while working_docs[0].adjacency < self._adjacency_start:
                _temp = []
                for working_doc in working_docs:
                    _temp.extend(working_doc.matches)
                working_docs = _temp
            _traverse(working_docs, 'chunks', None, 'granularity', self._granularity_start, self._granularity_end)

```

**Describe how you solve it**

```
       if 'chunks' in self._recur_on:
            working_docs = docs
            while working_docs[0].adjacency < self._adjacency_start:
                _temp = []
                for working_doc in working_docs:
                    if working_doc.matches:
                        _temp.extend(working_doc.matches)
                if not _temp:
                    break
                working_docs = _temp
            _traverse(working_docs, 'chunks', None, 'granularity', self._granularity_start, self._granularity_end)
```

- - -

**Environment**

**Screenshots**

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-234) by [Unito](https://www.unito.io/learn-more)

maximilianwerk:
Good catch. Should there be a warning logged as well? It would be nice for understanding, when one configured the ranges wrongly. On the other hand in the real world that might spam quite a lot of log messages.

I am not sure, how gracefully we should accept wrong ranges. Can you give an example, where this can happen with correct ranges, but just wrong documents?

maximilianwerk:
Sorry for not spotting this before: it should rather be a `return` then a `break`, since the defined ranges clearly do not match the document structure and thus there should be no more traversal. While accepting non-matching structures, it is not good to do some non-well-defined behavior with the ranges. `granularity_range=(1,1)` and `recur_on=['matches']` should always mean, that matches on the first layer of chunks are touched. If we break, then sometimes matches of the zeroes layer of chunks are touched.

PR #955, opened 2020-09-17
test(drivers): test the recursive traversal

nan-wang:
This PR is to illustrate the problem with the current implementation of `traverse_apply()`. It is related to #952 .

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -351,7 +351,10 @@ class BaseRecursiveDriver(BaseDriver):
                 while working_docs[0].granularity < self._granularity_start:
                     _temp = []
                     for working_doc in working_docs:
-                        _temp.extend(working_doc.chunks)
+                        if working_doc.chunks:
+                            _temp.extend(working_doc.chunks)
+                    if not _temp:
+                        break
                     working_docs = _temp
                 _traverse(working_docs, 'matches', None, 'adjacency', self._adjacency_start, self._adjacency_end)
 
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -331,7 +331,10 @@ class BaseRecursiveDriver(BaseDriver):
             while working_docs[0].adjacency < self._adjacency_start:
                 _temp = []
                 for working_doc in working_docs:
-                    _temp.extend(working_doc.matches)
+                    if working_doc.matches:
+                        _temp.extend(working_doc.matches)
+                if not _temp:
+                    break
                 working_docs = _temp
             _traverse(working_docs, 'chunks', None, 'granularity', self._granularity_start, self._granularity_end)
 
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -331,10 +331,7 @@ class BaseRecursiveDriver(BaseDriver):
             while working_docs[0].adjacency < self._adjacency_start:
                 _temp = []
                 for working_doc in working_docs:
-                    if working_doc.matches:
-                        _temp.extend(working_doc.matches)
-                if not _temp:
-                    break
+                    _temp.extend(working_doc.matches)
                 working_docs = _temp
             _traverse(working_docs, 'chunks', None, 'granularity', self._granularity_start, self._granularity_end)
 
@@ -351,10 +348,7 @@ class BaseRecursiveDriver(BaseDriver):
                 while working_docs[0].granularity < self._granularity_start:
                     _temp = []
                     for working_doc in working_docs:
-                        if working_doc.chunks:
-                            _temp.extend(working_doc.chunks)
-                    if not _temp:
-                        break
+                        _temp.extend(working_doc.chunks)
                     working_docs = _temp
                 _traverse(working_docs, 'matches', None, 'adjacency', self._adjacency_start, self._adjacency_end)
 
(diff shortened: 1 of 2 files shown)


tests/unit/drivers/test_recursive_traversal.py
```diff
@@ -198,3 +198,43 @@ def test_adjacency1_granularity1():
...
+    docs = [doc, ]
+    driver = SliceQL(
+        start=0,
+        end=1,
+        adjacency_range=(1, 1),
+        granularity_range=(1, 2),
+        recur_on=["chunks", ]
+    )
+    driver._traverse_apply(docs)
+    assert len(docs) == 1
```
maximilianwerk, line 215: Because of the upper `doc = docs[0]`, I wondered how this can be 1 and not 2.

jina/drivers/__init__.py
```diff
@@ -331,7 +331,10 @@ def pre_traverse(_docs, recur_on, context_doc=None,
             while working_docs[0].adjacency < self._adjacency_start:
                 _temp = []
                 for working_doc in working_docs:
-                    _temp.extend(working_doc.matches)
+                    if working_doc.matches:
+                        _temp.extend(working_doc.matches)
+                if not _temp:
+                    break
```
maximilianwerk, line 337: Sorry for not catching this beforehand, but this is wrong. If you do this, you will change the chunks of the root document, even thought you define `adjacency_range=(1,1)`, which explicitly states, that the root document's chunks should not be touched.

tests/unit/drivers/test_recursive_traversal.py
```diff
@@ -198,3 +198,43 @@ def test_adjacency1_granularity1():
...
+    assert len(docs[0].chunks) == 1
```
maximilianwerk, line 216: This should not be the case. `adjacency_range=(1,1)` says, that the chunks should only be touched beginning from the first adjacency layer and not from the original document.

tests/unit/drivers/test_recursive_traversal.py
```diff
@@ -198,3 +198,43 @@ def test_adjacency1_granularity1():
...
+    driver = SliceQL(
+        start=0,
+        end=1,
+        adjacency_range=(1, 2),
+        granularity_range=(0, 1),
+        recur_on=["matches", ]
+    )
+    driver._traverse_apply(docs)
+    assert len(docs) == 1
+    assert len(docs[0].matches) == 1
```
maximilianwerk, line 236: In contrast to the upper example: This is correct, since you set `granularity_range=(0,1)`, e.g. touch the first adjacency layer on the lowest chunk layer. Anyhow, the `break` statement in the `_traversal` function is not called, since the `_granularity_start` is `0` and thus there is no need for zooming into the structure.

maximilianwerk (commented):
I'd rather not do anything then doing non-well-defined behavior if document structure and traversal definition do not match.

tests/unit/drivers/test_recursive_traversal.py
nan-wang: I'm a bit confused. This means the setting of `adjacency_range` is coupled with `granularity_range` if though I just want to run either on `chunks` or `matches`.

tests/unit/drivers/test_recursive_traversal.py
nan-wang: `docs=[doc, ]` at line 206

tests/unit/drivers/test_recursive_traversal.py
nan-wang: By this setting, my expectation is that it works on `chunks` from `granularity_range=1` to `2`.

maximilianwerk: But on which `adjacency_depth`? So do you want to touch the chunks of the document with `adjacency_depth=0` or the chunks or the documents with `adjacency_depth=1`? Since we want to provide a possibility to iterate through the chunks of the documents with `adjacency_depth=1`, setting `adjacency_range=(1,1)` is the way to do that. If you now set `adjacency_range=(1,1)` and the chunks of the document with `adjacency_depth=0` are touched, that is confusing.

I think making this less confusing might be in decoupling the parameters further down (with example values):
- `recur_on=['chunks', 'matches']` same meaning as it has now
- `chunks_iteration_adjacency_start_depth=1` start iterating over chunks on the `adjacency_depth=1` and `granularity_depth=0` along the `chunks` edges (only has an effect if `chunks` is in `recur_on`)
- `granularity_range=(0,1)` touch all chunks on the `granularity_depth` zero and one (only has an effect if `chunks` is in `recur_on`)
- `matches_iteration_granularity_start_depth=2` start iterating over matches on the `granularity_depth=2` and `adjacency_depth=0` along the `matches` edges (only has an effect if `matches` is in `recur_on`)
- `adjacency_range=(1,2)` touch all matches on the `adjacency_depth` one and two (only has an effect if `matches` is in `recur_on`)

Beware the difference between iterating and touch in this description: 
- iterating: traverse through the document
- touch: call the `_apply` or `_apply_all` functions on the document

maximilianwerk: When you now take this variables and set `chunks_iteration_adjacency_start_depth=1`, even thought the root document has no matches, what should happen? I think it should (gracefully?) fail.

tests/unit/drivers/test_recursive_traversal.py
nan-wang: Yes, you are right. But still, the current implementation is limited. I've updated the test to illustrate my points.

nan-wang:
@maximilianwerk I think we should solve the problem in a more fundamental way, and therefore reduce this PR to just illustrate the problem.

nan-wang:
accidentally, I've added the submodules to the commit. Reopened in #970

PR #963, opened 2020-09-18
feat: add type annotations in jina executors

rameshwara:
feat: add type annotations in jina executors

### DIFF
--- a/jina/executors/compound.py
+++ b/jina/executors/compound.py
@@ -210,12 +210,12 @@ class CompoundExecutor(BaseExecutor):
         return (self.components and any(c.is_updated for c in self.components)) or self._is_updated
 
     @is_updated.setter
-    def is_updated(self, val: bool) -> bool:
+    def is_updated(self, val: bool):
         """Set :attr:`is_updated` for this :class:`CompoundExecutor`. Note, not to all its components """
         self._is_updated = val
 
     @is_trained.setter
-    def is_trained(self, val: bool) -> bool:
+    def is_trained(self, val: bool):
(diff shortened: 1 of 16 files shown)


rameshwara:
Due to the misplacement of the branch head, the commits made for adding annotations in drivers package were also included here. But all of them have now been reverted and only the type annotations for the executors are included in this pull request.

jina/executors/compound.py
```diff
@@ -210,12 +210,12 @@ def is_updated(self) -> bool:
         return (self.components and any(c.is_updated for c in self.components)) or self._is_updated
 
     @is_updated.setter
-    def is_updated(self, val: bool):
+    def is_updated(self, val: bool) -> bool:
```
JoanFM, line 213: this is not true, it is a setter so it does not have any return value, it returns `None`

jina/executors/compound.py
```diff
@@ -210,12 +210,12 @@ def is_updated(self) -> bool:
...
         """Set :attr:`is_updated` for this :class:`CompoundExecutor`. Note, not to all its components """
         self._is_updated = val
 
     @is_trained.setter
-    def is_trained(self, val: bool):
+    def is_trained(self, val: bool) -> bool:
```
JoanFM, line 218: same here

JoanFM (changes requested):

jina/executors/compound.py
rameshwara: Done

jina/executors/compound.py
rameshwara: Done

JoanFM (approved):

Issue #434, opened 2020-05-20
Add a ranker powered by text matching models


nan-wang:
**Describe the feature**

Add a ranker powered by text matching models 

**Your proposal**
- the ranker takes texts from query chunks and match chunks and call the model to calculate the matching scores

---

**Environment**

**Screenshots**

avhirupc:
Do you mean something like [Deep Rank](https://arxiv.org/pdf/1710.05649.pdf) , which takes in query and document and outputs a relevance score for ranking? 

If not, Isn't this already supported in one of the examples?
* [https://github.com/jina-ai/examples/tree/master/southpark-search](url) where matching between query and document based distilbert embedding is already implemented & ranked based on cosine similarity

celis:
.

celis:
Hi @nan-wang,  can you provide more details about what you call  text matching models?, I see already implemented some tf-idf and bm25 rankers

hanxiao:
i remove this feature from 0.4 as 
- not clear what does this mean
- 0.4 does not plan for this

PR #964, opened 2020-09-18
feat: recover argpartition at numpyindexer query

JoanFM:
**Changes introduced**
Since #434 , we have the capacity to rank documents at the same granularity level, so it is easy for a user to rerank documents after the 'matching' phase of the `VectorIndexer`. 

Therefore I propose to recover back this more efficient way of getting the `top_k` documents although they will not be directly sorted.

**Note**
It should be clarified under #885

### DIFF
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -235,6 +235,10 @@ class NumpyIndexer(BaseNumpyIndexer):
         .. note::
             Metrics other than `cosine` and `euclidean` requires ``scipy`` installed.
 
+            When querying with NumpyIndexer, the top_k closest documents to the embedding query are provided, but these top_k
+            documents are not guarantee to be sorted. If you want them to be sorted by the distance score you can use a SortQL
+            driver afterwards or use a downstream Ranker to fine-tune the ranking of the top_k matched documents.
+
         """
         super().__init__(*args, compress_level=compress_level, **kwargs)
         self.metric = metric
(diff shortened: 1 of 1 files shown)


jina/executors/indexers/vector.py
```diff
@@ -263,7 +267,7 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
         else:
             raise NotImplementedError(f'{self.metric} is not implemented')
 
-        idx = dist.argsort(axis=1)[:, :top_k]
+        idx = np.argpartition(dist, kth=top_k, axis=1)[:, :top_k]
```
maximilianwerk, line 270: You could also first do the `argpartition` and then the `argsort`. This does not break anything, but is still faster, since you only sort the `top_k` and not the whole list.

JoanFM: Smart move... did not occur to me!

JoanFM: or now that we are doing `batching` in the distance computation, we could maybe have some heap sort logic and do the top_k selection on the fly @hanxiao ?

hanxiao: looks good. but by benchmarking, the improvement is minor, 2% on avg.

time is measured as the sum of 5x 100-query
```
old, topk=256
    [7.548 secs]
old, topk=512
    [7.585 secs]
this PR, topk=256
    [7.367 secs]
old, topk=512
    [7.428 secs]
```

maximilianwerk: This also only start to be an issue, when the index becomes really large, not? The speed improvements depends upon the tail of the list and not upon the head.

hanxiao: no significant difference

hanxiao (changes requested):
im going to reject this unless it shows a significant speed advantage. Otherwise, it is really confusing how `top_k` only returns unsorted result. recommend to add an executor-level arg to control this behavior

JoanFM:
Let's close it then until someone raise an issue about it

JoanFM:
Early optimization is the source of all evil! Closing ...

PR #965, opened 2020-09-18
fix: issue with hyphen in random name and raise_error

deepankarm:
This is related to `jina hub build --test-uses`. We add a `random_name` which has `-` instead of `_`, hence we get the below error

`something wrong but it is probably not your fault. ValueError('name: empty-order is invalid, please follow the python variable name conventions')`

Also had an issue in `raises_error` to pop the right argument, when build fails & we raise error.

https://github.com/jina-ai/jina-hub/runs/1133291527?check_suite_focus=true


theUnkownName (approved):
Looks like it's good now!

PR #967, opened 2020-09-18
fix(import): fix module import for different usages

hanxiao:


JoanFM:
I love all these testing. I would suggest though to make more explicit that ImageResizer is just a dummy example to have a more agnostic `DummyExecutor` with some `dummy_helper` to emphasize that these tests are not at all about functionality.

hanxiao:
> I love all these testing. I would suggest though to make more explicit that ImageResizer is just a dummy example to have a more agnostic `DummyExecutor` with some `dummy_helper` to emphasize that these tests are not at all about functionality.

Yep, i renamed it to `DummyImageResizer`. In fact, without this renaming the real `ImageResizer` actually overrides the test, making the integration test fail

PR #968, opened 2020-09-18
test: add test for multitype messages without optimization

JoanFM:
**Changes introduced**
It adds tests for future TDD. Right now if `array_in_pb` is set to False (the default value), our custom protocol (intended to optimize bandwith) cannot handle chunks with multiple types. This PR does add some tests with some of them skipped to showcase the intention of having optimized message passing.


hanxiao:
btw, when we got time, this part is definitely going to be rewritten in C/Cython

JoanFM:
> btw, when we got time, this part is definitely going to be rewritten in C/Cython

It is a draft, by the way, not working yet, and PDFExtractorSegmenter in the hub will be waiting for this to pass

JoanFM:
Isn't there a way by `protobuf` to simply interpret a byte stream as a type of message? or is this a problem because of the `recursive` nature of our document?

jina/peapods/zmq.py
```diff
@@ -713,12 +714,11 @@ def _check_msg_version(msg: 'jina_pb2.Message'):
                                 'the message is probably sent from a very outdated JINA version')
 
 
-def _extract_bytes_from_msg(msg: 'jina_pb2.Message') -> Tuple:
+def _extract_bytes_from_msg(docs: Iterable['jina_pb2.Document']) -> Tuple:
```
maximilianwerk, line 717: That should be now `_extract_bytes_from_docs`, no? Why moving the document extraction outside of this method?

jina/peapods/zmq.py
```diff
@@ -748,20 +748,29 @@ def _extract_bytes_from_msg(msg: 'jina_pb2.Message') -> Tuple:
     return doc_bytes, chunk_bytes, chunk_byte_type
 
 
-def _fill_buffer_to_msg(msg: 'jina_pb2.Message', msg_data: List[bytes], offset: int = 2):
+def _fill_buffer_to_msg(msg_data: List[bytes], docs: Iterable['jina_pb2.Document'], offset: int = 3):
```
maximilianwerk, line 751: Similar `_fill_buffer_to_document`

jina/peapods/zmq.py
maximilianwerk: Ah I see, for testability. :+1:

tests/unit/peapods/test_zmq.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+from jina.peapods.zmq import _extract_bytes_from_msg
+from jina.proto import jina_pb2
+from jina.drivers.helper import array2pb, pb2array
+
+random_np_array = np.random.randint(10, size=(50, 10))
```
maximilianwerk, line 6: Why defining these outside the test? The only reason I see, it that multiple tests use them. Anyhow, there is only one test.

Nevertheless, if they are outside, I'd prefer them being constants, e.g. `RANDOM_NP_ARRAY`, `BUFFER` and `TEXT`

tests/unit/test_quant.py
```diff
@@ -29,36 +28,33 @@ def random_docs():
...
 
+        err = 0
+        for d in req.docs:
+            for c in d.chunks:
+                recv = pb2array(c.embedding)
+                send = np.random.random([embed_dim])
+                err += np.sum(np.abs(recv - send)) / embed_dim
 
-class MyTestCase(JinaTestCase):
+        print(f'reconstruction error: {err / num_docs:.6f}')
```
maximilianwerk, line 42: should this stay?

tests/unit/test_quant.py
```diff
@@ -29,36 +28,33 @@ def random_docs():
...
-    err = 0
-    for d in req.docs:
-        for c in d.chunks:
-            recv = pb2array(c.embedding)
-            send = np.random.random([embed_dim])
-            err += np.sum(np.abs(recv - send)) / embed_dim
-
-    print(f'reconstruction error: {err / num_docs:.6f}')
+def test_quant():
+    def get_output(req):
```
maximilianwerk, line 32: Just out of interest: Why did you make this an inner function instead of a top level function as `random_docs()`?

tests/integration/issues/github_969/test_messages_different_types.py
```diff
@@ -0,0 +1,291 @@
...
+import numpy as np
+from jina.proto import jina_pb2
+from jina.drivers.helper import array2pb, pb2array
+
+random_np_array = np.random.randint(10, size=(50, 10))
+buffer = 'text_buffer'.encode()
+text = 'text_content'
+
+
+@pytest.mark.skip(reason='with optimization, message cannot handle different chunk types in the same message')
```
maximilianwerk, line 12: If you expect an error to happen, I'd rather use `pytest.raises` instead of skipping the test. This guarantees, if we fix the problem, we don't forget to activate the test.

If it is not fixable: remove the test.

tests/integration/issues/github_969/test_messages_different_types.py
```diff
@@ -0,0 +1,291 @@
...
+        chunk2 = doc.chunks[2]
+        assert chunk2.id == 30
+        assert chunk2.buffer == buffer
+
+    with Flow().add(uses='_pass') as f:
+        f.search(input_fn=[input_doc()], output_fn=validate_fn)
+
+
+def test_message_docs_different_chunk_types_without_optimization():
+    def input_doc():
```
maximilianwerk, line 53: Not sure about this, since the build up documents are not always exactly the same, but: Can't we deduplicate a bit here?

maximilianwerk (commented):
I'd suggest some minor style changes. Even thought I looked the first time at this part of the code, I'd say the content looks good.

jina/peapods/zmq.py
JoanFM: for testability yes, but I am with u in changing the name, let's do it

tests/unit/test_quant.py
JoanFM: I would think no, but I want to not affect, and handle this in a different one

tests/unit/test_quant.py
JoanFM: I just like to keep it inside a test to have it explicit that is related to this test.

tests/integration/issues/github_969/test_messages_different_types.py
JoanFM: I think having a raised error and test against it can lead to the impression that the test must fail, this part is a preparation for the future

tests/integration/issues/github_969/test_messages_different_types.py
JoanFM: Personally, in tests I do not care about duplication much, I like them being self-contained

tests/integration/issues/github_969/test_messages_different_types.py
maximilianwerk: :+1:

tests/integration/issues/github_969/test_messages_different_types.py
maximilianwerk: :+1:

tests/unit/test_quant.py
maximilianwerk: Shouldn't the `random_docs` then also be inside? It just feels like it is not consistent, at least for this test.

maximilianwerk (approved):
LGTM

PR #979, opened 2020-09-22
test: proposal for the new flow traversal

maximilianwerk:
This is a new adaption of the traversal tests to do a test-driven development of the new traversal proposal. Furthermore, this allows us to decide whether this is the functionality, we want to achieve.

I choose `traversal_paths=['cmc', 'mcm']` over `traversal_paths=[['c', 'm', 'c'], ['m', 'c', 'm']`, since it easier to read, and especially in the `.yml` files it is much easier to write a list of strings instead of a list of lists of strings.

Furthermore, I added the root identifier `r` to also be able to apply a function to the roots of the trees.

### DIFF
--- a/tests/unit/drivers/test_recursive_traversal_tree.py
+++ b/tests/unit/drivers/test_recursive_traversal_tree.py
@@ -43,7 +43,7 @@ def iterate_build(d, current_granularity, max_granularity, current_adjacency, ma
             iterate_build(dc, dc.granularity, max_granularity, dc.adjacency, max_adjacency)
 
 
-@pytest.mark.skip('this is a test for the proposed new flow design')
+@pytest.mark.skip(reason='this is a test for the proposed new flow design')
 def test_only_root():
     docs = build_docs()
     driver = SliceQL(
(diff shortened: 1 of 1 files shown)


JoanFM:
I love this test @maximilianwerk , really clarifying to what we have been discussing

hanxiao:
i will come back to this after all tests are fixed

tests/unit/drivers/test_recursive_traversal_tree.py
```diff
@@ -0,0 +1,254 @@
...
+    if current_adjacency < max_adjacency:
+        for i in range(DOCUMENTS_PER_LEVEL):
+            dc = d.matches.add()
+            dc.granularity = current_granularity
+            dc.adjacency = current_adjacency + 1
+            dc.id = i
+            iterate_build(dc, dc.granularity, max_granularity, dc.adjacency, max_adjacency)
+
+
+@pytest.mark.skip('this is a test for the proposed new flow design')
```
JoanFM, line 46: pytest.mark.skip requries `reason` as kwargs, so u need to put `reason='this is ...`

JoanFM (changes requested):

hanxiao:
did I miss something here? where is `traversal_paths` come from? Can't see this args from our codebase or in this PR?

hanxiao (approved):

hanxiao (approved):

PR #981, opened 2020-09-23
Feat add mime types

theUnkownName:


tests/unit/drivers/test_segmenter_driver.py
```diff
@@ -24,56 +26,61 @@ def craft(self, text: str, *args, **kwargs) -> List[Dict]:
...
-        assert docs[0].chunks[2].id == 5
-        assert docs[0].chunks[2].parent_id == docs[0].id
-        assert docs[0].chunks[2].blob == array2pb(np.array([2.0, 2.0, 2.0]))
-        assert docs[0].chunks[2].weight == 2
-        assert docs[0].chunks[2].length == 3
-
-        with self.assertRaises(AttributeError) as error:
-            driver._apply(docs[1])
-        assert error.exception.__str__() == '\'Document\' object has no attribute \'non_existing_key\''
+def test_segment_driver():
```
JoanFM, line 37: Can u try to duplicate the test and have one specific to the feature we are enabling?

tests/unit/drivers/test_segmenter_driver.py
```diff
@@ -24,56 +26,61 @@ def craft(self, text: str, *args, **kwargs) -> List[Dict]:
...
+    assert valid_doc.chunks[1].blob == array2pb(np.array([1.0, 1.0, 1.0]))
+    assert valid_doc.chunks[1].weight == 1
+    assert valid_doc.chunks[1].length == 3
+    assert valid_doc.chunks[1].mime_type == "image/png"
+
+    assert valid_doc.chunks[2].id == 5
+    assert valid_doc.chunks[2].parent_id == valid_doc.id
+    assert valid_doc.chunks[2].blob == array2pb(np.array([2.0, 2.0, 2.0]))
+    assert valid_doc.chunks[2].weight == 2
+    assert valid_doc.chunks[2].length == 3
```
JoanFM, line 69: can u add the assertion that the other chunks inherit from their parent document their `mime_type`

JoanFM (changes requested):

tests/unit/drivers/test_segmenter_driver.py
theUnkownName: Sure, 
I'll have one that test that specifies the `mime_type` and one that inherits from the document

maximilianwerk: I think it would be nicer, if the parent document would not have an empty mime-type.

tests/unit/drivers/test_segmenter_driver.py
```diff
@@ -24,56 +26,60 @@ def craft(self, text: str, *args, **kwargs) -> List[Dict]:
...
+
+    invalid_doc = jina_pb2.Document()
+    invalid_doc.id = 2
+    invalid_doc.text = 'invalid'
+    invalid_doc.length = 2
+
+    assert invalid_doc.length == 2
+
+    with pytest.raises(AttributeError):
+        driver._apply(invalid_doc)
```
hanxiao, line 85: what does this test? cant see any meaningful `assert`

maximilianwerk: It checks whether and `AttributeError` is raised during running the code in the `with` block. If no exception is thrown, the test will fail.

This is the pytest way of checking, that exceptions are thrown. Pretty elegant, I'd say.

maximilianwerk: See https://docs.pytest.org/en/reorganize-docs/new-docs/user/pytest_raises.html

tests/unit/drivers/test_segmenter_driver.py
```diff
@@ -60,7 +59,7 @@ def test_segment_driver():
     assert valid_doc.chunks[1].blob == array2pb(np.array([1.0, 1.0, 1.0]))
     assert valid_doc.chunks[1].weight == 1
     assert valid_doc.chunks[1].length == 3
-    assert valid_doc.chunks[1].mime_type == "image/png"
+    assert valid_doc.chunks[1].mime_type == ""
```
maximilianwerk, line 62: Sorry for the confusion, but: Could you make this again "image/png"? I think it demonstrates better how the inheritance of this attribute works.

tests/unit/drivers/test_segmenter_driver.py
theUnkownName: If I switch it back, one of the documents still inherits from their doc parent and the doc parent doesn't have an empty mime-type, maybe that's better?

tests/unit/drivers/test_segmenter_driver.py
theUnkownName: Yeah Max told me about this pytest way yesterday, I think it looks cleaner

maximilianwerk (dismissed):
LGTM If tests run through, please merge

tests/unit/drivers/test_segmenter_driver.py
hanxiao: aha, okay, missing that

tests/unit/drivers/test_segmenter_driver.py
```diff
@@ -24,56 +26,61 @@ def craft(self, text: str, *args, **kwargs) -> List[Dict]:
...
+    assert valid_doc.chunks[0].mime_type == 'text/plain'
+
+    assert valid_doc.chunks[1].id == 4
+    assert valid_doc.chunks[1].parent_id == valid_doc.id
+    assert valid_doc.chunks[1].blob == array2pb(np.array([1.0, 1.0, 1.0]))
+    assert valid_doc.chunks[1].weight == 1
+    assert valid_doc.chunks[1].length == 3
+    assert valid_doc.chunks[1].mime_type == 'image/png'
+
+    assert valid_doc.chunks[2].id == 5
```
JoanFM, line 65: Just assert also the mime type of the 3rd chunk and good to go!!!

theUnkownName: Got it, the last two tests inherits `'image/png'` type from the doc parent

JoanFM (approved):

tests/unit/drivers/test_segmenter_driver.py
maximilianwerk: Yes. I like the final solution

PR #983, opened 2020-09-24
feat: new tree traversal structure

maximilianwerk:
This enables traversal via explicit tree path definition. The following changes happened:

- added tree traversal itself
-- no need for `granulartiy_range`, `adjacency_range`, `recur_on` or `recursion_order` with the new approach
- added feature toggle to switch between old traversal and new traversal
-- toggle is settable globally or on a per-driver level
- hello-world runs with mixed versions: for most parts use the old traversal, for the reduce functions use the new one.
- refactored `ReduceAllDriver` in order to be more explicit about what happens and have no internal state to reduce idling memory foodprint (and be stateless).

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -222,6 +222,7 @@ class BaseRecursiveDriver(BaseDriver):
                  adjacency_range: Tuple[int, int] = (0, 0),
                  apply_order: str = 'post',
                  traversal_paths: List[str] = [],
+                 use_tree_traversal: bool = False,
                  *args,
                  **kwargs):
         """
@@ -238,13 +239,14 @@ class BaseRecursiveDriver(BaseDriver):
         self._recur_on = recur_on
         self._granularity_start, self._granularity_end = granularity_range
         self._adjacency_start, self._adjacency_end = adjacency_range
+        self._traversal_paths = traversal_paths
+        self._use_tree_traversal = os.getenv('JINA_USE_TREE_TRAVERSAL', str(use_tree_traversal)) == "True"
         if apply_order in {'post', 'pre'}:
             self.recursion_order = apply_order
         else:
             raise AttributeError('can only accept oder={"pre", "post"}')
         self._is_apply = True
         self._is_apply_all = True
-        self._is_apply_path = False
 
     def _apply(self, doc: 'jina_pb2.Document', context_doc: 'jina_pb2.Document', field: str, *args, **kwargs):
         """ Apply function works on each doc, one by one, modify the doc in-place
@@ -269,26 +271,19 @@ class BaseRecursiveDriver(BaseDriver):
         self._traverse_apply(self.req.docs, *args, **kwargs)
 
     def _traverse_apply(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
-        if os.getenv('JINA_USE_TREE_TRAVERSAL', 'False') == 'True':
+        if self._use_tree_traversal:
             self._traverse_apply_explicit(docs, *args, **kwargs)
         else:
             self._traverse_apply_recur_on(docs, *args, **kwargs)
 
     def _traverse_apply_explicit(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
-        for path in self.traversal_paths:
+        for path in self._traversal_paths:
             if path[0] == 'r':
                 self._traverse_rec(docs, None, None, [])
             for doc in docs:
                 self._traverse_rec([doc, ], None, None, path)
 
     def _traverse_rec(self, docs, parent_doc, parent_edge_type, path):
-        if self._should_apply_all_pre(path):
-            self._apply_all(docs, parent_doc, parent_edge_type)
-
-        if self._should_apply_pre(path):
-            for doc in docs:
-                self._apply(doc, parent_doc, parent_edge_type)
-
         if path:
             next_edge = path[0]
             for doc in docs:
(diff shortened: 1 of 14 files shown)


jina/drivers/__init__.py
```diff
@@ -266,6 +269,54 @@ def __call__(self, *args, **kwargs):
...
 
     def _traverse_apply(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
+        if os.getenv('JINA_USE_TREE_TRAVERSAL', 'False') == 'True':
+            self._traverse_apply_explicit(docs, *args, **kwargs)
+        else:
+            self._traverse_apply_recur_on(docs, *args, **kwargs)
+
+    def _traverse_apply_explicit(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
+        for path in self.traversal_paths:
+            if path[0] == 'r':
```
JoanFM, line 279: according to this, the type of traversal Paths is a List[List[str]]?

tests/unit/drivers/test_recursive_traversal_tree.py
```diff
@@ -13,6 +13,13 @@
 DOCUMENTS_PER_LEVEL = 2
 
 
+@pytest.fixture(autouse=True)
+def set_environment():
+    os.environ['JINA_USE_TREE_TRAVERSAL'] = 'True'
+    yield
```
JoanFM, line 19: Smart move this one

jina/drivers/__init__.py
JoanFM: No I see in the tests, it is a str as array of chars, forget this comment

JoanFM (commented):
I have a question on how does this way of traversing play along with the possibility to input a query at a given granularity? Is it possible that with this new way of working absolute granularity and adjancency lose all their importance? It is all about granularity and adjacency with respect to the root document right?

jina/drivers/__init__.py
```diff
@@ -266,6 +269,54 @@ def __call__(self, *args, **kwargs):
         self._traverse_apply(self.req.docs, *args, **kwargs)
 
     def _traverse_apply(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
+        if os.getenv('JINA_USE_TREE_TRAVERSAL', 'False') == 'True':
```
JoanFM, line 272: Note for the future, if we have some toggle like this that can affect performance, I would put it as argument so that it can be used on docker containers and Remote Peas.

maximilianwerk: Ah yes, you can even do that in the `__init__` function and "hang in" the right function:
```
if os.getenv("JINA_USE_TREE_TRAVERSAL", "False"):
    self._traverse_apply = self._traverse_apply_explicit
else:
    self._traverse_apply = self._traverse_apply_recur
```

maximilianwerk:
> I have a question on how does this way of traversing play along with the possibility to input a query at a given granularity? Is it possible that with this new way of working absolute granularity and adjancency lose all their importance? It is all about granularity and adjacency with respect to the root document right?

That is true. All traversal is now defined from the root document and this is expected to be at level `0, 0`. Can you elaborate a use-case, where this is not the case? I think we should add this then as a test in order to have a well defined use-case.

It can even be, that a direct match of the `0,0` has the "coordinates" `1,2` (adjacency, granularity). In the `CollectMatches2DocRankDriver` you afterwards bubble up the information from the chunks to have then matches for the root documents which have "coordinates" `1, 1`.

jina/drivers/craft.py
```diff
@@ -45,6 +45,8 @@ class SegmentDriver(CraftDriver):
 
     def __init__(self, first_chunk_id: int = 0, random_chunk_id: bool = True,
                  level_names: List[str] = None, *args, **kwargs):
+        # Important to do the overwrite before calling `super()__init__(...)`
+        kwargs.setdefault('traversal_paths', ['r', 'c'])
```
JoanFM, line 49: isn't it more natural to simply pass this as named argument to the super().__init__()?

jina/drivers/reduce.py
```diff
@@ -82,26 +82,30 @@ class ReduceAllDriver(ReduceDriver):
         It uses the last request as a reference.
     """
 
-    def _apply(self, doc: 'jina_pb2.Document', context_doc: 'jina_pb2.Document', field: str, *args, **kwargs):
-        if doc.id not in self.doc_pointers:
-            self.doc_pointers[doc.id] = doc
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('traversal_paths', ['c'])
+        super().__init__(use_tree_traversal=True, *args, **kwargs)
```
JoanFM, line 87: same here, why do we choose to pass by named argument and another with this `setdefault`?

jina/drivers/craft.py
maximilianwerk: nope, because you want to be able to even overwrite it from the .yml files. This is just setting the defaults for the specific driver.

jina/drivers/reduce.py
maximilianwerk: see above.

jina/drivers/craft.py
maximilianwerk: `setdefault` will only set it, if it is not yet set.

maximilianwerk: Ah wait, you mean I could do:
```
def __init__(self, ..., traversal_paths=['r', 'c'], ...):
    super().__init__(traversal_paths=traversal_paths)
```
Yes that is better, I agree!

JoanFM: ah good point! I just learned something!!!

JoanFM: I did not mean that exactly, but I guess we had a good outcome, so yes it seems good!

JoanFM (approved):

Issue #929, opened 2020-09-08
Horizontal scaling for loading big datasets


jmferrete:
Since the resolution of #867, thanks a lot guys for your great work there, we can now index our test datasets of ~190 GB without problems. But now, we have a new issue when querying the indices, our search process gives timeout errors even with increasing the `timeout_ready` property value to 1 hour.

As we can see here:

```bash
$> PREFETCH=5 PREFETCH_ON_RECV=1 BATCH_SIZE=50 PARALLEL=1 SHARDS=10 python app.py -t query
           JINA@35452[W]:parser Flow can not recognize the following args: ['--read-only', '--port-expose', '45678'], they are ignored. if you are using them 
from a global args (e.g. Flow), then please ignore this message
           JINA@35452[W]:parser FlowPod can not recognize the following args: ['--reducing-uses', '_merge_all'], they are ignored. if you are using them from 
a global args (e.g. Flow), then please ignore this message
           Flow@35452[S]:successfully built Flow from a yaml config
2020-09-07 12:03:41.195354: W tensorflow/stream_executor/platform/default/dso_loader.cc:59] Could not load dynamic library 'libcudart.so.10.1'; dlerror: libcu
dart.so.10.1: cannot open shared object file: No such file or directory
2020-09-07 12:03:41.195392: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
        encoder@35508[I]:post initiating, this may take some time...
        encoder@35508[I]:post initiating, this may take some time takes 0.000 secs
        encoder@35508[S]:successfully built MyQueryEncoder from a yaml config
        encoder@35508[I]:setting up sockets...
        encoder@35508[I]:input tcp://0.0.0.0:48061 (PULL_BIND)   output tcp://0.0.0.0:47091 (PUSH_CONNECT)       control over tcp://0.0.0.0:46769 (PAIR_BIND)
        encoder@35508[S]:ready and listening
 vector_indexer@35512[I]:post initiating, this may take some time...
 vector_indexer@35512[I]:post initiating, this may take some time takes 0.000 secs
 vector_indexer@35512[S]:successfully built BaseExecutor from a yaml config
 vector_indexer@35512[I]:setting up sockets...
[...]

PR #984, opened 2020-09-24
feat: new binarypbindexer based on memmap

hanxiao:
related to #929


jina/executors/indexers/keyvalue.py
```diff
@@ -1,161 +1,72 @@
...
+    def query(self, key: int) -> Optional['jina_pb2.Document']:
 
-        :param obj: an object
-        """
-        for k in keys:
-            self.write_handler.write(k.SerializeToString() + __binary_delimiter__)
-            self._size += 1
+        pos_info = self.query_handler.get(key, None)
+        if pos_info is not None:
+            p, r, l = pos_info
```
JoanFM, line 64: Could it be possible to have more explicative variable names?

jina/executors/indexers/__init__.py
```diff
@@ -2,14 +2,17 @@
...
+from typing import Tuple, Union, List, Iterator, Optional
 
 import numpy as np
 
 from .. import BaseExecutor
 from ..compound import CompoundExecutor
 from ...helper import call_obj_fn, cached_property, get_readable_size
 
+if False:
+    from ...proto import jina_pb2
```
maximilianwerk, line 14: This is added for type hinting, right?

jina/executors/indexers/__init__.py
```diff
@@ -2,14 +2,17 @@
...
+
 
 class BaseIndexer(BaseExecutor):
     """``BaseIndexer`` stores and searches with vectors.
```
maximilianwerk, line 18: When you explicitly take out the vector docstring below, I'd also rewrite this to something like

"base class for storing and searching any kind of data strcuture"

jina/executors/indexers/__init__.py
```diff
@@ -190,6 +199,17 @@ class BaseKVIndexer(BaseIndexer):
     It can be used to tell whether an indexer is key-value indexer, via ``isinstance(a, BaseKVIndexer)``
     """
 
+    def add(self, docs: Iterator['jina_pb2.Document'], *args, **kwargs):
```
maximilianwerk, line 202: I'd add a docstring, which hints, that the key will always be the `id` of the documents.

jina/executors/indexers/keyvalue.py
```diff
@@ -1,161 +1,72 @@
...
 
 from . import BaseKVIndexer
-from ... import __binary_delimiter__
-from ...helper import call_obj_fn
 from ...proto import jina_pb2
 
 
 class BasePbIndexer(BaseKVIndexer):
-    """Storing and querying protobuf chunk/document using gzip and Python dict. """
+    """Depreciated, just for back-compat and for keeping the inheritance order unchanged"""
```
maximilianwerk, line 14: When would we remove this, if not now?

jina/executors/indexers/keyvalue.py
```diff
@@ -1,161 +1,72 @@
...
-    def _add(self, keys: Iterator['jina_pb2.Document'], *args, **kwargs):
-        """Add a object to the indexer
+    def query(self, key: int) -> Optional['jina_pb2.Document']:
 
-        :param obj: an object
-        """
-        for k in keys:
-            self.write_handler.write(k.SerializeToString() + __binary_delimiter__)
-            self._size += 1
+        pos_info = self.query_handler.get(key, None)
```
maximilianwerk, line 62: Personally, I don't like the magic, how this calls `get_query_handler`. I have to look at how and why this is implemented as it is. From my point of view it should either be a `@property` (if the invocation is cheap) or a direct call to the `get_query_handler` if invocation is expensive.

tests/unit/executors/indexers/test_shelfindexer.py
```diff
@@ -1,22 +1,23 @@
-from jina.executors.indexers.keyvalue import ShelfPbIndexer
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
```
maximilianwerk, line 1: Filename should be renamed to `test_binary_indexer.py`

tests/unit/executors/indexers/test_shelfindexer.py
```diff
@@ -31,3 +32,5 @@ def validate(req):
 
     with f:
         f.search([d], output_fn=validate)
+
+test_binarypb_in_flow()
```
maximilianwerk, line 36: This invocation is not necessary if you run it via `pytest`. If you just want to test this test locally, you can also do `pytest tests/unit/executors/indexers/test_shelfindexer.py::test_binarypb_in_flow` which only invokes exactly this test.

jina/executors/indexers/__init__.py
JoanFM: yes!

jina/clients/python/request.py
```diff
@@ -98,7 +99,7 @@ def _generate(data: Union[Iterator['jina_pb2.Document'], Iterator[bytes], Iterat
                 req.queryset.extend([top_k_queryset])
 
         for content in batch:
-            _add_document(request=req, content=content, mode=mode, doc_counter=doc_counter,
+            _add_document(request=req, content=content, mode=mode, doc_counter=doc_counter if override_doc_id else None,
```
maximilianwerk, line 102: That just became impossible to grasp when just skimming the function. Could you put one parameter per line in order to make it more obvious, that something else then just parameter passing is happening here?

tests/unit/executors/indexers/test_shelfindexer.py
hanxiao: its a draft for easing testing locally

jina/executors/indexers/keyvalue.py
hanxiao: its a draft for testing

PR #991, opened 2020-09-29
refactor: remove array_in_pb arg

hanxiao:


tests/integration/issues/github_969/test_messages_different_types.py
```diff
@@ -84,7 +84,7 @@ def validate_fn(resp):
         assert chunk2.id == 30
         assert chunk2.buffer == buffer
 
-    with Flow(array_in_pb=True).add(uses='_pass', array_in_pb=True) as f:
```
JoanFM, line 87: In this tests, I have some skipped because it was talking about optimization, we can remove entirely those and keep only ones

JoanFM (changes requested):

PR #992, opened 2020-09-29
refactor: move cli interface out of jina module

hanxiao:
very minor thing about CLI & autocomplete `\tab` in shell

Previously, ac triggers full `jina` module `__init__.py` given the previous module structure, making the whole ac slow and has some side-effect

By refactoring the module structure, `jina_cli` is now separated from `jina` and `jina_cli:main` is now used as the entrypoint in `setup.py`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,6 +9,7 @@
 !README.md
 !MANIFEST.in
 !tests/**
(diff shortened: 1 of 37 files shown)


hanxiao:
nope, this does not work

PR #993, opened 2020-09-29
feat: tree traversal default for SliceQL

maximilianwerk:
This PR should demonstrate, how the adaptation of the traversal will happen. `test_querylang_drivers.py` demonstrates nicely, how things become better understandable. `use_tree_traversal=True` should be set for now, since the tree traversal is not the default, yet. The overwrite of the default `traversal_paths` should only happen, if `['c', 'r']` is not a reasonable default.

### DIFF
--- a/jina/drivers/querylang/filter.py
+++ b/jina/drivers/querylang/filter.py
@@ -28,11 +28,12 @@ class FilterQL(QuerySetReader, BaseRecursiveDriver):
         ensures that the EncodeDriver will only get documents which modality field value is `mode2` by filtering
         those documents at the specific levels that do not comply with this condition
     """
+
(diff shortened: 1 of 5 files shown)


rutujasurve94 (approved):

rutujasurve94 (commented):
Should `traversal_paths` and the boolean also be used for `init` in `slice.py`?

maximilianwerk:
> Should `traversal_paths` and the boolean also be used for `init` in `slice.py`?

I'd say so, since slicing should usually only happen on one level and not on both: the top level and the chunks. And the boolean should be set in any driver we are currently changing. At some point, we'll switch e.g. make the tree traversal the default traversal and remove the old one. Then we'll remove all the boolean again.

PR #1001, opened 2020-09-30
feat: switched to tree traversal for index driver

maximilianwerk:
Switching to tree traversal for index driver.

### DIFF
--- a/tests/unit/flow/test_flow_multimode.py
+++ b/tests/unit/flow/test_flow_multimode.py
@@ -2,14 +2,12 @@ import os
 from typing import List, Dict
 
 import numpy as np
-import pytest
 
 from jina.executors.crafters import BaseSegmenter
 from jina.executors.encoders import BaseEncoder
 from jina.executors.indexers.keyvalue import BinaryPbIndexer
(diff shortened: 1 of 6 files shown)


tests/unit/flow/test_flow_multimode.py
```diff
@@ -79,11 +79,13 @@ def input_fn():
                                                   [1.0, 1.0, 1.0]]))
 
     chunkIndexer1 = BinaryPbIndexer.load('kvidx1.bin')
-    assert chunkIndexer1.size == 6
+    assert chunkIndexer1.size == 3
```
rutujasurve94, line 82: Why does the `chunkIndexer` size become half now?

rutujasurve94 (approved):

tests/unit/flow/test_flow_multimode.py
maximilianwerk: Before, the indexer indexed all documents twice, since the `recur_on` was not set to `chunks` for this indexer. So the real question is: why was it 6 beforehand ;)

rutujasurve94 (approved):

PR #1004, opened 2020-09-30
feat: enable fluentd logging

JoanFM:
**Changes introduced**

- Enable logging using `fluent-logging-handler`.

- Refactor Logging into a context manager class to properly handling release of resources.

- Profile logger should not be a different logger but the same with an extra handler


jina/peapods/zmq.py
```diff
@@ -46,7 +45,7 @@ def __init__(self, args: 'argparse.Namespace', logger: 'logging.Logger' = None):
         """
         self.args = args
         self.name = args.name or self.__class__.__name__
-        self.logger = logger or get_logger(self.name, **vars(args))
```
JoanFM, line 49: it always receives the logger from the outside, and having 2 owners of this logger could make it hard to release the handlers and close them

extra-requirements.txt
```diff
@@ -19,6 +19,7 @@
 scipy>=1.4.1:               index, numeric
 flask:                      devel, http, sse
 flask_cors:                 devel, http, sse
+fluent-logger:              logging
```
hanxiao, line 22:
```suggestion
fluent-logger:              logging, http, sse, dashboard
```

jina/logging/jina_logger.py
```diff
@@ -0,0 +1,252 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
```
hanxiao, line 1: we don't do recursive module naming, if you don't like `base.py` then rename it to `jina/logging/logger.py`, `jina` should not appear again in the filename

jina/peapods/zmq.py
```diff
@@ -187,10 +186,6 @@ def print_stats(self):
                          f'#recv: {self.msg_recv} '
                          f'sent_size: {get_readable_size(self.bytes_sent)} '
                          f'recv_size: {get_readable_size(self.bytes_recv)}')
-        profile_logger.debug({'msg_sent': self.msg_sent,
-                              'msg_recv': self.msg_recv,
-                              'bytes_sent': self.bytes_sent,
-                              'bytes_recv': self.bytes_recv})
```
hanxiao, line 193: note, removing this means dashboard lose all info in `TASK` panel

JoanFM: if the default logger has the profile handler?

JoanFM: when a logger can have multiple handlers, why to have an specific logger for Profile?

hanxiao: i don't remember the details, but it could be due to the easy toggle on/off, or `profile_logger` relies on some multi-thread/processing`Event`

JoanFM: Profile Logger is also using a `QueueHandler`, so it behaves as the sse_logger from behavior point of view

JoanFM: I will add a dedicated logger for PROFILE to avoid it getting every other emition

docs/chapters/logging/index.md
```diff
@@ -0,0 +1,85 @@
+# Logging configuration in jina
+
+**Logging** (beta)
+In order to better understand and monitor the running and lifetime of Jina's flow, peas and pods, jina logs messages
+in 6 different levels (DEBUG, INFO, WARNING, ERROR, CRITICAL, SUCCESS). The default level is controlled by 'JINA_LOG_VERBOSITY' environment variable.
+
+Jina uses loggers from logging python library with different Handlers to control where the messages are sent/stored:
+
+- `ConsoleHandler`. By default, logger uses a `ConsoleHandler` to print the logs in each Pea's local stdout.
+- `FileHandler`. Jina offers the possibility offers the solution to put the logs in local files either as simple text or as json format
```
maximilianwerk, line 10: Two times offers.

jina/peapods/pea.py
```diff
@@ -415,7 +415,9 @@ def close(self) -> None:
         self.send_terminate_signal()
         self.is_shutdown.wait()
         if not self.daemon:
+            # TODO: remove clear_queues when queues are totally removed
```
maximilianwerk, line 418: Should this TODO stay or be removed before merging the PR?

tests/unit/logging/test_logging.py
```diff
@@ -12,3 +12,4 @@ def test_logging_message():
     logger.warning('this is test warning message')
     logger.error('this is test error message')
     logger.critical('this is test critical message')
+    del os.environ['JINA_LOG_VERBOSITY']
```
maximilianwerk, line 15: Shouldn't this rather be reset to the former value, if it was set beforehands??

maximilianwerk (commented):
Some minor cosmetics. Could you add a test, that checks, whether the correct files are written? As far as I can see, this is missing.

jina/peapods/pea.py
JoanFM: no, because it is a TODO for the future

hanxiao (approved):

Issue #1000, opened 2020-09-30
🚨 BREAKING CHANGES 0.6.0 -> 0.7.0


hanxiao:
This issue is created to keep track breaking changes between `0.6.0` and `0.7.0` (scheduled release on Oct. 26, 2020)

## Breaking Changes & Migration Guide

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1005 #1034  | Not yet | |

- `Document.id` in Protobuf is now a string type variable, this applies also to `Document.parent_id` and `Relevance.ref_id`
- `id` is filled with the hexdigest of [blake2b algorithm](https://docs.python.org/3.7/library/hashlib.html#hashlib.blake2b), which takes the document content into consideration.
- `SortQL.field` is now using `dunder_get` syntax rather than `.` expansion, e.g. `a.b.c -> a__b__c`, `score.value -> score__value`. It now supports `dict` and `list` access.
- `first_doc_id`, `random_doc_id` and `override_doc_id` are removed from CLI. 

---

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #983 #995 #1000 #1001 #1003 #1006 #1007  #1011 #1027 #1039 #1036  | Not yet |  |

- `granularity_range`, `adjacency_range`, `recur_on` and `recursion_order` are deprecated and don't work anymore
- Replaced by `traversal_paths`, which allows to decribe the part of the data structure, where the traversal should happen.
-- `['r']` for example only applies an Executor on the root node
-- `['c', 'r']` applies an Executor on the chunks of the root node and the root node itself
-- `['m', 'r']` applies an Executor on the matches of the root node and the root node itself
- Removed the `_apply` function and only kept `_apply_all` in `BaseRecursiveDriver`. Rationale: The Executor itself should do the iteration over the documents, as most Executors did anyhow. Currently, there is no need for the added complexity of a dedicated `_apply` function.

---

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1031 |  | |
- a new log config YAML schema is introduced, see appendix below
- `--log-config` is added to `jina pea` CLI, by default it points to `logging.default.yml`.
- `--log-sse`, `--log-profile`, `--log-with-own-name` are removed from `jina pea` CLI. They are now mapped to different resource files as below:

| filename | logger in the code |
|  --- | --- |
| logging.default.yml | `default_logger` and any logger defined with `JinaLogger()` |
| logging.docker.yml | `logger` used in the `ContainerPea`|
[...]

PR #1005, opened 2020-09-30
feat: add hash proto class for gen unique id

hanxiao:
## Breaking Changes #1000 

- `Document.id` in Protobuf is now a string type variable, this applies also to `Document.parent_id` and `Relevance.ref_id`
- `id` is filled with the hexdigest of [blake2b algorithm](https://docs.python.org/3.7/library/hashlib.html#hashlib.blake2b), which takes the document content into consideration.
- `SortQL.field` is now using `dunder_get` syntax rather than `.` expansion, e.g. `a.b.c -> a__b__c`, `score.value -> score__value`. It now supports `dict` and `list` access.
- `first_doc_id`, `random_doc_id` and `override_doc_id` are removed from CLI.

## Features

- `jina.proto.uid` provides helper functions that convert between `int <> hexdigest (str) <> bytes`. Drivers can use them to prepare `doc.id` into the correct format that executor requires.

## Follow-up

Two tests are commented out due to #1014 
- tests/integration/level_depth/test_search_different_depths.py
- tests/unit/drivers/test_chunk2doc_rank_drivers.py


jina/drivers/rank.py
```diff
@@ -109,41 +118,6 @@ class CollectMatches2DocRankDriver(BaseRankDriver):
     def __init__(self, traversal_paths: Iterable[str] = ['m'], *args, **kwargs):
         super().__init__(traversal_paths=traversal_paths, *args, **kwargs)
 
-    def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2.Document', *args, **kwargs) -> None:
```
JoanFM, line 112: why are we removing this, it does not behave like `Chunk2DocRankDriver`?. It has some differences

JoanFM (changes requested):

jina/drivers/rank.py
hanxiao: ok my bad, yes, they have different loop structure

JoanFM: I will put it back

PR #1007, opened 2020-09-30
feat: slice is there and some test cleanups

maximilianwerk:
Since slice is refactored, there is no point having the old tests. That will decrease test coverage for code, that anyhow will be deleted very soon. Anyhow, let's stall this PR a little bit and do it as the last bit before finishing the whole refactoring with the code-removal of the old traversal.


tests/integration/level_depth/test_search_different_depths.py
```diff
@@ -24,31 +18,41 @@ def test_index_depth_0_search_depth_1():
     def validate_granularity_1(resp):
         assert len(resp.docs) == 3
         for doc in resp.docs:
-            assert doc.granularity == 1
-            assert len(doc.matches) == 1
+            assert doc.granularity == 0
+            assert len(doc.matches) == 3
```
JoanFM, line 22: why was it 1 and now 3? Not sure it was right in the first place?

tests/integration/level_depth/test_search_different_depths.py
```diff
@@ -24,31 +18,41 @@ def test_index_depth_0_search_depth_1():
...
+        ' I am chunk 1 of doc 1,',
         'I am chunk 0 of doc 2,',
-        'I am chunk 3 of doc 3',
+        ' I am chunk 3 of doc 3',
     ]
 
     search_flow = Flow().load_config('flow-query.yml')
     with search_flow:
-        search_flow.search(input_fn=search_data, output_fn=validate_granularity_1, callback_on_body=True, granularity=1)
+        search_flow.search(
```
JoanFM, line 52: We should remove then `granularity` from the `request` generator function. With the traversal path it makes no sense anymore right?

tests/integration/level_depth/test_search_different_depths.py
maximilianwerk: there was a use of `top_k` beforehand. But since `top_k` us currently buggy, it does not work and I completely removed it.

tests/integration/level_depth/test_search_different_depths.py
maximilianwerk: :+1:

JoanFM: u do it here or want to merge this and open a new one?

JoanFM (approved):

maximilianwerk:
Be aware of the code formatting change. That took me 1s, since I used black. Do you like it? Let's use black everywhere soon! ;)

JoanFM:
> Be aware of the code formatting change. That took me 1s, since I used black. Do you like it? Let's use black everywhere soon! ;)

As long as u tell me how to set it up in IntelliJ I do not care

maximilianwerk:
> > Be aware of the code formatting change. That took me 1s, since I used black. Do you like it? Let's use black everywhere soon! ;)
> 
> As long as u tell me how to set it up in IntelliJ I do not care

You don't even have to. Just setup a github action, that does it for you and you locally have to never are about it anymore...

PR #1008, opened 2020-09-30
Fix mypy errors and dunderkey bug 743

Arrrlex:
This PR is a combination of fixing mypy errors and fixing a bug in dunderkey.py.

Fixing mypy errors involves:

1. Fixing type annotations, when they're not correct
2. Making the code a little less "dynamically typed"

Although python is a dynamically typed language, it can help with correctness and readability to be as explicit as possible when exploiting python's dynamic features. For example, in the function dunder_partition I made it explicit that we split `key` into a 2-tuple, rather than an arbitrary-length list. This doesn't change the functionality of the code, it just makes it more explicit.

I came across a bug in undunder_keys while completing this work, and since it's a small change I figured I'd include it in this PR. I hope that's OK.

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -13,8 +13,8 @@ from argparse import ArgumentParser, Namespace
 from datetime import datetime
 from io import StringIO
 from itertools import islice
-from types import SimpleNamespace
-from typing import Tuple, Optional, Iterator, Any, Union, List, Dict, Set, TextIO
+from types import SimpleNamespace, ModuleType
+from typing import Tuple, Optional, Iterator, Any, Union, List, Dict, Set, TextIO, Sequence, Iterable
 
 import numpy as np
 from ruamel.yaml import YAML, nodes
@@ -134,7 +134,7 @@ def touch_dir(base_dir: str) -> None:
         os.makedirs(base_dir)
 
 
-def batch_iterator(data: Union[Iterator[Any], List[Any], np.ndarray], batch_size: int, axis: int = 0,
+def batch_iterator(data: Union[np.ndarray, Iterable[Any]], batch_size: int, axis: int = 0,
                    yield_slice: bool = False) -> Iterator[Any]:
     if not batch_size or batch_size <= 0:
         yield data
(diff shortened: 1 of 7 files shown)


jina/logging/base.py
```diff
@@ -136,14 +139,28 @@ def success(self, msg: str, **kwargs):
         if self.log_level <= LogVerbosity.SUCCESS:
             sys.stdout.write(f'W:{self.context}:{self._planify(msg)}')
 
+class SuccessLogger(Protocol):
```
JoanFM, line 142: Thank you for the work here, but since this part is undergoing a big refactor, I would appreciate if these changes can be left out of the PR

jina/helper.py
```diff
@@ -188,22 +188,23 @@ def parse_arg(v: str) -> Union[bool, int, str]:
         else:
             return []
     try:
-        v = int(v)  # parse int parameter
+        return int(v)  # parse int parameter
```
JoanFM, line 191: I prefer to keep a single return if possible (it can make much easier for debugging), plus now return in the end is never reached

JoanFM (changes requested):

JoanFM:
Hey @Arrrlex ,

Thank you for your PR, I would appreciate if you could split this in 2 PRs so that it can be clearly seen what is helping in type hinting and what is solving an issue (and to see clearly what is the issue).

Thank you very much

Arrrlex:
@JoanFM That makes sense. I'll close this PR and open 2 new ones.

hanxiao:
@Arrrlex hi thanks for the PR, before you making the new one. Please note you didn't set up your local git config correctly, the author of your commit does not match with your Github account. This poses problems:

- Your contribution can not be credited, and will not show up in the list of contributors (as bad git config = anonymous)
- We can not accept CLA for anonymous contribution.

Please note that this does not only affect your contribution to Jina, but also affects *all your contribution* to anything on Github. So better fix it, once for all.

Here is what you can do:
```bash
git config user.name "Correct new name"
git config user.email "correct@new.email" -> match with your Github registered email address
```

Then fix the commit message by
```bash
git commit --amend --author="Author Name <email@address.com>" --no-edit
git push --force
```

jina/helper.py
Arrrlex: The return at the end would be reached if both exceptions are raised, and each of the if clauses falls through (i.e. v is not an int, not a float, not length-0, and neither "true" nor "false")

PR #1009, opened 2020-09-30
Fix mypy errors 743

Arrrlex:
This PR fixes several mypy errors. This involves:

1. Fixing type annotations, when they're not correct
2. Making the code a little less "dynamically typed"

Although python is a dynamically typed language, it can help with correctness and readability to be as explicit as possible when exploiting python's dynamic features. For example, in the function `dunder_partition` I made it explicit that we split `key` into a 2-tuple, rather than an arbitrary-length list. This doesn't change the functionality of the code, it just makes it more explicit.

Running mypy can also uncover bugs. For example, in the function `get_full_version()` the mypy errors alerted me to the following behaviour:

- If the logger is an `NTLogger`, rather than `logging.Logger`, then attempting to log an error without first converting it to a string would result in another exception
- Since the exception is caught and logged but not re-raised, by default the return value would be None. This would most likely cause an error since the callsite would be expecting a 2-tuple


jina/helper.py
```diff
@@ -267,7 +267,7 @@ def _path_import(absolute_path: str) -> 'types.ModuleType':
             spec.loader.exec_module(module)
             return module
         except ModuleNotFoundError:
-            pass
+            return None
```
JoanFM, line 270: as per style use only `return`

jina/helper.py
```diff
@@ -662,7 +660,8 @@ def get_full_version() -> Tuple[Dict, Dict]:
         env_info = {k: os.getenv(k, '(unset)') for k in __jina_env__}
         return info, env_info
     except Exception as e:
-        default_logger.error(e)
+        default_logger.error(str(e))
+        raise
```
JoanFM, line 664: if it was not there, better avoid raising, I guess the process will continue working even when this error occured so do not want to interrupt it

jina/helper.py
```diff
@@ -179,7 +179,7 @@ def _get_yaml():
     return y
 
 
-def parse_arg(v: str) -> Union[bool, int, str]:
+def parse_arg(v: str) -> Union[bool, int, str, list, float, None]:
```
JoanFM, line 182: if None in Union, isn't better to use `Optional`?

jina/helper.py
```diff
@@ -156,13 +156,13 @@ def batch_iterator(data: Union[Iterator[Any], List[Any], np.ndarray], batch_size
                 yield tuple(sl)
             else:
                 yield data[tuple(sl)]
-    elif hasattr(data, '__len__'):
+    elif isinstance(data, Sequence):
```
JoanFM, line 159: Since we are here just changing type, let's avoid doing logical changes, what is the error identified here?

A `dict` for instance has attribute `length`, is an `Iterable` but not a `Sequence`, so according to new typing it is an "expected" input and we are changing its behavior here.

jina/helper.py
```diff
@@ -203,7 +203,7 @@ def parse_arg(v: str) -> Union[bool, int, str]:
     return v
 
 
-def countdown(t: int, reason: str = 'I am blocking this thread') -> None:
+def countdown(t: int, reason: str = 'I am blocking this thread'):
```
JoanFM, line 206: why is it here removed?

JoanFM (changes requested):

jina/helper.py
```diff
@@ -377,7 +372,7 @@ def get_random_identity() -> str:
 yaml = _get_yaml()
 
 
-def expand_env_var(v: str) -> str:
+def expand_env_var(v: str) -> Union[bool, int, str, list, float, None]:
```
JoanFM, line 375: if None in Union, isn't better to use Optional?

JoanFM (changes requested):

jina/helper.py
Arrrlex: The error here is that `dict`s don't support slicing. Thus the line `yield data[_:_ + batch_size]` will throw a `TypeError` for a `dict`.

jina/helper.py
Arrrlex: Apologies - turns out I misunderstood the recommended style for functions that have only an implicit return. (I thought you're supposed to omit the return, but in fact it's recommended to return None). I'll revert that change.

jina/helper.py
Arrrlex: Fair enough. I'll change this function to return `Optional[...]`.

PR #1011, opened 2020-10-01
Fix encoder driver for tree traversal

rutujasurve94:
- Fix encode driver for `tree traversal`.

- `mode yaml` files for encoder unit tests

### DIFF
--- a/jina/drivers/encode.py
+++ b/jina/drivers/encode.py
@@ -1,7 +1,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-from typing import Iterable
+from typing import Iterable, List
 
 from . import BaseExecutableDriver
 from .helper import extract_docs, array2pb
@@ -15,6 +15,7 @@ class BaseEncodeDriver(BaseExecutableDriver):
 
     def __init__(self, executor: str = None, method: str = 'encode', *args, **kwargs):
         super().__init__(executor, method, *args, **kwargs)
+        self._use_tree_traversal = True
 
 
 class EncodeDriver(BaseEncodeDriver):
--- a/tests/unit/drivers/yaml/mockencoder-mode1.yml
+++ b/tests/unit/drivers/yaml/mockencoder-mode1.yml
@@ -9,4 +9,4 @@ requests:
       - !EncodeDriver
         with:
           method: encode
-          granularity_range: [1, 2]
+          traversal_paths: ['c']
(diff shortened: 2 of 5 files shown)


tests/unit/drivers/yaml/mockencoder-mode1.yml
```diff
@@ -9,4 +9,4 @@ requests:
       - !EncodeDriver
         with:
           method: encode
-          granularity_range: [1, 2]
+          traversal_paths: ['r']
```
maximilianwerk, line 12: This should be `c` instead of `r`, since beforehand, the mockencoder also encoded the chunks (e.g. `granularity_level = 1`).

jina/drivers/encode.py
```diff
@@ -15,14 +15,15 @@ class BaseEncodeDriver(BaseExecutableDriver):
...
+        self._use_tree_traversal = True
 
 
 class EncodeDriver(BaseEncodeDriver):
     """Extract the chunk-level content from documents and call executor and do encoding
     """
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, traversal_paths: List[str] = ['r'], *args, **kwargs):
```
maximilianwerk, line 25: Since before, the default value was implicitly the same as in the `BaseRecursiveDriver`, I would not overwrite the value here and check if all tests pass afterwards. Just remove the `traversal_path` here and in the `super().__init__(...)` call.

tests/unit/drivers/yaml/mockencoder-mode2.yml
```diff
@@ -9,4 +9,4 @@ requests:
       - !EncodeDriver
         with:
           method: encode
-          granularity_range: [1, 2]
+          traversal_paths: ['r']
```
maximilianwerk, line 12: Same here.

maximilianwerk (changes requested):
Some places in the code are missing, where there is still a `granularity_range` used for the `EncodeDriver`, e.g. `tests/unit/flow/yaml/mockencoder-mode1.yml`.

Furthermore, there is one PR needed in the jina-hub for the encoder for the `ImageKerasEncoder`.

jina/drivers/encode.py
rutujasurve94: Done

tests/unit/drivers/yaml/mockencoder-mode2.yml
rutujasurve94: done

maximilianwerk (approved):

PR #1021, opened 2020-10-04
test: take out rm files from tests

bwanglzu:
Suggested by @maximilianwerk , @JoanFM 

> We should NOT create directories and delete them ourself, but rather use [tmpdir fixtures](https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture) of pytest.

This PR aims at taking out some usages of `rm_files` in [test](https://github.com/jina-ai/jina/blob/df7a7f494be7683a8377fdcf3bb8036ed7238e0a/tests/__init__.py#L55) to clean up the unit test a bit.

~The thing I don't like in this PR is I created the same fixture two times. Expecting we have a `conftest.py` to place some commonly used fixtures, not sure this `meta` one will be used a lot.~ Updated with `conftest.py`


tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -19,17 +18,17 @@
 cur_dir = os.path.dirname(os.path.abspath(__file__))
 
 
-@pytest.mark.parametrize('batch_size, compress_level', [(None, 0), (None, 1), (2, 0), (2, 1)])
-def test_numpy_indexer(batch_size, compress_level):
+@pytest.mark.parametrize('batch_size, compress_level', [(None, 0)])
+def test_numpy_indexer(batch_size, compress_level, tmpdir):
+    index_dir = tmpdir.join('NumpyIndexer-test.bin')
     with NumpyIndexer(index_filename='np.test.gz', compress_level=compress_level) as indexer:
```
JoanFM, line 24: In order to keep every possible file created by the Executor, it is better to set the workspace parameter in its metas to the tmpdir. See my PR on jina-hub for reference... Like this u are sure that the executor will "work" in the tmpdir

tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -19,17 +18,17 @@
...
         indexer.batch_size = batch_size
         indexer.add(vec_idx, vec)
-        indexer.save()
+        indexer.save(index_dir)
```
JoanFM, line 27: So u do not need to set this up because later may not be able to read from dump automatically

JoanFM (changes requested):

tests/unit/executors/indexers/test_numpyindexer.py
JoanFM: https://github.com/jina-ai/jina-hub/pull/424

tests/unit/executors/indexers/test_numpyindexer.py
bwanglzu: check, I created a fixture for meta

JoanFM: Love it!

bwanglzu:
The thing I don't like in this PR is I created the same fixture two times. Expecting we have a `conftest.file` to place some commonly used fixtures, not sure this `meta` one will be used a lot.

JoanFM (approved):

tests/unit/executors/encoders/numeric/test_transform_encoder.py
```diff
@@ -1,20 +1,26 @@
...
 
+from jina.executors.metas import get_default_metas
 from jina.executors.encoders.numeric import TransformEncoder
-from tests import rm_files
 
 input_dim = 28
 target_output_dim = 2
 
+@pytest.fixture(scope='function', autouse=True)
+def metas(request, tmpdir):
```
JoanFM, line 14: why request?

JoanFM (changes requested):

hanxiao (changes requested):
this PR looks good! but can we promote that fixture to `tests/__init__.py` so that there is only one fixture globally? when you need to use it then `from tests import metas`.

Plus `metas` does not look a good name for fixture, probably rename it to `cleanup` or something

tests/unit/executors/encoders/numeric/test_transform_encoder.py
```diff
@@ -1,20 +1,26 @@
...
+    os.environ['TEST_WORKSPACE'] = str(tmpdir)
```
maximilianwerk, line 15: I'd prefix it with `JINA_`, so it becomes `JINA_TEST_WORKSPACE` to guarantee, that there are no namespace collisions.

Furthermore, having the same environment variable for each and every test, might be a bad idea. What about:

```
@pytest.fixture()
def random_workspace_name():
    return f'JINA_TEST_WORKSPACE_{random.rand()}'
    # just make the random better...

@pytest.fixture(scope='function', autouse=True)
def metas(request, tmpdir, random_workspace_name):
    os.environ[random_workspace_name] = str(tmpdir)
    ...
```

This allows to explicitly give a workspace name in a test function:

```
@pytest.mark.parametrize('random_workspace_name', ['JINA_TEST_WORKSPACE_EXPLICIT_NAME'])
def my_test():
    assert metas['workspace'] == os.environ['JINA_TEST_WORKSPACE_EXPLICIT_NAME']
    # TRUE
```

tests/unit/executors/encoders/numeric/test_transform_encoder.py
maximilianwerk, line 14: As Han suggested, renaming would be nice. What about `test_metas` or `default_test_metas`?

maximilianwerk (changes requested):
Great idea to make this generic. Having a separate env variable per test might be the first step to run tests in parallel again.

tests/unit/executors/encoders/numeric/test_transform_encoder.py
bwanglzu: removed

tests/unit/executors/encoders/numeric/test_transform_encoder.py
JoanFM: randomizing may be problematic for tests where you need to set the workspace inside a `yml` file?

maximilianwerk: I accidentally deleted @JoanFM comment, that this might be a bad idea, since `.yml` files might need and explicit name.

But anyhow, then you can explicitly set it as shown in the `my_test` function.

tests/unit/executors/encoders/numeric/test_transform_encoder.py
bwanglzu: sure

bwanglzu:
@JoanFM @hanxiao @maximilianwerk is it good to put fixtures into [conftest](https://stackoverflow.com/q/34466027) rather than init?

JoanFM:
> @JoanFM @hanxiao @maximilianwerk is it good to put fixtures into [conftest](https://stackoverflow.com/q/34466027) rather than init?

If you consider it to be good practice, it seems okey for me!

hanxiao (approved):

tests/unit/executors/encoders/numeric/test_transform_encoder.py
bwanglzu: great suggestions! I applied most of the suggestions into this PR, please take a look :) @maximilianwerk

hanxiao (changes requested):
two tests are broken, please fix them:
```
FAILED tests/unit/executors/indexers/test_binary_indexer.py::test_binarypb_in_flow
FAILED tests/unit/flow/test_flow_example_faiss.py::test_index - jina.excepts....
```

bwanglzu:
The error is caused by `autouse=True`, according to [doc](https://docs.pytest.org/en/2.8.7/fixture.html#:~:text=Here%20is%20how%20autouse%20fixtures,directory%20will%20invoke%20the%20fixture.)
> if an autouse fixture is defined in a conftest.py file then all tests in all test modules below its directory will invoke the fixture.

PR #1024, opened 2020-10-05
fix: no early file closing

maximilianwerk:
A fix for a subtle bug introduced here: https://github.com/jina-ai/jina/pull/1009/files#diff-e47ef1c2a309b9935e22d5ff1e58d7dcR30

Beforehand, only the iterator was returned and thus the file closed. Since the iterator has not iterated the file yet, it would fail in the moment it starts iterating. With the `yield` solution, the file remains open.

@Arrrlex :)

### DIFF
--- a/jina/clients/python/io.py
+++ b/jina/clients/python/io.py
@@ -9,9 +9,14 @@ from typing import List, Union, Iterator, Any
 import numpy as np
 
 
-def input_lines(lines: Iterator[str] = None, filepath: str = None, size: int = None, sampling_rate: float = None,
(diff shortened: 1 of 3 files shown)


JoanFM (dismissed):

JoanFM:
I approved, but is there a way to add a test for this?

maximilianwerk:
> I approved, but is there a way to add a test for this?

:) Done

JoanFM (commented):
Can we move the `test_clients` that are related to io to this file?

JoanFM (approved):

Arrrlex:
thank you for catching this, and apologies for introducing it!

Issue #373, opened 2020-05-03
Refactor import to comply with PEP 8


maximilianwerk:
**Describe the feature**
Currently jina uses relative imports everywhere. Furthermore, there are a lot of import on function level and not at the top level of the files. jina guidelines say, PEP 8 should be followed. Anyhow, [PEP 8 is explicit about imports](https://www.python.org/dev/peps/pep-0008/#imports): They should be put on top of each file and absolute imports are preferred above relative imports. Doing this makes later refactoring and moving of modules less error prone. This might be especially useful, since jina is at the beginning of its lifetime and thus might undergo quite some refactoring in the future.

In case, this is wanted, I am happy to do the refactoring.

**Your proposal**
Follow best practices and refactor the relative imports to absolute imports and move imports to the top of each file. As an example, the `jina/main/__init__.py` file would look like:

```
__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
__license__ = "Apache-2.0"

import sys

from pkg_resources import resource_filename

from jina.helper import colored
from jina.logging import default_logger
from jina.main import api
from jina.main.autocomplete import ac_table
from jina.main.parser import get_main_parser


def _get_run_args(print_args: bool = True):
    parser = get_main_parser()
    if len(sys.argv) > 1:
        args = parser.parse_args()
        default_args = parser.parse_args([sys.argv[1]])
        if print_args:
            with open(
                resource_filename("jina", "/".join(("resources", "jina.logo")))
            ) as fp:
                logo_str = fp.read()
            param_str = []
            for k, v in sorted(vars(args).items()):
                j = f'{k.replace("_", "-"): >30.30} = {str(v):30.30}'
                if getattr(default_args, k) == v:
                    param_str.append("   " + j)
                else:
                    param_str.append("🔧️ " + colored(j, "blue", "on_yellow"))
            param_str = "\n".join(param_str)
            default_logger.info(
                f'\n{logo_str}\n▶️  {" ".join(sys.argv)}\n{param_str}\n'
            )
        return args
    else:
        parser.print_help()
        exit()


def _quick_ac_lookup():
    if len(sys.argv) > 1:
        if sys.argv[1] == "commands":
            for k in ac_table["commands"]:
                print(k)
            exit()
        elif sys.argv[1] == "completions":
            if sys.argv[2] in ac_table["completions"]:
                for k in ac_table["completions"][sys.argv[2]]:
                    if k not in sys.argv:
                        print(k)
            exit()


[...]

PR #1025, opened 2020-10-05
Fix drivers convert blob2pnguri

bhavsarpratik:


jina/drivers/convert.py
```diff
@@ -8,6 +8,7 @@
 import urllib.request
 import zlib
 
+from PIL import Image
```
JoanFM, line 11: Better move this import inside the function needing it

jina/drivers/convert.py
```diff
@@ -136,8 +138,27 @@ def png_pack(png_tag, data):
...
-        arr.uri = self.png_convertor(arr)
+    @staticmethod
+    def image_to_byte_array(image:Image, format:str):
+        import io
+        img_byte_arr = io.BytesIO()
+        image.save(img_byte_arr, format=format)
+        img_byte_arr = img_byte_arr.getvalue()
+        return img_byte_arr
+
+    def png_convertor_nd(self, arr: np.array):
```
JoanFM, line 149: It is weird that we are saying that it is nd but we are converting and interpreting as an RGB image

JoanFM: Better to say 3d and that is it

hanxiao:
unit test is required

JoanFM (changes requested):

jina/drivers/convert.py
bhavsarpratik: I think it handles greyscale 2d but need to confirm it. Will change it to 3d for now.

JoanFM: well then at least add exception if the shape is not properly handled and tested

jina/drivers/convert.py
bhavsarpratik: I have to keep it outside for argument type info for image_to_byte_array. Should I remove type hint from the argument of image_to_byte_array? 

I have kept PIL inside the function in png_convertor3d

JoanFM: if using for type hint you can do:

```python
if False:
   from PIL import Image
```

bhavsarpratik:
I think our test has failed due to the `timeout-minutes: 30` in actions yaml. Should we increase it? To what value? Should I push this change as a separate PR or in this PR?

JoanFM:
> I think our test has failed due to the `timeout-minutes: 30` in actions yaml. Should we increase it? To what value? Should I push this change as a separate PR or in this PR?

I think sometimes fails but it tends to be good. But we do not have a new test for the added cases right?

bhavsarpratik:
No new test is added for now

jina/drivers/convert.py
```diff
@@ -146,19 +146,28 @@ def image_to_byte_array(image:Image, format:str):
         img_byte_arr = img_byte_arr.getvalue()
         return img_byte_arr
 
-    def png_convertor_nd(self, arr: np.array):
+    def png_convertor(self, arr: np.array):
         from PIL import Image
         arr = arr.astype(np.uint8)
-        im = Image.fromarray(arr).convert('RGB')
+        if arr.shape == 1:
```
JoanFM, line 152: len(arr.shape)?

bhavsarpratik: Oopsie

jina/drivers/convert.py
```diff
@@ -139,35 +140,36 @@ def png_pack(png_tag, data):
...
-    def image_to_byte_array(image:Image, format:str):
+    def image_to_byte_array(image: 'Image', format: str):
         import io
         img_byte_arr = io.BytesIO()
         image.save(img_byte_arr, format=format)
         img_byte_arr = img_byte_arr.getvalue()
         return img_byte_arr
 
     def png_convertor(self, arr: np.array):
+        import numpy as np
```
JoanFM, line 151: Numpy is one of our key dependency and it can be put on top

bhavsarpratik: It is also kept at top. I have kept it again inside for performance as this function can be called many times. Should I remove it from the function?

JoanFM: why is it better for performance to keep it inside the function?

bhavsarpratik: Han told about this earlier. 
https://github.com/jina-ai/jina/issues/373#issuecomment-623483847

JoanFM: But this is about running simple `jina` commands (for instance doing `jina --help` and this kind of things) to run fast by not doing extra imports, but for the function to run fast it does not affect (or it could only be harmful). Plus, if it is alrady imported on the top.

bhavsarpratik: Okay, would like to understand why are we importing inside the functions over here?
https://github.com/jina-ai/jina-hub/blob/master/encoders/nlp/TransformerTorchEncoder/__init__.py

To confirm if importing inside helps, earlier I had tried a function with numpy import inside and outside. I think it was slightly faster when I kept it inside. I am also new to this concept of importing inside. 

As per this, it's cached the first time the function is used. So it doesn't make it slower by importing every time.
https://stackoverflow.com/questions/3095071/in-python-what-happens-when-you-import-inside-of-a-function

JoanFM: It should not affect, we import inside when it is not a basic dependency (so that the fact that only importing that file do not crash jina), `numpy` is a basic dependencya and therefore it is okey

JoanFM: but the main point here, is that u are importing both inside and outside. And since it is numpy let's keep it outside

bhavsarpratik: Yes I agree with the practice.

tests/unit/drivers/test_convert_driver.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.mark.parametrize('data', test_data)
+def test_blob2pnguri_driver(data):
+    doc, mode = data
+    width, height = 28, 32
+
+    driver = Blob2PngURI(target='uri', width=width, height=height)
+    driver._apply(doc)
+    Image.frombytes(mode, (width, height), doc.uri.encode()) # just to check if the data is enough for the image recreation
```
JoanFM, line 28: some assertion at least about the resulting size?

JoanFM (changes requested):

tests/unit/drivers/test_convert_driver.py
bhavsarpratik: I had actually kept the assertion but felt it useless since the shape is given for recreating the image. If it is not able to create it from the byte data, it throws the error. Also, I observed that this is not very reliable for exact checking. I can have input image of some shape and then have this recreated of some other shape and it still works. So this recreation is just to check if the uri data exists and is enough.

jina/drivers/convert.py
```diff
@@ -136,7 +139,35 @@ def png_pack(png_tag, data):
...
+        img_byte_arr = img_byte_arr.getvalue()
+        return img_byte_arr
+
+    def png_convertor(self, arr: np.array):
+        from PIL import Image
+
+        arr = arr.astype(np.uint8)
+
+        if len(arr.shape) == 1:
+            # im = Image.fromarray(arr.reshape((self.height, self.width))).convert('L')
```
JoanFM, line 156: remove comment

tests/unit/drivers/test_convert_driver.py
JoanFM: and testing that URI2Buffer returns the same value as in Blob would make sense?

jina/drivers/convert.py
bhavsarpratik: This is our replacement code for the png_converter_1d. Will remove and save it somewhere

tests/unit/drivers/test_convert_driver.py
bhavsarpratik: The blob is converted to numpy arr which is then converted to type uint8 in png_converter. I think due to type conversions, we might not be able to assert blobs.
arr = pb2array(d.blob)

The original blob is made from numpy array and the final blob will be made from png after doing URI2Buffer.

JoanFM: Ok have u tested in ur example then that it works at least?

JoanFM: maybe we can have an integration test for this one

bhavsarpratik: Yes tested it for certain values. The test throws error if the uri is empty or there is a huge mismatch of original image shape and final image shape. If original is 32,32 it works for like 32, 64 final but fails for 32, 128. Not able to understand this phenomenon. Throws ValueError: not enough image data.

Let me check integration test. Pushing code for now for testing current.

JoanFM (commented):
Where does this env come from?

bhavsarpratik:
I am trying out this 
https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d

.github/workflows/ci.yml
```diff
@@ -35,13 +35,17 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+      - uses: actions/cache@v2
```
JoanFM, line 38: if we want to have this I have nothing against, but let's put it in a different PR

JoanFM (approved):

PR #1031, opened 2020-10-05
refactor(logger): move all logger config into yaml

hanxiao:
This PR unifies the logger building in Jina and move all configurable part into YAML config. Each YAML file now corresponds to a `JinaLogger`. User can change/create new logger by customizing `--log-config`.

## Breaking Changes #1000 

- a new log config YAML spec is introduced, see appendix below
- `--log-config` is added to `jina pea` CLI, by default it points to `logging.default.yml`.
- `--log-sse`, `--log-profile`, `--log-with-own-name` are removed from `jina pea` CLI.

## New Features

- `Executor` will inherit `log-config` from its parent Pea/Pod/Flow
- `SysLogHandler` is added, one can use it to record log onto a remote host without `fluentd`
- four presets logger configs are (located in `resources/`):

| filename | logger in the code |
|  --- | --- |
| logging.default.yml | `default_logger` and any logger defined with `JinaLogger()` |
| logging.docker.yml | `logger` used in the `ContainerPea`|
| logging.profile.yml |  `profile_logger`|
| logging.remote.yml | `logger` used in the `RemotePea`|

## Appendix: New Log Config YAML Schema

```yaml
handlers:  # enabled handlers, order does not matter
  - StreamHandler
  - SysLogHandler
  - FluentHandler
level: INFO  # set verbose level
configs:
  FileHandler:
    format: '%(asctime)s:{name:>15}@%(process)2d[%(levelname).1s]:%(message)s'
    output: 'jina-{uptime}.log'
    formatter: JsonFormatter
  StreamHandler:
    format: '{name:>15}@%(process)2d[%(levelname).1s]:%(message)s'
    formatter: ColorFormatter
  SysLogHandler:
    ident: # this will be prepend to all messages
    format: '{name:>15}@%(process)2d[%(levelname).1s]:%(message)s'
    host: # when not given then record it locally, /dev/log on linux /var/run/syslog on mac
    port: # when not given then record it locally,  /dev/log on linux /var/run/syslog on mac
    formatter: PlainFormatter
  FluentHandler:
    # this configuration describes where is the fluentD daemon running and waiting for logs to be emitted.
    # FluentD then will have its own configuration to forward the messages according to its own syntax
    # prefix will help fluentD filter data. This will be prepended for FluentD to easily filter incoming messages
    tag: jina
    host: 0.0.0.0
    port: 24224
    format:
      host: '%(hostname)s'
      process: '%(process)s'
      type: '%(levelname)s'
```


jina/logging/__init__.py
```diff
@@ -1,13 +1,10 @@
...
+from .logger import JinaLogger
 
-default_logger = get_logger('JINA')  #: a logger at the global-level
-if 'JINA_LOG_PROFILING' in os.environ:
-    profile_logger = get_logger('PROFILE', log_profile=True)  #: a logger for profiling
-    profile_logger.success('profiling is enabled')
-else:
-    profile_logger = default_logger
+default_logger = JinaLogger('JINA')  #: a logger at the global-level
+profile_logger = JinaLogger('PROFILE', log_config=resource_filename('jina',
```
JoanFM, line 9: so it logs by default now right?

jina/logging/logger.py
```diff
@@ -150,85 +129,55 @@ def close(self):
         for handler in self.logger.handlers:
             handler.close()
 
+    def add_handlers(self, config_path: str = None, **kwargs):
```
JoanFM, line 132: using https://docs.python.org/3/library/logging.config.html woudln't directly work?

jina/logging/__init__.py
hanxiao: yes

jina/logging/logger.py
hanxiao: not immediately as there has to be some post tweak after building the handlers. could be done after some refactoring, I will leave it as a PR for others

JoanFM (approved):

PR #1034, opened 2020-10-06
fix: fixed the int float conversion collision between hash and idy

maximilianwerk:
Fixes the float int conversion error that happened for the new hashes.

### DIFF
--- a/jina/drivers/rank.py
+++ b/jina/drivers/rank.py
@@ -9,6 +9,7 @@ import numpy as np
 from . import BaseExecutableDriver
 from .helper import pb_obj2dict
 from ..proto import uid
+from ..executors.rankers import Chunk2DocRanker, Match2DocRanker
 
 if False:
     from ..proto import jina_pb2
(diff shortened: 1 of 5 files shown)


maximilianwerk:
FYI: here is the accompanied jina-hub fix: https://github.com/jina-ai/jina-hub/pull/434

JoanFM:
Do we have some figures of the 2 proposed versions difference in performance?

maximilianwerk:
> Do we have some figures of the 2 proposed versions difference in performance?

According to https://stackoverflow.com/questions/55493274/performance-of-sorting-structured-arrays-numpy it is much (100x) slower for huge arrays compared with the previous implementation (which is faulty).

Furthermore, sorting via `np.sort` instead of doing it by hand for all columns via `argsort` results in a 3x slower sorting.

Since ranking is for now not happening on super huge arrays, that should be fine.

JoanFM (approved):

PR #1035, opened 2020-10-06
feat: add evaluation flow and first ranking evaluators

JoanFM:
**Proposal for Evaluation Flow**
Addition to `jina.proto`:
  - Add EvaluateRequest
  - Add to the documents `Groundtruth` (similar to matches) and `evaluations` (multiple evaluations)

Addition to executors:
   - Add a new family of executor `evaluators`, special focus on `RankingEvaluators`

Addition to drivers:
   - Add a new driver `RankingEncodeDriver` that passes matches and groundtruth to executor

**Challenges**
- This can be useful to obtain single evaluation for a given document, how to return averaged and maybe reset the metrics of an executor.

**Future work**
As @maximilianwerk said, it would be cool to let the Evaluation Part not interfere with the search to be able to sometimes keep live stats, this would mean that the Evaluation Peas do not need to be merged at the end of the Flow (If we want to support this seems to deserve a separate PR)

**TODO**
Precision and RecallEvaluators are there to prove the intention of this work, a next step should involve moving them to the hub


jina/drivers/evaluate.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class EvaluateDriver(BaseExecutableDriver):
+    """Drivers inherited from this Driver will bind :meth:`evaluate` by default
+    """
+
+    def __init__(self, executor: str = None, traversal_paths: Tuple[str] = ('r',), method: str = 'evaluate', *args,
+                 **kwargs):
+        super().__init__(executor, method, traversal_paths=traversal_paths, *args, **kwargs)
+        self._is_apply = True
```
maximilianwerk, line 17: `False`?

jina/executors/evaluators/__init__.py
```diff
@@ -0,0 +1,28 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+from typing import Iterable
+from jina.proto import jina_pb2
+
+from .. import BaseExecutor
+
+
+class BaseEvaluator(BaseExecutor):
```
maximilianwerk, line 10: I think you have to adapt the `register_class` function in the `ExecutorType` meta class and add `evaluate` as another `prof_funcs`

maximilianwerk (commented):
I think we should differentiate between evaluation as suggested here and tracking live KPIs (e.g. for an A/B test or regular monitoring) which would be rather fire-and-forget messages to pods and therefor would need another ending as you suggested in *Future work*.

The in-memory count must not only be averages, but also the concrete scores. You cannot compute some KPIs with just averages (See NDCG).

jina/executors/evaluators/__init__.py
JoanFM: yes yes I have to do a lot of work, I just pretend to have a starting point of discussion

jina/drivers/evaluate.py
JoanFM: > I think we should differentiate between evaluation as suggested here and tracking live KPIs (e.g. for an A/B test or regular monitoring) which would be rather fire-and-forget messages to pods and therefor would need another ending as you suggested in _Future work_.
> 
> The in-memory count must not only be averages, but also the concrete scores. You cannot compute some KPIs with just averages (See NDCG).

When I mean average, is to aggregate all NDCG of all queries to have a unique value for the system. It can be flexible but I do not see where it is not applicable that u want the average on all queries

jina/executors/evaluators/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+        self.id_tag = id_tag
+
+    def evaluate(self, matches: Iterable[jina_pb2.Document],
+                 groundtruth: Iterable[jina_pb2.Document], *args, **kwargs) -> float:
+        """"
+        :param matches: the matched documents from the request as matched by jina indexers and rankers
+        :param groundtruth: the expected documents matches sorted as they are expected
+        :return the evaluation metric value for the request document
+        """
+        raise NotImplementedError
```
bwanglzu, line 28: some thoughts:
1.  I don't know how can we will implement the evaluate, will there be a metric@k attached to the `evaluate` method?
2.  `Iterable ` could be tricky, `set` is an `Iterable` object while not keep the order of the document. `Sequence` is something we expected.
3.  A bit curious about the reason why we don't use [abc](https://docs.python.org/3/library/abc.html)

bwanglzu (changes requested):

jina/executors/evaluators/__init__.py
JoanFM: Point 1. As I envision it, there will be multiple evaluator executors that wull receive k as parameter and will compute the metricm so one RecallEvaluator, One PrecisionEvaluator, on NDCGEvaluator ...

Point 2. Yes that makes sense it may not be the correct interface.

Point 3. I also would prefer using abc rather than raising exceptions

bwanglzu: @JoanFM great! Looks good if we can update the type hints. Point 3 could be discussed later :)

JoanFM: What is the highest abstraction that respects ordering?

bwanglzu: frankly I have never done multiple metaclass inherence before, might be a bad idea to do so. I see [this](https://stackoverflow.com/a/57351066) one, might be good for future reference.

bwanglzu:
[This](https://gist.github.com/bwhite/3726239/a8a107c5beab169c6efebcce24cd22ac2b5e7c73) could potentially be a time saver :) If you like I could contribute some evaluators

JoanFM:
> [This](https://gist.github.com/bwhite/3726239/a8a107c5beab169c6efebcce24cd22ac2b5e7c73) could potentially be a time saver :) If you like I could contribute some evaluators

Wait for when we move them to the hub ... but yes seems like a good resource

jina/executors/evaluators/recall.py
```diff
@@ -0,0 +1,38 @@
...
+    def evaluate(self, matches: Sequence[jina_pb2.Document],
+                 groundtruth: Sequence[jina_pb2.Document], *args, **kwargs) -> float:
+        """"
+        :param matches: the matched documents from the request as matched by jina indexers and rankers
+        :param groundtruth: the expected documents matches sorted as they are expected
+        :return the evaluation metric value for the request document
+        """
+        ret = 0.0
+        matches_ids = list(map(lambda x: x.tags[self.id_tag], matches))
+        groundtruth_ids = list(map(lambda x: x.tags[self.id_tag], groundtruth[:self.eval_at]))
```
bwanglzu, line 28: one thing might worth notice is we might have `IndexError: list index out of range` here.

JoanFM: Yes need to do more in depth testing

JoanFM: It does not do have IndexError, it is tested with `eval_at` 100 while only 5 documents as groundtruth

bwanglzu: ok

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, doc: 'jina_pb2.Document', groundtruth: 'jina_pb2.Document'):
+        self.doc = doc
+        self.groundtruth = groundtruth
+
+    def matches(self):
+        assert len(self.doc.matches) == len(self.groundtruth.matches)
+        return [DocGroundTruthTupleWrapper(doc, groundtruth) for doc, groundtruth in
+                zip(self.doc.matches, self.groundtruth.matches)]
+
+    def chunks(self):
```
JoanFM, line 27: This should be property

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,79 @@
...
+    It is important to note that it checks the matching structure of `docs` and `groundtruths`. It is important while
+    traversing to ensure that then the driver can be applied at a comparable level of granularity and adjacency.
+    This does not imply that you can't compare at the end a document with 10 matches with a groundtruth with 20 matches
+    """
+
+    def __init__(self, doc: 'jina_pb2.Document', groundtruth: 'jina_pb2.Document'):
+        self.doc = doc
+        self.groundtruth = groundtruth
+
+    def matches(self):
```
JoanFM, line 22: This should be property

jina/clients/python/__init__.py
```diff
@@ -187,26 +188,35 @@ def req_gen():
...
-    def index(self, input_fn: Union[Iterator['jina_pb2.Document'], Iterator[bytes], Callable] = None,
+    def index(self, input_fn: Union[Iterator[Union['jina_pb2.Document', bytes]], Callable]= None,
               output_fn: Callable[['jina_pb2.Message'], None] = None, **kwargs) -> None:
         self.mode = ClientMode.INDEX
         self.input_fn = input_fn
         if not self.args.skip_dry_run:
             self.dry_run(as_request='index')
         self.start(output_fn, **kwargs)
+
+    def eval(self, input_fn: Union[Iterator[Tuple[Union['jina_pb2.Document', bytes], Union['jina_pb2.Document', bytes]]], Callable] = None,
```
hanxiao, line 215: `eval` is Python built-in keywords, maybe change to `evaluate`?

jina/clients/python/__init__.py
```diff
@@ -187,26 +188,35 @@ def req_gen():
...
+             output_fn: Callable[['jina_pb2.Message'], None] = None, **kwargs) -> None:
+        self.mode = ClientMode.EVAL
+        self.input_fn = input_fn
+        if not self.args.skip_dry_run:
+            self.dry_run(as_request='eval')
+        print(f' EVAL hey')
```
hanxiao, line 221:
```suggestion
```

jina/clients/python/request.py
```diff
@@ -117,14 +117,35 @@ def _generate(data: Union[
...
+                               override_doc_id=override_doc_id
+                               )
+                groundtruth = getattr(req, str(mode).lower()).groundtruths.add()
+                _fill_document(document=groundtruth,
+                               content=content[1],
+                               docs_in_same_batch=batch_size,
+                               mime_type=mime_type,
+                               buffer_sniff=buffer_sniff,
+                               override_doc_id=override_doc_id
+                               )
```
hanxiao, line 148: this looks unnecessary to me so far (reviewing in top-down manner) and problematic:
- it calls `_fill_document` two times, may give the same doc different doc id.
- it should `_fill_document` once and the groundtruth is just a copy of it.

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,58 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+from typing import Iterable
+from .helper import DocGroundTruthPair
```
hanxiao, line 5: `Groundtruth` can be used as one word without capital `T` https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=groundtruth&btnG=

jina/drivers/helper.py
```diff
@@ -152,3 +152,30 @@ def guess_mime(uri):
         tmp = urllib.request.urlopen(page)
         m_type = tmp.info().get_content_type()
     return m_type
+
+
+class DocGroundTruthPair:
```
hanxiao, line 157: i see, pretty smart. but how about other attributes, `.embedding`, `.text`, etc?

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,58 @@
...
+    def _apply_all(self,
+                   docs_groundtruths: Iterable[DocGroundTruthPair],
+                   *args,
+                   **kwargs) -> None:
+        for doc_groundtruth in docs_groundtruths:
+            doc = doc_groundtruth.doc
+            groundtruth = doc_groundtruth.groundtruth
+            evaluation = doc.evaluations.add()
+            matches_ids = list(map(lambda x: x.tags[self.id_tag], doc.matches))
+            groundtruth_ids = list(map(lambda x: x.tags[self.id_tag], groundtruth.matches))
```
hanxiao, line 56: These two lines are basically query language: Given a path, extract all info.

jina/drivers/helper.py
JoanFM: As I see, this is helpful to accompany the pair during the traversal trip, (Then we can leave the responsability to the specific drivers to interpret the specific parts they want to extract?)

jina/drivers/evaluate.py
hanxiao: ref: jina/drivers/querylang/filter.py

jina/clients/python/request.py
JoanFM: note that the groundtruth is filled from different `content` (content[1]) since they come here in Tuples. I do agree it is quite ugly though

jina/drivers/evaluate.py
JoanFM: Adding a `mixin` class just to map from document to a single value seems overkilling?

jina/executors/evaluators/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+
+class BaseEvaluator(BaseExecutor):
+    """A :class:`BaseEvaluator` is used to evaluate different messages coming from any kind of executor
+    """
+
+    def __init__(self, *args, **kwargs):
+        """"
+        :param id_tag: the key in the tags of documents to identify uniquely a document
+        """
+        super().__init__(*args, **kwargs)
```
hanxiao, line 18:
```suggestion
    def __init__(self, *args, **kwargs):
        """"
        :param id_tag: the key in the tags of documents to identify uniquely a document
        """
        super().__init__(*args, **kwargs)
```

`id_tag` is missing or this part is unnecessary

jina/executors/evaluators/decorators.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+
+    @wraps(func)
+    def arg_wrapper(self, *args, **kwargs):
+        f = func(self, *args, **kwargs)
+        self.num_documents += 1
+        self.sum += f
+        return f
+
+    return arg_wrapper
```
hanxiao, line 19: I don't see how this decorator can be used in general. maybe it's better to just write them inline?

jina/drivers/evaluate.py
JoanFM: > ref: jina/drivers/querylang/filter.py

It is not filtering in this case, it is just getting identifiers so that the `executor` is abstracted from document complexity. (Ranking Evaluators tend to work only with list of `ids` regardless of what they represent.

jina/executors/evaluators/decorators.py
JoanFM: I would remove this part until we have clear vision on how we want the user to interact with `averages` and `running metrics`.  How would a user `reset` the count to 0? I would personally leave the metrics part on the client side

jina/flow/__init__.py
```diff
@@ -795,6 +795,47 @@ def my_reader():
         """
         self._get_client(**kwargs).search(input_fn, output_fn, **kwargs)
 
+    @deprecated_alias(buffer='input_fn', callback='output_fn')
+    def eval(self, input_fn: Union[Iterator[Tuple[Union['jina_pb2.Document', bytes], Union['jina_pb2.Document', bytes]]], Callable] = None,
```
hanxiao: also `eval_lines`, `eval_numpy`, ...

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,58 @@
...
+        self._traverse_apply(docs_groundtruths, *args, **kwargs)
+
+    def _apply_all(self, docs_groundtruths: Iterable[DocGroundTruthPair],
+                   context_doc_groundtruth: DocGroundTruthPair,
+                   *args,
+                   **kwargs) -> None:
+        pass
+
+
+class RankingEvaluationDriver(BaseEvaluationDriver):
```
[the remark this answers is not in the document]
JoanFM: We could say it is a `MatchingIdsComparationEvaluationDriver` if we want to abstract to functionality?

jina/flow/__init__.py
JoanFM: could we add them as we add new evaluations and executors for specific use cases? So that we can have better testing?

jina/proto/jina.proto
```diff
@@ -110,6 +110,19 @@ message Document {
...
+    // List of evaluations performed on the document
+    repeated Evaluation evaluations = 23;
+}
+
+/**
+ * Represents the result of an Evaluation.
+ */
+message Evaluation {
+    // identifier to identify how the evaluation was performed (several evaluations may come together)
+    string id = 1;
```
hanxiao, line 123: please refer it to `Relevance`, so as a score it can be self-contained and explainable

```proto
message Relevance {
    float value = 1; // value
    string op_name = 2; // the name of the operator/score function
    string description = 3; // text description of the score
    repeated Relevance operands = 4; // the score can be nested
    string ref_id = 5; // the score is computed between doc `id` and `ref_id`
}
```

hanxiao (changes requested):
in general, it's pretty good. 👍  One major thing is the "_given pb, extract path pattern_" in the evaluation driver. Two thoughts:
- make it more general? so `EvaluationDriver` differs only in `path`.
- reusing what we have in `jina/drivers/querylang`?

jina/executors/evaluators/decorators.py
hanxiao: >how we want the user to interact with averages and running metrics

kill the flow/pod and restart

jina/clients/python/request.py
hanxiao: but there is no point for feeding `content[0] & [1]`. You just feed groundtruth. That's it. The inference procedure won't use them and will override it. This is important.

jina/executors/evaluators/decorators.py
JoanFM: Won't it affect the user experience as seen as a user may want to keep the Flow running and evaluate just few requests and then keep searching and then evaluating and so....?

Would you agree to remove this part and handle it in a different PR?

jina/clients/python/request.py
JoanFM: Well,

If I feed a groundtruth with expected `matches` and copy to `docs` I see the problem that every SearchDriver calls `doc.matches.add()` without clearing the document matches. 

So what part of the groundtruth do I need to dismiss? I think it is safer to feed them separately

hanxiao (approved):

jina/clients/python/request.py
hanxiao: > If I feed a groundtruth with expected matches and copy to docs I see the problem that every SearchDriver calls doc.matches.add() without clearing the document matches.

Then this is the driver problem.

hanxiao: but let's keep it in this way. I think this may require quite some refactoring just for enabling this `single-input` feature

JoanFM: Well if we let the driver clean the matches then u cannot collect matches from different indexers

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def _apply_all(self,
+                   docs_groundtruths: Iterable[DocGroundtruthPair],
+                   *args,
+                   **kwargs) -> None:
+        for doc_groundtruth in docs_groundtruths:
+            doc = doc_groundtruth.doc
+            groundtruth = doc_groundtruth.groundtruth
+            evaluation = doc.evaluations.add()
+            matches_ids = list(map(lambda x: x.tags[self.id_tag], doc.matches))
```
maximilianwerk, line 55: This should rather be a list comprehension, then a `lambda` function. In my opinion 
```
matches_ids = [x.tags[self.id_tag] for x in doc.matches] 
```
is way better readable.

jina/drivers/evaluate.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def id(self):
+        if self.pea:
+            return self.pea.name
+        else:
+            return self.__class__.__name__
+
+    def _apply_all(self,
+                   docs_groundtruths: Iterable[DocGroundtruthPair],
```
maximilianwerk, line 48: Since it overwrites an super function, if should still be named `docs`, not? Imagine the iterator once will call `self._apply_all(docs=docs)` then this function will break.

jina/drivers/helper.py
```diff
@@ -152,3 +152,30 @@ def guess_mime(uri):
...
+    def __init__(self, doc: 'jina_pb2.Document', groundtruth: 'jina_pb2.Document'):
+        self.doc = doc
+        self.groundtruth = groundtruth
+
+    @property
+    def matches(self):
+        # TODO: Should we expect this assert to be done
+        #  (RankingEvaluation may work with a different lenght of groundtruth matches as the one returned)
+        assert len(self.doc.matches) == len(self.groundtruth.matches)
+        return [DocGroundtruthPair(doc, groundtruth) for doc, groundtruth in
```
maximilianwerk, line 174: Not sure, whether this is an issue, but in case the requests that are send are getting huge, we can heavily reduce the memory footprint by doing:
```
for doc, groundtruth in zip(...):
    yield DocGroundtruthPair(doc, groundtruth)
```

maximilianwerk (commented):
Sorry for the late review. Some hints to look out for on the next iteration on this topic.

jina/drivers/evaluate.py
JoanFM: Good point, or is thee also an alias?

jina/drivers/helper.py
JoanFM: It can be done yes

jina/drivers/evaluate.py
bhavsarpratik: Map with lambda is slower than comprehension as per a SO discussion

jina/drivers/evaluate.py
maximilianwerk: Alias? You mean an alias decorator? I have never seen one.

JoanFM: Similar to deprecated?

maximilianwerk: Sure, you could do that. Or you could just do
```
def _apply_all(self, docs: Iterable...):
    self.check_ground_truth(docs)

def check_ground_truth(self, docs_groundtruths: Iterable...):
   assert ...
```
Which would be more verbose and better readable, since it makes the mapping from `docs` to `docs_groundtruths` very explicit.

Issue #1030, opened 2020-10-05
Redirection of links in docs not working


bwanglzu:
I was looking at the change logs [here](https://docs.jina.ai/chapters/CHANGELOG.html), the redirection of links such as RELEASE NOTE 0.0.4 is not working in my browser.

This also happens [here](https://docs.jina.ai/chapters/CONTRIBUTING.html#-making-your-first-submission).

Tested on safari & chrome.

danielkomaz:
I updated the links within the CHANGELOG.md and CONTRIBUTING.md because they were i the wrong format "001" instead of "0-0-1"
Also fixed a few other broken links in hub/build-pod-image.md and hub/publish-your-pod-image.md

jancijen:
This remains a problem because there  is a GitHub worflkow (see commits with message `chore(docs): update TOC`) which automatically re-generate `CONTRIBUTING.md` and `CHANGELOG.md`.

PR #1032, opened 2020-10-06
docs: fix markdown links in CHANGELOG.md and CONTRIBUTING.md

jancijen:
This PR fixes issue #1030.

Before, the markdown links worked fine on GitHub and the ones in the documentation did not. I tried using a different format to identify markdown anchors, but then links did not work on GitHub and did work in the documentation. I kept the latter ones, because I found them easier to read (feel free to point out if you would like me to revert back to the old markdown anchor identifiers). And to make it work I manually inserted html hyperlink tags with matching names/identifiers (i.e. `<a name=...></a>`).

- fixes existing markdown links in `CHANGELOG.md` and `CONTRIBUTING.md`
- adjusts `realase.sh` script to add hyperlink tag when adding a "realease note" to `CHANGELOG.md`

PR #1037, opened 2020-10-06
docs: fix markdown html link names

jancijen:
Fixes issue #1030 and markdown html link names described in #1032 to match with the ones generated by https://github.com/technote-space/toc-generator in the GitHub action defined by [contributors.yml](https://github.com/jina-ai/jina/blob/master/.github/workflows/contributors.yml).

This anchor naming convention might be ambigious - e.g. `release-note-0111` might match be a heading with version `0.11.1` as well as `0.1.11`. This should be mitigated by appending `-1`, `-2`, ... to the anchor identifiers for the subsequent headings with the same anchor identifier. I tried it on a private repository and it seems that `toc-generator` generates those links correctly - i.e. it would be `release-note-0111`, `release-note-0111-1`, `release-note-0111-2`...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,7 +162,7 @@ Jina is released on every Friday evening. The PyPi package and Docker Image will
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-<a name="release-note-0-0-4"></a>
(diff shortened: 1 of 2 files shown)


maximilianwerk (approved):
LGTM

PR #1040, opened 2020-10-06
feat: try to see if syslog available

JoanFM:


hanxiao:
i think simply `try...except` this block would be sufficient?

```python
                if cfg['host'] and cfg['port']:
                    handler = SysLogHandlerWrapper(address=(cfg['host'], cfg['port']))
                else:
                    # a UNIX socket is used
                    if platform.system() == 'Darwin':
                        handler = SysLogHandlerWrapper(address='/var/run/syslog')
                    else:
                        handler = SysLogHandlerWrapper(address='/dev/log')
```

JoanFM:
> i think `try...except` this block would be sufficient?
> 
> ```python
>                 if cfg['host'] and cfg['port']:
>                     handler = SysLogHandlerWrapper(address=(cfg['host'], cfg['port']))
>                 else:
>                     # a UNIX socket is used
>                     if platform.system() == 'Darwin':
>                         handler = SysLogHandlerWrapper(address='/var/run/syslog')
>                     else:
>                         handler = SysLogHandlerWrapper(address='/dev/log')
> ```

I have seen inside the code and the constructor skips the error (it says that the port may not be available at construction time) and it is at the first log call that emthe error is raised. And @deepankarm exception is thrown when trying to log info and not when building the logger

hanxiao (approved):

PR #1051, opened 2020-10-08
Avoid passing whole doc to executor for BinaryPbIndexer

deepankarm:
- Changed invocation of `exec_fn` for `KVIndexDriver` & `KVSearchDriver`
- Test changes

### DIFF
--- a/jina/executors/indexers/keyvalue.py
+++ b/jina/executors/indexers/keyvalue.py
@@ -52,15 +52,19 @@ class BinaryPbIndexer(BaseKVIndexer):
         self._start = 0
         self._page_size = mmap.ALLOCATIONGRANULARITY
 
-    def add(self, docs: Iterator['jina_pb2.Document'], *args, **kwargs):
-        for d in docs:
-            s = d.SerializeToString()
-            l = len(s)  #: the length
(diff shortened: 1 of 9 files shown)


jina/executors/indexers/keyvalue.py
```diff
@@ -52,15 +52,19 @@ def __init__(self, *args, **kwargs):
         self._start = 0
         self._page_size = mmap.ALLOCATIONGRANULARITY
 
-    def add(self, docs: Iterator['jina_pb2.Document'], *args, **kwargs):
-        for d in docs:
-            s = d.SerializeToString()
-            l = len(s)  #: the length
+    def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
```
JoanFM, line 55: Keys is an iterator of string since change in doc id

JoanFM (changes requested):

jina/executors/indexers/keyvalue.py
deepankarm: Since we are passing `id2hash` as key, that would be an iterator of int

JoanFM: Good point, just ignore it then

JoanFM (approved):

PR #1062, opened 2020-10-11
feat: check whether a document has been already indexed

### LINKED ISSUE #1055: check whether a doc has already been indexed
**Describe the feature**
check whether a doc has already been indexed

**Your proposal**

---

**Environment**

**Screenshots**

nan-wang:
resolved #1055 

**Problem**
To support incremental indexing, we need a way to check whether a `Document` has been already indexed or not.

**Solution**
- Add a new driver `IncrementIndexDriver` together with a new executor `DuplicateChecker` for checking whether a `Document` has been already indexed or not. The `DuplicateChecker` maintains a set of indexed Document ids.

**Remaining Issues**
- [ ] Avoid the differences of using `DuplicateChecker` between single-shard and multi-shards usages. For the single-shard usage, one needs to construct a `CompoundIndexer` consist of `DuplicateChecker` and the `VectorIndexer`/`KVIndexer`. While for the multi-shards, one needs to put `DuplicateChecker` into the `uses_before`. This is because the `uses_before` does not play any role when `shards==1`. 
- [ ] Make `DuplicateChecker` scalable to the billion scales. Currently, a python `set()` is used to store the ids of the indexed Documents. This implementation takes too much memory.
- [ ] Fix the `RouteDriver` so that one can use `RouteDriver` without bounding to `ControlReqRequest`. Currently, change the `executors._route.yml` to the following will break the flow. 
```
!BaseExecutor
with: {}
metas:
  name: route
requests:
  on:
    IndexRequest:
      - !RouteDriver {}
    ControlRequest:
      - !RouteDriver {}
    SearchRequest:
      - !RouteDriver {}
```


JoanFM (commented):
Could you please explain the details of implementation, and how the feature is achieved?

jina/proto/jina.proto
```diff
@@ -113,6 +113,8 @@ message Document {
 
     // List of evaluations performed on the document
     repeated Evaluation evaluations = 23;
+
+    bool indexed = 24;
```
JoanFM, line 117: Since I guess this is internal fron jina and not suppoded to be exposed outside, we could use some reserved keys from tags like (__index__), we could do that with more things, for internal optimizations while hiding it from the external user

hanxiao: agree, in fact not necessary to change protobuf, just directly write in tags `tags['is_indexed'] = True`

jina/executors/indexers/helper.py
```diff
@@ -0,0 +1,52 @@
...
+        for id in id_list:
+            _indexed = (id in self.indexed)
+            is_indexed.append(_indexed)
+            if not _indexed:
+                self.indexed.add(id)
+        return is_indexed
+
+    def add(self, ids: Iterator[int], *args, **kwargs):
+        is_indexed = self._filter_ids(ids)
+        return is_indexed
```
hanxiao, line 36: I'm not a fan of this part. Can we use on-disk storage. Please note that `KVIndexer` and `BinaryPbIndexer` spend quite some effort on *not storing any data in memory*, hence their memory consumption is completely constant during indexing.

jina/proto/jina.proto
nan-wang: Good point!

jina/executors/indexers/helper.py
nan-wang: Definitely, I've listed this in the **Remaining Issues** and will start working on this.

jina/drivers/index.py
```diff
@@ -39,6 +40,28 @@ class KVIndexDriver(BaseIndexDriver):
...
     def _apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
-        keys = [uid.id2hash(doc.id) for doc in docs]
-        values = [doc.SerializeToString() for doc in docs]
+        keys = []
+        values = []
+        for doc in docs:
+            if doc.indexed:
+                continue
+            keys.append(uid.id2hash(doc.id))
+            values.append(doc.SerializeToString())
```
hanxiao, line 49: list comprehension, `[... for doc in docs if doc.tags.get('is_indexed', False)]`

hanxiao:
why this PR affects `RouteDriver`?

>  Fix the RouteDriver so that one can use RouteDriver without bounding to ControlReqRequest. Currently, change the executors._route.yml to the following will break the flow.

hanxiao:
I'm now confused by the overall idea. Why can't each (shard) indexer just say, _I have this key stored already, I will ignore it_? because this is basically what this PR does. @JoanFM @nan-wang

JoanFM:
> I'm now confused by the overall idea. Why can't each (shard) indexer just say, _I have this key stored already, I will ignore it_? because this is basically what this PR does. @JoanFM @nan-wang

Well to avoid other shards to index documents already present in another shard?

JoanFM:
I would rename this DuplicateChecker to something more generic because in the future it may evolve into helping in scalability, speed of the indexing, for instance it could in the future know in which shard a key is stored, or it could be used as a caching layer ....
 Does it make sense?

hanxiao:
yes, I think `CacheExecutor` makes sense, we can also introduce, MRU, LRU, FIFO cache

hanxiao:
> Well to avoid other shards to index documents already present in another shard?

okay yes, now I remember. because shards are working in different `workspace`.

hanxiao:
btw, if it is a Cache, then I would strongly go for driver, not the executor. With the on-disk `cache` path as an arg of the driver

JoanFM:
> btw, if it is a Cache, then I would strongly go for driver, not the executor. With the on-disk `cache` path as an arg of the driver

Well it can be seen as a cache, but for that I was thiking more as a potential evolution. 

I have no strong opinion abiut it being a Driver or Executor. It just feels strange for the Driver to persist data and with an Executor one can already profit of the mechanisms of dumping and persisting and reconstructing provided by Executors?

hanxiao:
how about this? https://github.com/jina-ai/jina/pull/1064

JoanFM:
> how about this? #1064

I like the idea, just curious about if it is easy to reconstruct? The logic to load from binary and stuff, is it properly handled at driver level?
 On the other hand having an executor in front of indexers may help to put a building block if we ever want to avoid dumping executors to disk '.bin'. This cohkd be a cache and metaindex executor holding general information of an indexer. (The part that I do not like of this executor idea is that makes 2 executors depenant on each other)

JoanFM:
> how about this? #1064

Removing the document from requst may be too much? Better to flag it in case it needs to be indexed or procrssed by downstream Pods?

hanxiao (changes requested):
I would recommend putting this PR on hold, as essentially we need a bloom filter. I will do it in #1064

hanxiao:
https://github.com/jina-ai/jina/pull/1064 is done. please consider using `BloomFilterDriver` as BloomFilter is the standard solution for `has_key` method in DB domain

hanxiao:
This PR is depreciated given #1070

hanxiao:
#1070 takes the tests from #1062 and improves it on the following aspects:
- #1070 does not change Protobuf, instead, it uses `tags` in our Protobuf schema
- #1070 does not change `KVIndexDriver` and `VectorIndexDriver` to unselect hit-cache-docs, instead, it uses `FilterQL` to filter them out

PR #1064, opened 2020-10-11
feat(driver): add CacheIndexer and its Driver

hanxiao:
Bloom filter to test whether a doc is observed or not based on its ``doc.id``

### DIFF
--- a/jina/drivers/cache.py
+++ b/jina/drivers/cache.py
@@ -1,6 +1,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
+import math
 import os
 import struct
 from typing import Iterable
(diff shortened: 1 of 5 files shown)


JoanFM:
Removing the document from requst may be too much? Better to flag it in case it needs to be indexed or procrssed by downstream Pods?

hanxiao:
> Removing the document from requst may be too much? Better to flag it in case it needs to be indexed or procrssed by downstream Pods?

yes, just an example, the `hit-cache` behavior can be changed.

hanxiao:
something like `on_hit`, `on_miss`

jina/drivers/cache.py
```diff
@@ -1,49 +1,88 @@
...
 if False:
     from ..proto import jina_pb2
 
 
-class UniqueDocDriver(BaseRecursiveDriver):
-    """ Remove docs that previously have seen in the cache.
-
-    :class:UniqueDocDriver: follows a simple rule: it stores all new doc.id in a cache file.
-    If a doc.id hits the cache, then it is removed from the request.
+class BloomFilterDriver(BaseRecursiveDriver):
```
JoanFM, line 15: Is the plan to have 2 implementations? Bloom Filters can lead to false positives, thus some documents may not be indexed. This should at least be warned?

hanxiao: add a property to show FP rate. note, this very simple naive bloomfilter, there are more advanced bloomfilters. the general direction is correct, I will leave the extensions to others

hanxiao: > Is the plan to have 2 implementations? 

second thought, you are right. I will add a `set`-based checker

hanxiao:
i'm suspending this PR for now, as it makes more sense to implement Cache as an IndexExecutor so that we can leverage existing index interface

PR #1065, opened 2020-10-11
feat: add pretrained model exception and accept it while test_uses

JoanFM:
**Changes introduced** 
Idea to enable passing PR in the hub such as https://github.com/jina-ai/jina-hub/pull/372 and https://github.com/jina-ai/jina-hub/pull/448 where the hub is not properly built because test fails, because of the incapacity to provide a default setting for pretrained models. 

How can we make sure that the Exception that the Flow sees is that one and not PeaFailsToStart?

**Future work**
Once agreed, all those executors on the hub depending on a pretrained model path should raise that exception when the path is not existing.

**Problem**
Exceptions during `start` or post_init are not well passed to the Flow and outside


JoanFM:
Tries to offer a solution for https://github.com/jina-ai/jina-hub/issues/452

nan-wang (approved):
Good catch!

hanxiao (approved):

PR #1068, opened 2020-10-12
fix: avoid the bounding RouteDriver to the ControlRequest

nan-wang:
resolved #1067 

Currently, only a unit test is added to illustrate the issue.

PR #1070, opened 2020-10-12
feat: check whether a document has been already indexed

hanxiao:
this PR is an improvement & replacement for #1062 

## Improvement over #1062 

#1070 takes the tests from #1062 and improves this PR on the following aspects:
- #1070 does not change Protobuf, instead, it uses `tags` in our Protobuf schema
- #1070 does not change `KVIndexDriver` and `VectorIndexDriver` to unselect hit-cache-docs, instead, it uses `FilterQL` and add it to the YAML level.
- #1070 adds `DocIdCache(KVIndexer)` and `BaseCacheDriver(BaseIndexDriver)`

## New high-level feature

- shortcut executor `_unique` is introduced, one can use it in `--uses-before` when indexing to prevent repeated indexing

## Caveats
#1068 test code is included in this PR but the problem is not solved. #1068 wrongly diagnosed the problem, real issue is posted in #1071

### DIFF
--- a/jina/resources/executors._unique.yml
+++ b/jina/resources/executors._unique.yml
@@ -1,0 +1,15 @@
+!DocIDCache
+with:
+  index_path: cache.tmp
+requests:
+  on:
+    [SearchRequest, TrainRequest, IndexRequest, ControlRequest]:
+      - !RouteDriver {}
+    IndexRequest:
+      - !TaggingCacheDriver
+        with:
+          tags:
+            is_indexed: true
+      - !FilterQL
+        with:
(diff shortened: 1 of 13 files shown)


jina/resources/executors._unique.yml
```diff
@@ -0,0 +1,15 @@
+!DocIDCache
+with:
+  index_path: cache.tmp
+requests:
+  on:
+    [SearchRequest, TrainRequest, IndexRequest, ControlRequest]:
```
JoanFM, line 6: Add EvaluateRequest

PR #1075, opened 2020-10-13
fix: slack links

alexcg1:
Previously I changed Slack link to redirect straight to our #general channel. However, that meant folks couldn't sign up (since they didn't have an invite)

I've now changed it back to an invite link to fix this


bwanglzu:
interesting .. code coverage increased..

JoanFM:
> interesting .. code coverage increased..

That's why I say that I am not convinced by our settings.

I think that overall code coverage is correct, but I think that the reports of individual increases and deltas has some problems

nan-wang (approved):
LGTM👍

PR #1084, opened 2020-10-14
style: autoblack on pr

maximilianwerk:
Testing automatically black on a PR.


.github/workflows/ci.yml
```diff
@@ -19,8 +19,34 @@ jobs:
...
+        if: failure()
+        run: |
+          black .
+          git config --global user.name 'autoblack'
+          git config --global user.email 'maximilianwerk@users.noreply.github.com'
+          echo $GITHUB_HEAD_REF
+          echo $GITHUB_REPOSITORY
+          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/jina-ai/jina
+          git checkout $GITHUB_HEAD_REF
+          git commit -am "style: automatic codeformating with black"
```
JoanFM, line 45: would it make sense just to ammend to last commit?

JoanFM: or it will be disturbing for the user to mess with their commits?

.github/workflows/ci.yml
```diff
@@ -19,8 +19,34 @@ jobs:
...
+      - name: Install Black
+        run: pip install black
+      - name: Run black --check .
+        run: black --check .
+      - name: If needed, commit black changes to the pull request
+        if: failure()
+        run: |
+          black .
+          git config --global user.name 'autoblack'
+          git config --global user.email 'maximilianwerk@users.noreply.github.com'
```
JoanFM, line 40: don't we have some `jina` account?

.github/workflows/ci.yml
maximilianwerk: I like the idea. I'll test it

.github/workflows/ci.yml
maximilianwerk: I'll have a look. On the other hand, if we amend we might skip this completely and keep whoever committed beforehand. I'll have a look

.github/workflows/ci.yml
maximilianwerk: I am not even sure if this works on a PR. If a user forks our repository, we cannot commit to their fork. So perhaps, it is better to only black on a merge.

.github/workflows/ci.yml
JoanFM: I would go for ammending but I would do it right before merge in CD.yml and not in CI.yml. (Appending commits in the middle of development process can be annoying for the developer?) Like they mayu continue working without pulling this commit and may get messy and conflicts they may not understand?

.github/workflows/ci.yml
JoanFM: yes black on a merge I think is the way to go

PR #1087, opened 2020-10-14
feat(flow): add eval() interface generates hanging pod

hanxiao:
`.eval()` will add an empty hanging Pod to the flow. No evaluation-related logic is implemented in this PR.

Examples:

```python
f = Flow().add(name='p1').add(name='p2', needs='gateway').needs(['p1', 'p2']).eval().plot(build=True)
```

image

```python
f = Flow().add(name='p1', parallel=3).eval().add(name='p2', needs='gateway').needs(['p1', 'p2']).eval().plot(build=True)
```

image

```python
f = Flow().add().eval().add().eval().add().eval().plot()
```

image

### DIFF
--- a/tests/unit/flow/test_flow_eval_pod.py
+++ b/tests/unit/flow/test_flow_eval_pod.py
@@ -49,7 +49,8 @@ def test_flow2():
         with open(f'tmp{j}.txt') as fp:
             assert fp.read() != ''
 
-# @pytest.mark.skip('can not support this topology for now')
+
+@pytest.mark.skip('can not support this topology for now')
 def test_flow3():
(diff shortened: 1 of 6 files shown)


tests/unit/flow/test_flow_eval_pod.py
```diff
@@ -23,14 +25,14 @@ class DummyEvaluator3(DummyEvaluator1):
 docs = list(random_docs(1))
 
 
-def test_flow1():
+def test_flow1(tmpdir):
```
JoanFM, line 28: tmpdir I think does nothing here. Because nothing forces the `BaseExecutor` to have `tmpdir` as workspace folder

hanxiao: yes

PR #1088, opened 2020-10-14
ci: add test job for examples in cd

bhavsarpratik:
Added a new job in cd.yml to run tests for each example in jina-ai/examples


.github/workflows/cd.yml
```diff
@@ -12,143 +15,183 @@ on:
...
+  #         JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
+  #         JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
+  #     - name: Upload coverage result from integration-test
+  #       if: ${{ matrix.python-version }} == 3.7 && ${{ matrix.os }} == ubuntu-latest
+  #       uses: actions/upload-artifact@v1
+  #       with:
+  #         name: integration-test-codecoverage
+  #         path: ./integration-test-coverage.xml
+
+  examples-test:
```
JoanFM, line 194: I would suggest to put first in the `ci.yml` so that at every commit or push forced it will be executed and u can see if the action works, then later it can be moved to the `cd.yml`. @maximilianwerk  was doing this for his black formating PR

bhavsarpratik: Yes same thoughts!

bhavsarpratik: Check the added `examples-test` job and let me know if looks ok for testing

hanxiao (dismissed):
please do not modify the current `cd.yml` in bulk, I can't tell the difference. Add a new workflow yaml file instead and bind to the correct events

.github/workflows/cd.yml
```diff
@@ -175,49 +218,43 @@ jobs:
           docker build  -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
           export JINA_LOG_VERBOSITY="ERROR"
           pip install .[test]
-          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=120 --ignore-glob='tests/integration/hub_usage/dummyhub*' -v tests/integration
-          mv coverage.xml integration-test-coverage.xml
-        timeout-minutes: 30
+          git clone https://github.com/jina-ai/examples.git --branch test-examples-cicd
```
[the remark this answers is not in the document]
bhavsarpratik: Yup

.github/workflows/ci-example.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[match-py-ver] --no-cache-dir
+      - name: Test with pytest
+        run: |
+          # make sure docker exist
+          docker ps
+          jina check
+          docker build  -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
```
JoanFM, line 33: u do not need this one

.github/workflows/ci-example.yml
```diff
@@ -0,0 +1,41 @@
...
+          export JINA_LOG_VERBOSITY="ERROR"
+          pip install .[test]
```
JoanFM, line 35: I am not sure this is needed either, but may be good to keep it

.github/workflows/ci-example.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          submodules: true
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[match-py-ver] --no-cache-dir
```
JoanFM, line 27: I do not think we need these dependencies

.github/workflows/ci-example.yml
JoanFM: Unless we test an example running based on a jina docker image but not the case now

.github/workflows/ci-example.yml
bhavsarpratik: I took these from the ci.yml and I think upgrading pip is a good idea. What is the last line doing?

.github/workflows/ci-example.yml
bhavsarpratik: Ok removing docker related lines

.github/workflows/ci-example.yml
JoanFM: I think is installing `jina` with `py37` or `py38` dependencies, I think this will be done later to install from master. So it is one or the other in my opinion

bhavsarpratik:
@JoanFM I will move the new ci-example job to cd.yml. There are some tests in some examples which fail. Is that ok? 
```
jobs:
  examples-test:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest] #[macos-latest, ubuntu-latest]
        python-version: [3.7, 3.8]

    steps:
      - uses: actions/checkout@v2
        with:
          submodules: true
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v1
        with:
          python-version: ${{ matrix.python-version }}
      - name: Install dependencies
        run: |
          pip install .[match-py-ver] --no-cache-dir
      - name: Test example with pytest
        run: |
          jina check
          export JINA_LOG_VERBOSITY="ERROR"
          pip install .[test]
          git clone https://github.com/jina-ai/examples.git
          pytest --force-flaky --min-passes 1 -n 1 --timeout=120 -v examples
        timeout-minutes: 60
        env:
          JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
          JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
```

.github/workflows/cd.yml
```diff
@@ -221,3 +221,34 @@ jobs:
           env_vars: OS,PYTHON
           name: codecov
           fail_ci_if_error: true
+
+  examples-test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest] #[macos-latest, ubuntu-latest]
```
JoanFM, line 229: I would just have them using `python3.7` for simplicy

.github/workflows/cd.yml
```diff
@@ -221,3 +221,34 @@ jobs:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[match-py-ver] --no-cache-dir
+      - name: Test example with pytest
+        run: |
+          jina check
+          export JINA_LOG_VERBOSITY="ERROR"
+          pip install .[test]
+          git clone https://github.com/jina-ai/examples.git
+          pytest --force-flaky --min-passes 1 -n 1 --timeout=0 -v examples
```
JoanFM, line 250: put some quite large timeout, but put something, otherwise it can be blocked for a really long time

bhavsarpratik: I kept it 0 so that pytest-timeout doesn't limit but actions will limit by timeout-minutes: 60

JoanFM (approved):

bhavsarpratik:
@hanxiao I think this requires your approval to merge

hanxiao:
> CI / unit-test (ubuntu-latest, 3.8) (pull_request) Failing after 16m — unit-test (ubuntu-latest, 3.8)
right now it's still failed

bhavsarpratik:
The changes in this PR were tested with a separate yml and now it's merged to cd.yml which will not be triggered in this PR. The current failure of test is as below.

FAILED tests/unit/test_container.py::MyTestCase::test_flow_with_replica_container_ext_yaml
====== 1 failed, 400 passed, 18 skipped, 6 warnings in 714.38s (0:11:54) =======

hanxiao:
ok then please retry or find out the problem unit it is merged

JoanFM (approved):

PR #1089, opened 2020-10-15
feat: potentially add argument of ordered indexes from batch

JoanFM:
**Changes introduced**
From a bug found in `NmsLibIndexer` I found that some indexers may not only need to have `batched` data but also would need information about which `ordinal` indexers correspond to each data in a batch.

This PR tries to enable https://github.com/jina-ai/jina-hub/pull/458

**Extra change**
I have not seen where the `label` is used in the `batching` decorator and therefore I try to remove it since it results in a lot of `if - else` logic. @hanxiao @nan-wang do you know in which context it is used?

### DIFF
--- a/jina/executors/decorators.py
+++ b/jina/executors/decorators.py
@@ -125,6 +125,7 @@ def batching(func: Callable[[Any], np.ndarray] = None, *,
              split_over_axis: int = 0,
              merge_over_axis: int = 0,
              slice_on: int = 1,
+             label_on: Optional[int] = None,
              ordinal_idx_arg: Optional[int] = None) -> Any:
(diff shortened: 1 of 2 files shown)


hanxiao:
label is useful for batching (x, y) training pair

JoanFM:
> label is useful for batching (x, y) training pair

Ok, will revert that part and add test about that, currently no test covers that

hanxiao (approved):

JoanFM:
I add an skipped test to prove the point and to have some kind of TDD approach for the time when this item will need to be added

hanxiao (approved):

PR #1093, opened 2020-10-15
perf: numpy indexer use partial sort at query time

clennan:
Hi, I have recently read your great blog article on your numpy indexer (https://hanxiao.io/2020/09/21/Numpy-Tricks-and-A-Strong-Baseline-for-Vector-Index/) and saw that at query time you do a full sort on the distances to retrieve the top k candidates.

This PR suggests using numpy's partial sort implementation `np.argpartition` which runs in O(n) versus O(n * log(n)) for `np.argsort`. 

With this code change the query method does not return the top k candidates ordered like with full sort before, but it seems the system is not relying on the top k candidates being ordered, at least the docstring does not imply it.

This little example illustrates the difference, the speedup for top 10 retrieval from a 100k array is around 12 fold
```python
# ipython 
import numpy as np

top_k = 10
dist = np.random.rand(100_000)

%%timeit 
idx_sort = dist.argsort()[:top_k]
# 9.4 ms ± 169 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
idx_partition = dist.argpartition(kth=top_k)[:top_k]
# 779 µs ± 2.85 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

idx_sort
# array([50384,  7687, 23873, 29129, 38318,   710, 51471, 81669, 47933, 61278])

idx_partition
# array([51471,   710, 47933, 38318, 50384, 61278, 81669, 29129, 23873, 7687])
```

I also ran the benchmarking script mentioned in the blog post (https://gist.github.com/hanxiao/43cad33b60cadd34f45236993689e5d9), the speedup is from 2.07s per run to 1.82s. When increasing the number of indexed documents from 10k to 100k and reducing their dimension from 10k to 1k, the speedup is from 4.05s per run to 3.14s.

Would be great to get your feedback on this :)

### DIFF
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -245,10 +245,10 @@ class NumpyIndexer(BaseNumpyIndexer):
         """ Find top-k smallest distances in ascending order.
 
         Idea is to use partial sort to retrieve top-k smallest distances unsorted and then sort these
-        in ascending order. Equivalent to full sort but faster for n >> k. If n == k revert to full sort.
+        in ascending order. Equivalent to full sort but faster for n >> k. If k >= n revert to full sort.
 
         """
-        if top_k == dist.shape[1]:
+        if top_k >= dist.shape[1]:
             idx = dist.argsort(axis=1)[:, :top_k]
             dist = np.take_along_axis(dist, idx, axis=1)
         else:
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -240,8 +240,28 @@ class NumpyIndexer(BaseNumpyIndexer):
         self.metric = metric
         self.backend = backend
 
+    @staticmethod
+    def _find_smallest_distances(dist: 'np.array', top_k: int) -> Tuple['np.ndarray', 'np.ndarray']:
+        """ Find top-k smallest distances in ascending order.
+
+        Idea is to use partial sort to retrieve top-k smallest distances unsorted and then sort these
+        in ascending order. Equivalent to full sort but faster for n >> k. If n == k revert to full sort.
+
+        """
+        if top_k == dist.shape[1]:
+            idx = dist.argsort(axis=1)[:, :top_k]
+            dist = np.take_along_axis(dist, idx, axis=1)
+        else:
+            idx_ps = dist.argpartition(kth=top_k, axis=1)[:, :top_k]
+            dist = np.take_along_axis(dist, idx_ps, axis=1)
+            idx_fs = dist.argsort(axis=1)
+            idx = np.take_along_axis(idx_ps, idx_fs, axis=1)
+            dist = np.take_along_axis(dist, idx_fs, axis=1)
+
+        return idx, dist
+
     def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
-        """ Find the top-k vectors with smallest ``metric`` and return their ids.
+        """ Find the top-k vectors with smallest ``metric`` and return their ids in ascending order.
 
         :return: a tuple of two ndarray.
             The first is ids in shape B x K (`dtype=int`), the second is metric in shape B x K (`dtype=float`)
@@ -263,8 +283,7 @@ class NumpyIndexer(BaseNumpyIndexer):
         else:
             raise NotImplementedError(f'{self.metric} is not implemented')
 
-        idx = dist.argpartition(kth=top_k, axis=1)[:, :top_k]
-        dist = np.take_along_axis(dist, idx, axis=1)
+        idx, dist = self._find_smallest_distances(dist, top_k)
         return self.int2ext_id[idx], dist
 
     def build_advanced_index(self, vecs: 'np.ndarray'):
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -263,7 +263,7 @@ class NumpyIndexer(BaseNumpyIndexer):
         else:
             raise NotImplementedError(f'{self.metric} is not implemented')
 
-        idx = np.argpartition(dist, kth=top_k, axis=1)[:, :top_k]
+        idx = dist.argpartition(kth=top_k, axis=1)[:, :top_k]
         dist = np.take_along_axis(dist, idx, axis=1)
         return self.int2ext_id[idx], dist
 
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -263,7 +263,7 @@ class NumpyIndexer(BaseNumpyIndexer):
         else:
             raise NotImplementedError(f'{self.metric} is not implemented')
 
-        idx = dist.argsort(axis=1)[:, :top_k]
+        idx = np.argpartition(dist, kth=top_k, axis=1)[:, :top_k]
         dist = np.take_along_axis(dist, idx, axis=1)
         return self.int2ext_id[idx], dist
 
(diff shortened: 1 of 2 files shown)


JoanFM:
Hey @clennan ,

Thank you very much for your proposal, unfortunately, we already thought about this and we decided not to go for this solution at this moment as it adds the requirement to have the Ranker enforce ordering with not so clear benefit.

You can see the conversation at https://github.com/jina-ai/jina/pull/964,

You are welcome to change the docstring to show in a more clear way this limitation.

JoanFM:
Hey @clennan,

I repoen the issue since I think maybe with this benchmarking of yours we could see if we decide to go for this direction. Sorry for the rushed reaction!

clennan:
ah, did not see the previous discussion, that's 1:1 to what I did here 😅

As mentioned in #964, the speedup only becomes significant when the index is large. Is there a reason why the benchmark was only done on a 10k index? I guess in practice the indexed data will be much larger and any optimisations would be targeted at those use cases anyways.

I agree that a full sort on the top k documents can be added to return sorted documents from the query method, should still be computationally much less expensive than a full sort on entire index.

JoanFM:
> ah, did not see the previous discussion, that's 1:1 to what I did here 😅
> 
> As mentioned in #964, the speedup only becomes significant when the index is large. Is there a reason why the benchmark was only done on a 10k index? I guess in practice the indexed data will be much larger and any optimisations would be targeted at those use cases anyways.
> 
> I agree that a full sort on the top k documents can be added to return sorted documents from the query method, should still be computationally much less expensive than a full sort on entire 

Adding first this argpartition and then sorting the top k in real order is a good solution, but then please provide some benchmark to ensure that this actually improve performance and that the extra cost of selecting the top_k is compensated by the sorting complexity gain.

JoanFM:
Another interesting point could be adding a mode parameter to the indexing to know if it is a `match` or `ranking` index, so that the user can choose depending on wether it has a ranker downstream or it heavily relies in semantic distance. Any thoughts @hanxiao @bhavsarpratik @nan-wang @deepankarm ?

clennan:
Good morning, I have added the "hybrid" solution where a partial sort is done first and then a full sort on the top k. Here are some benchmarks:

image
image


The y-axis measures the time difference between full sort (previous implementation) and partial sort in milliseconds, i.e. it is positive if the suggested new implementation is faster. The significant speedups occur if the number of documents (x-axis) in the index, and thus the number of calculated distances, is greater than 1million.

image
image
image


If the number of documents is below 5k the new implementation can be slightly slower with small batch sizes. I would argue that this performance loss is outweighed by the performance gains for a larger index or batch sizes.

The benchmark script can be found here 
https://gist.github.com/clennan/052115edf63a160ba160d081460df1d9

jina/executors/indexers/vector.py
```diff
@@ -263,8 +283,7 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
         else:
             raise NotImplementedError(f'{self.metric} is not implemented')
 
-        idx = dist.argsort(axis=1)[:, :top_k]
-        dist = np.take_along_axis(dist, idx, axis=1)
+        idx, dist = self._find_smallest_distances(dist, top_k)
```
JoanFM, line 286: like it, but I would call `_get_sorted_top_k`?

JoanFM (commented):
Looks good, just minor comment!

JoanFM:
Thanks for your contribution. Unfortunately, as you didn't set up your local git config correctly, the author of your commit does not match with your Github account. This poses problems:

- Your contribution can not be credited, and will not show up in the list of contributors (as bad git config = anonymous)

- We can not accept CLA for anonymous contribution.

Please note that this does not only affect your contribution to Jina, but also affects all your contribution to anything on Github. So better fix it, once for all.
Here is what you can do:
```
git config user.name "Correct new name"
git config user.email "correct@new.email" -> match with your Github registered email address
```
Then fix the commit message by
```
git commit --amend --author="Author Name <email@address.com>" --no-edit
git push --force
```

JoanFM:
> Good morning, I have added the "hybrid" solution where a partial sort is done first and then a full sort on the top k. Here are some benchmarks:
> 
> image
> image
> 
> The y-axis measures the time difference between full sort (previous implementation) and partial sort in milliseconds, i.e. it is positive if the suggested new implementation is faster. The significant speedups occur if the number of documents (x-axis) in the index, and thus the number of calculated distances, is greater than 1million.
> 
> image
> image
> image
> 
> If the number of documents is below 5k the new implementation can be slightly slower with small batch sizes. I would argue that this performance loss is outweighed by the performance gains for a larger index or batch sizes.
> 
> The benchmark script can be found here
> https://gist.github.com/clennan/052115edf63a160ba160d081460df1d9

good analysis, thanks!

bhavsarpratik:
Thanks @clennan for this addition! LGTM 👍 
@JoanFM Do we want to handle `n < k` - happens when we do testing by indexing a few docs but give a bigger k by mistake.

JoanFM:
> Thanks @clennan for this addition! LGTM 
> @JoanFM Do we want to handle `n < k` - happens when we do testing by indexing a few docs but give a bigger k by mistake.

Good point, we should definitely handle it, I think it is already handled by this, but we should definitely add a `test` covering this case as extra security!. (Not important in that case what branch it goes to)

jina/executors/indexers/vector.py
```diff
@@ -240,8 +240,28 @@ def __init__(self, metric: str = 'euclidean',
...
 
+    @staticmethod
+    def _find_smallest_distances(dist: 'np.array', top_k: int) -> Tuple['np.ndarray', 'np.ndarray']:
+        """ Find top-k smallest distances in ascending order.
+
+        Idea is to use partial sort to retrieve top-k smallest distances unsorted and then sort these
+        in ascending order. Equivalent to full sort but faster for n >> k. If n == k revert to full sort.
+
+        """
+        if top_k == dist.shape[1]:
```
JoanFM, line 251: changing to `top_k >= dist.shape[1]` should make it more clear that it works if `top_k` is greater than the indexed documents.

clennan:
> Thanks for your contribution. Unfortunately, as you didn't set up your local git config correctly, the author of your commit does not match with your Github account. This poses problems:
> 
> * Your contribution can not be credited, and will not show up in the list of contributors (as bad git config = anonymous)
> * We can not accept CLA for anonymous contribution.
> 
> Please note that this does not only affect your contribution to Jina, but also affects all your contribution to anything on Github. So better fix it, once for all.
> Here is what you can do:
> 
> ```
> git config user.name "Correct new name"
> git config user.email "correct@new.email" -> match with your Github registered email address
> ```
> 
> Then fix the commit message by
> 
> ```
> git commit --amend --author="Author Name <email@address.com>" --no-edit
> git push --force
> ```

Amended all three commits

jina/executors/indexers/vector.py
clennan: Very good point, will add this and also add test case

jina/executors/indexers/vector.py
clennan: fine for me :)

JoanFM (approved):

JoanFM:
recheckcla

JoanFM (changes requested):
Need to sign CLA

clennan:
> **Jina CLA check**
> 
> ❤️ Thank you for your pull request. It looks like this is your first contribution to an open source project maintained by Jina AI Limited. Before we can look at your pull request, we kindly ask that you sign our [Contributor License Agreement](https://github.com/jina-ai/jina/blob/master/CLA.md). You can sign it by commenting in format below.
> 
> _**I have read the CLA Document and I hereby sign the CLA**_

JoanFM:
recheckcla

JoanFM (approved):

PR #1098, opened 2020-10-16
feat: add evaluation encoder and crafter executors and drivers

JoanFM:
**Changes introduced**

- Added new `EncodeEvaluationDriver` and `CraftEvaluationDriver`

- Added new family of `BaseEncodingEvaluator` executors family with `CosineEvaluator` and `EuclideanEvaluator` (which sould be later moved to the `hub`)

- Added new family of `BaseCraftingEvaluator` executors family with `BaseNlpCraftingEvaluator` and `LengthStringEvaluator`  (which sould be later moved to the `hub`, or removed because it is just a dummy example)

### DIFF
--- a/jina/drivers/evaluate.py
+++ b/jina/drivers/evaluate.py
@@ -22,7 +22,7 @@ class BaseEvaluationDriver(BaseExecutableDriver):
         self._traverse_apply(docs_groundtruths, *args, **kwargs)
 
     @property
-    def id(self):
+    def metric(self):
         if self.pea:
             return self.pea.name
         else:
@@ -64,7 +64,7 @@ class RankingEvaluationDriver(BaseEvaluationDriver):
             matches_ids = [x.tags[self.id_tag] for x in doc.matches]
             groundtruth_ids = [x.tags[self.id_tag] for x in groundtruth.matches]
             evaluation.value = self.exec_fn(matches_ids, groundtruth_ids)
-            evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
+            evaluation.op_name = f'{self.metric}-{self.exec.metric}'
             evaluation.ref_id = groundtruth.id
 
 
@@ -86,7 +86,7 @@ class EncodeEvaluationDriver(BaseEvaluationDriver):
             groundtruth = doc_groundtruth.groundtruth
             evaluation = doc.evaluations.add()
             evaluation.value = self.exec_fn(pb2array(doc.embedding), pb2array(groundtruth.embedding))
-            evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
+            evaluation.op_name = f'{self.metric}-{self.exec.metric}'
             evaluation.ref_id = groundtruth.id
 
 
@@ -117,5 +117,5 @@ class CraftEvaluationDriver(BaseEvaluationDriver):
             if isinstance(gt_content, jina_pb2.NdArray):
                 gt_content = pb2array(gt_content)
             evaluation.value = self.exec_fn(doc_content, gt_content)
-            evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
+            evaluation.op_name = f'{self.metric}-{self.exec.metric}'
             evaluation.ref_id = groundtruth.id
--- a/jina/drivers/evaluate.py
+++ b/jina/drivers/evaluate.py
@@ -20,6 +20,13 @@ class BaseEvaluationDriver(BaseExecutableDriver):
                              zip(self.req.docs, self.req.groundtruths)]
         self._traverse_apply(docs_groundtruths, *args, **kwargs)
 
+    @property
+    def id(self):
+        if self.pea:
+            return self.pea.name
+        else:
+            return self.__class__.__name__
+
     def _apply_all(self, groundtruth_pairs: Iterable['DocGroundtruthPair'],
                    context_groundtruth_pair: 'DocGroundtruthPair',
                    *args,
@@ -28,7 +35,7 @@ class BaseEvaluationDriver(BaseExecutableDriver):
 
 
 class RankingEvaluationDriver(BaseEvaluationDriver):
-    """Drivers inherited from this Driver will bind :meth:`evaluate` by default
+    """Drivers used to pass `matches` from documents and groundtruths to an executor and add the evaluation value
     """
 
     def __init__(self,
@@ -43,13 +50,6 @@ class RankingEvaluationDriver(BaseEvaluationDriver):
         """
         super().__init__(*args, **kwargs)
         self.id_tag = id_tag
-
-    @property
-    def id(self):
-        if self.pea:
-            return self.pea.name
-        else:
-            return self.__class__.__name__
 
     def _apply_all(self,
                    groundtruth_pairs: Iterable[DocGroundtruthPair],
@@ -68,26 +68,13 @@ class RankingEvaluationDriver(BaseEvaluationDriver):
 
 
 class EncodeEvaluationDriver(BaseEvaluationDriver):
-    """Drivers inherited from this Driver will bind :meth:`evaluate` by default
+    """Drivers used to pass `embedding` from documents and groundtruths to an executor and add the evaluation value
     """
 
     def __init__(self,
                  *args,
                  **kwargs):
-        """
-
-        :param id_tag: the name of the tag to be extracted
-        :param args:
-        :param kwargs:
-        """
         super().__init__(*args, **kwargs)
-
-    @property
-    def id(self):
-        if self.pea:
-            return self.pea.name
-        else:
-            return self.__class__.__name__
 
     def _apply_all(self,
                    groundtruth_pairs: Iterable[DocGroundtruthPair],
--- a/jina/drivers/evaluate.py
+++ b/jina/drivers/evaluate.py
@@ -5,6 +5,7 @@ from typing import Iterable
 
 from . import BaseExecutableDriver
 from .helper import DocGroundtruthPair, pb2array
+from jina.proto import jina_pb2
 
 
 class BaseEvaluationDriver(BaseExecutableDriver):
@@ -85,5 +86,36 @@ class EncodeEvaluationDriver(BaseEvaluationDriver):
             groundtruth = doc_groundtruth.groundtruth
             evaluation = doc.evaluations.add()
             evaluation.value = self.exec_fn(pb2array(doc.embedding), pb2array(groundtruth.embedding))
+            evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
+            evaluation.ref_id = groundtruth.id
+
+
+class CraftEvaluationDriver(BaseEvaluationDriver):
+    """Drivers used to pass a content field from documents and groundtruths to an executor and add the evaluation value
+    """
+
+    def __init__(self,
+                 field: str = 'text',
+                 *args,
+                 **kwargs):
+        super().__init__(*args, **kwargs)
+        self.field = field
+
+    def _apply_all(self,
+                   groundtruth_pairs: Iterable[DocGroundtruthPair],
+                   *args,
+                   **kwargs) -> None:
+        for doc_groundtruth in groundtruth_pairs:
+            doc = doc_groundtruth.doc
+            groundtruth = doc_groundtruth.groundtruth
+            evaluation = doc.evaluations.add()
+
+            doc_content = getattr(doc, self.field)
+            gt_content = getattr(groundtruth, self.field)
+            if isinstance(doc_content, jina_pb2.NdArray):
+                doc_content = pb2array(doc_content)
+            if isinstance(gt_content, jina_pb2.NdArray):
+                gt_content = pb2array(gt_content)
+            evaluation.value = self.exec_fn(doc_content, gt_content)
             evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
             evaluation.ref_id = groundtruth.id
--- a/jina/drivers/evaluate.py
+++ b/jina/drivers/evaluate.py
@@ -4,7 +4,7 @@ __license__ = "Apache-2.0"
 from typing import Iterable
 
 from . import BaseExecutableDriver
-from .helper import DocGroundtruthPair
+from .helper import DocGroundtruthPair, pb2array
 
 
 class BaseEvaluationDriver(BaseExecutableDriver):
@@ -63,5 +63,40 @@ class RankingEvaluationDriver(BaseEvaluationDriver):
             matches_ids = [x.tags[self.id_tag] for x in doc.matches]
             groundtruth_ids = [x.tags[self.id_tag] for x in groundtruth.matches]
             evaluation.value = self.exec_fn(matches_ids, groundtruth_ids)
+            evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
+            evaluation.ref_id = groundtruth.id
+
+
+class EncodeEvaluationDriver(BaseEvaluationDriver):
+    """Drivers inherited from this Driver will bind :meth:`evaluate` by default
+    """
+
+    def __init__(self,
+                 *args,
+                 **kwargs):
+        """
+
+        :param id_tag: the name of the tag to be extracted
+        :param args:
+        :param kwargs:
+        """
+        super().__init__(*args, **kwargs)
+
+    @property
+    def id(self):
+        if self.pea:
+            return self.pea.name
+        else:
+            return self.__class__.__name__
+
+    def _apply_all(self,
+                   groundtruth_pairs: Iterable[DocGroundtruthPair],
+                   *args,
+                   **kwargs) -> None:
+        for doc_groundtruth in groundtruth_pairs:
+            doc = doc_groundtruth.doc
+            groundtruth = doc_groundtruth.groundtruth
+            evaluation = doc.evaluations.add()
+            evaluation.value = self.exec_fn(pb2array(doc.embedding), pb2array(groundtruth.embedding))
             evaluation.op_name = f'{self.id}-{self.exec.metric_name}'
             evaluation.ref_id = groundtruth.id
--- a/jina/executors/evaluators/craft/nlp/__init__.py
+++ b/jina/executors/evaluators/craft/nlp/__init__.py
@@ -1,0 +1,21 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+from typing import Any
+from .. import BaseEvaluator
+
+
+class BaseNlpCraftingEvaluator(BaseEvaluator):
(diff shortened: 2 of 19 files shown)


bwanglzu (commented):
This design makes me feel a bit consued. I'm supposed `evaluator` is a different concept compared with `metric`. Now we're kind of coupled them together.

To be more concrete,I think a evaluator should be able to contains multiple metrics, for me, it's something like a one-n relationship between evaluator and metric. Now we have
```python
class StringLengthEvaluator(BaseEvaluator):
    """A :class:`StringLengthEvaluator` evaluates the different lengths between doc and groundtruth text
    """

    @property
    def metric_name(self):
        return f'Length'

    @as_aggregator
    def evaluate(self, doc_content: str, groundtruth_content: str, *args, **kwargs) -> float:
        """"
        :param doc_content: the text of the document
        :param groundtruth_content: the expected text of the document
        :return the evaluation metric value for the request document
        """
        return abs(len(doc_content) - len(groundtruth_content))
```

While it is strange, because metric should be an independent object, for example:
```python
# We have a generalised evaluation driver
class EvaluationDriver(BaseExecutableDriver):
    # do something
    @property 
    def metrics(self):
        return [#some metrics here]

    @metrics.setter
    def metrics(self, metrics: List[Metric]):
       for metric in metrics:
           self.add_metric(metric)

    def add_metric(self, other_metric: Metric):
        if not isinstance(other_metric, Metric):
            raise TypeError(".....")
        self._metrics.append(metric)
    
    def evaluate(self, doc_content, groundtruth_content):
        eval_result = []
        for metric in self._metrics:
            eval_result.append(metric.mearesure())
        return eval_result



# Then we have a Metric object
class BaseMetric(ABCMeta):
    @property
    @abc.abstractmethod
    def name(self):
        pass
    
    @abc.abstractmethod
    def mearesure(self, doc, groundtruth):
        pass

# Then implement, for example, LengthMetric
class StringLengthMetric(BaseMetric):

    @property
    def name(self):
        return 'Length'
    
    def mearesure(self, doc_content, groundtruth_content):
        return abs(len(doc_content) - len(groundtruth_content))

```

JoanFM:
> This design makes me feel a bit consued. I'm supposed `evaluator` is a different concept compared with `metric`. Now we're kind of coupled them together.
> 
> To be more concrete,I think a evaluator should be able to contains multiple metrics, for me, it's something like a one-n relationship between evaluator and metric. Now we have
> 
> ```python
> class StringLengthEvaluator(BaseEvaluator):
>     """A :class:`StringLengthEvaluator` evaluates the different lengths between doc and groundtruth text
>     """
> 
>     def __init__(self, *args, **kwargs):
>         super().__init__(*args, **kwargs)
> 
>     @property
>     def metric_name(self):
>         return f'Length'  ## COMMENTS BY BO: this part we force to "merge" two concepts 
> 
>     @as_aggregator
>     def evaluate(self, doc_content: str, groundtruth_content: str, *args, **kwargs) -> float:
>         """"
>         :param doc_content: the text of the document
>         :param groundtruth_content: the expected text of the document
>         :return the evaluation metric value for the request document
>         """
>         return abs(len(doc_content) - len(groundtruth_content))
> ```
> 
> While it is strange, because metric should be an independent object, for example:
> 
> ```python
> # We have a generalised evaluation driver
> class EvaluationDriver(BaseExecutableDriver):
>     # do something
>     @property 
>     def metrics(self):
>         return [#some metrics here]
> 
>     @metrics.setter
>     def metrics(self, metrics: List[Metric]):
>        for metric in metrics:
>            self.add_metric(metric)
> 
>     def add_metric(self, other_metric: Metric):
>         if not isinstance(other_metric, Metric):
>             raise TypeError(".....")
>         self._metrics.append(metric)
>     
>     def evaluate(self, doc_content, groundtruth_content):
>         eval_result = []
>         for metric in self._metrics:
>             eval_result.append(metric.mearesure())
>         return eval_result
> 
> 
> 
> # Then we have a Metric object
> class BaseMetric(ABCMeta):
>     @property
>     @abc.abstractmethod
>     def name(self):
>         pass
>     
>     @abc.abstractmethod
>     def mearesure(self, doc, groundtruth):
>         pass
> 
> # Then implement, for example, LengthMetric
> class StringLengthMetric(BaseMetric):
> 
>     @property
>     def name(self):
>         return 'Length'
>     
>     def mearesure(self, doc_content, groundtruth_content):
>         return abs(len(doc_content) - len(groundtruth_content))
> ```
> 
> In the current design, we basically bind a metric to a driver, makes me feel the logic is not clear enough, and not flexible.

In this design, we keep the current `jina` design that one `Driver` is `driving` a single `executor`. Therefore one `driver` has only `one` metric at a singe point because it has only one `evaluator executor` (another thing is that u can chain different drivers and executors using `CompoundExectuors` or different `Pods`.

Linking one `evaluator` to run only a `metric` keeps our design `hub-friendly` in the sense that we can keep the evaluators as `hub executors` and users can develop their own evaluation logic.

bwanglzu:
ok, that make sense to me. It has nothing to do with right or wrong  :) just something in my mind on how to "model" the real world (entities).

jina/drivers/evaluate.py
```diff
@@ -65,3 +66,56 @@ def _apply_all(self,
...
+            groundtruth = doc_groundtruth.groundtruth
+            evaluation = doc.evaluations.add()
+
+            doc_content = getattr(doc, self.field)
+            gt_content = getattr(groundtruth, self.field)
+            if isinstance(doc_content, jina_pb2.NdArray):
+                doc_content = pb2array(doc_content)
+            if isinstance(gt_content, jina_pb2.NdArray):
+                gt_content = pb2array(gt_content)
+            evaluation.value = self.exec_fn(doc_content, gt_content)
```
bwanglzu, line 119: how can we handle the situation if one of the input is not `NdArray`? (I suppose we'll enocounter this situation because you did the check)

jina/executors/evaluators/craft/nlp/__init__.py
```diff
@@ -0,0 +1,21 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+from typing import Any
+from .. import BaseEvaluator
+
+
+class BaseNlpCraftingEvaluator(BaseEvaluator):
```
bwanglzu, line 8: what's the difference between `BaseCraftingEvaluator ` and `BaseNlpCraftingEvaluator `?

jina/executors/evaluators/craft/nlp/length.py
```diff
@@ -0,0 +1,26 @@
...
+class StringLengthEvaluator(BaseEvaluator):
+    """A :class:`StringLengthEvaluator` evaluates the different lengths between doc and groundtruth text
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    def metric_name(self):
+        return f'Length'
```
bwanglzu, line 17: you can remove the `f`

jina/drivers/evaluate.py
JoanFM: This is something that it happens all over jina, you will have to provide an `executor` that accepts the input the `driver` is providing. (I would like to have an early assert to check which drivers and executors are compatible, but it has been decided otherwise).

At this point, the driver just assumes that the `executor` has the function signature that accepts the provided arguments (See all drivers have a call to `self.exec_fn` without having any constraint on the kind of signature it contains. So there should be a family of `executors` accepting `buffer` another for `blob` and another for `text`.

I think you have a really good point but it should be considered in general for all jina drivers and in another scope.

tests/unit/drivers/test_craftevaluation_driver.py
```diff
@@ -0,0 +1,123 @@
...
+                gt.text = 'aaaa'
+            elif field_type == 'buffer':
+                doc.buffer = b'\x01\x02\x03'
+                gt.buffer = b'\x01\x02\x03\x04'
+            elif field_type == 'blob':
+                doc.blob.CopyFrom(array2pb(np.array([1, 1, 1])))
+                gt.blob.CopyFrom(array2pb(np.array([1, 1, 1, 1])))
+
+            pairs.append(DocGroundtruthPair(doc=doc, groundtruth=gt))
+        return pairs
```
bwanglzu, line 50: pairs should be a fixture

tests/unit/drivers/test_encodingevaluation_driver.py
```diff
@@ -0,0 +1,103 @@
...
+    def name(self):
+        return 'MockDiffEvaluator'
+
+    def evaluate(self, doc_embedding: 'np.array', groundtruth_embedding: 'np.array', *args, **kwargs) -> float:
+        """"
+        :param doc_embedding: the embedding of the document (resulting from an Encoder)
+        :param groundtruth_embedding: the expected embedding of the document
+        :return the evaluation metric value for the request document
+        """
+        return abs(sum(doc_embedding - groundtruth_embedding) / len(doc_embedding))
```
bwanglzu, line 22: this can be a fixture, because I see this class in three places.

bwanglzu (commented):
great work!

tests/unit/drivers/test_encodingevaluation_driver.py
bwanglzu: @JoanFM ok, because personally I think test should only consist code related to test, you should decoupke anything related to resources, instance construction outside the test itself. Fixture has two fold of benefits 1. reuse 2. more importantly, reflects the objective of **a function should do one thing, and only one thing. ** (quoted from clean code)

tests/unit/drivers/test_encodingevaluation_driver.py
```diff
@@ -0,0 +1,103 @@
...
+def test_encoding_evaluate_driver():
+    def create_document_ground_truth_pairs(num_docs):
+        pairs = []
+        for idx in range(num_docs):
+            doc = jina_pb2.Document()
+            gt = jina_pb2.Document()
+            doc.embedding.CopyFrom(array2pb(np.array([1, 1])))
+            gt.embedding.CopyFrom(array2pb(np.array([2, 2])))
+            pairs.append(DocGroundtruthPair(doc=doc, groundtruth=gt))
+        return pairs
```
bwanglzu, line 40: for example, here you're creating a test to test encoding evaluate driver, while put a nested function inside to create resources, I'd think create a fixture can be more clear (you decouple the resources from test itself). I'm not a big fan of nested functions.

tests/unit/drivers/test_encodingevaluation_driver.py
JoanFM: Ok, I see the point, makes a lot of sense. Thanks for the clarification !!! Good learning for me as well! Will change this!

tests/unit/drivers/test_encodingevaluation_driver.py
JoanFM: In the orher case, do I still need to define the class outside and pass an instance as fixture? or how would it be refactored?

tests/unit/drivers/test_rankingevaluation_driver.py
```diff
@@ -21,43 +22,51 @@ def evaluate(self, matches_ids, groundtruth_ids, *args, **kwargs) -> float:
```
bhavsarpratik, line 22: How about converting `groundtruth_ids` to `set` for faster lookup? It can be useful for high `eval_at`

tests/unit/drivers/test_rankingevaluation_driver.py
```diff
@@ -80,37 +89,45 @@ def req(self) -> 'jina_pb2.Request':
...
+            match.tags['id'] = idx
+
+    req = jina_pb2.Request.IndexRequest()
+    for idx in range(num_docs):
+        doc = req.docs.add()
+        gt = req.groundtruths.add()
+        chunk_doc = doc.chunks.add()
+        chunk_gt = gt.chunks.add()
+        chunk_doc.granularity = 1
+        chunk_gt.granularity = 1
```
bhavsarpratik, line 114: These lines are common in many tests

jina/executors/evaluators/craft/nlp/length.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @as_aggregator
+    def evaluate(self, doc_content: str, groundtruth_content: str, *args, **kwargs) -> float:
```
bhavsarpratik, line 20: Output can be `int`. Is float necessary?

jina/executors/evaluators/craft/nlp/length.py
```diff
@@ -0,0 +1,26 @@
...
+        """"
+        :param doc_content: the text of the document
+        :param groundtruth_content: the expected text of the document
+        :return the evaluation metric value for the request document
+        """
+        return abs(len(doc_content) - len(groundtruth_content))
```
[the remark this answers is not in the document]
JoanFM: It is just a `dummy evaluator`. The important point here is to have the infra and interfaces to have later these executors in the `hub` where we can add any logic as for encoders, crafters, etc ...

I did not want to add any dependency for the tests to run smoothly and this seemed easy to test. When we move it to hub we can even delete it

tests/unit/drivers/test_rankingevaluation_driver.py
JoanFM: it is important to keep the order which the set does not guarantee I think

jina/executors/evaluators/craft/nlp/length.py
JoanFM: It could be, but since the interface returns `float`

tests/unit/drivers/test_rankingevaluation_driver.py
JoanFM: I do not mind having `tests` self contained.

bwanglzu:
much better! while we can further improve, I'll give some concrete example

bwanglzu:
@JoanFM I would create test like this (pseudo code, not tested):
```python
@pytest.fixture(scope='function', params=['text', 'buffer', 'blob'])
def field_type(request):
    return request.param

@pytest.fixture(scope='function')
def doc_with_field_type(field_type):
    doc = jina_pb2.Document()
    if field_type == 'text':
        doc.text = 'aaa'
    elif field_type == 'buffer':
        doc.buffer = b'\x01\x02\x03'
    elif field_type == 'blob':
        doc.blob.CopyFrom(array2pb(np.array([1, 1, 1])))
    return doc

@pytest.fixture(scope='function')
def groundtruth_with_field_type(field_type):
    gt = jina_pb2.Document()
    if field_type == 'text':
        gt.text = 'aaa'
    elif field_type == 'buffer':
        gt.buffer = b'\x01\x02\x03'
    elif field_type == 'blob':
        gt.blob.CopyFrom(array2pb(np.array([1, 1, 1, 1])))
    return gt

@pytest.fixture(scope='function')
def doc_groundtruth_pair(request, doc_with_field_type, groundtruth_with_field_type):
    class DocGroundtruthPairFactory(object):
        def create(self):
            return DocGroundtruthPair(
                doc=doc_with_field_type,
                groundtruth=groundtruth_with_field_type
            )

@pytest.fixture(scope='function')
def ground_truth_pairs(doc_groundtruth_pair):
    doc_groundtruth_pairs = []
    for _ in range(10):
        doc_groundtruth_pairs.append(
            doc_groundtruth_pair.create()
        )
    return doc_groundtruth_pairs
```

JoanFM:
> @JoanFM I would create test like this (pseudo code, not tested):
> 
> ```python
> @pytest.fixture(scope='function', params=['text', 'buffer', 'blob'])
> def field_type(request):
>     return request.param
> 
> @pytest.fixture(scope='function')
> def doc_with_field_type(field_type):
>     doc = jina_pb2.Document()
>     if field_type == 'text':
>         doc.text = 'aaa'
>     elif field_type == 'buffer':
>         doc.buffer = b'\x01\x02\x03'
>     elif field_type == 'blob':
>         doc.blob.CopyFrom(array2pb(np.array([1, 1, 1])))
>     return doc
> 
> @pytest.fixture(scope='function')
> def groundtruth_with_field_type(field_type):
>     gt = jina_pb2.Document()
>     if field_type == 'text':
>         gt.text = 'aaa'
>     elif field_type == 'buffer':
>         gt.buffer = b'\x01\x02\x03'
>     elif field_type == 'blob':
>         gt.blob.CopyFrom(array2pb(np.array([1, 1, 1, 1])))
>     return gt
> 
> @pytest.fixture(scope='function')
> def doc_groundtruth_pair(request, doc_with_field_type, groundtruth_with_field_type):
>     class DocGroundtruthPairFactory(object):
>         def create(self):
>             return DocGroundtruthPair(
>                 doc=doc_with_field_type,
>                 groundtruth=groundtruth_with_field_type
>             )
> 
> @pytest.fixture(scope='function')
> def ground_truth_pairs(doc_groundtruth_pair):
>     doc_groundtruth_pairs = []
>     for _ in range(10):
>         doc_groundtruth_pairs.append(
>             doc_groundtruth_pair.create()
>         )
>     return doc_groundtruth_pairs
> ```

Let me check !

jina/drivers/evaluate.py
```diff
@@ -20,6 +21,13 @@ def __call__(self, *args, **kwargs):
                              zip(self.req.docs, self.req.groundtruths)]
         self._traverse_apply(docs_groundtruths, *args, **kwargs)
 
+    @property
+    def id(self):
```
hanxiao, line 25:
```suggestion
    def metric(self):
```

I refactored this name to metric in last PR

jina/drivers/evaluate.py
hanxiao: as Python does not do type check, what @bwanglzu proposed makes sense but has low priority. No need to bother adding type check now

jina/executors/evaluators/craft/nlp/length.py
```diff
@@ -0,0 +1,26 @@
...
+class StringLengthEvaluator(BaseEvaluator):
+    """A :class:`StringLengthEvaluator` evaluates the different lengths between doc and groundtruth text
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    def metric_name(self):
+        return 'Length'
```
hanxiao, line 17:
```suggestion
    def metric(self):
        return 'Length'
```

hanxiao (changes requested):
all in all it's good. but I don't see the necessity of differentiating `evaluator` per `executor`. This can be refactored in another PR.

JoanFM:
> all in all it's good. but I don't see the necessity of differentiating `evaluator` per `executor`. This can be refactored in another PR.

Would you do it by `field`?

hanxiao:
i would just do it by `metric`

JoanFM:
> i would just do it by `metric`

Ok, we can do when we move them to the `hub`?

hanxiao (approved):
👍 great job good to go

PR #1100, opened 2020-10-16
feat(driver): add collect evaluation driver

hanxiao:
- `.gather_inspect` now uses `_merge_eval` executor by default


JoanFM:
We should have 2 levels of grouping, one by `docID` and another by `evaluation` identifier. We must consider that every document may be evaluated by different evaluations in the same `Flow`

hanxiao:
do you mean by `op_name`? I think that should be straightforward by mocking what this driver does, will leave to the others

PR #1106, opened 2020-10-18
test: add docker helper test

bwanglzu:


codecov.yml
```diff
@@ -1,6 +1,3 @@
-codecov:
```
JoanFM, line 1: why this? should we isolate this in another PR?

bwanglzu: ok, I was trying to experiment what we discussed in triage codecov channel, I'll revert this one.

JoanFM: ah well, if it is for investigating is good, does it make any improvement?

bwanglzu: @JoanFM I didn't see, because python 3.9 ci failed and... coverage report not being uploaded

bwanglzu:
no..i messed up my pr, try to merge latest CI (without python 3.9 required)

PR #1111, opened 2020-10-20
feat: add file based evaluator

JoanFM:
**Changes introduced**
Enable the usage of Evaluator by loading the `groundtruth` from disk.

- Create a Driver `LoadGroundTruthDriver` which loads groundtruth from the keys obtained from `documents` and fills the `groundtruths` of the `requests`. These 2 Drivers with `KVIndexer` are intended to be used sequentially (with `CompoundExecutor` or `uses_before` together with some `evaluators`

- Crate `FileBasedEvaluator` shortcut to create a `CompoundExecutor` to provide this functionality as a unique executor (Useful mainly for documentation)


jina/drivers/evaluate.py
```diff
@@ -118,4 +119,57 @@ def _apply_all(self,
                 gt_content = pb2array(gt_content)
             evaluation.value = self.exec_fn(doc_content, gt_content)
             evaluation.op_name = f'{self.metric}-{self.exec.metric}'
-            evaluation.ref_id = groundtruth.id
\ No newline at end of file
+            evaluation.ref_id = groundtruth.id
+
+
+class GroundTruthKVIndexDriver(BaseIndexDriver):
```
hanxiao, line 125: i dont see the purpose of having a gtkvindexdriver, it is the same with KVindex

jina/drivers/evaluate.py
```diff
@@ -118,4 +119,57 @@ def _apply_all(self,
...
+        self._traversal_paths = ('r', )
+        self.id_tag = id_tag
+
+    def _apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
+        keys = [doc.tags[self.id_tag] for doc in docs]
+        values = [doc.SerializeToString() for doc in docs]
+        self.exec_fn(keys, values)
+
+
+class LoadGroundTruthDriver(BaseExecutableDriver):
```
hanxiao, line 139:
```suggestion
class LoadGroundTruthDriver(BaseExecutableDriver):
```

should inherit from KVSearchDriver

jina/drivers/evaluate.py
JoanFM: I agree that I can have a change to have the `BaseKVIndexDriver` access a specific `key`

jina/drivers/evaluate.py
```diff
@@ -118,4 +119,57 @@ def _apply_all(self,
...
+            evaluation.ref_id = groundtruth.id
+
+
+class GroundTruthKVIndexDriver(BaseIndexDriver):
+    """Serialize the documents/chunks in the request to key-value JSON pairs and write it using the executor
+    """
+    def __init__(self, id_tag: str = 'id', *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._traversal_paths = ('r', )
+        self.id_tag = id_tag
```
hanxiao, line 131: the motivation of using `tags[id]` instead of `.id` directly is you worry about `hash` gives them different id?

JoanFM: the motivation is that is hard for the `user` to have exact control on what `id` a specific document will receive, (plus he may evaluate with only some part of the document) (he may have stored just the embedding and not other parts). 

So I imagine the user may have a dataset such as:
ID, Ranking of IDs
1, [3, 5, 6

hanxiao: i feel there is different understanding on that, to me `actual` and `desired` are always eye-to-eye structural-alike. For example, 

```
actual:
    [doc.id: 1, matches: [doc.id: 45, doc.id 67, doc.id: 10],
    [doc.id: 3, matches: [doc.id: 32, doc.id 24, doc.id: 23]
    ...]
```

Then groundtruth should follow the same structure and *one only need to query for [1, 3]*:

```
desired:
    [doc.id: 1, matches: [doc.id: 45, doc.id 88, doc.id: 99],
    [doc.id: 3, matches: [doc.id: 32, doc.id 99, doc.id: 77]
    ...]
```

JoanFM: My idea was about allowing the `executor` not to know anything about the `document structure` so it can play with only basic types. And since he has no control on `ids` (maybe he just wants to index in the groundtruth his own `id` (but not the actual content))

hanxiao: > My idea was about allowing the executor not to know anything about the document structure so it can play with only basic types.

but this does not contradict the idea

JoanFM: so if we have `doc.id` it will be a `hash` that the `user` will have almost no control about (he will need to match exactly the same document).

(In many cases the groundtruth will come in form of user-related format and would be really hard for the user to convert into a document which could have the exact `hash-id`)

jina/drivers/evaluate.py
```diff
@@ -118,4 +119,36 @@ def _apply_all(self,
...
+        for idx, doc in enumerate(self.req.docs):
+            serialized_groundtruth = self.exec_fn(doc.tags[self.id_tag])
+            if serialized_groundtruth:
+                doc.tags['found'] = True
+                gt = self.req.groundtruths.add()
+                gt.ParseFromString(serialized_groundtruth)
+            else:
+                miss_idx.append(idx)
+            # delete non-existed matches in reverse
+        for j in reversed(miss_idx):
```
maximilianwerk, line 153: Is this really wanted? a match, that is not in the ground-truth might be important for evaluation, not?

jina/drivers/index.py
```diff
@@ -38,7 +38,18 @@ class KVIndexDriver(BaseIndexDriver):
...
+        """
+        super().__init__(*args, **kwargs)
+        self.id_tag = id_tag
+
     def _apply_all(self, docs: Iterable['jina_pb2.Document'], *args, **kwargs) -> None:
-        keys = [uid.id2hash(doc.id) for doc in docs]
+        if self.id_tag:
+            keys = [doc.tags[self.id_tag] for doc in docs]
+        else:
+            keys = [uid.id2hash(doc.id) for doc in docs]
```
hanxiao, line 53: hmm... changing the existing driver when adding new driver. something o avoid

jina/drivers/evaluate.py
hanxiao: but hash-id can be changed, it is not immutable.

jina/drivers/evaluate.py
JoanFM: I will clarify this

JoanFM: As you can see, it is not using `traversal` so it is only removing `documents` for which there is no `groundtruth`

jina/drivers/evaluate.py
JoanFM: Ok will revert this part and expect the user to provide an `id` they can control

jina/drivers/index.py
maximilianwerk: I agree. Might it be a good idea to add another `TagKVIndexDriver` instead?

JoanFM: I think it can be useful, but I will revert it to jsut using `id` and we can handle that in another PR

bhavsarpratik (dismissed):

jina/drivers/evaluate.py
JoanFM: Solved!

jina/drivers/evaluate.py
JoanFM: Solved!

jina/drivers/evaluate.py
JoanFM: Done!

jina/drivers/index.py
JoanFM: Reverted to only having `doc.id` and its hash as key, we can think about adding another driver in the future or another PR

nan-wang (changes requested):
Does it work for multi-shards?

jina/executors/evaluators/__init__.py
```diff
@@ -35,3 +35,50 @@ def std(self) -> float:
...
         return self._running_stats.variance
+
+
+class FileBasedEvaluator(CompoundExecutor):
+    """A Frequently used pattern for combining A :class:`BinaryPbIndexer` and :class:`BaseEvaluator`.
+     It will be equipped with predefined ``requests.on`` behaviors:
+
+         -  At evaluation time(query or index)
+             - 1. Checks for the incoming document, gets its value from the `BinaryPbIndexer` and fills the `groundtruth of the request
+             - 2. Filter the documents that do not have a corresponding groundtruth
```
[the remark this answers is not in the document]
JoanFM: > Does it work for multi-shards?

For `multishards` I think we need to use the `same` approach as for the `incremental indexing` ticket, but good point I will add a test for that.

JoanFM:
> Does it work for multi-shards?

Added tests for it

hanxiao (approved):

PR #1112, opened 2020-10-20
feat(pod): allow uses-before/after on single pea

hanxiao:
## New features

- Pod context manager now allows `--uses-before` and `--uses-after` when `--parallel = 1`. Previously `uses-before` and `uses-after` are only respected when `parallel > 1`.

With this new feature, you can build a sequential (microservice-like) quote `CompoundExecutor` unquote (`before -> exec -> after`)  simply via:

```python
f = Flow().add(uses_before='before.yml', uses_after='after.yml', uses='exec.yml')

with f:
    f.index(docs)
```

Another usage is to solve the `todo1` mentioned in #1062: 
> avoid the differences of using DuplicateChecker between single-shard and multi-shards usages. ...

This can now be implemented as 
```python
f = Flow().add(uses=os.path.join(cur_dir, 'docindexer.yml'), uses_before='_unique')
```
See `tests/integration/incremental_indexing/test_unique_indexing.py` for detailed examples.

## Changes (Non-breaking)

- `_pass` and  `_forward` use `RouteDriver` as default
- `ForwardDriver` is now a shortcut to `RouteDriver`

### DIFF
--- a/jina/peapods/pod.py
+++ b/jina/peapods/pod.py
@@ -77,7 +77,6 @@ class BasePod(ExitStack):
         if getattr(args, 'parallel', 1) > 1:
             # reasons to separate head and tail from peas is that they
             # can be deducted based on the previous and next pods
-            _set_after_to_pass(args)
             self.is_head_router = True
             self.is_tail_router = True
             peas_args['head'] = _copy_to_head_args(args, args.polling.is_push)
@@ -93,11 +92,9 @@ class BasePod(ExitStack):
                 peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args.get('head', None), peas_args.get('tail', None))
         else:
-            _set_after_to_pass(args)
             self.is_head_router = False
             self.is_tail_router = False
             peas_args['peas'] = [args]
-
 
         # note that peas_args['peas'][0] exist either way and carries the original property
         return peas_args
@@ -415,14 +412,6 @@ def _set_peas_args(args: Namespace, head_args: Namespace = None, tail_args: Name
             _args.host_out = _fill_in_host(bind_args=tail_args, connect_args=_args)
         result.append(_args)
     return result
-
-
-def _set_after_to_pass(args):
-    # TODO: I don't remember what is this for? once figure out, this function should be removed
-    # remark 1: i think it's related to route driver.
-    if hasattr(args, 'polling') and args.polling.is_push:
-        # ONLY reset when it is push
-        args.uses_after = '_pass'
 
 
 def _copy_to_head_args(args: Namespace, is_push: bool, as_router: bool = True) -> Namespace:
--- a/jina/peapods/pod.py
+++ b/jina/peapods/pod.py
@@ -38,11 +38,6 @@ class BasePod(ExitStack):
         self.deducted_head = None
         self.deducted_tail = None
 
-        # TODO: I don't remember what is this for???
-        # if hasattr(args, 'polling') and args.polling.is_push:
-        #     # ONLY reset when it is push
-        #     args.uses_after = '_pass'
-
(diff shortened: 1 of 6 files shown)


JoanFM:
I like the fact of unifying `uses_before` with a potential `CompoundExecutor` pattern! It must be however still be documented that both patterns are good, and the use of `CompoundExecutor` is still better for optimization purposes.

@rutujasurve94 this PR is interesting as per our earlier conversation

jina/peapods/pod.py
```diff
@@ -82,15 +80,27 @@ def _parse_args(self, args: Namespace) -> Dict[str, Optional[Union[List[Namespac
...
             peas_args['head'] = _copy_to_head_args(args, args.polling.is_push)
             peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args['head'], peas_args['tail'])
+            self.is_head_router = True
+            self.is_tail_router = True
+        elif getattr(args, 'uses_before', None) or getattr(args, 'uses_after', None):
+            args.scheduling = SchedulerType.ROUND_ROBIN
+            if getattr(args, 'uses_before', None):
+                peas_args['head'] = _copy_to_head_args(args, args.polling.is_push)
+                self.is_head_router = True
```
maximilianwerk, line 95: What about `self.is_tail_router` in case, `uses_after` is not set, but only `uses_before`? Should it be set explicitly to `False`?

maximilianwerk (commented):
:+1: removing hacks. Otherwise LGTM

jina/peapods/pod.py
```diff
@@ -82,15 +80,27 @@ def _parse_args(self, args: Namespace) -> Dict[str, Optional[Union[List[Namespac
...
+            if getattr(args, 'uses_after', None):
+                peas_args['tail'] = _copy_to_tail_args(args)
+                self.is_tail_router = True
+            peas_args['peas'] = _set_peas_args(args, peas_args.get('head', None), peas_args.get('tail', None))
         else:
             peas_args['peas'] = [args]
+            self.is_head_router = False
```
JoanFM, line 102: Moving these `self.is_head_router = False` and `self.is_tail_router = False` to the top of the function makes it clearer the default behavior and clarifies @maximilianwerk comment on wehter u can have a `head` without a `tail` and viceversa

JoanFM (commented):
Small comment from my side

tests/integration/incremental_indexing/test_unique_indexing.py
```diff
@@ -4,8 +4,7 @@
 from jina.executors.indexers.keyvalue import BinaryPbIndexer
 from jina.executors.indexers.vector import NumpyIndexer
 from jina.flow import Flow
-from tests.integration.incremental_indexing import random_workspace, get_duplicate_docs
```
maximilianwerk, line 7: It seems for pytest to accept this fixture, you should import `random_workspace` as well. I am voting for rather using 
```
@pytest.mark.parametrize('random_workspace_name', ['JINA_TEST_INCREMENTAL_INDEX_WORKSPACE'])
    def test_unique_indexing_vecindexers_before(test_metas):
    ...
```
here, since this is the way we started to use elsewhere as well.

maximilianwerk: You may chose a dedicated workspace per test or the same for all of them.

jina/peapods/pod.py
hanxiao: __init__ already has it

jina/peapods/pod.py
hanxiao: please note `__init__` already has `is_head_router ` `is_tail_router ` set

JoanFM (dismissed):

PR #1113, opened 2020-10-20
feat: add multimodal driver

bwanglzu:
**Changes introduced**
Added a new `MultiModalDriver` and a new executors interface `BaseMultiModalEncoder`.

- **MultiModalEncoder** has an `interface` encode that is a variable length argument of different batches of data. One `np.ndarray` per modality so that the executor can work in different ways.

- **MultiModalEncoder** has an argument that informs the `driver` which modality should be assigned to which index of the `variable length` argument.

- **MultiModalDriver** works on `chunks of documents` assuming that every `chunk` is from a different `modality` (otherwise it is an invalid document). So it assigns the `resulting embedding` to the `parent` document of the chunks with multiple modalities

- **MultiModalDriver** extracts from chunks content based on this logic: If it has an embedding, extract it, otherwise extract one of the actual content.

**Notes**
The traversal strategy looks quite similar as `Chunk2DocRankerDriver` but with `chunks`. This is related to the need to provide batching for `encoders`. Otherwise we may not be able to provide an efficient implementation for these executors.

**Next steps**
Work on the batching decorator to enable its capabilities to batch multiple inputs in a "zipped" way.

**Usage**
If modalities come from different `Flow` paths this driver needs to be combined with `ReduceAllDriver` in front of it. Just to showcase we have implemented for testing some `concatenation` executors. We could think on removing ConcatenateDriver because it also has too much coupling with `ReduceAllDriver`


JoanFM:
It looks like the idea is correct, let's add tests to prove it

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,89 @@
...
+                ]
+                context_doc.embedding.CopyFrom(
+                    array2pb(
+                        np.concatenate(
+                            embeddings_with_same_modality,
+                            axis=0)
+                    )
+                )
+            else:
+                embedding = self._extract_chunk_level_embedding(doc)
```
JoanFM, line 74: wrong method name

JoanFM:
After we have some tests, we can see how we decide which part of the document (content or embedding) must be extracted per modality

bwanglzu:
ok, while writing test I figured out a problem in the design. For each document, I grouped embeddings based on the same modality, and assign the concatenated embedding to `context_doc`. This means for each modality, I'm replacing the embedding with another modality embedding.

For example, we have 4 chunks with 2 modalities: 1 (visual), 2 (visual), 3 (textual), 4 (textual): embedding 1 and 2 were concatenated into embedding e1 and e1 was assigned as the embedding for `context_doc`. embedding3 and 4 were concatenated into embedding e2, and again, it was assigned as embedding for `context_doc`.

JoanFM:
> ok, while writing test I figured out a problem in the design. For each document, I grouped embeddings based on the same modality, and assign the concatenated embedding to `context_doc`. This means for each modality, I'm replacing the embedding with another modality embedding.
> 
> For example, we have 4 chunks with 2 modalities: 1 (visual), 2 (visual), 3 (textual), 4 (textual): embedding 1 and 2 were concatenated into embedding `e1` and `e1` was assigned as the embedding for `context_doc`. embedding3 and 4 were concatenated into embedding `e2`, and again, it was assigned as embedding for `context_doc`. and we lose `e1`..

In this case (only remember that is a temporary thing to have concatenation), u should be able to concatenate all the embeddings of every modality.

Actually I would assume that per modality there is only one chunk at this point.

JoanFM (commented):
I think in the scope of this PR and before merging, we should decide the interface of a Base class MultiModalEmbeddingExecutor so that we can have a test with a mock that does concatenation. To be done later

bwanglzu:
@JoanFM thanks for the input, while then how can we differentiate `MultimodalDriver` and `ReduceAllDriver`?

JoanFM:
> @JoanFM thanks for the input, while then how can we differentiate `MultimodalDriver` and `ReduceAllDriver`?

I do not quite understand the question, what is the issue?

bwanglzu:
@JoanFM my brain stopped running :) ignore my previous question

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,94 @@
...
+    ) -> None:
+        # docs are chunks of context_doc returned by traversal rec.
+        # Group chunks which has the same modality
+        modal_doc = {}
+        for doc in docs:
+            modal = doc.modality
+            if doc.id in modal_doc:
+                modal_doc[modal].append(doc.id)
+            else:
+                modal_doc[modal] = [doc.id]
```
maximilianwerk, line 60: This can be made easier with:
```
from collections import defaultdict
...
modal_doc = defaultdict(list)
for doc in docs:
    modal_doc[doc.modality].append(doc.id)
```

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,94 @@
...
+                context_doc.embedding.CopyFrom(
+                    array2pb(
+                        np.concatenate(
+                            embeddings_with_same_modality,
+                            axis=0)
+                    )
+                )
+            else:
+                embedding = self._extract_doc_embedding(doc)
+                if doc.id not in doc_pointers:
```
maximilianwerk, line 80: same here. `doc_pointers` can be `defaultdict(list)`

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,94 @@
...
+        # Group chunks which has the same modality
+        modal_doc = {}
+        for doc in docs:
+            modal = doc.modality
+            if doc.id in modal_doc:
+                modal_doc[modal].append(doc.id)
+            else:
+                modal_doc[modal] = [doc.id]
+
+        for modal, doc_ids in modal_doc.items():
```
maximilianwerk, line 62: Perhaps someone knows better than me: In the past python dictionaries where not order preserving. So when you iterate a dictionary, you can get the items in random order. If this is true, the concatenated embedding is not deterministic. I guess that is bad, right?

Anyhow, I am not sure, whether this is still true for current python version.

maximilianwerk (commented):
Some comments for cleaner code

jina/drivers/multimodal.py
bwanglzu: wow nice! thanks @maximilianwerk ! learned something new

jina/drivers/multimodal.py
bwanglzu: I'll leave the check here since `doc_pointer` is a parameter passed into the function.

jina/drivers/multimodal.py
bwanglzu: thanks for the comment, you're right, that's why we have `OrderedDict`. While concatenating is a temporary solution for this design (to be discussed further), so it's not an issue at the moment.

JoanFM: Yes concatrnation is a placeholder that needs to be changed by the call to the executor

JoanFM (commented):
Are we here considering that there is only one input per modality?

bwanglzu:
@JoanFM you mean one chunk per modality?

JoanFM:
> @JoanFM you mean one chunk per modality?

Yes, does it make sense this assumption?

bwanglzu:
@JoanFM yes it make sense (for most cases, e.g. one pic + text)

JoanFM:
> @JoanFM yes it make sense (for most cases, e.g. one pic + text)

Yes I think we can assume, every executor would expect a single input per modality. If it has a need to merge two modalities before we can expect it to do in a previous driver

JoanFM:
Since merging #1112 , I feel we should not have this driver inherit from ReduceDriver but have it a regular driver and assume that docs or chunks with different modalities come from different paths, then you need to add a ReduceDriver before it but if it comes from a previous crafter then you do not need it.

tests/unit/drivers/test_multimodal_driver.py
```diff
@@ -0,0 +1,73 @@
...
+        chunk1 = doc.chunks[0]
+        chunk2 = doc.chunks[1]
+        chunk3 = doc.chunks[2]
+
+        assert len(req.docs) == 1
+        assert len(doc.chunks) == 3
+        assert chunk1.embedding.shape == [embeddings[0].shape[0]]
+        assert chunk2.embedding.shape == [embeddings[1].shape[0]]
+        assert chunk3.embedding.shape == [embeddings[2].shape[0]]
+        assert chunk1.modality == chunk2.modality == 'visual'
```
JoanFM, line 64: We decided that the `driver` cannot handle 2 chunks of the same modality. They should be handled by an upstream Pod  to be merged into one

JoanFM (changes requested):

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,72 @@
...
+        Output:
+        document: (embedding: multimodal encoding)
+                |- chunk: {modality: mode1}
+                |
+                |- chunk: {modality: mode2}
+
+    .. note::
+        - It traverses on the ``documents`` for which we want to apply the ``multimodal`` embedding. This way
+        we can use the `batching` capabilities for the `executor`.
+        - It assumes that every ``chunk`` of a ``document`` belongs to a different modality.
```
JoanFM, line 36: Look at the documentation here. We should validate this and raise an `Exception` if not fullfilled.

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,72 @@
...
+        # docs are documents whose chunks are multimodal
+        # This is similar to ranking, needed to have batching?
+        data = []
+        for doc in docs:
+            data_by_modality = defaultdict(list)
+
+            for chunk in doc.chunks:
+                # Group chunks which has the same modality
+                # TODO: How should the driver know what does it need to extract per modality?
+                data_by_modality[chunk.modality].append(_extract_doc_content(chunk, self.field_by_modality[chunk.modality]))
```
JoanFM, line 67: We assume one `chunk` per modality, so no point to do append, if there is a duplicated one, raise `Error`. Otherwise, the interface for the `executor` will turn quite messy.

JoanFM (changes requested):

tests/unit/drivers/test_multimodal_driver.py
```diff
@@ -0,0 +1,73 @@
...
+    ]
+
+@pytest.fixture(scope='function')
+def doc_with_multimodal_chunks(embeddings):
+    doc = Document()
+    chunk1 = doc.chunks.add()
+    chunk2 = doc.chunks.add()
+    chunk3 = doc.chunks.add()
+    chunk1.modality = 'visual'
+    chunk2.modality = 'visual'
```
JoanFM, line 33: this is an invalid `document`. so good to test for an exception raised,

JoanFM:
@bwanglzu there is some problem with this branch, I think some rebase has gone wrong.

JoanFM:
@bwanglzu, I hope it is okey, I edited the description to explain in more detail the changes we have made

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+import numpy as np
+from collections import defaultdict
```
bhavsarpratik, line 5: Import sequence can be standard

jina/executors/encoders/multimodal/__init__.py
```diff
@@ -0,0 +1,36 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+
+import numpy as np
+from typing import Dict
```
bhavsarpratik, line 6: Import sequence can be standard

tests/integration/multimodal/test_multimodal_parallel.py
```diff
@@ -0,0 +1,90 @@
+import pytest
+import os
```
bhavsarpratik, line 2: Import sequence can be standard

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def _get_executor_input_arguments(self, content_by_modality: Dict[str, 'np.ndarray'], num_modalities: int):
+        """
+            From a dictionary ``content_by_modality`` it returns the arguments in the proper order so that they can be
+            passed to the executor.
+        """
+        input_args = [None] * num_modalities
+        for modality in self.position_by_modality.keys():
+            input_args[self.position_by_modality[modality]] = content_by_modality[modality]
+        return input_args
```
bhavsarpratik, line 74: input_args = [content_by_modality[modality] for modality in self.position_by_modality.keys()]

bhavsarpratik (changes requested):

jina/drivers/multimodal.py
JoanFM: what is the standard?

jina/drivers/multimodal.py
JoanFM: Done! Thanks!

jina/drivers/multimodal.py
bhavsarpratik: https://stackoverflow.com/questions/22722976/import-order-coding-standard

JoanFM:
Can you check again @bhavsarpratik

bhavsarpratik (approved):

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,112 @@
...
+        if not getattr(self._exec, 'position_by_modality', None):
+            raise RuntimeError('Could not know which position of the ndarray to load to each modality')
+        return self._exec.position_by_modality
+
+    def _get_executor_input_arguments(self, content_by_modality: Dict[str, 'np.ndarray']):
+        """
+            From a dictionary ``content_by_modality`` it returns the arguments in the proper order so that they can be
+            passed to the executor.
+        """
+        return [content_by_modality[modality] for modality in self.position_by_modality.keys()]
```
nan-wang, line 72: The position information is not used. 
```suggestion
        return [v for k, v in sorted(self.position_by_modality.items(), key=lambda item: item[1])]
```

nan-wang: After setting `self.position_by_modality = {'visual1': 1, 'visual2': 0, 'textual': 2}`, the following test will fail.
```
def test_multimodal_driver_with_shuffled_order(simple_multimodal_driver, mock_multimodal_encoder, doc_with_multimodal_chunks):
    simple_multimodal_driver.attach(executor=mock_multimodal_encoder, pea=None)
    simple_multimodal_driver._apply_all([doc_with_multimodal_chunks])
    doc = doc_with_multimodal_chunks
    assert len(doc.chunks) == 3
    visual1 = doc.chunks[1]
    visual2 = doc.chunks[0]
    textual = doc.chunks[2]
    control = np.concatenate([pb2array(visual2.embedding), pb2array(visual1.embedding), pb2array(textual.embedding)])
    test = pb2array(doc.embedding)
    np.testing.assert_array_equal(control, test)
```

nan-wang (changes requested):

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,112 @@
...
+        Output:
+        document: (embedding: multimodal encoding)
+                |- chunk: {modality: mode1}
+                |
+                |- chunk: {modality: mode2}
+
+    .. note::
+        - It traverses on the ``documents`` for which we want to apply the ``multimodal`` embedding. This way
+        we can use the `batching` capabilities for the `executor`.
+        - It assumes that every ``chunk`` of a ``document`` belongs to a different modality.
```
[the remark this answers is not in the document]
JoanFM: If you think about it, how else can you tell between two modalities by the executor? There must be some difference between every chunk, otherwise they must merge them in previous drivers.

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,112 @@
...
+            docs: Iterable['jina_pb2.Document'],
+            *args, **kwargs
+    ) -> None:
+        """
+        :param docs: the docs for which a ``multimodal embedding`` will be computed
+        :return:
+        """
+        # docs are documents whose chunks are multimodal
+        # This is similar to ranking, needed to have batching?
+        num_modalities = len(self.position_by_modality.keys())
```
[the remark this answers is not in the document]
JoanFM: I think the comment there is missplaced, what is needed for batching is the traversal logic.

jina/drivers/multimodal.py
JoanFM: will check!

JoanFM: Done Nan, added test, it was a changed we made trying to optimize, before was good

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def _extract_doc_content(doc: 'jina_pb2.Document'):
+    """Returns the content of the document with the following priority:
+    If the document has an embedding, return it, otherwise return its content.
+    """
+    if doc.embedding.buffer:
+        return pb2array(doc.embedding)
+    else:
+        return doc.text or doc.buffer or (doc.blob and pb2array(doc.blob))
```
hanxiao, line 21: is this somehow related to https://github.com/jina-ai/jina/blob/521436c17a3a807bb47c020a4527b209651624c4/jina/drivers/helper.py#L90-L93

if yes, can we refactor it to unify them into one?

JoanFM: it is kind of the same, it differs I think in the way documents are passed. Since here we are looping on chunks internall. I thought about it but did not want to overcomplicate.

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def _extract_modalities_from_document(doc: 'jina_pb2.Document'):
+    """Returns a dictionary of document content (embedding, text, blob or buffer) with `modality` as its key
+    """
+    doc_content = {}
+    for chunk in doc.chunks:
+        modality = chunk.modality
+        if modality in doc_content.keys():
```
hanxiao, line 30:
```suggestion
        if modality in doc_content:
```

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    if doc.embedding.buffer:
+        return pb2array(doc.embedding)
+    else:
+        return doc.text or doc.buffer or (doc.blob and pb2array(doc.blob))
+
+
+def _extract_modalities_from_document(doc: 'jina_pb2.Document'):
+    """Returns a dictionary of document content (embedding, text, blob or buffer) with `modality` as its key
+    """
```
hanxiao, line 26: not sure i get this, will keep reading

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+        Output:
+        document: (embedding: multimodal encoding)
+                |- chunk: {modality: mode1}
+                |
+                |- chunk: {modality: mode2}
+
+    .. note::
+        - It traverses on the ``documents`` for which we want to apply the ``multimodal`` embedding. This way
+        we can use the `batching` capabilities for the `executor`.
+.. warning::
```
hanxiao, line 54:
```suggestion
    .. note::
        - It traverses on the ``documents`` for which we want to apply the ``multimodal`` embedding. This way
        we can use the `batching` capabilities for the `executor`.
    .. warning::
```

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+        return self._exec.position_by_modality
+
+    def _get_executor_input_arguments(self, content_by_modality: Dict[str, 'np.ndarray'], num_modalities: int):
+        """
+            From a dictionary ``content_by_modality`` it returns the arguments in the proper order so that they can be
+            passed to the executor.
+        """
+        input_args = [None] * num_modalities
+        for modality in self.position_by_modality.keys():
+            input_args[self.position_by_modality[modality]] = content_by_modality[modality]
```
hanxiao, line 75:
```suggestion
        for modality, val in self.position_by_modality.items():
            input_args[val] = content_by_modality[modality]
```

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+    .. warning::
+        - It assumes that every ``chunk`` of a ``document`` belongs to a different modality.
+    """
+
+    def __init__(self,
+                 traversal_paths: Tuple[str] = ('r',), *args, **kwargs):
+        super().__init__(traversal_paths=traversal_paths, *args, **kwargs)
+
+    @property
+    def position_by_modality(self):
```
hanxiao, line 63:
```suggestion
    def position_by_modality(self) -> Dict[str, int]:
```

hanxiao: also need a docstring for this public property

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+            self,
+            docs: Iterable['jina_pb2.Document'],
+            *args, **kwargs
+    ) -> None:
+        """
+        :param docs: the docs for which a ``multimodal embedding`` will be computed, whose chunks are of different
+        modalities
+        :return:
+        """
+        num_modalities = len(self.position_by_modality.keys())
```
hanxiao, line 88:
```suggestion
        num_modalities = len(self.position_by_modality)
```

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+        content_by_modality = defaultdict(list)  # array of num_rows equal to num_docs and num_columns equal to
+
+        valid_docs = []
+        for doc in docs:
+            doc_content = _extract_modalities_from_document(doc)
+            if doc_content:
+                valid_docs.append(doc)
+                for modality in self.position_by_modality.keys():
+                    content_by_modality[modality].append(doc_content[modality])
```
hanxiao, line 97:
```suggestion
                for modality, val in doc_content.items():
                    content_by_modality[modality].append(val)
```

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+                self.logger.warning(f'Invalid doc {doc.id}. Only one chunk per modality is accepted')
+
+        if len(valid_docs) > 0:
+            # I want to pass a variable length argument (one argument per array)
+            for modality in self.position_by_modality.keys():
+                content_by_modality[modality] = np.stack(content_by_modality[modality])
+
+            # Guarantee that the arguments are provided to the executor in its desired order
+            input_args = self._get_executor_input_arguments(content_by_modality, num_modalities)
+            embeds = self.exec_fn(*input_args)
```
hanxiao, line 108: to me this part looks like a fancy `concat`, we do have a concat driver, not sure how they are related.

JoanFM: It is something that would be able to replace concatenating actually. You can see ConcatDriver as the combination of a `ReduceAll`plus a `MultiModalDriver` equipped with a `ConcatExecutor`

JoanFM: Concatenating is just another way of combining different modalities into one single vectorial representation. In the tests we have tested with such executors for instance

jina/executors/encoders/multimodal/__init__.py
```diff
@@ -0,0 +1,37 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
```
hanxiao, line 2: @bwanglzu btw, no need to add these two lines, they are added by our ci automatically

jina/executors/encoders/multimodal/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        :param position_by_modality: the map of arguments indicating in which order the modalities they need to come
+        for the encoding method
+        :return:
+        """
+        super().__init__(*args, **kwargs)
+        self.position_by_modality = position_by_modality
+
+    # @batching
+    @as_ndarray
```
hanxiao, line 31:
```suggestion
```

This can be removed as decorator is not inherited in Python.

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def _extract_modalities_from_document(doc: 'jina_pb2.Document'):
+    """Returns a dictionary of document content (embedding, text, blob or buffer) with `modality` as its key
+    """
+    doc_content = {}
+    for chunk in doc.chunks:
+        modality = chunk.modality
+        if modality in doc_content:
+            return None
```
hanxiao, line 31: i dont get this early return, what's the purpose here @bwanglzu @JoanFM

JoanFM: This None will make it an invalid document similar to what happens in the EncodeDriver. It works under the assumption that every chunk should have different modality.

jina/drivers/multimodal.py
hanxiao: also, isn't a List[str] enough for `position_by_modality`?

hanxiao: if `position_by_modality` is a simple `List[str]` then `input_args` is simply a reorder of `content_by_modality` via:

```python
input_args = [content_by_modality[modality] for modality in position_by_modality]
```

JoanFM: yes it can be done, it is more a matter of taste, I like it more explicit but we can for for this

hanxiao (changes requested):
i stop the review for now as I found `position_by_modality` can be a simple `List[str]`, no reason to make it a Dict

jina/drivers/multimodal.py
hanxiao: I went over the usage of `position_by_modality` and this really should be a list of str. This has to be fixed.

jina/drivers/multimodal.py
hanxiao: >  every chunk should have different modality

can you elaborate more on this? Say we have a PDF document, then the first granularity is
```
Doc
 |
 -- Chunk1 (image)
 |
 -- Chunk2 (text)
```
This assumption holds true only at this granularity, if we continue, this is not true

```
Doc
 |
 -- Chunk1 (image)
          |
          |-- ChunkChunk1 (image)
          |-- ChunkChunk2 (image)
 |
 -- Chunk2 (text)
```

jina/executors/encoders/multimodal/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+                 position_by_modality: Dict[str, int] = {},
+                 *args,
+                 **kwargs):
+        """
+        :param position_by_modality: the map of arguments indicating in which order the modalities they need to come
+        for the encoding method
+        :return:
+        """
+        super().__init__(*args, **kwargs)
+        self.position_by_modality = position_by_modality
```
hanxiao, line 28: is `self.position_by_modality` this really required at executor level? not seeing the executor using it, it is just used to inform the driver "please order in this way and give it to me". Perhaps just put `position_by_modality` directly into the driver?

tests/integration/multimodal/yaml/embedding-multimodal-concatenate.py
```diff
@@ -0,0 +1,25 @@
...
+        super().__init__(*args, **kwargs)
+
+    #@batching
+    @as_ndarray
+    def encode(self, *data: 'np.ndarray', **kwargs):
+        assert len(data) == 2
+        modality1 = data[0]
+        modality2 = data[1]
+        assert len(modality1) == len(modality2)
+        return np.concatenate((modality1, modality2), axis=1)
```
hanxiao, line 25: i see, so it is a fancy concat delegated to the executor. so one can add fancy ops, e.g. MLP, PCA to it

jina/executors/encoders/multimodal/__init__.py
JoanFM: It is not really required, I thought that the executor should have more knowledge than the driver on what is expected, but both are valid options.

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,114 @@
...
+        else:
+            doc_content[modality] = _extract_doc_content(chunk)
+    return doc_content
+
+
+class MultiModalDriver(BaseEncodeDriver):
```
hanxiao, line 37: hmm, now i see `MultiModalDriver` is a grouped version of `BaseEncodeDriver`, the grouping happens at `extract_docs`

jina/drivers/multimodal.py
JoanFM: it is true, but the user can have a previous multimodal driver to merge the `ChunkChunk1` and `ChunkChunk2` into `Chunk1`.

jina/executors/encoders/multimodal/__init__.py
bwanglzu: ok, thanks :)

jina/executors/encoders/multimodal/__init__.py
JoanFM: @bwanglzu I had a better thought and yes it makes more sense to have it at driver level. It is a quick change

jina/executors/encoders/multimodal/__init__.py
```diff
@@ -16,22 +15,30 @@ class BaseMultiModalEncoder(BaseExecutor):
...
-    # @batching
-    @as_ndarray
     def encode(self, *data: 'np.ndarray', **kwargs) -> 'np.ndarray':
         """
         :param: data: M arguments of shape `B x (D)` numpy ``ndarray``, `B` is the size of the batch, `M` is the number of modalities
         :return: a `B x D` numpy ``ndarray``
         """
         raise NotImplementedError
+
+    @property
```
JoanFM, line 37: Remove it from executor and move thr attribute to driver

jina/executors/encoders/multimodal/__init__.py
```diff
@@ -16,22 +15,30 @@ class BaseMultiModalEncoder(BaseExecutor):
...
+    def position_by_modality(self) -> List[str]:
+        """Get position per modality.
+        :return: the list of strings representing the name and order of the modality.
+        """
+        if not self._position_by_modality:
```
JoanFM, line 42: This check made sense in driver but not much sense here.

tests/integration/multimodal/yaml/embedding-multimodal-concatenate.py
```diff
@@ -2,21 +2,16 @@
 __license__ = "Apache-2.0"
 
 import numpy as np
-from jina.executors.decorators import batching, as_ndarray
 from jina.executors.encoders.multimodal import BaseMultiModalEncoder
 
 
 class ConcatenateMultiModalEncoder(BaseMultiModalEncoder):
 
-    batch_size = 10
```
JoanFM: Keep it here as it will be used in the future

tests/integration/multimodal/yaml/embedding-multimodal-concatenate.py
```diff
@@ -2,21 +2,16 @@
...
-
     def __init__(self,
                  *args,
                  **kwargs):
         super().__init__(*args, **kwargs)
 
-    #@batching
-    @as_ndarray
```
JoanFM: Leep them here as they will be used in the future

JoanFM: They did not make sense in the parent class but do here

JoanFM (changes requested):

jina/executors/encoders/multimodal/__init__.py
bwanglzu: ah I misunderstood.. I'll move it back

tests/integration/multimodal/yaml/embedding-multimodal-concatenate.py
bwanglzu: ok

tests/integration/multimodal/yaml/embedding-multimodal-concatenate.py
bwanglzu: ok

jina/drivers/multimodal.py
bwanglzu: updated

jina/drivers/multimodal.py
```diff
@@ -60,12 +60,22 @@ def __init__(self,
         super().__init__(traversal_paths=traversal_paths, *args, **kwargs)
 
 
+    @property
+    def position_by_modality(self) -> List[str]:
```
JoanFM, line 64: The point is to have it directly as an attribute of driver and not as a property reading from executor. So that the executor does not know anything about it

JoanFM (changes requested):

jina/executors/encoders/multimodal/__init__.py
JoanFM: @bwanglzu second thoughts. I would keep it as attribute of executor and access via property in the driver. This way the executor can use that to interpret the arguments of the encoders.

nan-wang: What about grouping them into `dict` and passing into the executors?

JoanFM: It makes a much worse interface. But also in my opinion mulimodal is about having N inputs that come from different  sources and that need N different treatments. It seems a fair assumption in my opinion

hanxiao (commented):
rename `position_by_modality` -> `positional_modality`? analogy to `positional args` which is well accepted

JoanFM:
> rename `position_by_modality` -> `positional_modality`? analogy to `positional args` which is well accepted

makes sense since is no longer a Dict

jina/drivers/multimodal.py
```diff
@@ -0,0 +1,113 @@
...
+    .. warning::
+        - It assumes that every ``chunk`` of a ``document`` belongs to a different modality.
+    """
+
+    def __init__(self,
+                 traversal_paths: Tuple[str] = ('r',), *args, **kwargs):
+        super().__init__(traversal_paths=traversal_paths, *args, **kwargs)
+
+    @property
+    def position_modality(self) -> List[str]:
```
hanxiao, line 63: should be `positional_modality`

hanxiao (approved):

JoanFM (approved):

nan-wang (approved):

maximilianwerk (approved):

PR #1114, opened 2020-10-20
refactor: flatten evaluator inheritance

hanxiao:
## Rationale

- `RunningStats` (prev aggregator_wrapper) should be implemented in a class, as its running stats are essentially private.
- `Evaluator` should not be grouped by the family of Executor, otherwise it creates a recursive logic structure:
```
A, B, C, D
              |
              |-> D4A, D4B, D4C
```
- `Evaluator` should be grouped by the type values in doc protobuf, e.g. string, list of id, embedding.
- `def evaluate(self, prediction: 'np.array', groundtruth: 'np.array')` this signature better be consistent across all `Evaluator`, otherwise users have to guess
- remove the usage of `scipy` in `cosine` and `euclidean` evaluator, as `scipy` is not in base dependencies of Jina


JoanFM (dismissed):
Like specially prediction as argument name 👍

hanxiao:
> Like specially prediction as argument name 👍

haha, just removed, now it's `actual` and `desired`, ref: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html

jina/executors/evaluators/embedding/__init__.py
```diff
@@ -0,0 +1,19 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+import numpy as np
+
+from .. import BaseEvaluator
+
+
+class BaseEmbeddingEvaluator(BaseEvaluator):
+    """A :class:`BaseEmbeddingEvaluator` evaluates the difference between doc and groundtruth embeddings
```
bwanglzu, line 10: maybe also change to: between actual and desired embeddings

hanxiao: done

jina/executors/evaluators/embedding/cosine.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def evaluate(self, actual: 'np.array', desired: 'np.array', *args, **kwargs) -> float:
+        """"
+        :param actual: the embedding of the document (resulting from an Encoder)
+        :param desired: the expected embedding of the document
+        :return the evaluation metric value for the request document
+        """
+        actual = expand_vector(actual)
+        desired = expand_vector(desired)
+        return _cosine(_ext_A(_norm(actual)), _ext_B(_norm(desired)))
```
bwanglzu, line 29: maybe it's better to turn `_cosine`, `_ext_A` and `_norm` into a public method?

> _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose names start with an underscore.

JoanFM, line 29: Since this would eventually be moved to the hub i would not depend on functions from the indexer, I did skip this import on purpose, we could duplicate if necessary

hanxiao: import aside. p/r is very essential, better keep them in the core.

JoanFM: Or we can move these functims to sme embedding_distances.py?

hanxiao: i will keep duplicate, they are not the core of this PR anyway

PR #1116, opened 2020-10-20
feat(flow): add plot for uses before after

hanxiao:
## Improved Feature

- `plot` now support before & after
- `plot(img_type=)` is removed, `img_type` is inferred from the suffix of the filename

```python
f = Flow().add(uses_before='_pass', uses_after='_pass', name='p1',).plot()
```

```python
f = Flow().add(uses_before='_pass', name='p1',).plot()
```

```python
f = Flow().add(uses_after='_pass', name='p1',).plot()
```

### DIFF
--- a/tests/unit/flow/test_flow_visualization.py
+++ b/tests/unit/flow/test_flow_visualization.py
@@ -5,43 +5,54 @@ from jina.flow import Flow
 cur_dir = os.path.dirname(os.path.abspath(__file__))
 
 
-def test_visualization_with_yml_file_img():
-    Flow.load_config(os.path.join(cur_dir, '../yaml/test_flow_visualization.yml')).plot(output='flow.svg')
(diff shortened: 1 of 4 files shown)


tests/unit/flow/test_flow_visualization.py
```diff
@@ -41,4 +41,7 @@ def test_visualization_plot_in_middle():
 
     assert os.path.exists('flow3.svg')
 
-test_visualization_with_yml_file_img()
\ No newline at end of file
+
+def test_flow_before_after_plot():
```
JoanFM, line 45: Use tmpdir fixture

hanxiao: will leave to others :)

hanxiao: fine i did it

PR #1118, opened 2020-10-21
refactor(hubio): remove the dependency of requests from list

hanxiao:
image

### DIFF
--- a/jina/docker/hubapi.py
+++ b/jina/docker/hubapi.py
@@ -1,5 +1,9 @@
 
 import json
+import urllib
+from urllib.parse import urlencode
+from urllib.request import Request, urlopen
+
 import requests
 from typing import Dict
 from pkg_resources import resource_stream
@@ -27,6 +31,10 @@ def _list(logger, name: str = None, kind: str = None, type_: str = None, keyword
         # The way lambda function handles params, we need to pass them comma separated rather than in an iterable 
         params['keywords'] = ','.join(keywords) if len(keywords) > 1 else keywords
     if params:
+        request = Request(f'{hubapi_url}{hubapi_list}', urlencode(params).encode())
+        urlopen(request)
+        json.load(resource)
+
         response = requests.get(url=f'{hubapi_url}{hubapi_list}',
                                 params=params)
         if response.status_code == requests.codes.bad_request and response.text == 'No docs found':
--- a/jina/docker/hubapi.py
+++ b/jina/docker/hubapi.py
@@ -1,11 +1,10 @@
-
 import json
-import urllib
+from typing import Dict
+from urllib.error import HTTPError
 from urllib.parse import urlencode
 from urllib.request import Request, urlopen
 
 import requests
-from typing import Dict
 from pkg_resources import resource_stream
 
 from .helper import credentials_file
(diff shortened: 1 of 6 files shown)


rutujasurve94 (approved):

jina/docker/hubapi.py
```diff
@@ -1,112 +1,104 @@
...
-    if keywords:
-        # The way lambda function handles params, we need to pass them comma separated rather than in an iterable 
-        params['keywords'] = ','.join(keywords) if len(keywords) > 1 else keywords
+        hubapi_url = hubapi_yml['hubapi']['url'] + hubapi_yml['hubapi']['list']
+
+    params = {
+        'name': image_name,
+        'kind': image_kind,
+        'type': image_type,
+        'keywords': ','.join(image_keywords) if image_keywords else None
```
rutujasurve94, line 34: `,'` should be prepended only if more than one keyword is specified. if it's just one keyword then its a single element list, no `,` needed

hanxiao: default behavior wont hurt, cleaner code

```
Python 3.7.5 (default, Jan 19 2020, 12:44:36)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> ','.join([])
''
```

hanxiao: background: keywords specified in cli with `nargs` will always result a list

jina/docker/hubapi.py
```diff
@@ -1,112 +1,104 @@
...
 from pkg_resources import resource_stream
 
 from .helper import credentials_file
 from ..helper import yaml, colored
+from ..logging.profile import TimeContext
 
 
-def _list(logger, name: str = None, kind: str = None, type_: str = None, keywords: tuple = ('numeric')):
-    """ Hub API Invocation to run `hub list` """
+def _list(logger, image_name: str = None, image_kind: str = None,
```
maximilianwerk, line 14: I wondered, why we don't use `**kwargs` here as we do all around the code. I guess, since it is not obvious, what the kwargs should be.

Anyhow, if this method would be a static method in the `HubIO` class using `**kwargs` would make way more sense. And since it is only used in `HubIO` it is also well placed there.

maximilianwerk (commented):
Looks good from a functional perspective. Just some cosmetic remarks.

deepankarm (approved):

jina/docker/hubapi.py
deepankarm: These arguments are highly specific to the hubapi, hence passing explicitly makes more sense. This method is kept out of `HubIO`, as we can keep all hubapi based invocations in one place. This file will evolve as we add more endpoints.

deepankarm:
@maximilianwerk we'll consider these remarks & refactor the `hubapi` code with further planned changes.

jina/docker/hubapi.py
maximilianwerk: But then it should not be "private" and we should remove the underscore in front.

jina/docker/hubio.py
```diff
@@ -51,18 +53,20 @@ def __init__(self, args: 'argparse.Namespace'):
...
         try:
             from cookiecutter.main import cookiecutter
         except (ImportError, ModuleNotFoundError):
-            self.logger.critical('requires "cookiecutter" dependency, please install it via "pip install cookiecutter"')
+            self.logger.critical('"jina hub new" requires "cookiecutter" dependency, '
+                                 'please install it via pip install "jina[cookiecutter]"')
             raise
 
-        import click
+        import click  # part of cookiecutter
```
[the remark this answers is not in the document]
rutujasurve94: We want to have minimal requirements, especially for cookiecutter

PR #1121, opened 2020-10-21
test: take out jinatestcase from protobuf test

bwanglzu:
A bit cleanup of `test_protobuf.py`.

### DIFF
--- a/tests/unit/proto/test_protobuf.py
+++ b/tests/unit/proto/test_protobuf.py
@@ -1,31 +1,29 @@
-import unittest
+import pytest
 
 from jina.proto.jina_pb2 import Document
-from tests import JinaTestCase
 from google.protobuf.json_format import MessageToJson, Parse
 
+@pytest.fixture(scope='function')
+def document():
+    d = Document()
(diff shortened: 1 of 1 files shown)


tests/unit/proto/test_protobuf.py
```diff
@@ -1,31 +1,29 @@
-import unittest
+import pytest
 
 from jina.proto.jina_pb2 import Document
-from tests import JinaTestCase
 from google.protobuf.json_format import MessageToJson, Parse
 
+@pytest.fixture(scope='function')
```
nan-wang, line 6: Is it necessary to set up the scope here? The default value is `function`. 
https://docs.pytest.org/en/stable/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

nan-wang (commented):
LGTM👍

tests/unit/proto/test_protobuf.py
bwanglzu: You're right. It's my personal habit to be explicit https://www.python.org/dev/peps/pep-0020/

hanxiao (approved):

PR #1128, opened 2020-10-22
fix: fixed the recall calcuation and tests

### LINKED ISSUE #1123: Bug in RecallEvaluator
**Describe the bug**
Given the relevant docs, `[0, 1, 2, 3, 4]`, and the retrieved docs, `[1, 0, 20, 30, 40]`, the recall@k is as below

|   | expected recall@k | current recall@k |
|---|---|---|
|1 | 0.2 | 1 |
|2 | 0.4 | 1 |
|3 | 0.4 | 0.667 | 
|4 | 0.4 | 0.5 |
|5 | 0.4 | 0.4 | 

The `Recall@k` should be the number of relevant results in the top k retrieved results divided by the number of all the relevant results.

> ... if the (k + 1)th document retrieved is nonrelevant then recall is the same as for the top k documents ...
> _from_  __Introduction to Information Retrieval__

**Describe how you solve it**

---

**Environment**

**Screenshots**

maximilianwerk:
That should fix #1123 . The for loop was mixed up.

### DIFF
--- a/jina/executors/evaluators/rank/recall.py
+++ b/jina/executors/evaluators/rank/recall.py
@@ -18,14 +18,13 @@ class RecallEvaluator(BaseRankingEvaluator):
         :param desired: the expected documents matches ids sorted as they are expected
         :return the evaluation metric value for the request document
         """
-        ret = 0.0
-        for doc_id in desired[:self.eval_at]:
-            if doc_id in actual:
-                ret += 1.0
-
-        divisor = min(self.eval_at, len(actual))
-        if divisor == 0.0:
+        if not desired:
             """TODO: Agree on a behavior"""
             return 0.0
-        else:
-            return ret / divisor
+
+        ret = 0.0
+        for doc_id in actual[:self.eval_at]:
+            if doc_id in desired:
+                ret += 1.0
+
+        return ret / len(desired)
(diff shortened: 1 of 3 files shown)


JoanFM:
Sorry, now I am not sure the bug is real,

Shouldn't recall be this?
Recall= (Relevant_Items_Recommended in top-k) / (Relevant_Items)

So we should see how many of the `top_k` in the desired are actually

jina/executors/evaluators/rank/recall.py
```diff
@@ -18,14 +18,13 @@ def evaluate(self, actual: Sequence[Any], desired: Sequence[Any], *args, **kwarg
...
         """
+        if not desired:
+            """TODO: Agree on a behavior"""
+            return 0.0
+
         ret = 0.0
-        for doc_id in desired[:self.eval_at]:
-            if doc_id in actual:
+        for doc_id in actual[:self.eval_at]:
+            if doc_id in desired:
```
JoanFM, line 27: Just to make sure, we agree then that all the matches here are considered as `relevant` right?

JoanFM: I think it should be documented expicitly that all the documents in `desired` will be considered as `relevant`

maximilianwerk: That is my understanding of `desired`, yes.

maximilianwerk: Then rename it to `relevant_docs`?

JoanFM: it is okey as it is I think, and we have `desired` all across the evaluators so it would be okey.

JoanFM (approved):

maximilianwerk:
Any guess, why python3.8 tests never finish on time for the PR?

PR #1132, opened 2020-10-22
chore: add maximilianwerk as codeowner for drivers

maximilianwerk:
I added me as owner for `jina/drivers`. Am I right, that @bwanglzu should be the owner for `tests` instead of `jina/tests`, since the latter does not exist?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,4 +9,5 @@
 setup.py  @hanxiao
 jina/peapods @JoanFM
 jina/proto @JoanFM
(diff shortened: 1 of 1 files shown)


JoanFM (approved):

JoanFM:
> I added me as owner for `jina/drivers`. Am I right, that @bwanglzu should be the owner for `tests` instead of `jina/tests`, since the latter does not exist?

you are really right!

bwanglzu (approved):
LGTM!

Issue #976, opened 2020-09-22
top_k not settable via .yml file


maximilianwerk:
**Describe the bug**

During setting of parameters for a driver, the order of defaults is not obvious. We have currently (at least) the following possibilities to set arguments (example is `VectorSearchDriver` and `top_k` argument):

1) Directly as a default for the class itself
```def __init__(self, top_k: int = 50, fill_embedding: bool = False, *args, **kwargs):```

2) As a command line argument
```    
gp1.add_argument('--top-k', type=int, default=10, help='top_k results returned in the search mode')
```

3) in the `.yaml` file
```
requests:
  on:
    ...
    SearchRequest:
      - !VectorSearchDriver
        with:
          executor: vecidx
          top_k: 5
    ...
```

4) As the calling client
```
search_flow.search(..., top_k=1,...)
```

Currently the order of importance is:
4 > 2 > 3 > 1

As a consequence, if a parameter has a default as a command line argument, it will never have any effect if 1 or 3 is used for setting them. This is bad, since consciously setting a value via 3) should have an effect, as long as no explicit command line argument is given.

**Describe how you solve it**
I see two options:

1) Removing the `default` for the `top_k` command line argument. Not sure, whether this has any bad side-effects. My perception is, that the command line arguments where really used in the early days of jina, but now that the `.yml` files become more and more prominent, shouldn't be to powerful anymore.

2) Change the order in which defaults argument are applied. This feels wrong, since running something with an explicit command line argument should also apply this command line argument.

3) Completely remove the command line argument in order to simplify the use of jina. Removing non-used features is important while the framework grows. So my question: Is this really a use-case for command-line arguments? Or won't users anyhow always set this via option 3 or 4?

I'd go for option 3, if we want to straighten up the command line interface or option 1, if we want to wait before removing arguments. Opinions?

JoanFM:
I am okey with removing top_k from command line arguments, but I still think that 4 should have priority to the other options, since it can be changed by request, and giving power to the client makes the behavior more flexible.

maximilianwerk:
Now that I have played with it, I am not sure about your suggestion. `top_k` is used for a lot of drivers and setting it via 4 sets it for all drivers. Think about the following example:

You have a query and want to get top-10 results: The algorithm does:

retrieval: retrieve top 50 documents

rerank via some custom ranker only the top 50

use QueryLang to cut down to top-10

[...]

PR #1133, opened 2020-10-22
fix: removed top_k default

bhavsarpratik:
Avoiding top_k overrides by removing it from `parser.py` and `request.py`


jina/peapods/gateway.py
```diff
@@ -282,6 +282,16 @@ def api(mode):
 
             content['mode'] = ClientMode.from_string(mode)
 
+            if 'top_k' in content:
```
JoanFM, line 285: Need to add the integration test calling the `rest api`

JoanFM (changes requested):

JoanFM:
WARNING! This will break examples, so let's not merge it before the 0.7.0!!!

maximilianwerk:
> WARNING! This will break examples, so let's not merge it before the 0.7.0!!!

Shouldn't we merge it exactly before 0.7.0 to break it in a minor release and not in a micro release?

tests/integration/issues/github_241/test_topk.py
```diff
@@ -0,0 +1,59 @@
+import os
+
+import numpy as np
+import pytest
+from jina.flow import Flow
```
bwanglzu, line 5: it would be nice if we can respect import sequence as in core, such as
```python
import os

import pytest
import numpy as np

from jina...
```

jina/parser.py
```diff
@@ -517,9 +517,6 @@ def set_client_cli_parser(parser=None):
     gp1.add_argument('--mode', choices=list(ClientMode), type=ClientMode.from_string,
                      # required=True,
                      help='the mode of the client and the server')
-    gp1.add_argument('--top-k', type=int,
```
hanxiao, line 520: why not remove `default=10` instead of removing the whole arg?

jina/clients/python/request.py
```diff
@@ -113,16 +112,6 @@ def _generate(data: Union[Iterator[Union['jina_pb2.Document', bytes]], Iterator[
...
 
-        if top_k and mode == ClientMode.SEARCH:
-            if top_k <= 0:
-                raise ValueError(f'"top_k: {top_k}" is not a valid number')
-            else:
-                top_k_queryset = jina_pb2.QueryLang()
-                top_k_queryset.name = 'VectorSearchDriver'
-                top_k_queryset.priority = 1
-                top_k_queryset.parameters['top_k'] = top_k
-                req.queryset.extend([top_k_queryset])
```
hanxiao, line 124: this can be kept, if my next suggestion is followed

tests/integration/issues/github_241/test_topk.py
```diff
@@ -0,0 +1,59 @@
...
+    os.environ['JINA_TOPK_OVERRIDE'] = '11'
+
+
+def random_docs(num_docs, embed_dim=10, jitter=1):
+    for j in range(num_docs):
+        d = jina_pb2.Document()
+        d.tags['id'] = j
+        d.text = b'hello'
+        d.embedding.CopyFrom(array2pb(np.random.random([embed_dim + np.random.randint(0, jitter)])))
+        yield d
```
bwanglzu, line 24: this could be a pytest [yield fixture](https://docs.pytest.org/en/2.9.2/yieldfixture.html)

jina/peapods/gateway.py
```diff
@@ -282,6 +282,16 @@ def api(mode):
...
+                top_k_queryset = jina_pb2.QueryLang()
+                top_k_queryset.name = 'VectorSearchDriver'
+                top_k_queryset.priority = 1
+                top_k_queryset.parameters['top_k'] = content['top_k']
+                if 'queryset' not in content:
+                    content['queryset'] = top_k_queryset
```
maximilianwerk, line 291: Not sure about the syntax, but the `.append` below hints, that it should be `= [top_k_queryset]`.

bwanglzu (commented):
nice!

jina/parser.py
JoanFM: Why do not we do this? 

we can keep that top_k  in the client (with a default to None) and move the change in the RESTGatewayPea to the search method in request.py

tests/integration/issues/github_241/test_topk.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def validate_results(resp):
+    assert len(resp.search.docs) == int(os.environ['JINA_NDOCS'])
+    for doc in resp.search.docs:
+        assert len(doc.matches) == int(os.environ['JINA_TOPK'])
+
+
+def test_topk(config):
```
maximilianwerk, line 33: Can you do the same test again with `shards = 2`. I would guess, this test will fail then. I think, the implementation in the `gateway.py` should use a `SliceQL` instead of a `VectorSearchDriver`.

tests/integration/issues/github_241/test_topk.py
```diff
@@ -0,0 +1,59 @@
...
+                          output_fn=validate_results)
+
+
+def validate_override_results(resp):
+    assert len(resp.search.docs) == int(os.environ['JINA_NDOCS'])
+    for doc in resp.search.docs:
+        assert len(doc.matches) == int(os.environ['JINA_TOPK_OVERRIDE'])
+
+
+def test_topk_override(config):
```
maximilianwerk, line 47: Same argument with shards applies here.

maximilianwerk (changes requested):
I am pretty sure, sharding will not work as expected. There should be a slice operation at the very end which slices all non-needed results away.

Or we just keep it as is and accept, that the `top-k` argument goes deep into the lookup. Anyhow, I'd argue that the client-API should be as obvious as possible and having more shards should not affect the results that are returned to a client.

jina/peapods/gateway.py
JoanFM: inside the `_generate` the type is checked and it can be a single one, but yes it may be more coherent to do this

tests/integration/issues/github_241/test_topk.py
JoanFM: Assuming, this is kind of a `hack`, `VectorSearchDriver` is almost always there in a `Flow` while `SliceQL` not so common (and `SliceQL` does not read `top_k` parameter). 

This is just a way to make it easier from the ouside, but is not really the best way to do things. (It is kind of a syntax sugar for a common pattern)

JoanFM:
> I am pretty sure, sharding will not work as expected. There should be a slice operation at the very end which slices all non-needed results away.
> 
> Or we just keep it as is and accept, that the `top-k` argument goes deep into the lookup. Anyhow, I'd argue that the client-API should be as obvious as possible and having more shards should not affect the results that are returned to a client.

Totally agreeing that it won't work, we cannot force the user to have a `SliceQL`. We should better document how `top_k` is accepted and so, but I think not in the scope of the #976

jina/parser.py
bhavsarpratik: As per our current changes, `top_k` parameter was not used and so we removed it

maximilianwerk:
> > I am pretty sure, sharding will not work as expected. There should be a slice operation at the very end which slices all non-needed results away.
> > Or we just keep it as is and accept, that the `top-k` argument goes deep into the lookup. Anyhow, I'd argue that the client-API should be as obvious as possible and having more shards should not affect the results that are returned to a client.
> 
> Totally agreeing that it won't work, we cannot force the user to have a `SliceQL`. We should better document how `top_k` is accepted and so, but I think not in the scope of the #976

OK, then keep it like that for now.

maximilianwerk (approved):

JoanFM:
Wait, @hanxiao 's point must be addressed

JoanFM:
There are 2 missing parts:

- Move logic from `RestGatewayPea` to python client
- Integration test where `top_k` is sent via rest

tests/integration/issues/github_241/flow.yml
```diff
@@ -0,0 +1,5 @@
+!Flow
```
JoanFM, line 1: github issue is `976` so can you rename the test folder?

JoanFM (changes requested):

tests/integration/issues/github_976/test_topk_rest_api.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import json
+from urllib import request
+
+from jina.flow import Flow
+
+
+@pytest.fixture
+def top_k():
```
JoanFM, line 9: Sorry for that, make this a constant `TOP_K`

tests/integration/issues/github_976/test_topk_rest_api.py
```diff
@@ -6,19 +6,19 @@
 
 
 @pytest.fixture
-def top_k():
+def TOP_K():
```
JoanFM, line 9: make it a `constant` not a fixture

JoanFM (changes requested):

jina/parser.py
JoanFM: @bhavsarpratik add the parameter back with no default

JoanFM (changes requested):

JoanFM (approved):

PR #1135, opened 2020-10-22
feat(cli): show local version side by side

hanxiao:
## New feature:

`jina hub list`

image

### DIFF
--- a/jina/docker/hubapi.py
+++ b/jina/docker/hubapi.py
@@ -63,7 +63,7 @@ def _list_local(logger) -> Optional[Dict[str, Any]]:
     """
     manifests = _load_local_hub_manifest()
     if manifests:
-        _print_hub_table(logger, manifests.values())
+        _make_hub_table(logger, manifests.values())
     return manifests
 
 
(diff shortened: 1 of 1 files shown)


jina/docker/hubapi.py
```diff
@@ -106,24 +108,58 @@ def _list(logger, image_name: str = None, image_kind: str = None,
...
+    for index, manifest in enumerate(manifests):
+        image_name = manifest.get('name', '')
+        ver = manifest.get('version', '')
+        desc = manifest.get('description', '')[:60].strip() + '...'
+        if image_name and ver and desc:
+            local_ver = ''
+            color = 'white'
+            if image_name in local_manifests:
+                local_ver = local_manifests[image_name].get('version', '')
+                _v1, _v2 = version.parse(ver), version.parse(local_ver)
```
maximilianwerk, line 135: I'd prefer rename to `hub_version` and `local_version`.

maximilianwerk:
Why does the `AnnoyIndexer` appears several times in the list? Can we order the list alphabetically? Especially, if double entries are OK, it would make it much better readable.

hanxiao:
> Why does the `AnnoyIndexer` appears several times in the list? Can we order the list alphabetically? Especially, if double entries are OK, it would make it much better readable.

@maximilianwerk this has to be done at https://github.com/jina-ai/cloud-ops/blob/master/lambda_handlers/hubapi_list.py 
the hub cli is a client, improving readability on the client side is useful but limited. especially the real UI is on the web. So server has to handle duplication, sorting, etc, not the client

Issue #1122, opened 2020-10-22
Jina can't run after installed by pip


matadorhong:
**Describe your problem**

           JINA@6835[E]:theses modules or classes can not be imported ['jina.drivers.querylang.queryset', 'jina.drivers.querylang.queryset.dunderkey', 'jina.drivers.cache', 'jina.drivers.convert', 'jina.drivers.craft', 'jina.drivers.helper', 'jina.drivers.querylang.queryset.lookup', 'jina.drivers.querylang', 'jina.drivers.querylang.filter', 'jina.drivers.querylang.slice', 'jina.drivers.search', 'jina.drivers.encode', 'jina.drivers.querylang.sort', 'jina.drivers.rank', 'jina.drivers.evaluate', 'jina.drivers.querylang.select', 'jina.drivers.querylang.reverse', 'jina.drivers.reduce', 'jina.drivers.control', 'jina.drivers.querylang.queryset.helper', 'jina.drivers.index']. You can use `jina check` to list all executors and drivers
           JINA@6835[E]:theses modules or classes can not be imported ['jina.executors.indexers.vector', 'jina.executors.evaluators', 'jina.executors.indexers.cache', 'jina.executors.evaluators.rank.precision', 'jina.executors.evaluators.decorators', 'jina.executors.rankers', 'jina.executors.crafters', 'jina.executors.requests', 'jina.executors.indexers', 'jina.executors.encoders.numeric', 'jina.executors.evaluators.rank.recall', 'jina.executors.devices', 'jina.executors.compound', 'jina.executors.decorators', 'jina.executors.encoders', 'jina.executors.encoders.tfserving', 'jina.executors.evaluators.rank', 'jina.executors.encoders.helper', 'jina.executors.indexers.keyvalue', 'jina.executors.metas', 'jina.executors.encoders.frameworks', 'jina.executors.clients']. You can use `jina check` to list all executors and drivers
[...]

PR #1139, opened 2020-10-22
test: add docker hello-world with no devel install

hanxiao:
## Changes to CI

- docker `pip install .` in fresh enviroment without `devel` deps, and then `docker run hello-world`. This make sure fresh pip install always work. this to check issue such as #1122

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics --exclude .git,__pycache__,docs/source/conf.py,old,build,dist,tests/,jina/hub/
       - name: Test docker install
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -97,7 +97,8 @@ jobs:
         run: |
           # make sure docker exist
           docker ps
-          docker build  -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
+          docker build -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
```
JoanFM, line 100: can we split this in step generating `test-pip` and doing this test?

hanxiao: what do you mean here? 1. this line does not change,  2. `test-pip` is not testing, it is a testbed for later tests

PR #1143, opened 2020-10-23
docs: change docstring of rank drivers

JoanFM:
**Changes introduced**

- Add some information in ranker drivers docstrings. 

- Change default `taversal_paths` of `Collect2Matches`

### DIFF
--- a/jina/drivers/rank.py
+++ b/jina/drivers/rank.py
@@ -1,7 +1,6 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-# lift the chunk-level topk to doc-level topk
 from typing import Dict, Iterable, List, Tuple
 
 import numpy as np
@@ -29,8 +28,6 @@ class Chunk2DocRankDriver(BaseRankDriver):
     """Extract matches score from chunks and use the executor to compute the rank and assign the resulting matches to the
     level above.
 
-    Note that it traverses on ``chunks`` not on ``matches``
-
     Input-Output ::
         Input:
         document: {granularity: k-1}
@@ -48,6 +45,10 @@ class Chunk2DocRankDriver(BaseRankDriver):
             |    |- matches: {granularity: k}
             |
             |-matches: {granularity: k-1} (Ranked according to Ranker Executor)
+
+        .. note::
+            - It traverses on ``chunks`` not on ``matches``. This is because ranker needs context information
+            from ``matches`` for several ``chunks``
     """
 
     def __init__(self, traversal_paths: Tuple[str] = ('c',), *args, **kwargs):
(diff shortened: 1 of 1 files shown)


jina/drivers/rank.py
```diff
@@ -64,15 +65,15 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2
         match_idx = []  # type: List[Tuple[int, int, int, float]]
         query_chunk_meta = {}  # type: Dict[int, Dict]
         match_chunk_meta = {}  # type: Dict[int, Dict]
-        for c in docs:
-            for match in c.matches:
+        for doc in docs:
```
maximilianwerk, line 68: I wondered, whether `chunk` would be the better naming here. There is one reason, that speaks against it: you could apply this function also against the matches of the `context_doc`. Anyhow, the driver is not appropriate for this task, since the `context_doc` would then have the old matches and the newly calculated matches mixed up. That is bad.

Thus, do we want to rename it `chunk`? (`context_doc_chunk` is to bulky)

jina/drivers/rank.py
```diff
@@ -191,7 +195,16 @@ def __init__(self, reverse: bool = False, traversal_paths: Tuple[str] = ('m',),
...
-        """ Call executer for score and sort afterwards here. """
+        """
+
+        :param docs: the matches of the ``context_doc``, they are at granularity ``k``
+        :param context_doc: the query document having ``docs`` as its matches, it is at granularity ``k``
+        :return:
+
+        .. note::
+            - This driver will change in place the ordering of ``matches`` of the ``context_doc`.
+            - Set the ``taversal_paths`` of this driver such that it traverses along the ``matches`` of the ``chunks`` at the level desired.
```
maximilianwerk, line 207: `traversal_paths`

jina/drivers/rank.py
```diff
@@ -127,9 +128,15 @@ def _apply_all(self, docs: Iterable['jina_pb2.Document'], context_doc: 'jina_pb2
...
-        :param docs: the chunks of the ``context_doc``, they are at depth_level ``k``
-        :param context_doc: the owner of ``docs``, it is at depth_level ``k-1``
+        :param docs: the matches of the ``context_doc``, they are at granularity ``k``
+        :param context_doc: the query document having ``docs`` as its matches, it is at granularity ``k``
         :return:
+
+        .. note::
+            - This driver will substitute the ``matches`` of `docs` to the corresponding ``parent documents`` of its current ``matches`` according
+            to the executor.
+            - Set the ``taversal_paths`` of this driver such that it traverses along the ``matches`` of the ``chunks`` at the granularity desired
```
maximilianwerk, line 138: typo: `traversal_paths`

maximilianwerk (changes requested):
minor things. Apart from that: Good job!

jina/drivers/rank.py
JoanFM: ok we can rename to `chunk` to make it more explicit that it should be at least someone's chunk

JoanFM:
@maximilianwerk  changed applied!

maximilianwerk (approved):

PR #1145, opened 2020-10-23
fix: random topk rest port

bhavsarpratik:


tests/integration/issues/github_976/test_topk_rest_api.py
```diff
@@ -1,22 +1,23 @@
...
+import pytest
+import random
 from urllib import request
 
 from jina.flow import Flow
 
 
 TOP_K = 2
-
+PORT = random.randint(1, 10000)
```
bwanglzu, line 10: we have a random port method built-in codebase 
```
from jina.helper import random_port
```

bwanglzu (changes requested):

bwanglzu:
It looks like an integration test thing how do you think @JoanFM

tests/integration/issues/github_976/test_topk_rest_api.py
hanxiao: yes, the reason of not randomint is many ports below 40000 are reserved by OS already. Example: 22, 80

bwanglzu:
hi @bhavsarpratik sorry I didn't realise that we're using `urllib.Request` here, do you think it's good to replace it with `requests`?  it's more readable and in our dependencies. Besides it's always better to respect our import logic

```
# python default
import json
import random
# community
import pytest
import requests
# Internal components
from jina.flow import Flow
from jina.helper import random_port
```
It's not necessary :) the rest lgtm!

JoanFM:
> hi @bhavsarpratik sorry I didn't realise that we're using `urllib.Request` here, do you think it's good to replace it with `requests`? it's more readable and in our dependencies. Besides it's always better to respect our import logic
> 
> ```
> # python default
> import json
> import random
> # community
> import pytest
> import requests
> # Internal components
> from jina.flow import Flow
> from jina.helper import random_port
> ```
> 
> It's not necessary :) the rest lgtm!

We decided to keep `urllib` to avoid external dependency, but yes I agree request is easier and more readable

tests/integration/issues/github_976/test_topk_rest_api.py
```diff
@@ -1,22 +1,25 @@
-import pytest
 import json
+import random
```
JoanFM, line 2: remove this import

JoanFM (changes requested):

bhavsarpratik:
I like requests too but I don't see it in our `extra-requirements.txt`. So we won't be able to use it.

bwanglzu (approved):
LGTM

PR #1146, opened 2020-10-23
style: black on merge

maximilianwerk:
This is a test to black one file first. It worked on CI and thus, I'd like to move it to CD. If that succeeded, I'll:

first: black the whole repository by hand, commit and merge it
second: enable autoblack on merge for the whole repository.

Why on a merge and not on a PR: That would mess with users commits and be not possible, when people have forked jina.


.github/workflows/cd.yml
```diff
@@ -12,7 +12,29 @@ on:
...
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install Black
+        run: pip install black
+      - name: Run black --check .
+        run: black --check -S jina/drivers/__init__.py
+      - name: If needed, commit black changes to the pull request
```
JoanFM, line 27: commit black changes to master?

JoanFM (commented):
Is this ready to review or merge? otherwise u can make it a draft

.github/workflows/cd.yml
```diff
@@ -12,7 +12,29 @@ on:
...
+        if: failure()
+        run: |
+          git checkout $GITHUB_HEAD_REF
+          black -S jina/drivers/__init__.py
+          git config --local user.email "dev-bot@jina.ai"
+          git config --local user.name "Jina Dev Bot"
+          git add . && git commit -m "update code-style ${{env.JINA_VERSION}} due to ${{github.event_name}} on ${{github.repository}}" && git push
```
hanxiao, line 34:
```suggestion
          git add . && git commit -m "update code-style ${{env.JINA_VERSION}} due to ${{github.event_name}} on ${{github.repository}}" && git push
```

this has two problems:
- recursive triggering CD flow
- commit msg is not styled

maximilianwerk: Fixed. Can you have another look? I took some more inspiration from `copyright.yml`

bhavsarpratik:
Do we want to keep `--line-length` default 88? This might reformat too many lines and sometimes long lines are better

maximilianwerk:
> Do we want to keep `--line-length` default 88? This might reformat too many lines and sometimes long lines are better

Since we usually want to PEP8 which is 79 chars line lengths, this is even bigger. And it you have to windows of code on one screen, more then 90 usually means, you can't see all. Also since we want to be inclusive, for people with are slightly limited vision it is more inclusive to have a not to long line-length, such that they can still easily read it when increasing font size.

Anyhow, if the majority of the team agrees, we can increase it.

.github/workflows/code-formatting.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install Black
+        run: pip install black
+      - name: Run black --check .
+        run: black --check -S jina/drivers/__init__.py
+      - name: If needed, commit black changes to master
+        if: failure()
```
JoanFM, line 30: does this failure could happen if `pip install black` fails too?

maximilianwerk: now it does. Thanks for pointing it out. `failure()` itself check for any previous failing step.

JoanFM (dismissed):

.github/workflows/code-formatting.yml
```diff
@@ -0,0 +1,37 @@
+name: Codeformatting
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  autoblack:
+    if: "!startsWith(github.event.head_commit.message, 'chore') && !startsWith(github.event.head_commit.message, 'build: hotfix')"
```
JoanFM, line 10: minor thing, but `hotfix` could also be formatted right?

.github/workflows/code-formatting.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Run black --check .
+        id: black_check
+        run: black --check -S jina/drivers/__init__.py
+      - name: If needed, commit black changes to master
+        if: "steps.black_check.outputs.exit_code != 0"
+        run: |
+          black -S jina/drivers/__init__.py
+          git add -u
+          git commit -m "chore: update code-style"
+          git status
```
JoanFM, line 36: minor non-blockinb, status could be avoided?

.github/workflows/code-formatting.yml
```diff
@@ -0,0 +1,37 @@
...
+          python-version: 3.7
+      - name: Install Black
+        run: pip install black
+      - name: Run black --check .
+        id: black_check
+        run: black --check -S jina/drivers/__init__.py
+      - name: If needed, commit black changes to master
+        if: "steps.black_check.outputs.exit_code != 0"
+        run: |
+          black -S jina/drivers/__init__.py
```
JoanFM, line 33: we could keep the path into a variable to make sure the `check` and the actual `change` is done based on the same path, but I guess it is also the first step and that is why we just put one file now

maximilianwerk: Later on it will be `.`. I think putting it in an env variable is overkill.

.github/workflows/code-formatting.yml
maximilianwerk: I think it is nice for logging purpose. Anyhow, after doing the full change, it can be removed.

.github/workflows/code-formatting.yml
maximilianwerk: I think the idea is to not interfere with hot fixes too much. It is the same behavior as for other github actions. Thus, I'd keep it in sync.

JoanFM (approved):

maximilianwerk:
Please review again @hanxiao

hanxiao (approved):

.github/workflows/code-formatting.yml
nan-wang: most of the `hotfix` is about releasing new versions.

PR #1150, opened 2020-10-23
chore: add packaging in extra requirements

JoanFM:
**Changes introduced**
Adding `packaging` as requirement for the `hub` so that https://github.com/jina-ai/jina-hub/pull/460 can pass.
It would affect any other `hub` build

### DIFF
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -60,4 +60,5 @@ pytest-repeat:              test
 flaky:                      test
 mock:                       test
 pymongo[srv]:               hub, devel, cicd
+packaging:                  hub
 


hanxiao:
i think `package` is part of `setuptools`, which user should have after install `pip`?

nan-wang (approved):
LGTM👍

hanxiao (changes requested):
for the reason i said

maximilianwerk:
> i think `packaging` is part of `setuptools`, which user should have after install `pip`?
> 
> i will take a close look tomorrow

I think that is not true. On a fresh `venv`:
```
> python
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
>>> import packaging
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'packaging'
>>> 
```

hanxiao:
hmm, then this makes `_is_latest_version` unusable, we also use `packaging` in it. need a workaround in another PR

hanxiao (approved):

PR #1154, opened 2020-10-24
feat: help multimodal provide batching

JoanFM:


tests/integration/multimodal/yaml/multimodal-concatenate.py
```diff
@@ -8,12 +8,14 @@
...
     def __init__(self,
                  *args,
                  **kwargs):
         super().__init__(*args, **kwargs)
 
     @batching
     @as_ndarray
-    def encode(self, data: Dict[str, Any], *args, **kwargs):
-        return np.concatenate((data['modality1'], data['modality2']), axis=1)
+    def encode(self, data: 'np.ndarray', *args, **kwargs):
```
JoanFM, line 20: this needs to be adapted to do concatenation ( I am not so handy with numpy magic)

PR #1155, opened 2020-10-24
fix(driver): fix route driver by singleton

minfun:


JoanFM:
Hello @leifanwang-aa ,

Could you please elaborate on why do you think the `RouteDriver` should be a `singleton`?

minfun:
Hi @JoanFM ,
I see the docstring in the RouteDriver, it should be A simple load balancer, so for the load balancer, I think it should be a global instance to do the load balance, and if it is a singleton it also can resolve this issue. But I am not really sure about this part, so do you have better advice for it?

JoanFM:
> Hi @JoanFM ,
> I see the docstring in the RouteDriver, it should be A simple load balancer, so for the load balancer, I think it should be a global instance to do the load balance, and if it is a singleton it also can resolve this issue. But I am not really sure about this part, so do you have better advice for it?

Hey @minfun ,

Although your reasoning makes sense. jina is a `framework` that provides the building blocks to build neural search applications, so the user can use them as they want. Right now, if the user just uses one in one `Pea` the proper behavior is guaranteed. 
So I would prefer not to merge this PR also not to start importing dependencies of `multithreading` and resource locking.

Thank you very much for the effort.

minfun:
Hi @JoanFM ,
Sure, also thank you for your review.

JoanFM:
I think the idea was very good, though, it is just that we are not yet looking at this.

PR #1165, opened 2020-10-26
refactor(driver): eval drivers focus on data type of field

hanxiao:


JoanFM (dismissed):
More coherent with the recent changes in `Executors` :+1

PR #1168, opened 2020-10-27
fix: disabled pip cache in docker

### LINKED ISSUE #1167: fix: disable pip cache in docker
**Describe the bug**
All of our docker containers contain pip cache which increases its size. This cache is useless.

**Describe how you solve it**
Add these to our jina debianx dockerfile. This will lead to smaller jina and hub images especially those with big dependencies like tensorflow and torch. This way we don't need to use flag `--no-cache-dir` in pip installs everywhere.
I also suggest disabling pip version check. 
```
ENV PIP_NO_CACHE_DIR=1
ENV PIP_DISABLE_PIP_VERSION_CHECK=1
```

bhavsarpratik:
Describe the bug
All of our hub docker containers contain pip cache which increases its size. This cache is useless.

Describe how you solve it
Add these to our jina dockerfiles. This will lead to smaller jina and hub images especially those with big dependencies like tensorflow and torch. This way we don't need to use flag `--no-cache-dir` in pip installs everywhere.
I also suggest disabling pip version check.

```
ENV PIP_NO_CACHE_DIR=1
ENV PIP_DISABLE_PIP_VERSION_CHECK=1
```

resolved #1167

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           jina check
           export JINA_LOG_VERBOSITY="ERROR"
           pip install .[test]
-          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=240 -v tests/unit
+          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=600 -v tests/unit
           mv coverage.xml unit-test-coverage.xml
         timeout-minutes: 30
         env:
@@ -102,7 +102,7 @@ jobs:
           jina check
           export JINA_LOG_VERBOSITY="ERROR"
           pip install .[test]
-          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=240 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' tests/integration
+          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=600 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' tests/integration
           mv coverage.xml integration-test-coverage.xml
         timeout-minutes: 30
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
           pip install .[test]
           pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=120 -v tests/unit
           mv coverage.xml unit-test-coverage.xml
-        timeout-minutes: 30
+        timeout-minutes: 40
         env:
           JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
           JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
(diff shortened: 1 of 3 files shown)


Dockerfiles/alpinex.Dockerfile
```diff
@@ -23,11 +23,13 @@ ADD cli ./cli/
 ADD jina ./jina/
 
 ENV PYTHONPATH=$PYTHONPATH:/usr/lib/python3.8/dist-packages:/usr/local/lib/python3.8/site-packages:/usr/lib/python3/dist-packages:/usr/local/lib/python3/site-packages
+ENV PIP_NO_CACHE_DIR=1
```
JoanFM, line 26: Isnt it possible to put all these in the same layer?

.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           jina check
           export JINA_LOG_VERBOSITY="ERROR"
           pip install .[test]
-          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=120 -v tests/unit
+          pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=600 -v tests/unit
```
nan-wang, line 63: Is there any reason why we need to increase the timeout for each test?

JoanFM: I guess is because we are suffering quite often from timeouts

bhavsarpratik: Earlier CI was failing due to timeout of test

hanxiao: please revert this change, the timeout is a false alarm. it is the port conflict problem caused by flaky tests.

hanxiao (changes requested):
Please revert the timeout, it is set to this number **empirically** for saving our time. Note, a timeout is almost always due to flaky test or bad implementation that causes thread/process trouble. **It is never because of we need more computation time for the test**. 

More discussion can be found from early slack discussion with Joan back in July.

bhavsarpratik:
@hanxiao Please confirm and merge

.github/workflows/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           pip install .[test]
           pytest --force-flaky --min-passes 1 --cov=jina --cov-report=xml -n 1 --timeout=120 -v tests/unit
           mv coverage.xml unit-test-coverage.xml
-        timeout-minutes: 30
+        timeout-minutes: 40
```
hanxiao, line 65:
```suggestion
        timeout-minutes: 30
```

hanxiao:
this looks good but let me explain why didn't I use `PIP` env variables. When someone uses `jina` as a base image, they expect it to provide a clean and standard Python enviroment. Those predefined env variables shipped with docker image could hide the bugs and our users could take some time to even realize the problem is about the image. So in general, this isn't the direction we should encourage. 

The two environmental variables you introduced here seems no harm at now. So good to go.

hanxiao (approved):
approve but please read my comment above @nan-wang @JoanFM @bhavsarpratik

hanxiao:
Just fyi, the following command would do the same:

```
pip --disable-pip-version-check --no-cache-dir
```

bhavsarpratik:
I agree. The idea to add ENV variable here is to disable cache in all downstream containers. Otherwise, all current hub executor dockerfiles pip install need to add `--disable-pip-version-check --no-cache-dir`.

PR #1169, opened 2020-10-27
fix: failure not combinable

maximilianwerk:
Sorry for the first try in a row. But `failure()` is not combinable @JoanFM . Anyhow, if any step before the `black --check...` fails, the last one will fail as well and nothing bad will be committed.

### DIFF
--- a/.github/workflows/code-formatting.yml
+++ b/.github/workflows/code-formatting.yml
@@ -25,10 +25,9 @@ jobs:
       - name: Install Black
         run: pip install black
       - name: Run black --check .
-        id: black_check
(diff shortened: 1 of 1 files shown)


JoanFM:
> Sorry for the first try in a row. But `failure()` is not combinable @JoanFM . Anyhow, if any step before the `black --check...` fails, the last one will fail as well and nothing bad will be committed.

Fair enough! I hope it was good learning at least!

JoanFM (approved):

nan-wang (approved):
LGTM👍

hanxiao (approved):

PR #1170, opened 2020-10-27
feat: allow zipped-like batching

JoanFM:
**Changes introduced**
Since the introduciton of multimodal driver, we need a new way of batching, were batching is done with a zipped set of datas.

Since `batching` decorator is growing to be very complex, I decided to create a new one for now.


jina/executors/decorators.py
```diff
@@ -255,11 +274,79 @@ def arg_wrapper(*args, **kwargs):
...
+    """
+
+    def _batching(func):
+        @wraps(func)
+        def arg_wrapper(*args, **kwargs):
+            # priority: decorator > class_attribute
+            # by default data is in args[1:] (self needs to be taken into account)
+            data_args = [args[slice_on + i] for i in range(0, num_data)]
+
+            data = list(zip(*data_args))
```
maximilianwerk, line 318: One of the reasons to do batching is, keeping the memory footprint low. explicit conversion to a `list` enforces whatever iterators are behind `data_args` to be converted to a list in memory.

I think batching should treat whatever input it gets as a stream and not instantiate the full stream directly. Or if it does, we should change the docstring, that the caller function should already care about only calling the to-be-batched function with in-memory data.

jina/executors/decorators.py
```diff
@@ -255,11 +274,79 @@ def arg_wrapper(*args, **kwargs):
...
+            args = list(args)
+
+            b_size = batch_size or getattr(args[0], 'batch_size', None)
+            # no batching if b_size is None
+            if b_size is None:
+                return func(*args, **kwargs)
```
maximilianwerk, line 324: This should be done first, before calculation of `data_args`

jina/executors/decorators.py
```diff
@@ -255,11 +274,79 @@ def arg_wrapper(*args, **kwargs):
...
+            default_logger.info(
+                f'batching enabled for {func.__qualname__} batch_size={b_size} '
+                f'num_batch={num_batch} axis={split_over_axis}')
+
+            # assume all datas have the same length
+            total_size = _get_total_size(data, b_size, num_batch, split_over_axis)
+            final_result = []
+
+            for multiple_batch in batch_iterator(data[:total_size], b_size, split_over_axis):
+                multiple_batch = list(zip(*multiple_batch))
```
maximilianwerk, line 335: Why doing that? Why not have one batch_iterator per data field? Something like
```
total_size = _get_total_size(args[slice_on], ...)
data_iterators = [batch_iterator(args[slice_on + i][:total_size], b_size, split_over_axis) for i in range(0, num_data)]

Now iterate over the first one and for each step call next() on all the others.
```

That would also help with the copy to memory issue.

jina/executors/decorators.py
```diff
@@ -255,11 +274,79 @@ def arg_wrapper(*args, **kwargs):
...
+
+                for idx, data_batch in enumerate(multiple_batch):
+                    args[slice_on + idx] = data_batch
```
maximilianwerk, line 338: I'd rather build a fresh copy of `args` instead of using the input variable.

jina/executors/decorators.py
```diff
@@ -255,11 +274,79 @@ def arg_wrapper(*args, **kwargs):
...
+
+                r = func(*args, **kwargs)
+
+                if r is not None:
+                    final_result.append(r)
```
maximilianwerk, line 343: append or extend? Not sure, was just wondering

tests/unit/executors/test_decorators.py
```diff
@@ -203,3 +203,30 @@ def f(self, data, labels):
...
+
+    instance = A(2)
+    data0 = np.random.rand(4, 2)
+    data1 = np.random.rand(4, 4)
+    data2 = np.random.rand(4, 6)
+    data = [data0, data1, data2]
+    result = instance.f(*data)
+    assert result.shape == (4, 12)
+    assert len(result) == 4
+    assert len(result[0]) == 12
```
maximilianwerk, line 229: These two are already covered by `assert result.shape == (4, 12)`.

maximilianwerk (commented):
Some opinionated comments. Ideally I would love to get rid of the copy iterator content to memory issue.

jina/executors/decorators.py
JoanFM: the data in args can be quite heavy (if u see in the simpler batching, we are also manipulating the `args` itself),  plus the complete point of the decorator is to silently manipulate the `arguments` passed to the decorated function.

jina/executors/decorators.py
JoanFM: Seems like a very good idea, let me try it out!

jina/executors/decorators.py
JoanFM: That was one of my first concerns when doing it, but did not see any point about it. With your solution in another comment it seems a better solution for it.

On the other hand, I was thinking that `requests` do live in memory, and not coming as stream. But I do definitely prefer to keep the abstraction!

JoanFM:
Really thankful @maximilianwerk ! You touched exactly the points I was more concerned about and more insecure when doing it!

jina/executors/decorators.py
JoanFM: I would break these 2 methods into smaller parts to share logic, but right now `single batching` has grown really into a big giant

jina/executors/decorators.py
JoanFM: Loved this, I was looking for this yesterday and did not come up with this solution!

JoanFM:
@maximilianwerk could you take a new look?

jina/executors/decorators.py
```diff
@@ -313,29 +313,30 @@ def _batching(func):
...
-                    args[slice_on + idx] = data_batch
+            data_iterators = [batch_iterator(args[slice_on + i][:total_size], b_size, split_over_axis) for i in
+                              range(0, num_data)]
+
+            for batch in data_iterators[0]:
+                for idx in range(0, num_data):
+                    if idx == 0:
+                        args[slice_on + idx] = batch
+                    else:
+                        args[slice_on + idx] = next(data_iterators[idx])
```
maximilianwerk, line 339: That is heavily opinionated, but I find the following more explicit:
```
for batch in data_iterators[0]: 
    args[slice_on] = batch
    for idx in range(1, num_data): 
        args[slice_on + idx] = next(data_iterators[idx])
```

jina/executors/decorators.py
```diff
@@ -313,29 +313,30 @@ def _batching(func):
...
-
-            data = list(zip(*data_args))
-            args = list(args)
-
             b_size = batch_size or getattr(args[0], 'batch_size', None)
             # no batching if b_size is None
             if b_size is None:
                 return func(*args, **kwargs)
 
+            data_args = [args[slice_on + i] for i in range(0, num_data)]
```
maximilianwerk, line 321: Not needed anymore

jina/executors/decorators.py
```diff
@@ -313,29 +313,30 @@ def _batching(func):
...
+            args = list(args)
```
maximilianwerk, line 322: Why that? I always thought `*args` must be a list.

maximilianwerk (changes requested):
Some more comments. But looks much better now!

jina/executors/decorators.py
JoanFM: it seems it is a tuple

JoanFM: Yes, it is a tuple and we pass it to a list to make it mutable

JoanFM: So it is something done already in the previous `batching`

maximilianwerk: Cool interesting insight. Thanks for clarification

maximilianwerk (dismissed):
LGTM

jina/executors/decorators.py
```diff
@@ -255,11 +274,76 @@ def arg_wrapper(*args, **kwargs):
...
+        def arg_wrapper(*args, **kwargs):
+            # priority: decorator > class_attribute
+            # by default data is in args[1:] (self needs to be taken into account)
+            b_size = batch_size or getattr(args[0], 'batch_size', None)
+            # no batching if b_size is None
+            if b_size is None:
+                return func(*args, **kwargs)
+
+            args = list(args)
+            default_logger.info(
```
nan-wang, line 322: This looks quite verbose for info in production. 
```suggestion
            default_logger.debug(
```

jina/executors/decorators.py
```diff
@@ -255,11 +274,76 @@ def arg_wrapper(*args, **kwargs):
...
+    , collect the merged result and return. This is useful when the input is too big to fit into memory
+
+    :param func: function to decorate
+    :param batch_size: size of each batch
+    :param num_batch: number of batches to take, the rest will be ignored
+    :param split_over_axis: split over which axis into batches
+    :param merge_over_axis: merge over which axis into a single result
+    :param slice_on: the location of the data. When using inside a class,
+            ``slice_on`` should take ``self`` into consideration.
+    :param num_data: the number of data inside the arguments (starting from ``slice_on``)
```
nan-wang, line 294: > starting from ``slice_on``
This should goes to `.. warning:: `
```suggestion
    :param num_data: the number of data inside the arguments (starting from ``slice_on``)
```

tests/unit/executors/test_decorators.py
```diff
@@ -203,3 +203,28 @@ def f(self, data, labels):
...
+    instance = A(2)
+    data0 = np.random.rand(4, 2)
+    data1 = np.random.rand(4, 4)
+    data2 = np.random.rand(4, 6)
+    data = [data0, data1, data2]
+    result = instance.f(*data)
+    assert result.shape == (4, 12)
+    assert len(instance.batching) == 2
+    for batch in instance.batching:
+        assert batch.shape == (2, 12)
```
nan-wang, line 230: Make the test easier to interpret. 

```
def test_batching_multi():
    num_data = 3

    class A:
        def __init__(self, batch_size):
            self.batch_size = batch_size
            self.batching = []

        @batching_multi_input(num_data=num_data)
        def f(self, *datas):
            assert len(datas) == num_data
            concat = np.concatenate(datas, axis=1)
            self.batching.append(concat)
            return concat

    num_docs = 4
    batch_size = 2
    instance = A(batch_size)
    data0 = np.random.rand(num_docs, 2)
    data1 = np.random.rand(num_docs, 4)
    data2 = np.random.rand(num_docs, 6)
    data = [data0, data1, data2]
    result = instance.f(*data)
    from math import ceil
    result_dim = sum([d.shape[1] for d in data])
    assert result.shape == (num_docs, result_dim)
    assert len(instance.batching) == ceil(num_docs / batch_size)
    for batch in instance.batching:
        assert batch.shape == (batch_size, result_dim)
```

tests/integration/multimodal/yaml/all-types-multimodal-encoder.py
```diff
@@ -13,8 +14,13 @@ def __init__(self,
                  **kwargs):
         super().__init__(*args, **kwargs)
 
+    @batching_multi_input(num_data=4)
+    @as_ndarray
     def encode(self, *data: 'np.ndarray', **kwargs):
         assert len(data) == 4
+        for d in data:
+            assert self.batch_size == 10
```
nan-wang, line 22:
```suggestion
            assert self.batch_size == batch_size
```

nan-wang (changes requested):
Minor changes on the tests. In general, lgtm

jina/executors/decorators.py
```diff
@@ -255,11 +274,76 @@ def arg_wrapper(*args, **kwargs):
...
-        total_size = data.shape[axis]
-    elif hasattr(data, '__len__'):
-        total_size = len(data)
+def batching_multi_input(func: Callable[[Any], np.ndarray] = None, *,
+                         batch_size: Union[int, Callable] = None,
+                         num_batch: Optional[int] = None,
+                         split_over_axis: int = 0,
+                         merge_over_axis: int = 0,
+                         slice_on: int = 1,
+                         num_data: int) -> Any:
```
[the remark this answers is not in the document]
JoanFM: Which pattern, do not get it much?

tests/integration/multimodal/yaml/all-types-multimodal-encoder.py
JoanFM: Done!

tests/unit/executors/test_decorators.py
JoanFM: Done!

jina/executors/decorators.py
JoanFM: Done!

jina/executors/decorators.py
JoanFM: Done!

jina/executors/decorators.py
```diff
@@ -189,13 +229,7 @@ def arg_wrapper(*args, **kwargs):
...
                 f'num_batch={num_batch} axis={split_over_axis}')
 
-            full_data_size = _get_size(data, split_over_axis)
-            batched_data_size = b_size * num_batch if num_batch else None
-
-            if full_data_size is not None and batched_data_size is not None:
-                total_size = min(full_data_size, batched_data_size)
-            else:
-                total_size = full_data_size or batched_data_size
+            total_size = _get_total_size(data, batch_size, num_batch, split_over_axis)
```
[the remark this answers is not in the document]
JoanFM: Done!

jina/executors/decorators.py
```diff
@@ -133,6 +133,46 @@ def arg_wrapper(self, *args, **kwargs):
...
+    if isinstance(data, np.ndarray):
+        total_size = data.shape[axis]
+    elif hasattr(data, '__len__'):
+        total_size = len(data)
+    else:
+        total_size = None
+    return total_size
+
+
+def _get_total_size(data, batch_size, num_batch, split_over_axis):
```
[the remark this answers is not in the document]
JoanFM: Done!

JoanFM:
@nan-wang, changes have been addressed

jina/executors/decorators.py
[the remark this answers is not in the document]
maximilianwerk: You added a `, *,` at the end of line 277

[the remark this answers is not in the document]
JoanFM: oh, thanks @maximilianwerk I did not see it, it is the same signature as batching almost. Let me see what happens when removed.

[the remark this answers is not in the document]
JoanFM: Ok, removed @nan-wang, I see I guess it was introduced when some parameters were mixed with and without default.

hanxiao (approved):

nan-wang (approved):
LGTM👍

PR #1172, opened 2020-10-27
refactor(hubio): remove duplicate logic on hub push

hanxiao:


jina/docker/hubio.py
```diff
@@ -258,8 +265,6 @@ def _docker_login(self) -> None:
                 self.logger.debug(f'successfully logged in to docker hub')
             except APIError:
                 raise DockerLoginFailed(f'invalid credentials passed. docker login failed')
-        else:
-            raise DockerLoginFailed('no username/password specified, docker login failed')
```
maximilianwerk, line 262: At least log something in the case of failure?

jina/docker/hubio.py
```diff
@@ -118,25 +119,21 @@ def login(self) -> None:
...
-                        self.logger.info(f'successfully logged in!')
-                        break
+            for j in range(login_max_retry):
+                response = requests.post(url=access_token_url,
+                                         headers=headers,
+                                         data=access_request_body)
+                access_token_response = response.json()
+                if 'error' in access_token_response and access_token_response['error'] == 'authorization_pending':
+                    countdown(10, reason=colored('re-fetch access token', 'cyan', attrs=['bold', 'reverse']))
+                elif 'access_token' in access_token_response:
```
maximilianwerk, line 129: Can this `elif` also fail? If so, what should happen in the `else` case?

jina/docker/hubio.py
hanxiao: no credentials will follow docker host auth context, therefore this `raise` is unnecessary in the first place

hanxiao:
related https://github.com/jina-ai/cloud-ops/pull/30

PR #1173, opened 2020-10-27
test: add test ranking evaluator with score field

JoanFM:
**Changes introduced**
Have test that will give more confidence that https://github.com/jina-ai/jina-hub/pull/461 will work

### DIFF
--- a/tests/unit/drivers/test_rankingevaluation_driver.py
+++ b/tests/unit/drivers/test_rankingevaluation_driver.py
@@ -51,6 +51,7 @@ def ground_truth_pairs():
         for idx in range(num_matches):
             match = doc.matches.add()
             match.tags['id'] = idx
+            match.tags['score'] = idx
 
     pairs = []
     for idx in range(num_docs):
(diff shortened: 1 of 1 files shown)


tests/unit/drivers/test_rankingevaluation_driver.py
```diff
@@ -170,3 +171,66 @@ def test_evaluate_assert_doc_groundtruth_structure(simple_chunk_evaluate_driver,
...
+        actual_at_k = actual[:self.eval_at]
+        desired_at_k = desired[:self.eval_at]
+        if len(actual) < 2:
+            raise ValueError(f'Expecting gains with minimal length of 2, {len(actual)} received.')
+        dcg = _compute_dcg(gains=actual_at_k)
+        idcg = _compute_idcg(gains=desired_at_k)
+        if idcg == 0.0:
+            return 0.0
+        else:
+            return dcg / idcg
```
bhavsarpratik, line 214:
```suggestion
        return 0.0 if idcg == 0.0 else dcg / idcg
```

tests/unit/drivers/test_rankingevaluation_driver.py
```diff
@@ -170,3 +171,66 @@ def test_evaluate_assert_doc_groundtruth_structure(simple_chunk_evaluate_driver,
...
+    ret = 0.0
+    for score, position in zip(gains[1:], range(2, len(gains) + 1)):
+        ret += score / log(position, 2)
+    return gains[0] + ret
+
+
+def _compute_idcg(gains):
+    """Compute ideal discounted cumulative gain."""
+    sorted_gains = sorted(gains, reverse=True)
+    return _compute_dcg(sorted_gains)
```
bhavsarpratik, line 188: Why do we have NDCG logic in the test?

JoanFM: just for a test, I would like to have something more stupid there, can u think of some other stupid measure we can take based on scores? I also do not like having this logic here.

bhavsarpratik: I think precision_at_k is the dumbest and can be taken from [here](https://gist.github.com/bwhite/3726239)

JoanFM: The point is that precision is based on ids of matches while ndcg works on relevancy scores. That is the difference I want to test, but anyway is an unnecesssry test. Will close

PR #1180, opened 2020-10-28
refactor(encoder): change get_model to get_cell mindspore idiom

hanxiao:
`cell` is the idiom in Mindspore context

### DIFF
--- a/jina/docker/hubio.py
+++ b/jina/docker/hubio.py
@@ -336,7 +336,7 @@ class HubIO:
 
             if is_build_success:
                 if self.args.test_uses:
-                    from jina.excepts import PretrainedModelFileDoesNotExist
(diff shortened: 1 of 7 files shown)


jina/docker/hubio.py
```diff
@@ -336,7 +336,7 @@ def build(self) -> Dict:
 
             if is_build_success:
                 if self.args.test_uses:
-                    from jina.excepts import PretrainedModelFileDoesNotExist
+                    from jina.excepts import ModelCheckpointNotExist
```
JoanFM, line 339: It is a problem, because a lot of `executors` in the hub will be able to be built again.

hanxiao: sure, let me inherit it

Issue #1171, opened 2020-10-27
feat: add hub component when running docker


theUnkownName:
**Describe the feature**
Add the hub component when building docker from jina:devel

**Your proposal**
When building a docker image currently the hub is being skipped. 
This behaviour is intentional and was previously [introduced](https://github.com/jina-ai/jina/pull/883) due to this [problem](https://github.com/jina-ai/jina-hub/pull/295#issuecomment-683279092). However, it would be nice to bring it back when using the `devel` tag.

nan-wang:
Do you mean to run all the hub tests in jina?

JoanFM:
> Do you mean to run all the hub tests in jina?

The point is that to build `docker` examples we cannot be based on `FROM jinaai:jina` because the jina image itself do not contain the `hub executors`. This is intentional for the `jinaai/jina:latest` but we could have them for `devel`

PR #1184, opened 2020-10-29
fix(docker): include submodule when building devel images

hanxiao:
`devel` image is generated in two places:
- `cd.yml` on master merging, building x64 image
- `nightly.yml` midnight on building multi-arch devel images

Fixing these two places solves #1171


JoanFM (approved):

nan-wang:
For `cicd`, we can set another docker image tag because hub is not used there and we can resources.

PR #1187, opened 2020-10-29
fix: rely on control request to set ready docker pea

JoanFM:
**Changes introduced**
Instead of relying on reading messages from the logs to set ready signal to the Pod, it relies in network connection with the Pea running inside docker.

To ensure that the PretrainedModelException is properly passed when building executors in the hub, we are still basing ourselves on reading log messages. Not sure if there is a solution for it

**Discussion**
Since there are 2 processes where the `ContainerPea` lives right now, if we start the `docker` container in a `detach=False` way, we may have better control on how its resources are used, and maybe we can avoid all this inner communication? I am not an expert but in non-detached mode we may be able to extract exceptions and errors?

### DIFF
--- a/jina/peapods/container.py
+++ b/jina/peapods/container.py
@@ -5,6 +5,8 @@ import os
 from pathlib import Path
 
 from .pea import BasePea
+from .zmq import send_ctrl_message
+from ..proto import jina_pb2
 from .. import __ready_msg__, __unable_to_load_pretrained_model_msg__
 from ..helper import is_valid_local_config_source, kwargs2list, get_non_defaults_args
 from ..logging import JinaLogger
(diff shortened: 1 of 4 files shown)


jina/peapods/container.py
```diff
@@ -109,6 +112,18 @@ def loop_teardown(self):
...
 
+    @property
+    def status(self):
+        """Send the control signal ``STATUS`` to itself and return the status """
+        if getattr(self, 'ctrl_addr'):
+            return send_ctrl_message(self.ctrl_addr, jina_pb2.Request.ControlRequest.STATUS,
+                                     timeout=self.args.timeout_ctrl)
+
+    @property
+    def get_ready(self) -> bool:
```
hanxiao, line 123:
```suggestion
    def is_ready(self) -> bool:
```

jina/peapods/container.py
```diff
@@ -65,37 +67,50 @@ def post_init(self):
...
+        def check_ready():
+            import asyncio
+            while not self.is_ready:
+                asyncio.sleep(0.1)
+            self.is_ready_event.set()
+            self.logger.success(__ready_msg__)
+            return True
+
+        async def _loop_body():
+            import asyncio
```
deepankarm, line 93: Can we move the import to the top of the file, since this is a standard Python import?

JoanFM: Done!

deepankarm (dismissed):

hanxiao (approved):

PR #1189, opened 2020-10-30
fix: pod argument overriden by flow

maximilianwerk:
The explicitly defined values on pod level should have higher priority, then the ones on flow level.

### DIFF
--- a/jina/flow/__init__.py
+++ b/jina/flow/__init__.py
@@ -293,7 +293,10 @@ class Flow(ExitStack):
 
         needs = op_flow._parse_endpoints(op_flow, pod_name, needs, connect_to_last_pod=True)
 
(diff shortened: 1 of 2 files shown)


jina/flow/__init__.py
```diff
@@ -293,7 +293,10 @@ def add(self,
 
         needs = op_flow._parse_endpoints(op_flow, pod_name, needs, connect_to_last_pod=True)
 
-        kwargs.update(op_flow._common_kwargs)
+        for key, value in op_flow._common_kwargs.items():
```
JoanFM, line 296: missing some tests to show the problem and the feature

JoanFM (changes requested):

JoanFM (approved):

hanxiao (approved):

Issue #1174, opened 2020-10-27
🚨 BREAKING CHANGES 0.7.0 -> 0.8.0


nan-wang:
This issue is created to keep track of breaking changes between `0.7.0` and `0.8.0` (scheduled release on Nov. 23, 2020)

## Breaking Changes & Migration Guide

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1190  | Yes |  #1283  |
| #1283  | Yes |   |

Add support to sparse `ndarray`.

- Update `jina.proto`
    - Add `SparseNdArray`
    - Rename `NdArray` to `DenseNdArray`
    - Add `NdArray` to provide generic interface for `SparseNdArray` and `DenseNdArray`

- Add `GenericNdArray()` function in the `jina.proto.ndarray.generic` for accessing `.blob` and `.embedding`. 

<table stype="margin-left:auto;margin-right:auto;">
<tr>
<td>  </td>
<td> v0.7.0 </td>
<td> v0.8.0 </td>
</tr>
<tr>
<td> dense array </td>
<td> <sub> 

```python
from jina.proto import jina_pb2
from jina.proto import jina_pb2
from jina.drivers.helper import array2pb
a = jina_pb2.jina_pb2.NdArrayProto()
a.CopyFrom(array2pb(np.ndarray([2, 17])))
```

</sub> </td>
<td> <sub> 

```python
from jina.types.ndarray.generic import NdArray
a = NdArray()
a.value = np.ndarray([2, 17])
```

</sub> </td>
</tr>

<tr>
<td> sparse array </td>
<td> <sub> 

```python
not support
```

</sub> </td>
<td> <sub> 

```python
from jina.types.ndarray.generic import NdArray
from .sparse.scipy import SparseNdArray
from scipy.sparse import coo_matrix
row = np.array([20, 0])
col = np.array([0, 20])
data = np.array([2, 17])
a = NdArray(is_sparse=True, sparse_cls=SparseNdArray)
a.value = coo_matrix((data, (row, col)), shape=(21, 21))

```

</sub> </td>
</tr>
</table>

---

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1210  | Yes | #1283  |
| #1283  | Yes |   |

Add `ProtoMessage`, `LazyRequest` to replace the original `jina_pb2.Message` and `jina_pb2.Request` so that the protobuf message is deserialized in a lazy way.

<table stype="margin-left:auto;margin-right:auto;">
<tr>
[...]

PR #1190, opened 2020-10-30
feat(proto): introduce sparse ndarray

hanxiao:
## Breaking Change #1174 
- `.blob`, `.embedding` should not be accessed directly, they should be accessed via `GenericNdArray(doc.blob).value` and `GenericNdArray(doc.embedding).value`
- detailed info and usage can be found in `jina.proto.ndarray.generic.GenericNdArray`

## Content
- proto: Add `SparseNdArray`
- proto: Old `NdArray` is renamed to `DenseNdArray`
- proto: Add `NdArray` to provide generic interface for `SparseNdArray` and `DenseNdArray`
- python API: Add `GenericNdArray` to `jina.proto.ndarray.generic`, corresponds to `NdArray`
- python API: Add tensorflow, scipy, pytorch Sparse ndarray backend
- Add testing
- Refator old test due to breaking changes on access of `.blob` & `.embedding`

## Out-of-scope
- Executor supporting sparse ops 

## Reference

Main refs come from Tensorflow, Pytorch, Scipy, ONNX and their data representation on Sparsity.
- tf: https://www.tensorflow.org/api_docs/python/tf/sparse/SparseTensor
- pytorch: https://pytorch.org/docs/stable/sparse.html
- scipy: https://docs.scipy.org/doc/scipy/reference/sparse.html
- onnx: https://github.com/onnx/onnx/blob/2cee65c7eaa4b4fa8f6e23ef7e4b546395b9b7f1/onnx/onnx.in.proto#L600-L622


jina/proto/jina.proto
```diff
@@ -53,7 +53,7 @@ message SparseNdArray {
     DenseNdArray values = 2;
 
     // A 1-D int64 tensor of shape [ndims], which specifies the dense_shape of the sparse tensor.
-    repeated uint64 dense_shape = 3;
+    repeated int64 dense_shape = 3;
```
JoanFM, line 56: why not uint?

hanxiao: i had the same question, but after looking at all references i have onnx, tf, they all use int64

hanxiao: my guess is that `int64` is more natural to join the computation when deploying on multi-arch machines, cpu, gpu, embedded device.

JoanFM: so more for being conservative?

JoanFM: or that we suspect they must have their own reasons and better be conservative?

hanxiao: my experience is that `shape` sometimes join the computation flow of those frameworks, explicitly casting probably better.

JoanFM: ok did not see second message yet

jina/ndarray/sparse/scipy.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
JoanFM, line 1: I would move this module ndarray into proto. Or even create one data where you have ndarray and proto. But I would not have it at the same level as executors, peapods, drivers, etc...

tests/unit/test_ndarray.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
```
JoanFM, line 1: Also better put tests in their module

PR #1192, opened 2020-10-31
test: refactor workspace test

bwanglzu:
During testing, discovered error in `CompoundExecutor` not saving it's ~components~ and itself to `filename`.

1. Take out `JinaTestCase` from `test_workspace`
2. Clean up `test_workspace`
3. Save `CompoundExecutor` itself to `filename`, as it's base class (if `filename` exist).

### DIFF
--- a/jina/executors/compound.py
+++ b/jina/executors/compound.py
@@ -231,8 +231,8 @@ class CompoundExecutor(BaseExecutor):
         """
 
         for c in self.components:
-            c.save()
-        super().save()  # do i really need to save the compound executor itself
+            c.save(filename=filename)
+        super().save(filename=filename)  # do i really need to save the compound executor itself
         return True
 
     @property
--- a/jina/executors/compound.py
+++ b/jina/executors/compound.py
@@ -231,7 +231,7 @@ class CompoundExecutor(BaseExecutor):
         """
 
         for c in self.components:
-            c.save(filename=filename)
+            c.save()
         super().save(filename=filename)  # do i really need to save the compound executor itself
         return True
 
--- a/tests/unit/test_workspace.py
+++ b/tests/unit/test_workspace.py
@@ -1,88 +1,72 @@
 import os
 
+import pytest
 import numpy as np
 
 from jina.executors import BaseExecutor
-from tests import JinaTestCase
 
-cur_dir = os.path.dirname(os.path.abspath(__file__))
+@pytest.mark.parametrize('replica_id', [1,2,3])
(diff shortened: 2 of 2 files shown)


jina/executors/compound.py
```diff
@@ -231,8 +231,8 @@ def save(self, filename: str = None) -> bool:
         """
 
         for c in self.components:
-            c.save()
-        super().save()  # do i really need to save the compound executor itself
+            c.save(filename=filename)
```
JoanFM, line 234: It seems weird to store the executors formingn the compound executor and the compound executor itself in the same file

tests/unit/test_workspace.py
```diff
@@ -1,88 +1,72 @@
...
 
+import pytest
 import numpy as np
 
 from jina.executors import BaseExecutor
-from tests import JinaTestCase
 
-cur_dir = os.path.dirname(os.path.abspath(__file__))
+@pytest.mark.parametrize('replica_id', [1,2,3])
+def test_share_workspace(tmpdir, replica_id):
```
JoanFM, line 9: The term replica_id should be replaces by pea_id.

JoanFM (dismissed):

tests/unit/test_workspace.py
bwanglzu: this is what we have in the definition:
```python
    @classmethod
    def load_config(cls: Type[AnyExecutor], source: Union[str, TextIO], separated_workspace: bool = False,
                    replica_id: int = 0) -> AnyExecutor:
```

JoanFM: I see, we will refactor then

bwanglzu: do you have idea that I always get an error: `E           AttributeError: 'NoneType' object has no attribute 'header'` when calling `add`, such as
```python
def test_compound_indexer_rw():
    all_vecs = np.random.random([6, 5])
    for j in range(3):
        with BaseExecutor.load_config('yaml/test-compound-indexer2.yml', True, j) as a:
            assert a[0] == a['test_meta']
            assert not a[0].is_updated
            assert not a.is_updated
            a[0].add([j, j * 2, j * 3], [bytes(j), bytes(j * 2), bytes(j * 3)])
```
The last test case never passed locally

bwanglzu: wired.. passed in CI..

JoanFM: Did u clean ur workspace?

hanxiao (approved):

PR #1193, opened 2020-11-01
Refactor(executors):calculate f1 score of rank evaluators

pgiank28:


JoanFM:
Hello @pgiank28 , thank you for your contribution.

 We are trying to move these new implementations of evaluators into out jina-hub. Would you mind moving these classes there? There are already some open PR related to evaluators that you can take as reference.

pgiank28:
Moving the implementation to jina-hub

JoanFM:
Thanks!

PR #1194, opened 2020-11-01
feat(executor): add classifier as new family of executor

hanxiao:
This PR introduces a new family of executor: `Classifier`, which can be used to generated ML/DL-based tags. This is considered as document enrichment; it can enable future features such as smart sub-flow routing.

## Scope

- Add `BaseClassifier` and corresponding drivers to support: binary, multi-class, multi-label, regression (considered as soft label) prediction
- Add testing

## Out of Scope
- Add any concrete classification algorithm


jina/drivers/predict.py
```diff
@@ -0,0 +1,163 @@
...
+        :param output_tag: output label will be written to ``doc.tags``
+        :param args:
+        :param kwargs:
+        """
+        super().__init__(*args, **kwargs)
+        self.output_tag = output_tag
+
+    def _apply_all(
+            self,
+            docs: Iterable['jina_pb2.Document'],
```
bwanglzu, line 34: i think `Sequence` is a better type hints since we expect the interable object to be ordered

bwanglzu (commented):
some other thoughts, maybe we can expose `self.labels` as a property,

jina/drivers/predict.py
hanxiao: Then this should be changed globally in a separate PR

hanxiao: raised #1196

PR #1209, opened 2020-11-02
docs: cleanup crafters docstring

cristianmtr:
cleanup docs on crafters

- crafters **craft**. circular definition
- remove dated `seealso`

### DIFF
--- a/jina/executors/crafters/__init__.py
+++ b/jina/executors/crafters/__init__.py
@@ -8,11 +8,10 @@ from .. import BaseExecutor
 
 
 class BaseCrafter(BaseExecutor):
-    """A :class:`BaseCrafter` craft the content of `Document` or `Chunk`. It can be used for preprocessing,
-    segmenting etc.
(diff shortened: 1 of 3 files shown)


jina/executors/crafters/__init__.py
```diff
@@ -8,11 +8,10 @@
 
 
 class BaseCrafter(BaseExecutor):
-    """A :class:`BaseCrafter` craft the content of `Document` or `Chunk`. It can be used for preprocessing,
-    segmenting etc.
+    """A :class:`BaseCrafter` encodes the content of `Document` or `Chunk`. It can be used for preprocessing,
```
JoanFM, line 11: `encodes` is not the right word, for that we use `encoders`. May we use `transforms`?

JoanFM (changes requested):

jina/executors/crafters/__init__.py
cristianmtr: Updated to `transforms`

JoanFM (approved):

PR #1210, opened 2020-11-03
feat(zmq): lazy access to request body

hanxiao:
## Rationale

(de)serialization & (de)compression only when necessary. 

For example, `_route, _pass, _merge` they don't read the request, therefore no (de)serialization & (de)compression is required. Normal `HeadPea` and `TailPea` equipped with non-executor drivers do no need access to requests. Hence, this feature can significantly improve the scalability especially when Flow contains multiple parallel Pods.

The blue path in the following depicts where the optimization kicks in:

image

## Changes

- Add `ProtoMessage`, `LazyRequest` interface, unifying all access to original `jina_pb2.Message` and `jina_pb2.Request`.  `LazyRequest` provides lazy (de)serialization & (de)compression, which is only triggered after the first R/W access on `Request` body. (This trigger is delegated to `Pea -> Driver -> Executor`)
- Refactor the low-level zmq, pea, gateway to use `ProtoMessage`; making sure they leverage the lazy-access feature and do not randomly & unnecessarily access the request body
- Refactor Protobuf schema to make `Envelope` self-contained.
- Add testing on `LazyRequest` & `ProtoMessage`, all existing Flow/Pea/Pod-involved tests help to validate this feature

#### New Features
- Add all compression methods that Python builtin: `zlib`, `gzip`, `bz2`, `lzma`

#### Breaking Changes  #1174 
- **_Nothing_ above zmq & protobuf layer, including `Driver` need to comply to the change**. This PR is at low-level ZMQ & protobuf network layer. `ProtoMessage` persist the same way as using `jina_pb2.Message`, 
- For low-level programming, please notice `ProtoMessage` should be now the interface of `jina_pb2.Message`, use the former not the latter.
- For low-level programming, please be fully aware of the lazy-access feature and leverage them as much as possible. Do not introduce unnecessary access.

## Out of Scope
- Full benchmark on before vs. after, different comparison methods and hwm/lwm configs.


jina/proto/message.py
```diff
@@ -0,0 +1,364 @@
...
+_trigger_req_fields = set(jina_pb2.Request.DESCRIPTOR.fields_by_name.keys()).difference(
+    {'train', 'index', 'search', 'control'})
+
+_trigger_fields = _trigger_req_fields.union(_trigger_body_fields)
+_empty_request = jina_pb2.Request()
+
+__all__ = ['ProtoMessage', 'ControlMessage']
+
+
+class LazyRequest:
```
JoanFM, line 27: can't we split this into another file? I think is better to keep smaller self-contained files

jina/proto/message.py
```diff
@@ -0,0 +1,364 @@
...
+        return self._deserialized is not None
+
+    def __getattr__(self, name: str):
+        # https://docs.python.org/3/reference/datamodel.html#object.__getattr__
+        if (name in _trigger_fields) or hasattr(_empty_request, name):
+            self._deserialized = self.as_pb_object()
+        if name in _trigger_body_fields:
+            req = getattr(self._deserialized, self._deserialized.WhichOneof('body'))
+            return getattr(req, name)
+        elif hasattr(_empty_request, name):
```
JoanFM, line 54: this branch will never be triggered. Since the same option is in an `or` statement in the first `if`

jina/proto/message.py
```diff
@@ -0,0 +1,364 @@
...
+            data = lzma.decompress(data)
+        elif ctag == CompressAlgo.ZLIB:
+            import zlib
+            data = zlib.decompress(data)
+        elif ctag == CompressAlgo.GZIP:
+            import gzip
+            data = gzip.decompress(data)
+        return data
+
+    def as_pb_object(self) -> 'jina_pb2.Request':
```
JoanFM, line 81: what if we make `_deserialized` a `cached_property`?

jina/proto/message.py
```diff
@@ -0,0 +1,364 @@
...
+
+    @property
+    def is_used(self) -> bool:
+        """Return True when request has been r/w at least once """
+        return self._deserialized is not None
+
+    def __getattr__(self, name: str):
+        # https://docs.python.org/3/reference/datamodel.html#object.__getattr__
+        if (name in _trigger_fields) or hasattr(_empty_request, name):
+            self._deserialized = self.as_pb_object()
```
JoanFM, line 50: it is weird to assign himself, I would use some kind of property for this?

jina/proto/message.py
```diff
@@ -0,0 +1,364 @@
...
+        :param num_part: the total parts of the message, 0 and 1 means single part
+        :param check_version: turn on check_version 
+        :return: the resulted protobuf message
+        """
+        envelope = jina_pb2.Envelope()
+        envelope.receiver_id = identity
+        if isinstance(self.request, jina_pb2.Request) or (request_id and request_type):
+            # not lazy request, so we can directly access its request_id without worrying about
+            # triggering the deserialization
+            envelope.request_id = request_id or self.request.request_id
```
JoanFM, line 194: being it a `private` function, and only used in one place, why is there the possibility that `request_id` is different than `self.request.request_id`?

jina/proto/message.py
```diff
@@ -0,0 +1,364 @@
...
+            envelope.request_type = request_type or \
```
JoanFM, line 195: same here?

jina/proto/message.py
hanxiao: i want to manually control the cache, not just an one-time thing.

jina/proto/message.py
hanxiao: remember protobuf do not init any default value and there is no way to force them equal. 
also note that, envelope do not exist in client side, on client side, only `request.request_id` is available

jina/proto/message.py
hanxiao: it is triggered

jina/proto/message.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def as_pb_object(self) -> 'jina_pb2.Request':
+        if self._deserialized is None:
+            r = jina_pb2.Request()
+            _buffer = self._decompress(self._buffer)
+            r.ParseFromString(_buffer)
+            self._deserialized = r
+
+            # # Though I can modify back the envelope, not sure if it is a good design:
+            # # My intuition is: if the content is changed dramatically, e.g. fron index to control request,
```
nan-wang, line 89:
```suggestion
            # # My intuition is: if the content is changed dramatically, e.g. from index to control request,
```

jina/proto/jina.proto
```diff
@@ -183,6 +183,25 @@ message Envelope {
 
     repeated uint32 num_part = 8; // the number of partial messages
 
+    string request_type = 9; // type of the request: TrainRequest, IndexRequest, SearchRequest, ControlRequest
+
+    bool check_version = 10; // check local Protobuf version on every Pod that this message flows to
```
nan-wang, line 188: What about putting this into the `Flow`. Adding this to the `Envelop` does not seem to be necessary.
```suggestion
    bool check_version = 10; // check local Protobuf version on every Pod that this message flows to
```

jina/proto/message.py
```diff
@@ -0,0 +1,370 @@
...
+            data = bz2.decompress(data)
+        elif ctag == CompressAlgo.LZMA:
+            import lzma
+            data = lzma.decompress(data)
+        elif ctag == CompressAlgo.ZLIB:
+            import zlib
+            data = zlib.decompress(data)
+        elif ctag == CompressAlgo.GZIP:
+            import gzip
+            data = gzip.decompress(data)
```
nan-wang, line 78: Put this also into `try...except...`

jina/proto/message.py
hanxiao: let me first pass all tests and replace it to `cached_property`

jina/proto/message.py
```diff
@@ -0,0 +1,370 @@
...
+            _size_after = sys.getsizeof(c_data)
+            _c_ratio = _size_after / _size_before
+
+            if _c_ratio < self.envelope.compression.low_watermark:
+                data = c_data
+            else:
+                # compression rate is too bad, dont bother
+                # save time on decompression
+                default_logger.debug(f'compression rate {(_size_after / _size_before * 100):.0f}% '
+                                     f'is lower than low_watermark '
```
nan-wang, line 278:
```suggestion
                                     f' >= low_watermark '
```

tests/unit/test_echostream.py
```diff
@@ -51,4 +53,12 @@ def test_flow_with_jump():
...
-        f.dry_run()
+        f.index(random_docs(10))
+
+
+def test_flow_with_parallel():
+    f = (Flow().add(name='r1', uses='_pass')
+         .add(name='r2', uses='_pass', parallel=3))
+
+    with f:
+        f.index(random_docs(100))
```
nan-wang, line 64: This is expected to be an integration test.

nan-wang (commented):
Minor comments

jina/proto/message.py
hanxiao: Raise on purpose. Note that, **unlike compression**, if decompress is errored, then there is **nothing** we can save.

jina/proto/message.py
hanxiao: it is lower

jina/proto/jina.proto
hanxiao: see `ProtoMessage` for the reason of this design

PR #1216, opened 2020-11-03
feat(evaluator): add average precision

kilsenp:
Adding Average Precision (AP) as an evaluator.
See tests for some sanity checks.

I considered adding this to jina-hub, but thought this can fit here as well as there is no other dependencies besides numpy.

The `BaseRankingEvaluator` class expects a eval_at argument which is not necessary for AP as it evaluates on the whole ranking.
It could be used to limit the AP to only consider the first x results though.

### DIFF
--- a/jina/executors/evaluators/rank/average_precision.py
+++ b/jina/executors/evaluators/rank/average_precision.py
@@ -1,0 +1,38 @@
+from typing import Sequence, Any
+from ..rank import BaseRankingEvaluator
+import numpy as np
+
+
+class AveragePrecisionEvaluator(BaseRankingEvaluator):
+    """A :class:`AveragePrecisionEvaluator` evaluates the Average Precision of the search.
+       https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Average_precision
+    """
+
+    @property
+    def metric(self):
+        return 'AveragePrecision'
+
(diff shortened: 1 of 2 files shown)


JoanFM:
Hey @kilsenp, it is true that it does not need extra dependencies but we would like to keep these executors in the hub nevertheless. Could you please submit it there?

jina/executors/evaluators/rank/average_precision.py
```diff
@@ -0,0 +1,38 @@
...
+        :return the evaluation metric value for the request document
+        """
+        matches = []
+        for idx, doc_id in enumerate(desired):
+            matches.append(doc_id in actual[:idx+1])
+        matches = np.array(matches)
+        cumsum = np.cumsum(matches)
+        precision = cumsum / np.arange(1, len(matches) + 1)
+
+        # Does not match behaviour in
```
JoanFM, line 28: What does this mean? Is there any bug in Precision Evaluaror?

JoanFM (changes requested):
Looks really good, but needs to be moved to jina-hub

kilsenp:
> Hey @kilsenp, it is true that it does not need extra dependencies but we would like to keep these executors in the hub nevertheless. Could you please submit it there?

Okay, sure will move tomorrow!

PR #1220, opened 2020-11-04
Style code commits in one workflow

maximilianwerk:
Consolidating all action that do a `git commit` to `jina-ai/jina` into one workflow in order to avoid conflicts like https://github.com/jina-ai/jina/runs/1313814830?check_suite_focus=true


.github/workflows/test-examples.yml
```diff
@@ -1,10 +1,13 @@
 name: Test Examples
 
-on:
-  push:
-    branches:
-      - master
+# on:
+#   push:
+#     branches:
+#       - master
```
cristianmtr, line 6: Should these be left uncommented, or why can't we just remove them?

maximilianwerk: Ouch that is a commit that leaked in from another branch. My bad, I used git wrongly. Good catch

hanxiao (dismissed):

.github/workflows/test-examples.yml
maximilianwerk: fixed this one.

JoanFM (approved):

hanxiao (approved):

PR #1222, opened 2020-11-05
refactor: remove replica naming

JoanFM:
**Changes introduced**
Changes `replica_id` to `pea_id` and `replica_workspace` to `pea_workspace`


maximilianwerk (approved):
LGTM. Like the Ranker test refactoring, even thought I am not sure, that it belongs here. Merge away if tests pass.

JoanFM:
> LGTM. Like the Ranker test refactoring, even thought I am not sure, that it belongs here. Merge away if tests pass.

The point I realised about it as a side effect, but yes should not be here

PR #1227, opened 2020-11-05
feat: add identifier to logger

JoanFM:
**Changes introduced**
Pass identifier to logger that should identify the pod so that `fluentd` can group the messages.

Also pass the pod_id if possible to the executor logger so that all the messages from `post_init` can be handled by `fluentd` and eventually the dashboard could receive them.

**TODO**
- Have identifier for `Flow`
- Pass proper identifiers from jinad.
- Move logger sse to `jinad` reading from the wanted `ids`


jina/executors/__init__.py
```diff
@@ -52,6 +52,8 @@ def __call__(cls, *args, **kwargs):
 
         obj = type.__call__(cls, *args, **kwargs)
 
+        obj.logger = JinaLogger(obj.__class__.__name__, id=m['pod_id'] if 'pod_id' in m else None)
```
hanxiao, line 55:
```suggestion
        obj.logger = JinaLogger(obj.__class__.__name__, id=m.get('pod_id', None))
```

jina/executors/crafters/__init__.py
```diff
@@ -17,8 +17,11 @@ class BaseCrafter(BaseExecutor):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.required_keys = {k for k in inspect.getfullargspec(self.craft).args if k != 'self'}
+        
+    def post_init(self):
         if not self.required_keys:
             self.logger.warning(f'{self.__class__} works on keys, but no keys are specified')
+        super(BaseCrafter, self).post_init()
```
hanxiao, line 24:
```suggestion
        super(BaseCrafter, self).post_init()
```
but why put it into `post_init`?

JoanFM: It is an undesired effect. If I want to enforce the identifier to be present from the metas when the logger is instantiated. I need to call it after the init and before the post init. 

My idea is that post init will potentially have business logic that will be good to have properly caught and handled by fluentd.

My first idea but did not see how was to inject someway eith the yaml constructor the pea logger thought

JoanFM: This particular warning is to keep the same messsge as before, so no impact from the outside

jina/logging/logger.py
```diff
@@ -82,7 +82,15 @@ class SysLogHandlerWrapper(logging.handlers.SysLogHandler):
 class JinaLogger:
     supported = {'FileHandler', 'StreamHandler', 'SysLogHandler', 'FluentHandler'}
 
-    def __init__(self, context: str, log_config: Optional[str] = None, **kwargs):
+    def __init__(self, context: str, log_config: Optional[str] = None, group_id: Optional[str] = None, **kwargs):
```
hanxiao, line 85: I don't see how this `group_id` is related to `id` in `m.get('pod_id', None)`

JoanFM: We can change the name, but the idea is to allow the fluentd to group the logs by an identifier (which could be pod_id or flow_id) provided by the jinad. This way once the dashboard requests specific logs to jinad it will stream by reading from these folders.

jina/executors/__init__.py
hanxiao: where is this `id` used in `JinaLogger`? I don't see any reference

JoanFM: Wrong argument name will fix

jina/executors/crafters/__init__.py
nan-wang: One reason we don't want to touch the `post_init` is that it is highly likely to be overridden without calling `super().post_init()`

hanxiao:
can we simply use `os.environ`, and refer it in `logging.yml`?

image

JoanFM:
> can we simply use `os.environ`, and refer it in `logging.yml`? Our YAML support environment expansion
> 
> image

I actually did not consider the possibility, but I think there would be a problem with `different` Pods living in the same environment

hanxiao:
> but I think there would be a problem with different Pods living in the same environment

They never do, as `os.environ`  is process-specific and `Pod` is living in process by default

JoanFM:
> > but I think there would be a problem with different Pods living in the same environment
> 
> They never do, as `os.environ` is process-specific and `Pod` is living in process by default

will consider, not sure if it may pose problems when distributed Peas in the same Pod, but will think about

JoanFM:
> > but I think there would be a problem with different Pods living in the same environment
> 
> They never do, as `os.environ` is process-specific and `Pod` is living in process by default

JoanFM:
> > but I think there would be a problem with different Pods living in the same environment
> 
> They never do, as `os.environ` is process-specific and `Pod` is living in process by default

If working locally, Pods do not start a new process, what does spin new processes are Peas.

JoanFM:
Closed to open #1232

Issue #1215, opened 2020-11-03
`num_part` behavior is obfuscated, making `gather_inspect` useless


hanxiao:
**Describe your problem**

in the very very early version (< 0.1 & even before),`num_part` has two meanings:

1. after this Pod, the message will be divided into number of parts to the sequel Flow
2. at this Pod, we are expecting number of parts message to be collected and then contiune.

Somehow during the evolution the second meaning is lost. The `ReduceDriver` now fully depends on `num_part` in the envelope. This makes the `gather_inspect` useless as it won't block all inspect nodes due to missing `num_part` in the `Envelope`.

```python
f = Flow(inspect='COLLECT').add().inspect().add().inspect().add().inspect().plot()
```

[...]

PR #1228, opened 2020-11-05
refactor(driver): move reduce logic to BaseDriver

hanxiao:
WIP, otherwise `ReduceDriver` is orthogonal to all other Drivers.
also fixing `wrap_func` in `register_class`

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -156,6 +156,7 @@ class BaseDriver(metaclass=DriverType):
 
     @property
     def expect_parts(self) -> int:
+        """The expected number of partial messages before trigger :meth:`__call__` """
         return self._expect_parts or self.msg.num_part
 
     def attach(self, pea: 'BasePea', *args, **kwargs) -> None:
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -2,14 +2,16 @@ __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
 import inspect
+from collections import defaultdict
 from functools import wraps
-from typing import Any, Dict, Callable, Tuple, Iterable, Iterator
-from google.protobuf.struct_pb2 import Struct
+from typing import Any, Dict, Callable, Tuple, Iterable, Iterator, List, Optional, Sequence
 
 import ruamel.yaml.constructor
+from google.protobuf.struct_pb2 import Struct
 
 from ..enums import OnErrorSkip
 from ..executors.compound import CompoundExecutor
+from ..executors.decorators import as_reduce_method, wrap_func
 from ..helper import yaml
 from ..proto import jina_pb2
 from ..proto.message import ProtoMessage, LazyRequest
@@ -86,10 +88,10 @@ class QuerySetReader:
         if getattr(self, 'queryset', None):
             for q in self.queryset:
                 if (
-                    not q.disabled
-                    and self.__class__.__name__ == q.name
-                    and q.priority > self._priority
-                    and key in q.parameters
+                        not q.disabled
+                        and self.__class__.__name__ == q.name
+                        and q.priority > self._priority
+                        and key in q.parameters
                 ):
                     ret = q.parameters[key]
                     return dict(ret) if isinstance(ret, Struct) else ret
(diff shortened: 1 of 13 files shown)


jina/drivers/reduce.py
```diff
@@ -113,42 +50,31 @@ def _apply_all(
...
-            doc_pointers[context_doc.id] = context_doc.evaluations
+        if context_doc.id not in self.doc_pointers:
+            self.doc_pointers[context_doc.id] = context_doc.evaluations
         else:
-            doc_pointers[context_doc.id].extend(context_doc.evaluations)
+            self.doc_pointers[context_doc.id].extend(context_doc.evaluations)
 
 
-class ConcatEmbedDriver(ReduceDriver):
+class ConcatEmbedDriver(ReduceAllDriver):
```
JoanFM, line 61: I think `ConcatEmbedDriver` should either disappear and changed by a `ConcatEncoder` which would be a `MultiModalEncoder` or at least separate it from the `ReduceAllDriver` (as u say the logic is orthogonal), one may want to concat the `embedding` of 2 chunks?

JoanFM: If moved to a `MultiModalEncoder` then it would need to be combined with a `uses_before` = reduce

hanxiao: make sense, but right now in the core I can only find `ConcatEncoder` in test: `tests/integration/multimodal/yaml/embedding-multimodal-concatenate.py`? what's the plan here?

JoanFM:
Is it needed? Since we added the option to use `uses_before` it is quite flexible and easy to use separatedly.  I think is nice to have only one class for testing, or add it as another `mixin`

jina/drivers/reduce.py
JoanFM: We added this just for testing, we could add as a real executor in the `hub`, it can be interesting.

JoanFM: Actually we were desigining `MultiModalDriver` the same way `ConcatEmbedDriver` is done, but then we realised we needed to separate these concerns.

hanxiao:
> Is it needed? Since we added the option to use `uses_before` it is quite flexible and easy to use separatedly. I think is nice to have only one class for testing, or add it as another `mixin` maybe?

the idea here is to allow all drivers to access `partial_reqs` and apply their own logic to all `partial_reqs`, providing a uniform interface to all drivers.

If you look at the `__apply_all__` and `traverse_` logic before, there are some redundancy and inconsistency to other drivers.

jina/drivers/reduce.py
hanxiao: ConcatEmbed is one level, group by `doc.id` only, `MultiModalDriver` group by `doc.id` and `doc.modality` if I understand correctly

JoanFM: yes, it has some slight differences right

jina/drivers/__init__.py
```diff
@@ -134,15 +137,27 @@ class BaseDriver(metaclass=DriverType):
 
     store_args_kwargs = False  #: set this to ``True`` to save ``args`` (in a list) and ``kwargs`` (in a map) in YAML config
 
-    def __init__(self, priority: int = 0, *args, **kwargs):
+    def __init__(self, priority: int = 0, expect_parts: Optional[int] = 1, *args, **kwargs):
```
JoanFM, line 140: Remember to add docstring for num parts

JoanFM: Now a user needs to know how many parts it expects? It seems worse than before when user did not have to worry

JoanFM (commented):
I think maybe my messages come from my lack of clarity on the difference between ReduceDriver and ReduceAllDriver, I will check in more detail

jina/drivers/__init__.py
hanxiao: no need, but this is for preparing #1215

jina/drivers/reduce.py
hanxiao: no, they are irrelevant. `ConcatEmbedDriver` is a `ReduceDriver` which works on multiple requests. Complete different task than `MultiModalDriver`

JoanFM: But it could be able to concatenate chunks coming from the same segmenter without the need of reducing.

JoanFM: Anyway no need to change now

PR #1230, opened 2020-11-06
style(jinad): fixing jinad code styling

hanxiao:


JoanFM (dismissed):

jina/peapods/jinad.py
```diff
@@ -82,112 +99,102 @@ def __init__(self,
...
-
-    def create(self, pea_args: Dict, pod_type: str = 'flow'):
-        return super().create(kind='pod', pea_args=pea_args, pod_type=pod_type)
-
-    def log(self, pod_id: uuid.UUID):
-        super().log(kind='pod', remote_id=pod_id)
-
-    def delete(self, pod_id: uuid.UUID):
-        return super().delete(kind='pod', remote_id=pod_id)
+    kind = 'pod'
```
deepankarm, line 200: I thought I was missing something. This is so elegant!

jina/peapods/remote.py
```diff
@@ -47,17 +47,24 @@ def spawn_remote(self, host: str, port: int, **kwargs) -> Optional[str]:
...
-    def loop_teardown(self):
+    def delete_remote(self):
         if hasattr(self, 'api') and self.api.is_alive and self.remote_id:
             self.api.delete(self.remote_id)
 
+    def loop_teardown(self):
+        self.delete_remote()
+
+    def close(self):
+        # close is `pass` here to avoid sending a terminate signal to the remote via zmq
```
JoanFM, line 65: Why should we avoid sending this signal?

hanxiao: not sure, @deepankarm

deepankarm: The remote Pea/Pod context is closed using `api.delete()`, which calls `.close()` on remote.  We're not relying on zmq to send the terminate signal

JoanFM: why the `self.delete_remote` is not called in `close`?

PR #1235, opened 2020-11-06
fix: adapt log info after lazy reading

JoanFM:
**Changes introduced**
Since #1210 the `LogInfoDriver` is broken and here I fix it and allow to extract the wanted fields of a message via `dunder_get`


JoanFM:
Tested locally !

hanxiao (approved):

PR #1239, opened 2020-11-07
test: refactor helloworld

bwanglzu:
take out `JinaTestCase` from hello world IT.


hanxiao:
somehow this PR gives a higher fail rate, flakier than before.

hanxiao:
close due to #1263

PR #1243, opened 2020-11-09
chore: add new banner

BingHo1013:
New banner for readme


hanxiao:
this single gif is 23MB, and too complicated.

PR #1245, opened 2020-11-09
docs: add queryset usage info

bhavsarpratik:


docs/chapters/flow/index.md
```diff
@@ -444,3 +444,68 @@ with:
 ```
 
 In this case, this construction lets the `FaissIndexer` use the `vectors` stored by the indexer named `wrapidx`. 
+
+
+### Using Queryset
```
JoanFM, line 449: I would expect a little explanation on how `QuerySetReader` mixin will be used to override params in the `driver` and use `top_k` as an example

JoanFM (changes requested):

docs/chapters/flow/index.md
bhavsarpratik: @JoanFM Can you please push the changes you feel correct

JoanFM: Sure I will do a PR to ur branch

hanxiao (approved):

PR #1247, opened 2020-11-09
refactor(num_part): rewrite num_part logic at pea, driver and proto

hanxiao:
## Background

This PR is motivated by #1215 

To understand the problem, the following Flow is not runnable at the current master.

```python
f = (Flow().add(name='p1').add(name='p2', needs='gateway')
         .add(name='p3', needs='gateway')
         .add(name='p4', needs='gateway')
         .add(name='p5', needs='gateway')
         .needs(['p2', 'p4'], name='r1')
         .needs(['p3', 'p5'], name='r2')
         .needs(['p1', 'r1'], name='r3')
         .needs(['r2', 'r3'], name='r4')).plot()
```

The reason:
- in the master, `num_part` is purely determined by the upstream. It works like a FILO queue. The downstream Pods has no control this `num_part` in `Flow.needs()` but following what is written in the envelope.
- when the flow has hanging pods (introduced in #1087 ), collecting pods at downstream requires flexibile control of `num_part`.

## Changes

The change is about letting downstream to decide how to collect. This affects the logic of `ReduceDriver`, `BasePea`, `Protobuf`, `Flow.needs`. **No breaking changes at the Flow-level user experience.**

- Change the semantic of `num_part` to `the number of messages expected from upstream` as written in the `parser.py`
- Remove `num_part` from envelope & FILO logic, not required anymore.
- Move `num_part` gather logic to `BasePea` so that `ReduceDriver` simplified to `_pass`. And all `Drivers` bind to that executor now have access to `partial_reqs`.
- `_merge` is now `_pass`. All basic routing such as prev `_merge`, `_route`, `_pass` are simply `_pass`.

### DIFF
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -173,7 +173,16 @@ class BaseDriver(metaclass=DriverType):
     @property
     def partial_reqs(self) -> Sequence['LazyRequest']:
         """The collected partial requests under the current ``request_id`` """
-        return self.pea.partial_requests
+        if self.expect_parts > 1:
+            return self.pea.partial_requests
+        else:
+            raise ValueError(f'trying to access all partial requests, '
+                             f'but {self.pea} has only one message')
+
+    @property
+    def expect_parts(self) -> int:
+        """The expected number of partial messages """
+        return self.pea.expect_parts
 
     @property
     def msg(self) -> 'ProtoMessage':
@@ -267,7 +276,7 @@ class BaseRecursiveDriver(BaseDriver):
 
     @property
     def docs(self):
-        if len(self.partial_reqs) > 1:
+        if self.expect_parts > 1:
             return (d for r in reversed(self.partial_reqs) for d in r.docs)
         else:
             return self.req.docs
--- a/jina/drivers/__init__.py
+++ b/jina/drivers/__init__.py
@@ -2,7 +2,6 @@ __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
 import inspect
-from collections import defaultdict
 from functools import wraps
 from typing import (
     Any,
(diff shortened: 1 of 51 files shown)


jina/drivers/__init__.py
```diff
@@ -187,13 +173,16 @@ def req(self) -> 'LazyRequest':
     @property
     def partial_reqs(self) -> Sequence['LazyRequest']:
         """The collected partial requests under the current ``request_id`` """
-        if self.expect_parts <= 1:
-            raise ValueError(
-                f'trying to get collected requests even though expect_parts={self.expect_parts},'
-                f'maybe you want to use self.req instead?'
-            )
+        if self.expect_parts > 1:
```
JoanFM, line 176: is it always there this `attribute`? It is not set in `__init__`?

jina/flow/builder.py
```diff
@@ -83,14 +91,21 @@ def _build_two_connections(flow: 'Flow', start_node_name: str, end_node_name: st
         elif end_node_name == 'gateway':
             first_socket_type = SocketType.PUSH_BIND
         FlowPod.connect(start_node, end_node, first_socket_type=first_socket_type)
-        flow.logger.debug(f'Connect {start_node_name} ({start_node.tail_args.num_part}) '
-                          f'with {end_node_name} {str(end_node.role)} {end_node.head_args.num_part}')
+        # flow.logger.debug(f'Connect {start_node_name} ({start_node.tail_args.num_part}) '
```
JoanFM, line 94: why?

jina/drivers/__init__.py
hanxiao: no need to set
```
    @property
    def expect_parts(self) -> int:
        """The expected number of partial messages before trigger :meth:`handle` """
        return self.args.num_part if self.message.is_data_request else 1

```

PR #1260, opened 2020-11-11
docs: add note on 'score' field

### LINKED ISSUE #1255: clarify meaning of 'score'
**Describe the bug**

From a discussion with @JoanFM , it seems like there is a bit of confusion around the `score` field.

- In the context of chunk comparisons (at the Indexer level) it means `distance`. Thus, **smaller is better**
- In the context of documents (as parents of `chunks`) it means `similarity`. Or at least, in the context of `MinRanker` (https://github.com/jina-ai/jina-hub/blob/f92276c1c4a38ef9ac34a16c4065b5d360805600/rankers/MinRanker/__init__.py#L12), **larger is better**. 

Thus there is nothing wrong with the output of the below example (from multires lyrics example):

image

However, this is **confusing**. The problem, to me, is a matter of documentation / conflicting definitions for the `score` field. 

As originally discussed in https://github.com/jina-ai/jina/issues/1224

**Describe how you solve it**

Possible solutions:

- documentation 
- changing the name it by document / chunk. Document would be "similarity" (?), while chunk should be "distance". AFAIK, protobuf represents them with the same object, so that wouldn't work
- adding another field to the list of `match` and list of `chunk` that represents the method use to generate the `score` field. 

Something like 

```python
 "method": "MinRanker"  # or "Indexer
```

in the output

image

**Environment**

```
jina                          0.7.8
jina-proto                    0.0.75
jina-vcs-tag                  (unset)
libzmq                        4.3.2
pyzmq                         1.19.4
protobuf                      3.13.0
proto-backend                 cpp
grpcio                        1.33.2
ruamel.yaml                   0.16.12
python                        3.8.5
platform                      Linux
platform-release              5.4.0-52-generic
platform-version              #57-Ubuntu SMP Thu Oct 15 10:57:00 UTC 2020
architecture                  x86_64
processor                     x86_64
jina-resources                /home/cristian/code/jina/jina/resources
JINA_ARRAY_QUANT              (unset)
JINA_BINARY_DELIMITER         (unset)
JINA_CONTRIB_MODULE           (unset)
JINA_CONTRIB_MODULE_IS_LOADING(unset)
JINA_CONTROL_PORT             (unset)
JINA_DB_COLLECTION            (unset)
JINA_DB_HOSTNAME              (unset)
JINA_DB_NAME                  (unset)
JINA_DB_PASSWORD              (unset)
JINA_DB_USERNAME              (unset)
JINA_DEFAULT_HOST             (unset)
JINA_DISABLE_UVLOOP           (unset)
JINA_EXECUTOR_WORKDIR         (unset)
JINA_FULL_CLI                 (unset)
JINA_IPC_SOCK_TMP             (unset)
JINA_LOG_CONFIG               (unset)
JINA_LOG_NO_COLOR             (unset)
JINA_POD_NAME                 (unset)
JINA_PROFILING                (unset)
JINA_RANDOM_PORTS             (unset)
JINA_SOCKET_HWM               (unset)
JINA_TEST_GPU                 (unset)
JINA_TEST_PRETRAINED          (unset)
JINA_VCS_VERSION              (unset)
JINA_WARN_UNNAMED             (unset)
```

**Screenshots**

cristianmtr:
Closes #1255 

Short overview of the "score" field based on those discussions and a look through the code

### DIFF
--- a/docs/chapters/flow/index.md
+++ b/docs/chapters/flow/index.md
@@ -530,3 +530,25 @@ Note that more than one `queryset` can be passed with any request.
     with Flow().load_config('flow.yml') as search_flow:
         search_flow.search(input_fn=docs, output_fn=print_results, queryset=[top_k_queryset])
 ```
+
+## The Score Field
+
+In ranking document matches, Jina uses several algorithms to compute a `score` field. This field can mean either *distance* (with "smaller is better") or *similarity* (with "higher is better"). As, an example, the `NumpyIndexer` uses k-NN (with various distance metrics) to calculate a *distance*. On the other hand, the `MinRanker` uses a the function *1/(1+s)* (where *s* is the min. score from all `chunks`) in order to "bubble up" the score from child `chunks` to the parent, thus "larger is better".
+
+This can pe problematic when deciding when sorting the vector of results. Fortunately, Jina provides the name of the operator that has performed the scoring in the field `"score"`:
+
+```
+/**
+ * Represents the relevance model to `ref_id`
+ */
+message NamedScore {
+    float value = 1; // value
+    string op_name = 2; // the name of the operator/score function
+    ...
+}
+```
+
+In the above we can see `"op_name"`. As the comment suggests, this is the name of the operator that has performed the sorting. Based on this, the `"value"` field is either a *distance* ("smaller is better"), or a *similarity* ("higher is better"). 
+
+**Conclusion**: When sorting results, make sure you check the `document.score.op_name` in order to understand the direction of the score.
+
--- a/docs/chapters/flow/index.md
+++ b/docs/chapters/flow/index.md
@@ -533,9 +533,11 @@ Note that more than one `queryset` can be passed with any request.
 
 ## The Score Field
 
-In ranking document matches, Jina uses several algorithms to compute a `score` field. This field can mean either *distance* (with "smaller is better") or *similarity* (with "higher is better"). As, an example, the `NumpyIndexer` uses k-NN (with various distance metrics) to calculate a *distance*. On the other hand, the `MinRanker` uses a the function *1/(1+s)* (where *s* is the min. score from all `chunks`) in order to "bubble up" the score from child `chunks` to the parent, thus "larger is better".
+In ranking document matches, Jina uses several algorithms to compute the *relevance* of a document given a specific query. This is stored as a numeric value in the `score` field of a `match`, both at a `document` level and at the `chunk` level. This field can mean either "smaller is better" (*distance*) or "larger is better" (*similarity*, *relevance*).  
 
-This can pe problematic when deciding when sorting the vector of results. Fortunately, Jina provides the name of the operator that has performed the scoring in the field `"score"`:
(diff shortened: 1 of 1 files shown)


alexcg1:
LGTM. Any feedback from others?

docs/chapters/flow/index.md
```diff
@@ -529,4 +529,26 @@ Note that more than one `queryset` can be passed with any request.
 ```python
     with Flow().load_config('flow.yml') as search_flow:
         search_flow.search(input_fn=docs, output_fn=print_results, queryset=[top_k_queryset])
-```
\ No newline at end of file
+```
+
+## The Score Field
+
+In ranking document matches, Jina uses several algorithms to compute a `score` field. This field can mean either *distance* (with "smaller is better") or *similarity* (with "higher is better"). As, an example, the `NumpyIndexer` uses k-NN (with various distance metrics) to calculate a *distance*. On the other hand, the `MinRanker` uses a the function *1/(1+s)* (where *s* is the min. score from all `chunks`) in order to "bubble up" the score from child `chunks` to the parent, thus "larger is better".
```
JoanFM, line 536: It is not only `distance` or `similarity`, it can also mean `relevance`. It is just a minor detail, but needs to be included, because `score` can mean whatever is given

JoanFM: I would describe that a lot of rankers can describe the `score` as a measure of `relevance` of a `document` in relation to a given query, and then describe the `MinRanker` (although it will be deprecated soon to be changed by SimpleAggregateRanker`)

docs/chapters/flow/index.md
```diff
@@ -529,4 +529,26 @@ Note that more than one `queryset` can be passed with any request.
...
+ * Represents the relevance model to `ref_id`
+ */
+message NamedScore {
+    float value = 1; // value
+    string op_name = 2; // the name of the operator/score function
+    ...
+}
+```
+
+In the above we can see `"op_name"`. As the comment suggests, this is the name of the operator that has performed the sorting. Based on this, the `"value"` field is either a *distance* ("smaller is better"), or a *similarity* ("higher is better").
```
JoanFM, line 551: Again here, similarity is not the only option. Similarity or distance are measures mostly related to `vectors` or `objects` relation to another, but `relevance` can be something more arbitrary

JoanFM (changes requested):

JoanFM:
Just minor comments, good description though!!!

docs/chapters/flow/index.md
cristianmtr: But isn't **relevance** another type of **similarity** in the sense that they are both positive (i.e. "higher is better") ?

JoanFM: yes in the sense of `higher the better`, but it does not relate to "comparision" so directly

cristianmtr: Updated

docs/chapters/flow/index.md
cristianmtr: Updated

JoanFM (approved):
Great and clear explanation I would say!

PR #1263, opened 2020-11-11
test: helloworld without JinaTestCase and local directories

maximilianwerk:
Removed `JinaTestCase` and added tmpdir in tests.


JoanFM (approved):

bwanglzu (approved):
lgtm, somehow we are doing the same thing #1239

PR #1265, opened 2020-11-11
feat(client): add callback_on and continue_on_error

hanxiao:


tests/unit/clients/python/test_on_err.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.mark.parametrize('cb_on, x_type', [('DOCS', RepeatedCompositeContainer),
+                                           ('GROUNDTRUTHS', RepeatedCompositeContainer),
+                                           ('REQUEST', jina_pb2.Request),
+                                           ('BODY', jina_pb2.Request.IndexRequest),
+                                           ])
+def test_diff_field(cb_on, x_type):
+    def validate(x):
+        isinstance(x, x_type)
```
JoanFM, line 17: you need to add an assert? or the assert will make the index fail and should fail?

PR #1267, opened 2020-11-11
test: change incremental indexing

JoanFM:
This test fails and shows that something is broken in incremental indexing


JoanFM:
Fix #1268

JoanFM:
@cristianmtr if you want to take a look

jina/executors/indexers/__init__.py
```diff
@@ -59,6 +58,10 @@ def add(self, *args, **kwargs):
     def post_init(self):
         """query handler and write handler can not be serialized, thus they must be put into :func:`post_init`. """
         self.index_filename = self.index_filename or self.name
+        self.handler_mutex = True  #: only one handler at a time by default
+        self.is_handler_loaded = False
+
+    def reset(self):
         self.is_handler_loaded = False
```
hanxiao, line 65: aren't these two lines doing the same thing? so you say `is_handler_loaded` is saved to True in dump?

JoanFM: I did not check if it was saved to dump, but it seems to me that having it in `post_init` would make that more explicit.

reset in this case could be the same as `post_init` but I did not want to use `post_init` to avoid unwanted effects in other classes. This solution I do not like much, I would rather go to change the inspection of methods for `compound`. 

But would it be easy or a little overkill?

hanxiao: how about set it to `False` before serialization?

JoanFM: the reset is not used before serialization, the naming might be wrong. And what you describe is exactly what declaring in `post_init` achieves right?

hanxiao (approved):

PR #1273, opened 2020-11-12
refactor: remote break into mix-ins

JoanFM:
**Changes introduced**
@deepankarm just an idea on how we could break this ugly inheritance from BasePea in remote?


deepankarm:
These still need to be executed in a thread/process. We need to move away from BasePea but would require a separate `Runtime` class that these inherit/have. I think, it would be better to change these only after BasePea is refactored.

JoanFM:
> These still need to be executed in a thread/process. We need to move away from BasePea but would require a separate `Runtime` class that these inherit/have. I think, it would be better to change these only after BasePea is refactored.

Yes, yes, no intention to ever merge this

PR #1274, opened 2020-11-12
Ci update hub

cristianmtr:
* add new step to `tag` flow that
* clones jina hub
* attempts to bump version of jina to current
* opens PR to jina hub
* performs checks
* if checks pass commit with new version
* bumps minor version of hub module too

Need input on 
* how should we test the workflow from this PR? The workflow is only triggered on a `tag` or if it's part of master? It works locally using [https://github.com/nektos/act](https://github.com/nektos/act) @deepankarm

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -99,3 +99,24 @@ jobs:
           dockerfile: Dockerfiles/debianx.Dockerfile
           buildargs: BUILD_DATE, VCS_REF, JINA_VERSION
           tags: "latest, ${{env.JINA_VERSION}}"
+
+  update-jina-hub:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          token: ${{ secrets.JINA_DEV_BOT }}
+      - run: |
+          apt-get update -y
+          apt-get install python3.7 git python3-distutils -y
+          curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+          python3.7 get-pip.py
+          pip install gitpython ruamel.yaml PyGithub semver
+          git clone git@github.com:jina-ai/jina-hub.git --depth=1
+          output=$(python3.7 scripts/jina-hub-update.py)
+          output="${output//'%'/'%25'}"
+          output="${output//$'\n'/'%0A'}"
+          output="${output//$'\r'/'%0D'}"
+          echo "::set-output name=changelog::$output"
+        id: step1
--- a/scripts/jina-hub-update.py
+++ b/scripts/jina-hub-update.py
@@ -1,0 +1,97 @@
+""" Script to change versioning of files (eg. manifest.yml) for executors
+[encoders, crafters, indexers, rankers, evaluators, classifiers etc.].
+It also adds the required jina version.
+Commits the change in the branch and raises a PR for the executor.
+"""
+import glob
+import os
+
+import git
+import semver
+from github import Github
+from ruamel.yaml import YAML
+
+# TODO what env var do we export?
+g = Github(os.environ["GIT_TOKEN"])
+
+yaml = YAML()
+
+
+def main():
+    hub_repo = git.Repo('jina-hub')
+    hub_origin = hub_repo.remote(name='origin')
+    hub_origin_url = list(hub_origin.urls)[0]
+    assert 'jina-hub' in hub_origin_url, f'hub repo was not initialized correctly'
+    gh_hub_repo = g.get_repo('jina-ai/jina-hub')
+
+    jina_core_repo = git.Repo('.')
+    core_origin_url = list(jina_core_repo.remote(name='origin').urls)[0]
+    assert 'jina.git' in core_origin_url, f'core repo was not initialized correctly'
+
+    jina_core_version = jina_core_repo.tags[-1].tag.tag[1:]  # remove leading 'v'
+
+    print(f'cur. dir. is "{os.getcwd()}"')
+    print(f'got jina core v: "{jina_core_version}"')
+
+    modules = glob.glob(f'jina-hub/**/manifest.yml', recursive=True)
+    print(f'got {len(modules)} modules to update')
+    # traverse list of modules in jina-hub
+    for fpath in modules:
+        dname = fpath.split('/')[-2]
+        print(f'handling {dname}...')
+        with open(fpath) as fp:
+            info = yaml.load(fp)
+            # make sure the (possibly) existing version is older
+            if 'jina-version' in info.keys():
+                existing_jina_version = info['jina-version']
+                if semver.VersionInfo.parse(existing_jina_version) >= semver.VersionInfo.parse(jina_core_version):
+                    print(f'existing jina-core version for {dname} was greater or equal than version to update '
+                          f'({existing_jina_version} >= '
+                          f'{jina_core_version}). Skipping...')
+                    continue
+
+            old_ver = info['version']
+            new_ver = '.'.join(old_ver.split('.')[:-1] + [str(int(old_ver.split('.')[-1]) + 1)])
+            info['version'] = new_ver
+            print(f'bumped to {new_ver}')
+            info['jina-version'] = jina_core_version
+        with open(fpath, 'w') as fp:
+            yaml.dump(info, fp)
+
+        br_name = ''
+        try:
+            print('preparing the branch ...')
+            br_name = f'chore-{dname.lower()}-{new_ver.replace(".", "-")}-core-{jina_core_version.replace(".", "-")}'
+            new_branch = hub_repo.create_head(br_name)
+            new_branch.checkout()
+
+            print(f'bumping version to {new_ver} and committing to {new_branch}...')
+            hub_repo.git.add(update=True)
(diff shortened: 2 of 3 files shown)


scripts/jina-hub-update.py
```diff
@@ -0,0 +1,77 @@
...
+            hub_repo.git.add(update=True)
+            hub_repo.index.commit(f'chore: bump {dname} version to {new_ver}')
+            hub_repo.git.push("--set-upstream", hub_origin, hub_repo.head.ref)
+
+            # make PR using `gh`
+            print('making a PR ...')
+            title_string = f'bumping version for {dname} to {new_ver}'
+            body_string = f'bumping version from {old_ver} to {new_ver}'
+            pr_command = f'gh pr create -f'
+            subprocess.call(pr_command, shell=True)
```
JoanFM, line 67: the `git` client does not allow to do `PR`? It seems odd to combine `subprocess call` with `git` client?

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,77 @@
...
+from ruamel.yaml import YAML
+
+repo_dir = os.getcwd()
+yaml = YAML()
+
+version_file = 'manifest.yml'
+
+
+def main():
+    hub_repo = git.Repo('jina/hub')
```
JoanFM, line 21: the PR need to be made to `https://github.com/jina-ai/jina-hub/` and not to change directly the files under the submodule.

We will update the `jina-hub` repo and then jina will update its `hub` submodule folder but those are independent processes.

JoanFM (changes requested):

scripts/jina-hub-update.py
cristianmtr: The subprocess call is to call the `gh` CLI tool for GitHub. That's the one opening the PR

scripts/jina-hub-update.py
cristianmtr: Ah ok I see

scripts/jina-hub-update.py
JoanFM: I know, but the `python client` does not have this feature?

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,77 @@
...
+    assert 'jina.git' in core_origin_url, f'core repo was not initialized correctly'
+
+    jina_core_version = jina_core_repo.tags[-1].tag.tag[1:]  # remove leading 'v'
+    previous_branch = hub_repo.active_branch
+
+    print(f'cur. dir. is "{os.getcwd()}"')
+    print(f'got jina core v: "{jina_core_version}"')
+
+    # TODO remove limit set during dev.
+    for fpath in glob.glob(f'./jina/hub/**/{version_file}', recursive=True)[:1]:
```
JoanFM, line 37: so this recursion would need to be done  inside the cloned `jina-hub` repo I guess or whatever API u cna use

cristianmtr: Yep, updated and it seems to work. There are still some problems with conflicting branch names (if there is one that wasn't merged yet) and still need to figure out how to properly set up auth in GH actions

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,78 @@
+""" Script to change versioning of files (eg. manifest.yml) for executors
+[encoders, crafters, indexers, rankers].
```
JoanFM, line 2: also evaluators and classifiers

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,78 @@
...
+    assert 'jina.git' in core_origin_url, f'core repo was not initialized correctly'
+
+    jina_core_version = jina_core_repo.tags[-1].tag.tag[1:]  # remove leading 'v'
+    previous_branch = hub_repo.active_branch
+
+    print(f'cur. dir. is "{os.getcwd()}"')
+    print(f'got jina core v: "{jina_core_version}"')
+
+    # TODO remove limit set during dev.
+    os.chdir('jina-hub')
```
JoanFM, line 35: dangerous thing to do I think

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,78 @@
...
+    modules = glob.glob(f'**/manifest.yml', recursive=True)
+    print(f'got {len(modules)} modules to update')
+    for fpath in modules[:1]:
+        dname = fpath.split('/')[-2]
+        print(f'handling {dname}...')
+        with open(fpath) as fp:
```
JoanFM, line 41: not sure but I think u can just open once as 'w' and read it

scripts/jina-hub-update.py
cristianmtr: Dangerous to `chdir`? What do you mean?

scripts/jina-hub-update.py
cristianmtr: That one, no, but I did find [https://pygithub.readthedocs.io/en/latest/examples/PullRequest.html#create-a-new-pull-request](https://pygithub.readthedocs.io/en/latest/examples/PullRequest.html#create-a-new-pull-request) Will try

scripts/jina-hub-update.py
cristianmtr: got it

scripts/jina-hub-update.py
JoanFM: It is not a good practice to use `os.chdir` I think. It is confusing and is never clear why things happen afterwards. As much as possible I strongly suggest to avoid.

scripts/jina-hub-update.py
cristianmtr: Moved to using the lib mentioned

scripts/jina-hub-update.py
cristianmtr: I see. Removed

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,87 @@
...
+import os
+
+import git
+from github import Github
+from ruamel.yaml import YAML
+
+# TODO what env var do we export?
+g = Github(os.environ["GIT_TOKEN"])
+
+repo_dir = os.getcwd()
```
JoanFM, line 16: not used?

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,87 @@
...
+yaml = YAML()
+
+
+def main():
+    hub_repo = git.Repo('jina-hub')
```
JoanFM, line 21: what is this used for? how does git know where to get this `repo`?

cristianmtr: It's the folder name. It is pulled in the yml workflow

scripts/jina-hub-update.py
cristianmtr: Oh yeah, good catch

cristianmtr: For some reason PyCharm doesn't detect it as unused

scripts/jina-hub-update.py
cristianmtr: Nope, getting an error from yaml lib. Seems it's expecting the file to be in `r`?

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,96 @@
...
+            body_string = f'bumping version from {old_ver} to {new_ver}'
+
+            gh_hub_repo.create_pull(
+                title=title_string,
+                body=body_string,
+                head=br_name,
+                base='master'
+            )
+        except git.GitCommandError as e:
+            if 'tip of your current branch is behind' in str(e) \
```
JoanFM, line 84: this is quite fragile, isn't there a specific sub error for that case?

JoanFM: or I would just print the `repr(e)`

cristianmtr: There doesn't seem to be a specific sub error, unfortunately. But I did add the `repr(e)`

scripts/jina-hub-update.py
```diff
@@ -0,0 +1,96 @@
...
+                    or 'the remote contains work that you do' in str(e):
+                print(f'warning: Branch "{br_name}" already existed. Original exception: {repr(e)}. Skipping...')
```
JoanFM, line 86: I would remove the `if`, just print the `exception`.

JoanFM (changes requested):

hanxiao (dismissed):

hanxiao:
This looks good to go.

deepankarm (approved):
Minor comment

scripts/jina-hub-update.py
cristianmtr: Done

.github/workflows/tag.yml
```diff
@@ -99,3 +99,24 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          token: ${{ secrets.JINA_DEV_BOT }}
+      - run: |
+          apt-get update -y
+          apt-get install python3.7 git python3-distutils -y
+          curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+          python3.7 get-pip.py
```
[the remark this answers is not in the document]
cristianmtr: Didn't get to change that

PR #1275, opened 2020-11-12
feat(remote): add ssh remote peapod

hanxiao:
Reason for this PR:
- we can have multiple remote access methods
- having an SSH-based remote peapod requires no extra dependency, long-term: ssh peapod is a poorman solution for novice
- help us to design better RemotePea/Pod lifecycle

Not trying to overengineer anything at the moment.

### DIFF
--- a/jina/peapods/ssh.py
+++ b/jina/peapods/ssh.py
@@ -1,0 +1,67 @@
+from subprocess import Popen, PIPE
+
+from .. import __ready_msg__, __stop_msg__
+from ..helper import get_non_defaults_args, kwargs2list
+from ..logging import JinaLogger
+from ..peapods.pea import BasePea
+
+
+class RemotePea(BasePea):
+    """Simple SSH based RemotePea for remote Pea management
+
+    .. note::
+        It requires one to upload host public key to the remote
+        1. ssh-keygen -b 4096
+        2. scp ~/.ssh/id_rsa.pub username@hostname:~/.ssh/authorized_keys
+    """
+
+    @property
+    def remote_command(self) -> str:
+        from ..parser import set_pea_parser
+        non_defaults = get_non_defaults_args(self.args, set_pea_parser(), taboo={'host'})
+        _args = kwargs2list(non_defaults)
+        return f'jina pea {" ".join(_args)}'
+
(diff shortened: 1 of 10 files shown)


jina/peapods/ssh.py
```diff
@@ -0,0 +1,67 @@
+from subprocess import Popen, PIPE
+
+from .. import __ready_msg__, __stop_msg__
+from ..helper import get_non_defaults_args, kwargs2list
+from ..logging import JinaLogger
+from ..peapods.pea import BasePea
+
+
+class RemotePea(BasePea):
```
JoanFM, line 9: Isnt this name already used in another file? Can we have RemoteSSHPea?

jina/peapods/ssh.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @property
+    def remote_command(self) -> str:
+        from ..parser import set_pod_parser
+        non_defaults = get_non_defaults_args(self.args, set_pod_parser(), taboo={'host'})
+        _args = kwargs2list(non_defaults)
+        return f'jina pod {" ".join(_args)}'
+
+
+class RemoteMutablePod(RemotePod):
```
JoanFM, line 60: Same here. It can lead to confusion?

JoanFM:
I think that this PR at this moment when we are stabilizing our distribution vision (and when potentially we may rearrange the Pods and Peas) adds not so much value and can add maintenance cost and may limit some redesign.

hanxiao:
> and when potentially we may rearrange the Pods and Peas

as the refactoring work has not started yet, it does not block any existing work. Having SSH peapod helps us to design a better refactoring on RemotePea/Pod.

jina/peapods/ssh.py
hanxiao: they are in different module

JoanFM: Yes but leads to confusion

hanxiao:
One thing I'd like to highlight about RemotePod is that we don't need to manually implement close/kill method on the remote. Simply send `terminate` signal via control port, this is exactly why it is inherited from `BasePea`.

PR #1283, opened 2020-11-13
refactor: introduce jina primitive data type

hanxiao:
WIP!

## Changes

- New top-level module: `jina.types`
- Refactor `jina/__init__.py` to make alias to the primitive type, Flow and executors
- All protobuf message definition now end with `Proto` to differ from Python classes

## Impact

Mostly drivers

## Out-of-scope

- Add `Document` as new data type
- Refactoring all drivers so that they work with primitive data type instead of protobuf directly


JoanFM:
after this change the goal is to slowly keep `proto` hidden from the drivers so that they only interact with classes under `types` right?

PR #1289, opened 2020-11-16
feat(types): Document as new primitive type

hanxiao:
This PR is on the low-level, it does not change how drivers (and all layers above) behave.

## Content

- Introduced new primitive type `Document` and `QueryLang`
- Completed their tests and docstring
- Refactored `client/` generation part using the new `Document` type
- Interconnect five primitive types `Document`, `QueryLang`, `Message`, `Request`, `NdArray` 

## Out of scope

No driver is using the new `Document` & `QueryLang` primitive types at the moment, they still working directly with Protobuf.

- [ ] refactoring all Drivers to use all Jina primitive types

## Todo (@jina-ai/engineering )
- [ ] use the new `Document` type in future tests


jina/types/document/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+            NdArray(getattr(self._document, k)).value = v
+        elif isinstance(v, NdArray):
+            NdArray(getattr(self._document, k)).is_sparse = v.is_sparse
+            NdArray(getattr(self._document, k)).value = v.value
+        else:
+            raise TypeError(f'{k} is in unsupported type {typename(v)}')
+
+    @property
+    def matches(self) -> Iterator['Document']:
+        for d in self._document.matches:
```
JoanFM, line 112: may it be more performant to avoid having a `Generator` here? After all these are already in memory right?

jina/types/document/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+            self._document.uri = value
+            self._document.mime_type = guess_mime(value)
+        else:
+            raise ValueError(f'{value} is not a valid URI')
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.update_id()
```
JoanFM, line 194: I think this is a little dangerous, people may uncounciously update the `id` of a document which is a very sensitive thing to do. I would leave it as something to be done explicitly.

hanxiao: will expose an interface. this is nothing near final version. I will adapt all interfaces according to our real usage in the code.

jina/types/document/__init__.py
hanxiao: nothing final, will adapt all interfaces accordingly during refactoring driver and requests. not there yet

alexcg1:
Will this be a breaking change?

nan-wang:
> Will this be a breaking change?

Nope, this only makes life easier 😎

PR #1294, opened 2020-11-17
fix: hub update token

cristianmtr:


.github/workflows/tag.yml
```diff
@@ -110,8 +110,8 @@ jobs:
       - run: |
           sudo apt-get update -y
           sudo apt-get install python3.7 git python3-distutils python3-pip -y
-          pip install gitpython ruamel.yaml PyGithub semver
-          git clone git@github.com:jina-ai/jina-hub.git --depth=1
+          pip3 install gitpython ruamel.yaml PyGithub semver
+          git clone --depth=1 https://$DEV_BOT_TOKEN@github.com/jina-ai/jina-hub.git
           output=$(python3.7 scripts/jina-hub-update.py)
           output="${output//'%'/'%25'}"
```
deepankarm, line 116: are we removing the url-parsing here? please check [this](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote)

cristianmtr: I don't think GH tokens have those characters. At least mine doesn't. And from what I've read, you're supposed to use it like this, right?

deepankarm (approved):
Looks good

cristianmtr:
Hm for some reason I cannot merge. 

It says

> Merging can be performed automatically with 1 approving review.

even though there is one now

@hanxiao can you have a look as well?

hanxiao (approved):

PR #1297, opened 2020-11-17
refactor: remote default should be jinad

JoanFM:
I think the remote default should be jinad.

SSHRemote does not handle proper closure of `pod` and executors, just finished the `ssh` connection ?


hanxiao (approved):

hanxiao:
please make the usage visiable from the core docs/readme/tutorials. Otherwise this change makes no difference at all.

PR #1300, opened 2020-11-17
fix: adapt to new remote log endpoint

JoanFM:
To work when https://github.com/jina-ai/jinad/pull/29 merged


hanxiao (dismissed):

hanxiao:
btw, https://github.com/jina-ai/jinad/issues/20 can this be solved real quick?

JoanFM:
@deepankarm can we have a fallback to an empty log iterator when `websockets` is not installed? Otherwise this is adding an extra main dependency

jina/peapods/jinad.py
```diff
@@ -159,26 +164,51 @@ def create(self, args: Dict, pod_type: str = 'flow', **kwargs) -> Optional[str]:
         except requests.exceptions.RequestException as ex:
             self.logger.error(f'couldn\'t create {pod_type} with remote jinad {repr(ex)}')
 
+    async def wslogs(self, remote_id: 'str', current_line: int = 0):
+        with ImportExtensions(required=True):
```
JoanFM, line 168: what happens if this is not installed?

JoanFM:
> @deepankarm can we have a fallback to an empty log iterator when `websockets` is not installed? Otherwise this is adding an extra main dependency

Ah I see, not so important since it is under http tag, which is already needed to work remotely I understand, although I still think in the future we should allow to bypass the remote logging

jina/peapods/remote.py
```diff
@@ -58,6 +60,12 @@ def loop_body(self):
         else:
             self.logger.error(f'fail to create {typename(self)} remotely')
 
+    def send_terminate_signal(self) -> None:
+        """Gracefully close this pea and release all resources """
+        if self.is_ready_event.is_set() and hasattr(self, 'ctrl_addr'):
+            send_ctrl_message(address=self.ctrl_addr, cmd='TERMINATE',
```
JoanFM, line 66: isn't this command an enum?

deepankarm: https://github.com/jina-ai/jina/blob/remote-log-endpoint/jina/peapods/pea.py#L407

JoanFM: did it change recently?

jina/peapods/remote.py
```diff
@@ -85,13 +95,7 @@ def __init__(self, args: Union['Namespace', Dict]):
...
 
     def spawn_remote(self, host: str, port: int, pod_type: str = 'cli', **kwargs) -> Optional[str]:
-        return super().spawn_remote(host, port, pod_type=pod_type)
-
-    def send_terminate_signal(self) -> None:
-        """Gracefully close this pea and release all resources """
-        if self.is_ready_event.is_set() and hasattr(self, 'ctrl_addr'):
-            send_ctrl_message(self.ctrl_addr, jina_pb2.RequestProto.ControlRequestProto.TERMINATE,
-                              timeout=self.args['peas'][0].timeout_ctrl)
+        return super().spawn_remote(host=host, port=port, pod_type=pod_type)
```
JoanFM, line 98: why do we need this, just to override the default `pod_type`?

jina/peapods/remote.py
JoanFM: ah ok I guess is the new type redesign

jina/peapods/remote.py
deepankarm: for now, yes. we've separated into 2 classes, as `PeaAPI` and `PodAPI` might have different endpoints.

deepankarm (approved):

Issue #1304, opened 2020-11-17
test_helloworld.py::test_helloworld is misleading 


bio-howard:
**Describe your problem**
I think that test
`def test_helloworld(tmpdir):
    subprocess.check_call(['jina', 'hello-world', '--workdir', str(tmpdir)])`
https://github.com/jina-ai/jina/blob/master/tests/integration/test_helloworld.py
Doesn't "really" test if we are calling hello-world. The result tells if jina hello-world was called and any option which accepts --workdir parameter was executed properly. It makes a risk that other script will be executed, it won't be hello-world and it will still pass the test.

**What is your guess?**
Test which is executing this command is a must. I would suggest to change name or add more meaningful description and add  new test which checks if it's "really" hello world script being executed.

---

**Environment**

n/a
**Screenshots**

n/a

bio-howard:
I have solution for this problem. Should I commit to this task or create new issue?

nan-wang:
hi, @bio-howard Thanks for your issue. You can make a PR and link the PR with this issue. You might want to check out the  [Github official doc](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords).

bio-howard:
It's here https://github.com/jina-ai/jina/pull/1305

PR #1305, opened 2020-11-17
tests(integration): improve test_hello_world

bio-howard:


nan-wang:
Bad commit message. Instead of `tests(...): ...`, `test(...): ...` is preferred by lint. Please refer to:
- [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
- [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)

nan-wang (commented):
I didn't get the point of this test. The original `test_helloworld` is to check whether `jina hello-world` can be successfully executed. If the subprocess failed, a `CalledProcessError` will be thrown.

bio-howard:
@nan-wang as it's described here: https://github.com/jina-ai/jina/issues/1304

I think that test test_helloworld(tmpdir) doesn't "really" test if we are calling hello-world. The result tells if jina hello-world was called and any option which accepts --workdir parameter was executed properly. It makes a risk that other script will be executed, it won't be hello-world and it will still pass the test.
One cannot assume that is impossible to make a mistake in the script call.

Test which is added by me stops after it will recognize that hello world script was being executed. It's not a big time cost.

tests/integration/test_helloworld.py
```diff
@@ -13,10 +13,29 @@
 
 
 @pytest.mark.timeout(360)
-def test_helloworld(tmpdir):
+def test_helloworld_call(tmpdir):
```
cristianmtr, line 16: But then why keep this one as well?

tests/integration/test_helloworld.py
```diff
@@ -13,10 +13,29 @@
...
+        #  'cli = hello-world' is in stdoutput of hello-world script
+        #  this should be an indicator that hello-world script is executed
+        if 'cli = hello-world' in stdout_line:
+            is_hello_world = True
+            popen.terminate()
+            break
+    popen.stdout.close()
+    popen.wait()
+    assert is_hello_world, 'No cli = hello-world in stdoutput,' \
+                           'probably hello-world wasnt executed'
```
bwanglzu, line 36: Hi @bio-howard thanks for the test, I like it!
several points:
1. Can you rename `popen` to something like `p` or `proc`?
2. Can you use context manager, i.e. `with subprocess.Popen(..) as proc:` to ensure the proc being killed after running?
3. maybe rename `is_hellow_world` to `is_hello_world_in_stdout`?

bwanglzu (changes requested):

tests/integration/test_helloworld.py
bio-howard: I thought that it would be better to don't have two asserts in one test, so every test will test other functionality. Now, I don't see any benefit for such approach in this case. I will make one test.

tests/integration/test_helloworld.py
bio-howard: Thanks!
1. yes, proc is more meaningful
2. I will try to do it with context manager.
3. yes, it's more  accurate name

bhavsarpratik (approved):

bhavsarpratik:
recheckcla

bhavsarpratik:
@bio-howard Thanks for the contribution! To merge it you need to sign the CLA as stated above.

bwanglzu (approved):
LGTM

bio-howard:
I have read the CLA Document and I hereby sign the CLA.

bhavsarpratik:
recheckcla

Issue #1290, opened 2020-11-16
deadlinks on documentation


pswu11:
**Describe your problem**

I am working on README Learn section and found out that some tba links / dead links are still on below pages: 

https://docs.jina.ai/chapters/remote/run-remote-pod-flow.html
https://docs.jina.ai/chapters/remote/run-remote-pod-jina.html
https://docs.jina.ai/chapters/remote/run-remote-pod-console.html

For **Flow API**, maybe the link should be https://docs.jina.ai/chapters/flow/index.html, but for **understanding Peas & Pods** I didn't see the section with the exact name.

nan-wang:
Good catch. I've updated the Flow links. As for the __understanding Peas & Pods__, we still need one document for that. For the time being, I've redirected it to [https://docs.jina.ai/chapters/101/.sphinx.html#pea](https://docs.jina.ai/chapters/101/.sphinx.html#pea). Let's keep this ticket open as a reminder.

PR #1306, opened 2020-11-17
docs(remote): fix the dead links

nan-wang:
docs(remote): fix the dead links


nan-wang:
This partly resolved #1290

PR #1309, opened 2020-11-18
feat: random port ranges are configurable

maximilianwerk:
Making the random port range configurable via environment variables. I did not add tests, since this is heavily tested in a lot of integration tests and just testing the setting of environment variables is a bit shallow.

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -12,6 +12,7 @@ import time
 import urllib.parse
 import urllib.request
 import uuid
+import numpy as np
 from argparse import ArgumentParser, Namespace
 from datetime import datetime
 from io import StringIO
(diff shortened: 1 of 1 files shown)


theUnkownName (approved):
Nice! just out of curiosity, that range is arbitrary right? we could put whatever ports?

hanxiao:
> Nice! just out of curiosity, that range is arbitrary right? we could put whatever ports?

https://github.com/jina-ai/jina/pull/1145#discussion_r510731915

theUnkownName:
> > Nice! just out of curiosity, that range is arbitrary right? we could put whatever ports?
> 
> [#1145 (comment)](https://github.com/jina-ai/jina/pull/1145#discussion_r510731915)

Makes sense

maximilianwerk:
> > Nice! just out of curiosity, that range is arbitrary right? we could put whatever ports?
> 
> [#1145 (comment)](https://github.com/jina-ai/jina/pull/1145#discussion_r510731915)

It is a bit more subtle:

The default ephemeral port range from 49153 through 65535 is always used for Docker versions before 1.6.0. Which matches the port mappings that `ecs` uses by default when mapping random ports. See: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html

theUnkownName (approved):

JoanFM:
I would rather not use `os.environment`

PR #1312, opened 2020-11-18
fix: send terminate signal instead of delete jinad endpoint call

JoanFM:
**Changes introduced**

Avoid sending delete to jinad REST API. We should use terminate signal. Not sure this is the right place to do it and if it will work in every case.

@deepankarm let's have this draft to discuss this issue

**Problems**
Beside the obvious design problems (we will tackle some refactoring of Peas and Pods), there is a problem inside jinad that it will keep references of Pods and Peas remotely


JoanFM:
> **Changes introduced**
> 
> Avoid sending delete to jinad REST API. We should use terminate signal. Not sure this is the right place to do it and if it will work in every case.
> 
> @deepankarm let's have this draft to discuss this issue
I think the challenge is to properly set the `control_port` of the real remote Pod. or Pea.

On the other hand, how would then jinad delete them ?

hanxiao:
> I think the challenge is to properly set the control_port of the real remote Pod. or Pea.

why is this challenging?

> On the other hand, how would then jinad delete them ?

when it is terminated the process lifecycle ends, this is automatic. Does `jinad` bookkeep them so one has to `delete` it manually?

JoanFM:
> > I think the challenge is to properly set the control_port of the real remote Pod. or Pea.
> 
> why is this challenging?
> 
> > On the other hand, how would then jinad delete them ?
> 
> when it is terminated the process lifecycle ends, this is automatic. Does `jinad` bookkeep them so one has to `delete` it manually?

Sorry, challenging is the wrong word, It is just that I am not so certain about it.

Second, yes jinad keeps a reference to every Flow, Pod, or Pea to be later deleted

JoanFM:
We may need to have a logic in the `jinad` store to check if Pods and Peas are still alive. We may need to differentiate which ones need to be managed and deleted manually and which should not?

JoanFM:
Another point is that `sending_terminate_signal` does send the signal to the `Peas` but the `Pod` itself may still continue living

jina/peapods/remote.py
```diff
@@ -47,21 +47,20 @@ def spawn_remote(self, host: str, port: int, **kwargs) -> Optional[str]:
...
             self.logger.error(f'fail to create {typename(self)} remotely')
             self.is_shutdown.set()
 
     def delete_remote(self):
         if hasattr(self, 'api') and self.api.is_alive and self.remote_id:
             self.api.delete(self.remote_id)
-
-    def close(self):
-        self.delete_remote()
-        self.join()
```
JoanFM, line 63: I think this join would never work right? or it would work when the log will end?

jina/peapods/remote.py
```diff
@@ -58,9 +65,9 @@ def delete_remote(self):
         if hasattr(self, 'api') and self.api.is_alive and self.remote_id:
             self.api.delete(self.remote_id)
 
-    def close(self):
-        self.delete_remote()
-        self.join()
+    def close(self) -> None:
+        self.send_terminate_signal()
+        # TODO: I would rather wait for loop_body to finish but it seems to go on forever
```
rutujasurve94, line 70: Do we intend on waiting for timeout from Send terminate request for exit instead of waiting forever? We could adjust the timeout experimentally around the time taken by loop body to ideally exit.
We are injecting the timeout info in the first pea

JoanFM: the timeout is to sent the signal (it is a connection parameter), but right now we do not have any info of reply back knowing that something has been properly closed

jina/peapods/zmq.py
```diff
@@ -75,26 +75,29 @@ def resume_pollin(self):
         self.poller.register(self.in_sock)
 
     @staticmethod
-    def get_ctrl_address(args: 'argparse.Namespace') -> Tuple[str, bool]:
+    def get_ctrl_address(host: str, port_ctrl: str, ctrl_with_ipc: bool) -> Tuple[str, bool]:
         """Get the address of the control socket
 
-        :param args: the parsed arguments from the CLI
+        :param host: the host int he arguments
```
nan-wang, line 81:
```suggestion
        :param host: the host in the arguments
```

jina/peapods/remote.py
```diff
@@ -49,29 +53,52 @@ def loop_body(self):
...
             self.set_ready()
-            self.api.log(self.remote_id)
+            self.api.log(self.remote_id, self.is_shutdown)
         else:
             self.logger.error(f'fail to create {typename(self)} remotely')
-            self.is_shutdown.set()
 
     def delete_remote(self):
         if hasattr(self, 'api') and self.api.is_alive and self.remote_id:
             self.api.delete(self.remote_id)
```
nan-wang, line 62: After this fix, this function is deprecated.

nan-wang (changes requested):
minor changes

jina/peapods/zmq.py
JoanFM: Done! Thx

jina/peapods/remote.py
JoanFM: Done! Thx

nan-wang (approved):

PR #1313, opened 2020-11-19
refactor(driver): all drivers now use primitive types

hanxiao:
🔴 WIP, changing constantly.

Initiated from #1190, #1210, follow-up of #1283 #1284 #1289 we finally come to this step: changing all drivers to use our Jina Primitive Type!

Content:
- [x] Removing all reference of `jina_pb` modules from `drivers` module. Drivers (non-control ones) now only talk to the primitive types.
- [x] Refactor all drivers to use primitive types in the most efficient way
- [ ] Fix all broken tests


jina/types/document/__init__.py
```diff
@@ -307,7 +374,7 @@ def buffer(self) -> bytes:
     @buffer.setter
     def buffer(self, value: bytes):
         self._document.buffer = value
-        if self._document.buffer:
+        if value:
```
JoanFM, line 377: put comment here, but I'd rather have `mime_type` as separate `property` as well

PR #1317, opened 2020-11-20
test: skeleton new jinad integ test

JoanFM:
**Changes introduced**
I want to add the next small integration test with jinad that proves the capability to persist an index and query it when closing properly a Flow.

It also tests #1331


maximilianwerk:
I see now, that `test_integration.sh` is not really readable. I'll think about how to move the whole script to python in order to improve readability and re-usability.

JoanFM:
> I see now, that `test_integration.sh` is not really readable. I'll think about how to move the whole script to python in order to improve readability and re-usability. And we should better do this, before more and more similar script exist.

Then we will need to move these tests to the same level of integrstion to avoid pytest running them?

hanxiao (dismissed):

rutujasurve94 (approved):

hanxiao (approved):

PR #1319, opened 2020-11-20
test: help fix kv search driver test

JoanFM:
**Changes introduced**
- Adapt for **test_kv_search_driver.py**
- Adapt **test_segment_driver.py** and add weights property to document primitive type 
- Fix **test_rankingevaluation_driver.py** as per get back assert
- Fix **test_multimodal_driver.py**

### DIFF
--- a/jina/drivers/multimodal.py
+++ b/jina/drivers/multimodal.py
@@ -24,7 +24,7 @@ def _extract_modalities_from_document(doc: 'Document'):
         if modality in doc_content:
             return None
         else:
-            doc_content[modality] = chunk.embedding or chunk.content
+            doc_content[modality] = chunk.embedding if chunk.embedding is not None else chunk.content
     return doc_content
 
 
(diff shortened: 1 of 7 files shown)


hanxiao (approved):
awesome!

jina/drivers/multimodal.py
```diff
@@ -24,7 +24,7 @@ def _extract_modalities_from_document(doc: 'Document'):
         if modality in doc_content:
             return None
         else:
-            doc_content[modality] = chunk.embedding or chunk.content
+            doc_content[modality] = chunk.embedding if chunk.embedding is not None else chunk.content
```
hanxiao, line 27: the shortcut doesn't work because I didn't implement `None` interface for `embedding`?

hanxiao: or I see, ndarray doesn't like `or`

PR #1320, opened 2020-11-21
test: help refactor tests

JoanFM:
**Changes introduced**
Fix 2 more tests.

### DIFF
--- a/jina/drivers/search.py
+++ b/jina/drivers/search.py
@@ -128,7 +128,7 @@ class VectorSearchDriver(QuerySetReader, BaseSearchDriver):
 
             topk_embed = fill_fn(topks) if (self._fill_embedding and fill_fn) else [None] * len(topks)
             for match_hash, score, vec in zip(topks, scores, topk_embed):
-                r = doc.add_match(doc_id=uid.hash2id(match_hash),
+                r = doc.add_match(doc_id=match_hash,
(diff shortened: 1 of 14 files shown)


jina/drivers/search.py
```diff
@@ -127,9 +127,8 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
         for doc, topks, scores in zip(doc_pts, idx, dist):
 
             topk_embed = fill_fn(topks) if (self._fill_embedding and fill_fn) else [None] * len(topks)
-
             for match_hash, score, vec in zip(topks, scores, topk_embed):
-                r = doc.add_match(doc_id=match_hash,
+                r = doc.add_match(doc_id=uid.hash2id(match_hash),
```
hanxiao, line 131: i remember I add shortcut for hash-value `doc_id` in `add_match`, let me check

hanxiao: image
yes,  I think this is not needed

hanxiao, line 131:
```suggestion
                r = doc.add_match(doc_id=match_hash,
```

nan-wang (commented):
We need to rename the helper function after the `random_docs` is deprecated. `random_docs_new_api` does not sounds a good name.  🤓

PR #1322, opened 2020-11-21
test: help refactoring tests

JoanFM:


jina/types/document/__init__.py
```diff
@@ -318,6 +319,8 @@ def add_match(self, doc_id: Union[str, int], score_value: float, **kwargs) -> 'D
         :param kwargs: other key-value parameters written to the ``score`` object
         """
         r = self._document.matches.add()
+        if isinstance(doc_id, np.int64) or isinstance(doc_id, np.int32):
+            doc_id = doc_id.item()
         if isinstance(doc_id, int):
             r.id = uid.hash2id(doc_id)
```
hanxiao, line 325:
```suggestion
        if isinstance(doc_id, (int, np.integer)):
            r.id = uid.hash2id(int(doc_id))
```

hanxiao: first comment in https://stackoverflow.com/a/37727662

PR #1324, opened 2020-11-22
refactor: add set types to top-level import

hanxiao:


jina/resources/logging.default.yml
```diff
@@ -1,7 +1,7 @@
 handlers:  # enabled handlers, order does not matter
   - StreamHandler
   - SysLogHandler
-  - FluentHandler
+#  - FluentHandler TODO(Han): I temporally disable it for suppress the warning and better ux for newbies
```
deepankarm, line 4: `jinad` is now bundled with fluentd to privide the `log streaming` endpoint. how will this handler be enabled?

hanxiao: by setting different environment variable `JINA_LOG_CONFIG`
image

hanxiao: i revert the change and suppress the warning for this particular import now

PR #1335, opened 2020-11-23
feat: add multimodal document

bwanglzu:
Add `MultimodalDocument` to primitive types, and apply changes to `MultimodalDriver`.

Example usage of a `MultimodalDocument`?
```python
# With a list of documents (chunks)
md = MultimodalDocument(chunks=[chunk1, chunk2])
md = MultimodalDocument.from_chunks(chunks=[chunk1, chunk2])
# With modality content mapping (dict representation of modality and document content)
mapping = {'visual': 'visual content', 'textual': 'textual content'}
md = MultimodalDocument(modality_content_mapping=mapping)
md = MultimodalDocument.from_modality_content_mapping(modality_content_mapping=mapping)
# Exposed the modality content mapping as a property
md.modality_content_mapping
>>> {'visual': 'visual content', 'textual': 'textual content'}
# Exposed modalities (all modalities from chunk level) as a property
md.modalities
>>> ['visual', 'textual']
# Implemented a method to extract content from modality (used in driver)
md.extract_content_from_modality('visual')
>>> 'visual content'
```

I personally think `from_modality_content_mapping` should be discussed further (keep or not) since it brings a bit inconsistency for extracting `embedding` or `content`.


jina/types/document/multimodal.py
```diff
@@ -0,0 +1,39 @@
+from typing import List
+
+from . import Document
+from ...excepts import LengthMismatchException, BadDocType
+
+class MultimodalDocument(Document):
+    """Each :class:`MultimodalDocument` should have at least 2 chunks (represent as :class:`DocumentSet`)
+    and len(set(doc.chunks.modality)) == len(doc.chunk)
+    """
+    def __init__(self, document = None,
```
JoanFM, line 10: Maybe there should be the possibility to build it from N documents with different modalities and to merge them into one multimodal document?

Would that also be useful?

JoanFM: Another useful interface would be to extract the embedding or content by modality (or the chunk) given a modality name.

CHANGELOG.md
```diff
@@ -188,6 +188,14 @@ Jina is released on every Friday evening. The PyPi package and Docker Image will
 - [Release Note (`0.7.2`)](#release-note-072)
 - [Release Note (`0.7.3`)](#release-note-073)
 - [Release Note (`0.7.4`)](#release-note-074)
+- [Release Note (`0.7.5`)](#release-note-075)
```
JoanFM, line 191: remember to rebase on master

jina/drivers/multimodal.py
```diff
@@ -83,11 +68,11 @@ def _apply_all(
 
         valid_docs = []
         for doc in docs:
-            doc_content = _extract_modalities_from_document(doc)
```
JoanFM, line 86: Maybe we can adapt the interface to be `docs: MultiModalDocumentSet`, would that make sense?

jina/types/document/multimodal.py
```diff
@@ -0,0 +1,44 @@
...
+        self._modalities = set()
+        self._doc_content = {}
+        if len(self._document.chunks) < 2:
+            raise BadDocType(
+                f'A MultimodalDocument should have at least 2 modalities, {len(self.document.chunks)} received.')
+        if len(self._document.chunks) != len(self.modalities):
+            raise LengthMismatchException(
+                f'Document has {len(self._document.chunks)} chunks and {len(self.modalities)} modalities')
+
+    def extract_content(self) -> List[str]:
```
JoanFM, line 22: just to make sure it is understood, the original code is returning a `Dict` and not a `List` (maybe it is better a List) (Just raising awareness)

jina/drivers/multimodal.py
```diff
@@ -83,11 +68,11 @@ def _apply_all(
...
+            doc_content = doc.extract_content()
```
JoanFM, line 71: This naming is ambiguous, because it seems that it only extracts the content for a document (and this extracts content from chunks grouped by modality)

jina/types/document/multimodal.py
JoanFM: We will need that interface, otherwise we do not remove any heavylifting from the user when creating this, no?. Or we will need some kind of MultiModalDocumentBuilder?

We need to decide what experience we want to offer when building a document like this

JoanFM: I think we can offer a Builder interface or design this class as a Builder. And delegate the checks of correctness at the last step of the build step

jina/types/document/multimodal.py
bwanglzu: you're right, I updated the annotation

jina/drivers/multimodal.py
bwanglzu: my understanding is that `DocumentSet` should handle both `Document` and `MultimodalDocument`.

jina/types/document/multimodal.py
bwanglzu: I had some thought before, it should be something like an observer. The assurance of the correctness should happens at when we call `chunks.add` or `chunks.append` at `DocumentSet` or `ChunkSet`. But meanwhile I'm afraid it's a bit "over engineering". But it's good to have some discussion over that.

JoanFM: For now I would be happy just by adding thr interface to build directly from chunks. So that some boilerplatr can be added. What we would need to know is maybe about the granularities to be assigned and so on

jina/types/document/multimodal.py
```diff
@@ -0,0 +1,80 @@
...
+                parse a :class:`DocumentProto` from it; ``str`` is given, then consider
+                it as a JSON string and parse a :class:`DocumentProto` from it; finally,
+                one can also give `DocumentProto` directly, then depending on the ``copy``,
+                it builds a view or a copy from it.
+        :param copy: when ``document`` is given as a :class:`DocumentProto` object, build a
+                view (i.e. weak reference) from it or a deep copy from it.
+        :param kwargs: other parameters to be set
+        """
+        super().__init__(document=document, copy=copy, **kwargs)
+        self._modality_content_mapping = {}
```
nan-wang, line 39: Enable the user to create `MultimodalDocument` with the following codes will be awesome
```python
with Document() as c_text:
    c_text.text = 'jina'
    c_text.embedding = np.random.random(2)

with Document() as c_img:
    c_img.blob = np.random.random((28, 28))
    c_img.embedding = np.random.random(4)

d = MultimodalDocument(modality_content={'text': c_text, 'image': c_img})
```

JoanFM: yes something like this, and we sill need to take care of the `granularity fields`.

bwanglzu: @nan-wang @JoanFM I updated the PR with the following interface:
```python
c_text = Document()
c_text.text = 'jina'
c_text.embedding = np.random.random(2)
c_text.modality = 'textual'

c_visual = Document()
c_visual.embedding = np.random.random(2)
c_visual.modality = 'visual'
# Two ways to initialize
# from constructor
d = MultimodalDocument(chunks=[c_textual, c_visual])
# from classmethod
d = MultimodalDocument.from_chunks(chunks=[c_textual, c_visual])
```
The above suggestion poposed by @nan-wang  looks good, while user do not actually need to provide modality information like this: `modality_content={'text': c_text, 'image': c_img}` , since `modality` is a property of chunk, it should be attached to `chunk` directly.

JoanFM: I like it because it respects non-mutability of Documents as much as possible, but it could be handy to provide Nan's interface. But let's see what the community demmands regarding this. 

What do u think @nan-wang  it is true that if you offer that interface, u internally need to change those chunks. (What you could do is an interface of the type:

```python
d = MultiModalDocument({'modalitya': contentA, 'modalityB': contentB})
```
And inside it manages the creation of both the chunks and the multimodal document?

jina/types/document/multimodal.py
bwanglzu: the interface add as `from_chunks`. for check of the correctness, now I'm using an internal method called `_validate`, and chunks were validated at 2 places: 1. if we build `MultimodalDocument` using `from_chunks` or constructor, it will be validated inside the constructor. 2. Validate the chunks at `modality_content_mapping` inside `_build_modality_content_mapping`, since `modality_content_mapping` is the common entrance of other methods and properties. and we avoid overkill/change  `DocumentSet.add` and `ChunkSet.append`

jina/types/document/multimodal.py
bwanglzu: shall we provide a classmethod called `MultimodalDocument.from_modality_content_mapping` to allow user build the document from a dictionary of embedding/contents?

jina/types/document/multimodal.py
```diff
@@ -0,0 +1,96 @@
...
+                received a list of :class:`Document`, with different modalities.
+        :param copy: when ``document`` is given as a :class:`DocumentProto` object, build a
+                view (i.e. weak reference) from it or a deep copy from it.
+        :param kwargs: other parameters to be set
+        """
+        super().__init__(document=document, copy=copy, **kwargs)
+        self._modality_content_mapping = {}
+        if chunks:
+            self._validate(chunks)
+            self.chunks.clear()
```
JoanFM, line 44: Isn't it weird to clear something that has not been initialized?

jina/types/document/multimodal.py
```diff
@@ -0,0 +1,96 @@
...
+        """Get the mapping of modality and content, the mapping is represented as a :attr:`dict`, the keys
+        are the modalities of the chunks, the values are the corresponded content of the chunks.
+
+        :return: the mapping of modality and content extracted from chunks.
+        """
+        if not self._modality_content_mapping:
+            self._build_modality_content_mapping()
+        return self._modality_content_mapping
+
+    def extract_content_by_modality(self, modality: str) -> DocumentContentType:
```
JoanFM, line 73: I would rather use `from-modality` since `by` seems like we are grouping

JoanFM (changes requested):
Before merging I would like to have exposed an `interface`  that lets the user build a `MultiModalDocument` from different data like:

```python
d = MultiModalDocument({'modalitya': contentA, 'modalityB': contentB})
```

Also I think that this class should have better care of setting the right `granularity` parameters for the `MultiModalDocument` and the `chunks`. (Not so important but needed for the sake of coherence)

bwanglzu:
@JoanFM how about
```python
MultimodalDocument.from_content_modality_mapping({'visual': xxx, 'textual': xxx}) 
```
Where xxx could be content or embedding. this classmethod has the same naming convention as the property `MultimodalDocument.content_modality_mapping`

JoanFM:
> @JoanFM how about
> 
> ```python
> MultimodalDocument.from_content_modality_mapping({'visual': xxx, 'textual': xxx}) 
> ```
> 
> Where xxx could be content or embedding. this classmethod has the same naming convention as the property `MultimodalDocument.content_modality_mapping`

This is what I meant yes. The problem we have is that is not easy to tell if the content when is a numpy array if it is embedding or content. So we need a flag to say that.

bwanglzu:
> Also I think that this class should have better care of setting the right granularity parameters for the MultiModalDocument and the chunks. (Not so important but needed for the sake of coherence)

Agreed, I'll create a getter & setter in `:class:Document` to expose `granularity`, now we can only access the property from `:class:DocumentProto`

JoanFM:
> > @JoanFM how about
> > ```python
> > MultimodalDocument.from_content_modality_mapping({'visual': xxx, 'textual': xxx}) 
> > ```
> > 
> > 
> > Where xxx could be content or embedding. this classmethod has the same naming convention as the property `MultimodalDocument.content_modality_mapping`
> 
> This is what I meant yes. The problem we have is that is not easy to tell if the content when is a numpy array if it is embedding or content. So we need a flag to say that.

Well i thought it twice and no, I think is better to assume they arr created from content so u assume the chunks are filled by document. But make sure it is documented that if one chunk is created from embeddings they will need to create from the other interface

jina/types/document/multimodal.py
bwanglzu: Some fresh thoughts, I think `from_chunkset` would be a better interface compared with `from_chunks` since `ChunkSet` is a primitive data type.

nan-wang: Yes, `from_chunkset` sounds better

nan-wang (approved):
LGTM👍

jina/types/document/multimodal.py
JoanFM: chunk_set is not the right name though, because a `ChunkSet` is a set of `Documents` that are already have a parent. In this case we receive a `docset` and make it a `ChunkSet`.

jina/types/document/multimodal.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _build_modality_content_mapping(self) -> Dict:
+        for chunk in self.chunks:
+            modality = chunk.modality
+            self._modality_content_mapping[modality] = chunk.embedding \
+                if chunk.embedding is not None \
+                else chunk.content
+        self._validate(chunks=self.chunks)
+
+    def _validate(self, chunks: List[Document]):
```
JoanFM, line 55: just being a little picky, but we either make it static or we do not pass `self.chunks` and extract them inside the function right? I would vote for the second

JoanFM (commented):
Still missing the handling of granularity level, but looks really good!

jina/types/document/multimodal.py
bwanglzu: true, we expect `self` as the `reference_doc`, I'll stick with the implementation for now and to see if we can receive some feedback from users.

jina/types/document/multimodal.py
```diff
@@ -0,0 +1,108 @@
...
+        modalities = set([chunk.modality for chunk in chunks])
+        if len(chunks) < 2:
+            raise BadDocType('MultimodalDocument should consist at least 2 chunks.')
+        if len(modalities) != len(chunks):
+            raise LengthMismatchException(f'Length of modality is not identical to length of chunks.')
+
+    def _handle_chunk_attributes(self):
+        """Handle chunk attributes, such as :attr:`granularity` and :attr:`mime_type`.
```
JoanFM, line 64: What if the chunks come already with the proper granularity but the document has None?

JoanFM: We should at lesst add a warning in the constructor about how the granularity behaves. Would not change it but sure add docstring

bwanglzu: @JoanFM I did check it out, the document will have a default `granularity ` as 0.
image

If you take a look at our code logic in `DocumentSet` or `MatchSet`, you'll see something similar:
```python
    def append(self, document: Optional['Document'] = None, **kwargs) -> 'Document':
        """Add a sub-document (i.e chunk) to the current Document

        :return: the newly added sub-document in :class:`Document` view

        .. note::
            Comparing to :attr:`DocumentSet.append()`, this method adds more safeguard to
            make sure the added chunk is legit.
        """
        c = self._docs_proto.add()
        if document is not None:
            c.CopyFrom(document.as_pb_object)

        from ..document import Document
        with Document(c) as chunk:
            chunk.set_attrs(parent_id=self._ref_doc.id,
                            granularity=self._ref_doc.granularity + 1,
                            **kwargs)
            if not chunk.mime_type:
                chunk.mime_type = self._ref_doc.mime_type
            return chunk
```

JoanFM: Yes, my point is what happens if the user creates a MultiModalDocument with chunks of granularity 2. Then is not the chunks that mist have document.granularity + 1, but this new multimodaldocumet shluld have chunk.granularity - 1

JoanFM: What I suggest:
1. Validate all chunks have same granularity.
2. If chunk.granularity is 0. (This means a user without caring for granularity wants to merge N documents into a multimodal document, therefore we do what u have here of increasing their granularity inside this set) Well documented please
3. If the chunk comes with granularity > 0, then it means that someone has cared to chunk already the document or that we have some driver that generates muktimodal documents in the future. Then, have document.granularity = chunk.granularity - 1. 
3. Proper testing of this
Does it make sense? What we have here does not allow for deeper levels of multimodal documents.

jina/drivers/multimodal.py
```diff
@@ -83,11 +68,10 @@ def _apply_all(
 
         valid_docs = []
         for doc in docs:
-            doc_content = _extract_modalities_from_document(doc)
-            if doc_content:
+            if doc.modality_content_mapping:
```
JoanFM, line 71: should we change the `docs` type to `MultiModalDocumentSet`?

JoanFM (approved):

PR #1338, opened 2020-11-24
test: add integ test about non blocking gateway

JoanFM:
**Changes introduced**
Add test showing that when parallel is larger than 1, the gateway is non blocking and faster requests can be returned. Also shown that this does not happen when parallel is 1


nan-wang:
This test only send one request to gateway.

JoanFM:
> This test only send one request to gateway.

It sends 2 requests, see the `batch_size=1`

hanxiao (approved):

Issue #1229, opened 2020-11-06
crash when sharding without batching 


cristianmtr:
**Describe the bug**

When attempting to upgrade the multires lyrics examples to use sharding, I encountered a bug that produced the following error at query time:

```
Traceback (most recent call last):
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/peapods/pea.py", line 292, in msg_callback
    self.zmqlet.send_message(self._callback(msg))
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/peapods/pea.py", line 266, in _callback
    self.pre_hook(msg).handle(msg).post_hook(msg)
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/peapods/pea.py", line 160, in handle
    self.executor(self.request_type)
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/executors/__init__.py", line 570, in __call__
    d()
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/drivers/__init__.py", line 251, in __call__
    self._traverse_apply(self.req.docs, *args, **kwargs)
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/drivers/__init__.py", line 268, in _traverse_apply
    **kwargs,
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/drivers/__init__.py", line 281, in _traverse_rec
    doc.chunks, doc, 'chunks', path[1:], *args, **kwargs
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/drivers/__init__.py", line 284, in _traverse_rec
    self._apply_all(docs, parent_doc, parent_edge_type, *args, **kwargs)
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/drivers/search.py", line 124, in _apply_all
    idx, dist = self.exec_fn(embed_vecs, top_k=int(self.top_k))
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/executors/indexers/vector.py", line 279, in query
    dist = self._euclidean(_keys, self.query_handler)
  File "/home/cristian/envs/lyrics/lib/python3.7/site-packages/jina/executors/decorators.py", line 241, in arg_wrapper
    for b in batch_iterator(data[:total_size], b_size, split_over_axis, yield_slice=yield_slice):
[...]

PR #1340, opened 2020-11-24
fix: handle empty shards

cristianmtr:
Fixes https://github.com/jina-ai/jina/issues/1229

### DIFF
--- a/jina/executors/decorators.py
+++ b/jina/executors/decorators.py
@@ -244,7 +244,7 @@ def batching(func: Callable[[Any], np.ndarray] = None,
 
             b_size = (batch_size(data) if callable(batch_size) else batch_size) or getattr(args[0], 'batch_size', None)
             # no batching if b_size is None
-            if b_size is None:
+            if b_size is None or data is None:
                 return func(*args, **kwargs)
 
             default_logger.debug(
@@ -337,11 +337,12 @@ def batching_multi_input(func: Callable[[Any], np.ndarray] = None,
     def _batching(func):
         @wraps(func)
         def arg_wrapper(*args, **kwargs):
+            data = args[slice_on]
             # priority: decorator > class_attribute
             # by default data is in args[1:] (self needs to be taken into account)
             b_size = batch_size or getattr(args[0], 'batch_size', None)
             # no batching if b_size is None
-            if b_size is None:
+            if b_size is None or data is None:
                 return func(*args, **kwargs)
 
             args = list(args)
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -260,7 +260,7 @@ class NumpyIndexer(BaseNumpyIndexer):
 
         return idx, dist
 
-    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
+    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple[Optional['np.ndarray'], Optional['np.ndarray']]:
         """ Find the top-k vectors with smallest ``metric`` and return their ids in ascending order.
 
         :return: a tuple of two ndarray.
@@ -272,6 +272,8 @@ class NumpyIndexer(BaseNumpyIndexer):
             Distance (the smaller the better) is returned, not the score.
 
         """
+        if self.size == 0:
+            return None, None
         if self.metric not in {'cosine', 'euclidean'} or self.backend == 'scipy':
             dist = self._cdist(keys, self.query_handler)
         elif self.metric == 'euclidean':
--- a/tests/integration/issues/github_1229/test_sharding_empty_index.py
+++ b/tests/integration/issues/github_1229/test_sharding_empty_index.py
@@ -1,0 +1,73 @@
+import os
+import numpy as np
+import pytest
+
+from jina import Flow, Document
+
+callback_was_called = False
+
+
+def get_index_flow():
+    num_shards = 2
+    f = Flow() \
+        .add(
+        uses='vectorindexer.yml',
+        shards=num_shards,
+        separated_workspace=True,
+    )
+    return f
+
+
+def get_search_flow():
+    num_shards = 2
+    f = Flow(read_only=True) \
+        .add(
+        uses='vectorindexer.yml',
+        shards=num_shards,
+        separated_workspace=True,
+        uses_after='_merge_all',
+        polling='all',
+        timeout_ready='-1'
+    )
+    return f
+
+
+# required because we don't know the order of the pod returning
+# and, when the test failed, we still some time didn't see the error
+@pytest.mark.parametrize('execution_number', range(10))
+def test_sharding_empty_index(tmpdir, execution_number):
+    os.environ['JINA_TEST_1229_WORKSPACE'] = os.path.abspath(tmpdir)
+
+    f = get_index_flow()
+
+    num_docs = 1
+    data = []
+    for i in range(num_docs):
+        with Document() as doc:
+            doc.content = f'data {i}'
+            doc.embedding = np.array([i])
+            data.append(doc)
+
+    with f:
+        f.index(data)
+
+    f = get_search_flow()
+
+    num_query = 10
+    query = []
+    for i in range(num_query):
+        with Document() as doc:
+            doc.content = f'query {i}'
+            doc.embedding = np.array([i])
+            query.append(doc)
+
+    def callback(result):
+        global callback_was_called
+        callback_was_called = True
+        assert len(result.docs) == num_query
+        for d in result.docs:
+            assert len(list(d.matches)) == num_docs
+
+    with f:
+        f.search(query, output_fn=callback)
+    assert callback_was_called
--- a/tests/unit/executors/indexers/test_numpyindexer.py
+++ b/tests/unit/executors/indexers/test_numpyindexer.py
@@ -160,3 +160,25 @@ def test__get_sorted_top_k(batch_size, num_docs, top_k):
 
         np.testing.assert_equal(idx, expected_idx)
         np.testing.assert_equal(dist, expected_dist)
+
+
+@pytest.mark.parametrize('batch_size, compress_level', [(None, 0), (None, 1), (2, 0), (2, 1)])
+def test_numpy_indexer_empty_data(batch_size, compress_level, test_metas):
+    np.random.seed(500)
+    num_dim = 64
+    num_query = 10
+    query = np.array(np.random.random([num_query, num_dim]), dtype=np.float32)
+
+    idx_file_path = os.path.join(test_metas['workspace'], 'np.test.gz')
+    with NumpyIndexer(index_filename=idx_file_path, compress_level=compress_level, metas=test_metas) as indexer:
+        indexer.batch_size = batch_size
+        indexer.touch()
+        indexer.save()
+        assert os.path.exists(indexer.index_abspath)
+        save_abspath = indexer.save_abspath
+
+    with BaseIndexer.load(save_abspath) as indexer:
+        assert isinstance(indexer, NumpyIndexer)
+        idx, dist = indexer.query(query, top_k=4)
+        assert idx == None
+        assert idx == None
--- a/tests/integration/issues/github_1229/vectorindexer.yml
+++ b/tests/integration/issues/github_1229/vectorindexer.yml
@@ -1,0 +1,6 @@
+!NumpyIndexer
+with:
+  index_filename: vec.gz
+metas:
+  workspace: $JINA_TEST_1229_WORKSPACE
+  name: vec_idx


jina/executors/decorators.py
```diff
@@ -247,6 +247,10 @@ def arg_wrapper(*args, **kwargs):
             if b_size is None:
                 return func(*args, **kwargs)
 
+            # catch empty shards
+            if data is None:
```
JoanFM, line 251: I think is not the place to add the fix, I am not sure the return expected is an `array`. I think it needs to be added at `indexer`

cristianmtr: Well the crash happens at the batch_iterator level

```
    for b in batch_iterator(data[:total_size], b_size, split_over_axis, yield_slice=yield_slice):
TypeError: 'NoneType' object is not subscriptable
```

And I assume it will be several indexers that will call this.

And when looking at the indexer query https://github.com/jina-ai/jina/blob/efdb11342536426b78d1c5d448c64a8f3634027a/jina/executors/indexers/vector.py#L263

```
        if self.metric not in {'cosine', 'euclidean'} or self.backend == 'scipy':
            dist = self._cdist(keys, self.query_handler)
        elif self.metric == 'euclidean':
            _keys = _ext_A(keys)
            dist = self._euclidean(_keys, self.query_handler)
        elif self.metric == 'cosine':
            _keys = _ext_A(_norm(keys))
            dist = self._cosine(_keys, self.query_handler)
        else:
```

I see that I would need to add it for each `if` substatement, right?

Also: still a draft. Need to add the tests.

As for return type: if you look at the function that would return after the `for` we can look at this function https://github.com/jina-ai/jina/blob/f0809e32db805401b2768aef7e8eec9c4ebae83d/jina/executors/decorators.py#L179, which seems to return an array

JoanFM: yes, maybe some change also needs to be there for `batch_iterator` but it will also affect the `indexer` executor.

JoanFM: but it is not necessaritly an `np.array`, it may also a simple `array` I think

tests/unit/executors/test_decorators.py
```diff
@@ -9,6 +9,8 @@
 cur_dir = os.path.dirname(os.path.abspath(__file__))
 
 
+# TODO add test empty index with search
```
JoanFM, line 12: this TODO should not be here right?

JoanFM: the tests would belong somewhere else I think?

jina/executors/decorators.py
JoanFM: I think I was wrong it may be actually a bug at batching level. I am howevrr not sure we can assume it returns an np.ndarray. batching decorator has too many things, if you find a pattern to abstract in another function it can also be good

cristianmtr: I think you're looking at an older version. After discussing it with @maximilianwerk we said the batching should just raise the error. And it's up to the Indexers / modules to know what sort of data and format to default to

tests/unit/executors/test_decorators.py
cristianmtr: A very basic version is here https://github.com/jina-ai/jina/pull/1340/files#diff-f72cb602d29fda8028f46b7b7d9940710c62d5b7fb95966f1fa58f6c924cc2d2R242

jina/executors/decorators.py
JoanFM: I think it is not the way to go, because this is not the data passed but the data returned. I think what batching is responsible is to handle data that comes and goes, if it does not receive any data it should just return None.

raising an error is a big statement in my opinion, it should be a natural thing for batching to handle empty data.

maximilianwerk: I disagree. Batching has no clue about the type of the returned value and cannot guess it. If the type is `Option[Any]`, `None` is fine. Anyhow, if the type is not `Option`, `None` does not make any sense, so either we call the underlying function with an empty `np.ndarray` and let it be handled there or we raise an `Exception`.

jina/executors/indexers/vector.py
```diff
@@ -306,3 +310,9 @@ def _cdist(self, *args, **kwargs):
             return cdist(*args, **kwargs, metric=self.metric)
         except ModuleNotFoundError:
             raise ModuleNotFoundError(f'your metric {self.metric} requires scipy, but scipy is not found')
+
+    def _is_empty(self):
```
JoanFM, line 314: I think the emptiness is extracted from the `_size` and not from the `query_handler` being None

tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -160,3 +160,29 @@ def test__get_sorted_top_k(batch_size, num_docs, top_k):
...
+
+@pytest.mark.parametrize('batch_size, compress_level', [(None, 0), (None, 1), (2, 0), (2, 1)])
+def test_numpy_indexer_empty_data(batch_size, compress_level, test_metas):
+    np.random.seed(500)
+    num_data = 100
+    num_dim = 64
+    num_query = 10
+    vec_idx = np.random.randint(0, high=num_data, size=[num_data])
+    # empty data
+    vec = np.empty((0, 0))
```
JoanFM, line 173: Are we sure this reflects the problem we faced? Empty shards are empty because they indexed nothing not because they received an empty index query right?

JoanFM (changes requested):

JoanFM:
I think this ticket requires to add an integration test when a `Flow` with an `Index Pod` with `sharding=2` indexes only one document, and then trying to retrieve that document still works.

jina/executors/indexers/vector.py
cristianmtr: Got it. The problem was that we were passing keys even though there wasn't any data. Fixed that now

tests/unit/executors/indexers/test_numpyindexer.py
cristianmtr: Got it

JoanFM: it may still be an interesting usecase to test though

jina/executors/indexers/vector.py
```diff
@@ -306,3 +310,10 @@ def _cdist(self, *args, **kwargs):
...
```
JoanFM, line 314: do we still need `_is_empty`?

JoanFM (changes requested):
Missing integration test showcasing issue is fixed

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_sharding_empty_index(random_workspace_name):
+    print("WORKSPACE =", random_workspace_name)
+    os.environ['JINA_TEST_1229_WORKSPACE'] = os.path.abspath(random_workspace_name)
+    num_shards = 2
+
+    f = Flow() \
+        .add(
+        uses='vectorindexer.yml',
+        shards=1,
```
JoanFM, line 19: u need to index with 2 shards. I guess it is still on progress right?

tests/unit/executors/test_decorators.py
```diff
@@ -5,6 +5,7 @@
 
 from jina.executors.decorators import as_update_method, as_train_method, as_ndarray, batching, \
     require_train, store_init_kwargs, batching_multi_input
+from jina.executors.indexers.vector import _ext_B
```
JoanFM, line 8: I guess this is not needed anymore?

jina/executors/indexers/vector.py
cristianmtr: Removed

tests/integration/issues/github_1229/test_sharding_empty_index.py
maximilianwerk: Yes it was to shove the code to me for debugging.

cristianmtr: Yep, done now.

tests/unit/executors/test_decorators.py
cristianmtr: Removed.

cristianmtr:
> I think this ticket requires to add an integration test when a `Flow` with an `Index Pod` with `sharding=2` indexes only one document, and then trying to retrieve that document still works.

Yep, added now

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,47 @@
...
+    data = []
+    for i in range(num_docs):
+        doc = Document()
+        doc.content = f'random {i}'
+        doc.embedding = np.array([i])
+        doc.update_id()
+        data.append(doc)
+
+    with f:
+        f.index(data)
```
JoanFM, line 32: to make more explicit that only one shard will have data, add `f.index(data, batch_size=2)` like this u are sure that the request is not split in 2

JoanFM: Just to make sure, have you locally tested this fails without ur changes in current master?

cristianmtr: Hm good point. It seems like the integration test is passing even without my change. However, the multires lyrics example at this commit (https://github.com/jina-ai/examples/blob/96863bec5de24b589fab943bad92fde5328a3ef0/multires-lyrics-search/tests/test_multireslyrics.py#L18) does fail without it and pass with it. That's the original experiment where we noticed the issue

JoanFM: Try with `batch_size=2`

cristianmtr: I reduced the nr of docs to 1 and it still passed

JoanFM: then we need to understand why this is passing in master

cristianmtr:
Just a FYI I am getting empty content on the matches for the docs:

```python
        for d in result.docs:
            print(f'document: {d.content}')
            print(f'matches = {[m.content for m in d.matches]}')
```

```
document: query 0
matches = [None]
document: query 1
matches = [None]
document: query 2
matches = [None]
document: query 3
matches = [None]
document: query 4
matches = [None]
document: query 5
matches = [None]
document: query 6
matches = [None]
document: query 7
matches = [None]
document: query 8
matches = [None]
document: query 9
matches = [None]
```

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,51 @@
+import os
+
+import numpy as np
+
+from jina import Flow, Document
+
+PORT = 45888
```
bwanglzu, line 7:
```python
from jina.helper import random_port
```

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,51 @@
...
+        f.index(data)
+
+    num_query = 10
+    query = []
+    for i in range(num_query):
+        doc = Document()
+        doc.content = f'query {i}'
+        doc.embedding = np.array([i])
+        doc.update_id()
+        query.append(doc)
```
bwanglzu, line 42: maybe you can use `with Document() as doc` and you do not necessarily to call `doc.update_id()`

bwanglzu (changes requested):
not sure, do you see it's related to #1372 ?

cristianmtr:
The integration test in `test_sharding_empty_index.py` does seem to be passing even without the changes introduced in the Indexer. But it's passing with the empty results from above. 

However, the unit test `test_numpy_indexer_empty_data` does fail without the changes introduced in this PR.

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def callback(result):
+        for d in result.docs:
+            print(f'document: {d.content}')
+            print(f'matches = {[m.content for m in d.matches]}')
+        assert result is not None
```
maximilianwerk, line 48: That `assert` is never reached on master. Better might be something along:
```
    callback_was_called = False

    def callback(result):
        callback_was_called = True
        for d in result.docs:
            print(f'document: {d.content}')
            print(f'matches = {[m.content for m in d.matches]}')
        assert result is not None

    with f:
        f.search(query, output_fn=callback)
    assert callback_was_called
```
I am not sure, whether there is a more pytesty way to do this. But that's what needs testing.

tests/integration/issues/github_1229/test_sharding_empty_index.py
maximilianwerk: See my comment below

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,51 @@
+import os
+
+import numpy as np
+
+from jina import Flow, Document
+
+PORT = 45888
+
+
+def test_sharding_empty_index(random_workspace_name):
```
maximilianwerk, line 10: I showed you the `random_workspace_name`, but apparently, either we use plain pytest `tmpdir` fixture or we use `test_metas`. Since the second one is a bit overkill, I'd say, let' go for `tmpdir`. Main reason: The created directories are in the operating system temporary directory and not in the jina main directory.

cristianmtr: Done

tests/integration/issues/github_1229/test_sharding_empty_index.py
cristianmtr: It is reached, because it prints the documents and it is triggered in debugger.

The problem in the case above is that `callback_was_called` is not referenced as `global` within the callback. See my new commit.

```python
callback_was_called = False


def test_sharding_empty_index(tmpdir):
...

    def callback(result):
        global callback_was_called
        callback_was_called = True
        for d in result.docs:
            print(f'document: {d.content}')
            print(f'matches = {[m.content for m in d.matches]}')

    with f:
        f.search(query, output_fn=callback)
    assert callback_was_called

```

This passes, but I still get `None` in the matches content.

JoanFM: Well u were printing the `document` and `matches` right? so it was called

cristianmtr: Yes

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,63 @@
...
+    num_query = 10
+    query = []
+    for i in range(num_query):
+        with Document() as doc:
+            doc.content = f'query {i}'
+            doc.embedding = np.array([i])
+            query.append(doc)
+
+    def callback(result):
+        global callback_was_called
```
JoanFM, line 53: This is something that I find interesting:
1. On one side I hate to see the `global` variable, but I like this idea that guarantees that we do not have false positivies

However, 

we could pass to the callback some `container` to be switched. (We could wrap a boolean in a class that it can be switched in the callback). This could even be made a `fixture` available for every test (which may not be available for the global variable)

cristianmtr:
Adding `polling='all'` fixes it. But then, when I remove the changes from decorator and vector, I cannot reproduce the error from multires lyrics

jina/executors/indexers/vector.py
```diff
@@ -272,6 +272,10 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
             Distance (the smaller the better) is returned, not the score.
 
         """
+        if self.size == 0:
+            query_count = len(keys)
+            result_count = 0
+            return np.empty((query_count, result_count)), np.empty((query_count, result_count))
```
hanxiao, line 278: this is pretty dangerous, `numpy.empty` is like `memalloc` it assigns a memory but not initializing it, resulting in a *random* array. How can one tell if this is legit or not later?

hanxiao:
```
>>> import numpy as np
>>> np.empty([3, 3])
array([[ 2.00000000e+000,  2.00000000e+000, -3.95252517e-323],
       [ 0.00000000e+000,  2.12199579e-314,  0.00000000e+000],
       [ 0.00000000e+000,  0.00000000e+000,  1.77229088e-310]])
>>>
```

hanxiao (changes requested):
Not sure why `numpy.empty` is used here

jina/executors/indexers/vector.py
cristianmtr: Well in this case, since one of the dims is `0` it won't be anything. But I understand your general point. I guess it should be `zeros`?

tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -160,3 +160,25 @@ def test__get_sorted_top_k(batch_size, num_docs, top_k):
...
+        indexer.touch()
+        indexer.save()
+        assert os.path.exists(indexer.index_abspath)
+        save_abspath = indexer.save_abspath
+
+    with BaseIndexer.load(save_abspath) as indexer:
+        assert isinstance(indexer, NumpyIndexer)
+        idx, dist = indexer.query(query, top_k=4)
+        assert idx.shape == dist.shape
+        assert idx.shape == (num_query, 0)
```
hanxiao, line 184: so is the previous `numpy.empty` designed for complying with this?

cristianmtr: That's just the `.shape`, but yes.

hanxiao: in #1229 you mentioned
> If a shard is empty, is should still return an empty list of results.

so the desired behavior is to return an empty ndarray with the shape `[num_query, 0]`?

hanxiao: is there some reason to keep the `axis=0`? or what the choice between
```python
>>> np.empty([0, 0])
array([], shape=(0, 0), dtype=float64)
>>> np.empty([3, 0])
array([], shape=(3, 0), dtype=float64)
>>> np.empty([0])
array([], dtype=float64)
```

cristianmtr: > in #1229 you mentioned
> 
> > If a shard is empty, is should still return an empty list of results.
> 
> so the desired behavior is to return an empty ndarray with the shape `[num_query, 0]`?

Yes

cristianmtr: I think we want this, just so that the shapes still asign (nr of queries - first dim)

```python
>>> np.empty([3, 0])
array([], shape=(3, 0), dtype=float64)
```

hanxiao: can we `return None, None`?

hanxiao: what I don't like about this part is: as a user of this API, you don't know if the indexer is empty *until you inspect at axis=1*, and then say "oh, axis=1 is 0, so it must be empty!" - which isn't an obvious connection unless you documented in that way.

Return `None, None` would be much more direct.

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,78 @@
...
+    for i in range(num_docs):
+        with Document() as doc:
+            doc.content = f'data {i}'
+            doc.embedding = np.array([i])
+            data.append(doc)
+
+    with f:
+        f.index(data)
+
+    f.close()
```
nan-wang, line 55: Not necessary. `with` will call the context manager to close it
```suggestion
```

tests/unit/executors/indexers/test_numpyindexer.py
cristianmtr: But, for example, in the case of sharding, where we want to merge results from two different shards, if one is empty, it's OK, they can merge. If one is `None, None`, error. Right?

tests/integration/issues/github_1229/test_sharding_empty_index.py
```diff
@@ -0,0 +1,76 @@
...
+            doc.content = f'query {i}'
+            doc.embedding = np.array([i])
+            query.append(doc)
+
+    def callback(result):
+        global callback_was_called
+        callback_was_called = True
+        assert len(result.docs) == num_query
+        for d in result.docs:
+            print(f'document: {d.content} with {len(d.matches)} matches')
```
JoanFM, line 70: may we remove all these prints?

JoanFM (approved):

JoanFM (changes requested):
The empties should be handled differently?

tests/integration/issues/github_1229/test_sharding_empty_index.py
cristianmtr: Done

tests/unit/executors/indexers/test_numpyindexer.py
hanxiao: then this logic should be put into driver. which merge are we talking about? can you point out the function?

hanxiao (changes requested):
the `None, None` handling should be put into driver, which merge function are we talking about, can someone points out the function?

cristianmtr:
> the `None, None` handling should be put into driver, which merge function are we talking about, can someone points out the function?

How about returning something like

```python
        if self.size == 0:
            query_count = len(keys)
            result = np.array([[] for _ in range(query_count)])
            return result, result
```

So we maintain compatibility with the interface?

Not sure what would break if we switch to `None, None`, running all the tests now to see

maximilianwerk:
> > the `None, None` handling should be put into driver, which merge function are we talking about, can someone points out the function?
> 
> How about returning something like
> 
> ```python
>         if self.size == 0:
>             query_count = len(keys)
>             result = np.array([[] for _ in range(query_count)])
>             return result, result
> ```
> 
> So we maintain compatibility with the interface?
> 
> Not sure what would break if we switch to `None, None`, running all the tests now to see

The way `_merge_all` is constructed, returning `None, None` should be fine. Haven't thought about it that way beforehand. I rather thought we want to stick with the function interface. But I agree, that if it just works, it makes sense on this level.

cristianmtr:
> > > the `None, None` handling should be put into driver, which merge function are we talking about, can someone points out the function?
> > 
> > 
> > How about returning something like
> > ```python
> >         if self.size == 0:
> >             query_count = len(keys)
> >             result = np.array([[] for _ in range(query_count)])
> >             return result, result
> > ```
> > 
> > 
> > So we maintain compatibility with the interface?
> > Not sure what would break if we switch to `None, None`, running all the tests now to see
> 
> The way `_merge_all` is constructed, returning `None, None` should be fine. Haven't thought about it that way beforehand. I rather thought we want to stick with the function interface. But I agree, that if it just works, it makes sense on this level.

Done, returning `None, None` now. Adapted interface to be `Optional[...]`

hanxiao (approved):
👍

hanxiao:
the most discussed PR so far?

tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -160,3 +160,25 @@ def test__get_sorted_top_k(batch_size, num_docs, top_k):
...
+        indexer.touch()
+        indexer.save()
+        assert os.path.exists(indexer.index_abspath)
+        save_abspath = indexer.save_abspath
+
+    with BaseIndexer.load(save_abspath) as indexer:
+        assert isinstance(indexer, NumpyIndexer)
+        idx, dist = indexer.query(query, top_k=4)
+        assert idx == None
+        assert idx == None
```
nan-wang, line 184: Fix the duplicates.
```
assert dist == None
```

cristianmtr: https://github.com/jina-ai/jina/pull/1396

PR #1342, opened 2020-11-24
refactor: moving log parse to core from jinad

deepankarm:


jina/peapods/jinad.py
```diff
@@ -189,6 +189,7 @@ async def wslogs(self, remote_id: 'str', stop_event: Event, current_line: int =
                             log_line = json.loads(log_line)
                             current_line = int(list(log_line.keys())[0])
                             log_line_dict = list(log_line.values())[0]
+                            log_line_dict = json.loads(log_line_dict.split('\t')[-1].strip())
```
JoanFM, line 192: is this safe? can we add a try to default to simply the `log_line_dict` as it is?

JoanFM (approved):

jina/peapods/jinad.py
deepankarm: We've added a `json.decoder.JSONDecodeError`. Should be safe for now.

PR #1343, opened 2020-11-24
refactor(types): move add_chunk add_match to Set

hanxiao:


jina/types/document/uid.py
```diff
@@ -88,3 +91,35 @@ def is_valid_id(value: str) -> bool:
...
+
+        return str.__new__(cls, seq)
+
+    @property
+    def to_hash(self):
+        """The document id in the integer form of bytes, as 8 bytes map to int64.
+        This is useful when sometimes you want to use key along with other numeric values together in one ndarray,
+        such as ranker and Numpyindexer
+        """
+        return id2hash(self)
```
bwanglzu, line 117: can we do something like this? `to_hash` sounds like a behaviour, not a property
```python
def __hash__(self):
    return id2hash(self)
```

jina/types/document/uid.py
```diff
@@ -88,3 +91,35 @@ def is_valid_id(value: str) -> bool:
...
+
+    @property
+    def to_bytes(self):
+        """The document id in the binary format of str, it has 8 bytes fixed length,
+        so it can be used in the dense file storage, e.g. BinaryPbIndexer,
+        as it requires the key has to be fixed length.
+        """
+        return id2bytes(self)
```
bwanglzu, line 125: same as above

jina/drivers/rank.py
```diff
@@ -63,15 +62,15 @@ def _apply_all(self, docs: 'DocumentSet', context_doc: 'Document', *args,
...
-            query_chunk_meta[chunk.id_in_hash] = chunk.get_attrs(*self.exec.required_keys)
+            query_chunk_meta[chunk.id.to_hash] = chunk.get_attrs(*self.exec.required_keys)
             for match in chunk.matches:
                 match_idx.append(
-                    (match.parent_id_in_hash,
-                     match.id_in_hash,
-                     chunk.id_in_hash,
+                    (match.parent_id.to_hash,
+                     match.id.to_hash,
+                     chunk.id.to_hash,
```
bwanglzu, line 70: with the comment blow, I would do something like this:
```python
query_chunk_meta[hash(chunk.id)] = chunk.get_attrs(...)
# or 
query_chunk_meta[chunk.id.to_hash()] = chunk.get_attrs(...)
```

jina/types/document/uid.py
hanxiao: `hash` is built_in tho, can change it to `as_hash`

hanxiao: change to `hash_value`

bwanglzu: @hanxiao no, the point is. `hash_value` is an "action"/"behaviour", not a "property".  the name `to_hash` or `hash_value` both sounds good to me. while it's a method performing some action on an instance, can we rewrite the default magic method? (i.e. `__hash__`)

hanxiao: it is a computational property, so a property

hanxiao: i will override __bytes__ and __hash__

nan-wang:
Right now, there are two functions for adding a chunk/match to the document. I'd suggest removing `add()`. 
> There should be one-- and preferably only one --obvious way to do it.
>      _The Zen of Python_

```python
with Document() as d:
    c1 = d.chunks.append()
    c1.id = f'1:0>16'
    c2 = d.chunks.add()
    c2.id = f'2:0>16'
```

Issue #1277, opened 2020-11-13
Jupyter notebook & lab kernel restarting without obvious reason when working with Jina


hanxiao:
**Describe your problem**

Orginal post at here:
https://github.com/jupyter/notebook/issues/5863#issue-741473325

Answer from Jupyter maintainer here:
https://github.com/jupyter/notebook/issues/5863#issuecomment-726067963

**What is your guess?**

Pretty hardcore. Seems related to the `asyncio` eventloop lifecycle override.

**Related issues**

https://github.com/ipython/ipython/issues/11303/#issuecomment-420566168
https://github.com/ipython/ipykernel/issues/548

---

**Environment**

**Screenshots**

hanxiao:
https://github.com/ipython/ipython/issues/11303/#issuecomment-420566168

hanxiao:
https://github.com/ipython/ipykernel/issues/548

hanxiao:
related PR #1450

PR #1348, opened 2020-11-25
feat: native async based grpc & fastapi based rest for gateway

### LINKED ISSUE #1287: Refactoring async gRPC servlet from ThreadPoolExecutor to native asyncio support
**Describe the feature**

gRPC asyncio support graduated since this Aug. https://github.com/grpc/grpc/issues/6046#issuecomment-675620819
Doc: https://grpc.github.io/grpc/python/grpc_asyncio.html

The current way of async on gRPC introduce compatibility problem with ipykernel. see #1277 

**Migrating to native asyncio support may or may not solve #1277**. Speedwise it may improve as I expect native asyncio support has better implementation than ThreadExecutorPool.

The gRPC docs have not included the usage of async yet. One can learn from the unit test here: https://github.com/grpc/grpc/blob/c47bc1a7093fa2202ac28cdf2d374ea9680cbec1/src/python/grpcio_tests/tests_aio/unit/call_test.py#L218

Major contributor of this async feature on the grpc(/google) side is @lidizheng

**Your proposal**

---

**Environment**

**Screenshots**

deepankarm:
### Following changes are done in this PR

- Enable native async based gRPC for gateway server (fixes #1287)
- Change `PyClient` to read from `grpc.aio.channel`
- Move from `Flask + gevent` to `FastAPI + uvicorn` based REST gateway
- Move the gateway logic to a separate process for both gRPC & REST (fixes https://github.com/jina-ai/jinad/issues/9)
- Move PyClient to a separate process/thread

#### gRPC
- Moved from custom `AsyncExecutor` to [native asyncio for gRPC](https://grpc.github.io/grpc/python/grpc_asyncio.html)
- Previously server was async, whereas `PyClient` was sync. With this PR, `PyClient` becomes async (`call`, `call_unary`, `dry_run`, `index`, `search`, `train` are now coroutines.
 

#### REST
- `FastAPI` & `uvicorn` support native async capabilities, rather than solving it in a hacky way with `gevent.WSGIServer`.
- FastAPI also claims much higher performance benefits (performance comparison to be done)
- With gRPC, we have native streaming + prefetch ability, which was not there with REST. This can now be done wih the help of `websockets` on `FastAPI` server.

#### PyClient
- `PyClient` (and hence `py_client`) is now an awaitable.
- We enable `PyClientRuntime`, which executes `PyClient` to a separate process to avoid an event loop in the main process.
- `PyClient` (optionally) sends the gRPC response to a zmq socket, from which main process can read & execute callbacks. (This is the default way of when we do `flow.index`, `flow.search` etc)

#### JinaD
- Since gateway now lives on a separate process for both gRPC & REST, flow invocation logic now can move from `def` to `async def` for jinad [(Difference)](https://fastapi.tiangolo.com/async/#path-operation-functions)
- Deleting a Flow wouldn't kill `jinad` now

#### CtrlZmqlet
- Zmqlet with only PAIR socket (capable of BIND or CONNECT) to be tied with all runtimes that don't have any Peas related logic (e.g. - `gateway-grpc`, `gateway-rest`, `remote`, `client`) 

#### Jupyter Notebook
- We can now build & run Flows in Jupyter. This is possible as we avoid `event loops` in the main process (avoiding collision with Jupyter's default event loop). Fixes https://github.com/jina-ai/jina/issues/1277 

image

#### Pending
- `pytest.asyncio` tests for `PyClient` would still cause segmentation faults, as that brings `grpc.aio` channel (hence the event loop) on main process. This needs to be handled.
- Do performance tests for gRPC + REST based changes
- Make changes to `jinad` (`def ` to `async def`)
- Discuss about async propagation in core (whereto `await` vs `run_until_complete` a coroutine)


jina/peapods/gateway.py
```diff
@@ -144,6 +144,44 @@ def prefetch_req(num_req, fetch_to):
                     prefetch_task = [j for j in onrecv_task]
 
 
+class AsyncGatewayPea:
+    def __init__(self, args):
+        if not args.proxy and os.name != 'nt':
+            os.unsetenv('http_proxy')
```
JoanFM, line 150: This feels like users may get very mad at us?

jina/peapods/pod.py
```diff
@@ -556,6 +556,25 @@ def start(self) -> 'GatewayPod':
         return self
 
 
+class AsyncGatewayPod(BasePod, AsyncExitStack):
```
JoanFM, line 559: Will this play nice with inheriting both `ỀxitStack` and `AsyncExitStack` (I think is a good point I may try to have a mixin in the refactor for `ExitStack` so that it is more separated from `BasePod`.

jina/peapods/gateway.py
JoanFM: Why is this needed?

jina/peapods/pod.py
deepankarm:
```

class DualContext(ExitStack, AsyncExitStack):
    def __init__(self) -> None:
        super(ExitStack, self).__init__()
        super(AsyncExitStack, self).__init__()

    def start(self):
        raise NotImplementedError

    def close(self):
        raise NotImplementedError

    async def astart(self):
        raise NotImplementedError

    async def aclose(self):
        raise NotImplementedError

    def __enter__(self):
        return self.start()

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.close()
        super(ExitStack, self).__exit__(exc_type, exc_val, exc_tb)

    async def __aenter__(self):
        return await self.astart()

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        await self.aclose()
        await super(AsyncExitStack, self).__aexit__(exc_type, exc_val, exc_tb)


class SyncContext(DualContext):
    def start(self):
        print('Sync')

    def close(self):
        print('Sync close')


class AsyncContext(DualContext):
    async def astart(self):
        print('Async')

    async def aclose(self):
        print('Async close')


with SyncContext():
    pass

async def blah():
    async with AsyncContext():
        pass

asyncio.run(blah())
```

deepankarm: This works, maybe we can use something similar

jina/peapods/gateway.py
nan-wang: Agree. Unsetting `http_proxy` and `https_proxy` sounds really bad.

hanxiao: it is needed since GNES @deepankarm @nan-wang @JoanFM please 
1. read the old code
2. when you running grpc with http and https proxy you will find grpc won't work anymore.
3. always keep in mind os.environ is set on process base not globally, there is nothing people get mad at us, because nothing change in global 
4. anytime, use warning.warn

hanxiao: please always understand grpc is on HTTP2 layer and that's why http proxy setting affects it. if you don't know about this, read grpc doc.

if you really want to solve this elegantly, i suggest looking for issue help/posting issue on grpc github.

also please fix the misconception that os.environ changes anything globally.

until then, keep this lines. Many companies have internal proxy, without this line, Jina won't work for them.

JoanFM: Are we sure this init lives in another process?

JoanFM: Shouldn't it be moved into start?

JoanFM:
> Zmqlet with only ctrl socket (`AsyncCtrlZmqlet`) -
> 
> * main process fetches the ctrl address
> * `loop_body` in the event loop can await `AsyncCtrlZmqlet` (which awaits the terminate signal on the ctrl socket)
> * upon termination signal, loop_termination kicks in (killing grpc/rest servers, sending delete signal to jinad etc)
> * every jina-process which is not a local pea with `zmqstreamlet` can inherit this (remote, gateway-grpc, gateway-rest)
> * ctrl address can be used for termination
> * doesn't need input/output sockets
> * i'm assuming, this will always be local. we should be good with ipc only (to avoid clash)?

I think assuming is always local w.r.t the Flow is a valid assumption right now

hanxiao:
the test is broken

tests/unit/clients/python/test_client.py
```diff
@@ -60,12 +61,14 @@ def test_check_input_fail(input_fn):
 def test_gateway_ready(port_expose, route, status_code):
     p = set_gateway_parser().parse_args(['--port-expose', str(port_expose)])
     with RESTGatewayPea(p):
+        time.sleep(0.5)
```
theUnkownName, line 64: Why are we having a sleep?

deepankarm: `uvicorn` takes a small amount of time to get started. This is to allow the threads to get started.

jina/clients/python/grpc.py
```diff
@@ -27,49 +29,46 @@ def __init__(self, args: 'argparse.Namespace'):
...
         self._stub = jina_pb2_grpc.JinaRPCStub(self._channel)
 
         # attache response handler
         self.logger.success(f'connected to the gateway at {self.args.host}:{self.args.port_expose}!')
         self.is_closed = False
 
-    def call(self, *args, **kwargs):
+    def configure_event_loop(self):
+        try:
+            self.loop = asyncio.get_running_loop()
```
JoanFM, line 51: what is the meaning of this? we try to see if there is an eventloop and if not we go for `uvloop`? In which context may this happen?

tests/unit/clients/python/test_client.py
theUnkownName: Ah ok, got it!

jina/clients/python/grpc.py
deepankarm: This is an attempt to make the client work with an already running loop (e.g.- Jupyter). We'd need an additional loop-patching to make this work on Jupyter though

JoanFM: would be good to add a comment to avoid someone removing since I am sure it won't be covered by tests

jina/clients/__init__.py
```diff
@@ -18,16 +31,130 @@ def py_client(**kwargs) -> 'PyClient':
...
+
+    with JinaLogger(context='PyClientRuntime') as logger:
+        with CtrlZmqlet(args=args, logger=logger, is_bind=False, is_async=False, timeout=10) as zmqlet:
+            with PyClientRuntime(args, mode=mode, input_fn=input_fn, output_fn=output_fn,
+                                 address=zmqlet.address, **kwargs):
+                counter = 0
+                while True:
+                    try:
+                        msg = zmqlet.sock.recv()
+                        if msg == b'TERMINATE':
```
JoanFM, line 76: when is the client sending `TERMINATE` I haven't seen this happen, it is done between Peas but not from client right?

jina/clients/__init__.py
```diff
@@ -18,16 +31,130 @@ def py_client(**kwargs) -> 'PyClient':
...
+                        if counter == 5:
+                            logger.warning('Waited for 5 seconds for zmq BIND socket before exiting')
+                            break
+                        logger.debug('PyClient\'s BIND socket is not open yet. waiting for some time!')
+                        time.sleep(1)
+                        counter += 1
+                        continue
+
+
+class PyClientRuntime(BasePea):
```
JoanFM, line 101: could you call it PyClientPea? Like this it won't affect the naming that will come with the `BasePea` refactor

jina/clients/python/__init__.py
```diff
@@ -24,7 +26,7 @@
     import argparse
 
 
-class PyClient(GrpcClient):
+class PyClient(AsyncGrpcClient):
```
JoanFM, line 29: What is the difference between this `client` and `PyClientRuntime` or `PyClientPea`?

JoanFM: Is this the `real` client used in regular Grpc calls like in examples when using `input_lines` etc.. and the other is internal and instantiated when we expose a RestGateway?

deepankarm: `PyClient` is the actual client that does the `index` etc operations. But every operation now is `async`. To avoid these in the main process, `PyClientRuntime` allows us to run `PyClient` in a separate process.

jina/clients/__init__.py
deepankarm: I was hoping to change such Processes which don't have any `Pea` kind of logic to a different naming convention after the refactoring - e.g. - gateway-grpc, gateway-rest, remote. client etc

jina/clients/__init__.py
deepankarm: `PyClient` would send a terminate signal to `PyClientRuntime`

JoanFM: it should be the other way around right? PyClient does not know it lives inside a `PyClientRuntime`? Why don't u use `multiprocessing Events`?

jina/clients/python/__init__.py
JoanFM: Ok I see, PyClient is kind of the executor of the regular BasePea (in terms of where is initialized and so on)

jina/clients/__init__.py
deepankarm: `PyClient` receives a stream of responses from gRPC servicer. To make sure the `callback` functions are executed back in the main process, `PyClient` sends each such response to `PyClientRuntime`. Once the stream is completed, `PyClient` sends a `TERMINATE` signal to `PyClientRuntime`. But you're right, this can be handled using an `Event.

jina/clients/__init__.py
JoanFM: Sure I will

jina/clients/python/grpc.py
```diff
@@ -27,52 +29,62 @@ def __init__(self, args: 'argparse.Namespace'):
...
+            self.logger.critical(f'can not connect to the server at {self.args.host}:{self.args.port_expose} after '
+                                 f'{self.args.timeout_ready} ms, please double check the ip and grpc port number'
+                                 f' of the server')
+            raise GRPCServerError(f'can not connect to the server at {self.args.host}:{self.args.port_expose}')
+        return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        self.close()
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        await self.close()
```
JoanFM, line 78: This aexit is not tested?

jina/clients/python/grpc.py
```diff
@@ -84,21 +96,24 @@ def start(self, *args, **kwargs) -> 'GrpcClient':
...
                 raise BadClientRequestGenerator(f'{msg}\n'
                                                 'often the case is that you define/send a bad input iterator to jina, '
                                                 'please double check your input iterator') from rpc_ex
             else:
                 raise BadClient(msg) from rpc_ex
         finally:
-            self.close()
+            # avoid closing a client after a single `index`, `search` or `train` operation
+            if 'close' in kwargs:
+                await self.close()
```
JoanFM, line 108: Test for this part exists?

jina/peapods/gateway/grpc.py
```diff
@@ -0,0 +1,106 @@
...
+        self._bind_address = f'{args.host}:{args.port_expose}'
+        self._server.add_insecure_port(self._bind_address)
+
+    async def start(self):
+        await self._server.start()
+        self.logger.success(f'gateway (gRPC) is listening at: {self._bind_address}')
+        await self.is_gateway_ready.wait()
+        return self
+
+    async def __aenter__(self):
```
JoanFM, line 98: No test using it as ctxt manager

JoanFM (commented):
Codecov reports missing test for some cases specially when using ctxt management

deepankarm:
> Codecov reports missing test for some cases specially when using ctxt management

Testing the context manager for `AsyncGrpcClient` (and `PyClient`) is not possible until we fix issues with `pytest.asyncio`. This is because, `grpc.aio` channel comes to the main process, and creates issues with `segmentation faults` intermittently. I'll raise an issue for this to keep track.

hanxiao (approved):

PR #1351, opened 2020-11-25
feat: containerpea communication

JoanFM:
**Changes introduced**
Change the communication between `ContainerPea` and the `BasePea in container` to be based on control_port. Use `bridge network` for this communication when possible (to ensure it can work when `docker from docker` is used (as proven by the `hubapp` test)

Add test for hubapp
Added skipped test for hubpods in docker_compose that showcase the limitations to test these features


nan-wang:
I guess we need an integrating test that has a flow container with a pod container.

JoanFM:
> I guess we need an integrating test that has a flow container with a pod container.

Yes, I plan to add 2 tests:

1. Use jinad with Hub pods as pods.
2. Have a complete hubapp contaneirized (separate github action)

tests/integration/jinad/test_simple_hub_pods/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+version: "3.3"
```
JoanFM, line 1: This test does not work, and not sure if we can make it work (I think we would need to pass parameters for special cases where docker pods are used in a docker-compose).

So the first point is:

- We need to expose `/var/run/docker.sock:/var/run/docker.sock` which allows the docker containers created via the docker-compose to communicate with the `docker daemon`. (Thus, ContainerPea can start peas in Pods via docker SDK). First thing to notice. These new docker containers won't leave inside the docker images hidden from the outside, they are docker containers in the same host and visible via host.

- The problem is that right now this new docker is using network mode host, while the containers started by docker-compose have their own bridge network. I haven't been able to successfully connect from the inner `bridge network` to the new dockers in host mode.

- As a test exercise I started some new docker containers attached to this new `bridge` network, and I could communicate to the other containers.  (This could lead to the conclusion that we could have the new `Peas in containers` attached to this network. If we do so, these containers will be assigned a new IP Address that we can't control and the rest of the pods won't be able to rediscover them).

jina/peapods/container.py
```diff
@@ -69,29 +72,46 @@ def post_init(self):
...
-                                                      # activate
-                                                      )
+                                                      entrypoint=self.args.entrypoint)
+
+        # Related to potential docker-in-docker communication. If `ContainerPea` lives already inside a container.
+        # it will need to communicate using the `bridge` network.
+        if platform in ('linux', 'linux2'):
+            bridge_network = self._client.networks.get('bridge')
+            if bridge_network:
+                self.ctrl_addr, self.ctrl_with_ipc = Zmqlet.get_ctrl_address(
```
JoanFM, line 82: apparently using `bridge` gateway seems to allow the `ContainerPea` to always connect with the Pea in container.

tests/integration/jinad/test_simple_hub_pods/docker-compose.yml
JoanFM: I think it is not necessarily a problem not to enable this specific usecase, the problem I see is that it does not let us test the `ContainerPea` in `remote` configuration.

tests/integration/hubapp-dind/test_integration.sh
```diff
@@ -0,0 +1,30 @@
+set -e
+
+if [ "${PWD##*/}" != "jina" ]
+  then
+    echo "test_integration.sh should only be run from the jina base directory"
+    exit 1
+fi
+
+CONTAINER_ID=$(docker run -v /var/run/docker.sock:/var/run/docker.sock --network=host -d jinaai/test_hubapp_dind)
```
JoanFM, line 9: As we can see here, in order for `hubapp` to be able to connect to `hubpod` they need to be in the same network, meaning that `hubapp` needs to be started with `--network==host`

nan-wang (commented):
`tests/integration/hubapp-dind` -> `tests/integration/hubapp_dind`

.github/workflows/test-hubapp-dind.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          submodules: true
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Build jina test-pip
+        run: |
+          docker build -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
+          docker build --build-arg PIP_TAG="[devel]" -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
```
nan-wang, line 31: Is it a typo here? I can't see the reason for building the docker image twice with different pip tags and the same image tag.

JoanFM: I had the same question, just copied from our `CI.yml` actually!

tests/integration/hubapp-dind/requirements.txt
```diff
@@ -0,0 +1,3 @@
+# seems that BiT only works in 2.0.0 not 2.2.0, but hey it's google's problem!
+tensorflow_cpu==2.1.0
+jina[cv,http]==0.6.4
```
nan-wang, line 3: Do we need this?

JoanFM: Just removed!

jina/peapods/container.py
nan-wang: ## Background

```
hubapp (container#1)
    |- gateway
    |- pod1
    |    |- pea1_1
    |        |- hubpod (container#2)
    |- ...
```
## Goal
What is the goal here? 
Do we want to pass a message from `gateway` to `container#2`?
Or do we want to send the ctrl signal directly to `container#2`?

tests/integration/hubapp-dind/test_integration.sh
nan-wang: By setting `--network==host` for `container#1`, the `container#1` will share all its port with the physical machine. As for the alternative solution you mentioned above, do you mean `container#2` also need `--network==host`?

```
hubapp (container#1)
    |- gateway
    |- pod1
    |    |- pea1_1
    |        |- hubpod (container#2)
    |- ...
```

.github/workflows/test-hubapp-dind.yml
JoanFM: Try to remove in both places

nan-wang (commented):
Maybe we can split the two integration tests into separated PRs?

JoanFM:
> I guess we need an integrating test that has a flow container with a pod container.

You can see it here, the trick to enable this is to start `hubapp` container with `network='host'`

jina/peapods/container.py
JoanFM: We want to be able to send control message from `pea_1` to `hubpod (container #2) (I think I understood the schema.

tests/integration/hubapp-dind/test_integration.sh
JoanFM: I do not understand which alternative solution you refer to?

JoanFM:
> Maybe we can split the two integration tests into separated PRs?

We could, but since these changes potentially affected all these cases...

However, as I state above, I do not think that the test of having `docker_compose` simulating `ContainerPea` in remote will work (unless we do a lot of changes and maybe we need to provide the user the ability to provide more settings on what network and what arguments they will need to provide to the containerPea. (Which I would consider for now out of scope and maybe see if the community provides better ideas on how they may want to handle it)

tests/integration/hubapp-dind/test_integration.sh
JoanFM: But `container #2` is the one spawned by `ContainerPea` and for now we always start it with `--network=host` (because we expect to be available from the outside). I think this is a good thing to have. (I do not think we should do (or even if it is possible) ) to do the "reflection" that this `pea1_1` is also living inside a `Container` and therefore `pea1_1` should be attached to the `bridge` network where I live and with a specific IP that I know.

hanxiao (approved):

PR #1352, opened 2020-11-25
chore(flow): no constructor for class method needed

florian-hoenicke:
Since `load_config` is a `classmethod`, we don't need to call it on a Flow object. Instead we can call the `classmethod` directly.

### DIFF
--- a/docs/chapters/flow/index.md
+++ b/docs/chapters/flow/index.md
@@ -527,7 +527,7 @@ Passing `top_k_queryset` to `flow.search` will override `top_k` value of `10` wi
 Note that more than one `queryset` can be passed with any request.
 
 ```python
-    with Flow().load_config('flow.yml') as search_flow:
+    with Flow.load_config('flow.yml') as search_flow:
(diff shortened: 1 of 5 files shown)


bwanglzu (approved):
Good catch!

JoanFM (approved):

cristianmtr:
Nice, but then should't it be a `staticmethod`? I can't see that the `cls` is being used.

JoanFM:
> Nice, but then should't it be a `staticmethod`? I can't see that the `cls` is being used.

yes, u are true, it should be a static

nan-wang:
@florian-hoenicke Good job! A minor comment. `chore(xxx):` will skip our unit tests. For the changes like this PR, it's better to use `refactor(xxx):`

hanxiao:
> @florian-hoenicke Good job! A minor comment. `chore(xxx):` will skip our unit tests. For the changes like this PR, it's better to use `refactor(xxx):`

correct: skip our cd not skip unit test

florian-hoenicke:
@nan-wang  Thanks, I will use refactor in the future.

PR #1353, opened 2020-11-25
fix(helper): search file error handling

florian-hoenicke:
When running `f = Flow.load_config('flow-index.yml')`, I received the following error:
`FileNotFoundError: [Errno 2] No such file or directory: '/Library/TeX/texbin/flow-index.yml'`
This is because the file `flow-index.yml` did not exist. However, the error message tells me that the file does not exist in my LaTeX folder, which is weird.
It turned out that the current helper function takes the last path in the `PATH` variable as default. For sure, this is not intended.
I changed the implementation to fire an exception instead of taking the last path as default.

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -562,28 +562,34 @@ def get_now_timestamp():
     return int(datetime.timestamp(now))
 
 
-def complete_path(path: str) -> str:
+def search_file_in_paths(path):
+    '''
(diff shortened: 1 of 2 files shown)


cristianmtr:
But won't this affect performance? Recursively checking all these folders?

JoanFM:
Performance is not an issue at this stage because this is all working at creation time and spin-up

hanxiao (approved):

PR #1354, opened 2020-11-25
fix: make load config static

cristianmtr:


JoanFM (approved):
I think we can use many more staticmethods in many plaes

PR #1356, opened 2020-11-25
test: replace all random docs with new api

bwanglzu:
1. Replace all `random_docs` with `random_docs_new_api`, and raname `random_docs_new_api` to `random_docs`. (didn't turn it into a conftest fixture since it has a lot of parameters)
2. Add workspace fixture to some test cases, ensure stored in `tmpdir`.
3. Complete & refactor `DocumentSet` test. During refactoring, fixed some bugs in `DocumentSet`.

What has been fixed in `DocumentSet`:
1. `__setitem__` calls `CopyFrom`, which expect to receive a `Document`, passed `Document.as_pb_object`.
2. `__setitem__` for string object not setting, but returning.
3. `__getitem__` when `isinstance(key, str)`, we convert the value into a string.


tests/unit/yaml/test-index.yml
```diff
@@ -3,7 +3,7 @@ with:
   index_filename: tmp2
 metas:
   name: test2
-  workspace: $JINA_TEST_INDEX
+  workspace: $TEST_WORKDIR
```
JoanFM, line 6: We arr trying to set the habit to prefix env variablws with JINA so please keep it as it was

JoanFM (changes requested):

jina/types/sets/document_set.py
```diff
@@ -49,7 +49,7 @@ def __getitem__(self, item):
         if isinstance(item, int):
             return Document(self._docs_proto[item])
         elif isinstance(item, str):
-            return Document(self._docs_map[str(item)])
```
hanxiao, line 52: this is required, as `UniqueId` can not be directly used as a string

hanxiao: oh but the test doesn't complain 🤔

bwanglzu: image
@hanxiao this is what I got in terminal

tests/unit/yaml/test-index.yml
bwanglzu: I'll change it back

hanxiao (approved):

PR #1358, opened 2020-11-26
fix: fix top_k

### LINKED ISSUE #1350: Flow search always return 50 results with random top_k set
**Describe your problem**

Hi all, I have setup a jinabox.js environment and create several image search apps, but I found that no matter how I set `top_k`, the search result would always be 50.

**What is your guess?**

After digging out the source code, I found that the default value of top_k for `Flow.search()` method is 50, so I'm wondering if the `top_k` parameter hasn't been passed to Flow gateway when setting up Flask server.

---

**Environment**

v0.7.11

**Screenshots**

image

BTW, I think the `time_out` parameter is also not working here.

JoanFM:
Fixes #1350


hanxiao:
it should be `SliceQL`: `end` to modify

JoanFM:
> it should be `SliceQL`: `end` to modify

Yes it is known issue; but it is a big assumption to think that SliceQL is present in the Flow while VectorSearchDriver is almost mandatory

PR #1361, opened 2020-11-26
fix: changes to test levels for hubio

rutujasurve94:
- Changes to logic of `_test_build` in `HubIO` : returns failed test levels

- Determine if build is successful or not depending on the level of test failure returned.


jina/docker/hubio.py
```diff
@@ -344,8 +344,9 @@ def build(self) -> Dict:
                     p_names = []
                     try:
                         is_build_success = False
-                        p_names = self._test_build(image)
-                        is_build_success = True
+                        p_names, failed_test_levels = self._test_build(image)
+                        if BuildTestLevel.POD_DOCKER or BuildTestLevel.EXECUTOR in failed_test_levels:
```
JoanFM, line 348: this is wrong, the logic should be negative, is more about which errors we do not allow rather than what we accept

JoanFM: Now if every level fails, we will consider it a success

JoanFM (changes requested):

jina/docker/hubio.py
```diff
@@ -344,8 +344,13 @@ def build(self) -> Dict:
...
                         is_build_success = False
-                        p_names = self._test_build(image)
-                        is_build_success = True
+                        p_names, failed_test_levels = self._test_build(image)
+                        if BuildTestLevel.POD_DOCKER or BuildTestLevel.FLOW in failed_test_levels:
+                            is_build_success = False
+                            self.logger.error(f'build unsuccessful, failed at {str(failed_test_levels).strip('[]')} level')
+                        else:
+                            is_build_success = True
+                            self.logger.warning(f'test failed at : {str(failed_test_levels).strip('[]')} levels')
```
JoanFM, line 353: Clarify the message maybe? 'Build successful. Tests failed at this level ... (This could be do to the fact that `executor` has non-installed external dependencies)

jina/docker/hubio.py
```diff
@@ -405,33 +410,47 @@ def build(self) -> Dict:
...
 
         # test uses at Flow level
         if self.args.test_level >= BuildTestLevel.FLOW:
             p_name = random_name()
-            with Flow().add(name=random_name(), uses=image.tags[0], daemon=self.args.daemon):
-                pass
-            p_names.append(p_name)
+            try:
+                with Flow().add(name=random_name(), uses=image.tags[0], daemon=self.args.daemon):
+                    pass
```
nan-wang, line 448: FYI, `pass` won't run the flow. Is it what we expected?

JoanFM: I think it is intended, it would be cool to be able to send some data to ensure it does not break in the `exec` method though. But it may be hard to provide a solution for every executor family, although we could provide some predefined set of input for tests. (Like text for nlpcrafters, docs with `matches` for rankers etc... etc...). It would be out of the scope of this PR however...

tests/unit/docker/test_hub_build_level.py
```diff
@@ -0,0 +1,37 @@
...
+import pytest
+
+from jina.docker.hubio import HubIO
+from jina.helper import yaml
+from jina.enums import BuildTestLevel
+from jina.peapods import Pod
+from jina.executors import BaseExecutor
+from jina.parser import set_hub_build_parser
+
+cur_dir = os.path.dirname(os.path.abspath(__file__))
```
JoanFM, line 12: can u use Pathlib since we have switched to it?

tests/unit/docker/test_hub_build_level.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_hub_build_pull():
```
JoanFM, line 14: u may parametrize these tests?

JoanFM (changes requested):

tests/unit/docker/test_hub_build_level.py
```diff
@@ -0,0 +1,38 @@
...
+from jina.helper import yaml
+from jina.enums import BuildTestLevel
+from jina.peapods import Pod
+from jina.executors import BaseExecutor
+from jina.parser import set_hub_build_parser
+
+cur_dir = Path.cwd()
+
+def test_hub_build_pull():
+    args = set_hub_build_parser().parse_args([os.path.join(cur_dir, 'hub-mwu'), '--push', '--host-info', '--test-level', 'FLOW'])
```
JoanFM, line 16: os.path.join can be changed to a pathlib way?

JoanFM (changes requested):

JoanFM (commented):
I miss a test that `checks` the logic of setting `build_success` depending on the `failed_levels`

JoanFM (dismissed):

JoanFM (approved):

tests/unit/docker/test_hub_build_level.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.fixture
+def mock_load_config():
+    return BaseExecutor
+
+def test_hub_build_level_pass(monkeypatch, mock_load_config):
+
+    monkeypatch.setattr(BaseExecutor, "load_config", mock_load_config)
+    args = set_hub_build_parser().parse_args(['path/hub-mwu', '--push', '--host-info', '--test-level', 'EXECUTOR'])
+
+    p_names, failed_levels = HubIO._test_build("jinahub/pod.dummy_mwu_encoder", BuildTestLevel.EXECUTOR, "sample/yaml.yaml", True)
```
hanxiao, line 24: this is a completely misleading test, it tests nothing.
`_test_build` does not accept image as `str`, it accepts a `Docker Image Object`

hanxiao (commented):
this test is completely misleading and it is unclear what this PR is about.

PR #1365, opened 2020-11-26
fix: dont overwrite docker image names

cristianmtr:
:construction: 

As part of the Hub Updater, we want to make sure NOT to overwrite images with the same name.

### DIFF
--- a/tests/unit/docker/test_hub_build.py
+++ b/tests/unit/docker/test_hub_build.py
@@ -65,6 +65,39 @@ def test_hub_build_push():
     assert manifests[0]['name'] == summary['manifest_info']['name']
 
 
+def test_hub_build_push_push_again():
+    args = set_hub_build_parser().parse_args([os.path.join(cur_dir, 'hub-mwu'), '--push', '--host-info'])
+    summary = HubIO(args).build()
+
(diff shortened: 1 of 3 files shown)


hanxiao (approved):

tests/unit/docker/test_hub_build.py
```diff
@@ -65,6 +65,39 @@ def test_hub_build_push():
...
+    response = HubIO(args).list()
+    manifests = response
+
+    assert len(manifests) >= 1
+    assert manifests[0]['name'] == summary['manifest_info']['name']
+
+    # try and push same version again should fail
+    args = set_hub_build_parser().parse_args([os.path.join(cur_dir, 'hub-mwu'), '--push', '--host-info'])
+    summary = HubIO(args).build()
+    print(summary['is_build_success'])
```
nan-wang, line 98: An assertion for the exception is expected here.

cristianmtr: Is there a way to restore the PR and edit it? Or should I just create a new one?

JoanFM: create a new one better

PR #1367, opened 2020-11-27
fix: assert with var

JoanFM:


tests/integration/jinad/test_index_query_with_shards/test_integration.sh
```diff
@@ -56,7 +56,7 @@ docker-compose -f tests/integration/jinad/test_index_query/docker-compose.yml --
 
 EXPECTED_TEXT='"text:hey, dude"'
 
-if [ "$EXPECTED_TEXT" = "TEXT_MATCHED" ]; then
+if [ "$EXPECTED_TEXT" = "$TEXT_MATCHED" ]; then
```
deepankarm, line 59: It is always better to use bash instead of default sh, and `if [[ "$a" == "$b" ]]` instead of `[ "$a" = "$b" ]`

deepankarm (dismissed):

hanxiao (approved):

PR #1368, opened 2020-11-28
refactor(multimodal): refactor multimodal document and set

bwanglzu:
Blocked by #1372 when adding chunk level test cases.

PR #1372, opened 2020-11-29
refactor chunkset and matchset append method

bwanglzu:
blocking #1368 .
When refactoring `traverse_apply`, need a lot of chunk level test cases. While `ChunkSet.append` is not working when append document on `Sequence[Document]`. 
image

Why?
`docs_proto` was defined as `Union['RepeatedCompositeContainer', Sequence['Document']]`, the current implementation does not support operations on sequence of documents.

What has been changed?
1. Refactored `MatchSet` and `ChunkSet` , allow `append` operation on a list of document.
2. Add tests.

Didn't add unit test case to initialise a `DocumentSet` or `MatchSet` when `self._docs_proto`  is `RepeatedCompositeContainer`. (I was not able to initialise `RepeatedCompositeContainer` manually, while it has been covered when testing a driver).

### DIFF
--- a/jina/types/sets/chunk_set.py
+++ b/jina/types/sets/chunk_set.py
@@ -37,7 +37,9 @@ class ChunkSet(DocumentSet):
         chunk.set_attrs(parent_id=self._ref_doc.id,
                         granularity=self._ref_doc.granularity + 1,
                         **kwargs)
+
         if not chunk.mime_type:
             chunk.mime_type = self._ref_doc.mime_type
+        chunk.update_id()
 
         return chunk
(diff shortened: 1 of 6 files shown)


JoanFM:
Why does it exist thr class ChunkSet and is not merged with DocumentSet?

bwanglzu:
@JoanFM as far as know, in `ChunkSet` we overwrite the `append` method in `DocumentSet` to set the `parent_id` and `granularity`.

hanxiao:
> Why does it exist thr class ChunkSet and is not merged with DocumentSet?

ChunkSet and MatchSet are inherited from DocumentSet, I did it on purpose to abstract the logic of `add_chunk` and `add_match`

JoanFM:
Makes sense ! Thanks for clarification

hanxiao (approved):
👍

PR #1375, opened 2020-11-30
chore(style): update the Dockerfile style

hongchhe:
It's just an suggestion. I think this style is more readable than the old one.

### DIFF
--- a/Dockerfiles/alpinex.Dockerfile
+++ b/Dockerfiles/alpinex.Dockerfile
@@ -1,45 +1,45 @@
-FROM python:3.7.6-alpine
+FROM      python:3.7.6-alpine
 
-ARG VCS_REF
-ARG BUILD_DATE
-ARG JINA_VERSION
+ARG       VCS_REF
+ARG       BUILD_DATE
+ARG       JINA_VERSION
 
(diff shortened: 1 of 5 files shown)


bwanglzu:
are there any style guide or automatic formatting tools when writing a docker file?

bwanglzu:
I was looking at the docker [official style guide](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/), seems they did not use tabs or four spaces?

hongchhe:
> are there any style guide or automatic formatting tools when writing a docker file?

Sorry, there are no style guide. I just find from someone good example and I think it's better. Therefore , it's just a suggestion.

hongchhe:
> I was looking at the docker [official style guide](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/), seems they did not use tabs or four spaces.

The whitespace is just used for keeping the contents in line up and down and each command will take one line

bwanglzu:
@hongchhe get it, thanks. I'll let others to decide.

JoanFM:
Hello @hongchhe, Thank you very much for the suggestions, but we prefer to stick with the current version.

PR #1377, opened 2020-11-30
refactor: split pea logic into pea and runtime support

JoanFM:
**Changes introduced**

- [ ] Add **RuntimeSupport** class to remove complexity from **BasePea**. **RuntimeSupport** will spin a **BasePea** and ensure the connection with this Pea (both using `control_port` and `multiprocessing Events`

**Would like to have**

- [ ] Split **Remote** logic into mix-in classes to simplify them


jina/peapods/pea.py
```diff
@@ -7,22 +7,22 @@
...
 from ..enums import PeaRoleType, SkipOnErrorType
-from ..excepts import NoExplicitMessage, ExecutorFailToLoad, MemoryOverHighWatermark, DriverError, PeaFailToStart, \
+from ..excepts import RequestLoopEnd, NoExplicitMessage, ExecutorFailToLoad, MemoryOverHighWatermark, DriverError, \
     ChainedPodException
 from ..executors import BaseExecutor
-from ..helper import is_valid_local_config_source, typename
+from ..helper import is_valid_local_config_source
 from ..logging import JinaLogger
 from ..logging.profile import used_memory, TimeDict
 from ..proto import jina_pb2
```
theUnkownName, line 24: So at this level we are still using the `jina_pb2`?

JoanFM: At this level is still a little leftove (just checking a status code, I guess we could easily refactor that too), it should still be lower in the code.

JoanFM:
Will start over in another branch

PR #1380, opened 2020-12-01
feat(kvindexers): update method for binary pb indexer

florian-hoenicke:
Related ticket: 
https://jinaai.atlassian.net/browse/JINACORE-483

Adding update and delete operations for all key value indexers.


theUnkownName (dismissed):
Looks good!

jina/executors/indexers/keyvalue.py
```diff
@@ -75,6 +75,25 @@ def query(self, key: int) -> Optional[bytes]:
             with mmap.mmap(self.query_handler.body, offset=p, length=l) as m:
                 return m[r:]
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        for key, new_value in zip(keys, values):
+            pos_info = self.query_handler.header.get(key, None)
+            p, r, l = pos_info
```
theUnkownName, line 81: Maybe just something more explicit on what `p, r, l `are?

hanxiao: fun fact: when I wrote something like this (often intentionally), it's usually a sign this code is hardcore and at low-level, to remind you Data Structure & Algorithms 101

hanxiao (changes requested):
- this does not work, as discussed in https://jinaai.slack.com/archives/C018QC78L84/p1606823311131600?thread_ts=1606820889.125300&cid=C018QC78L84
- missing unit test

jina/executors/indexers/keyvalue.py
```diff
@@ -75,6 +75,25 @@ def query(self, key: int) -> Optional[bytes]:
...
+            with mmap.mmap(self.query_handler.body, offset=p, length=l) as m:
+                old_value = m[r:]
+                if old_value != new_value:
+                    # update
+                    l_end = len(new_value)  #: the length
+                    p_end = int(self._start / self._page_size) * self._page_size  #: offset of the page
+                    r_end = self._start % self._page_size  #
+                    self.write_handler.header.write(key, p_end, r_end, r_end + l_end)
+                    self.write_handler.body.write(new_value)
```
hanxiao, line 90: as far as i understand, this tries to write the updated values in the *same* position but with varied length, if the length is longer than the old one, then the next entry in the `body` is overflowed.

cristianmtr: Ah yes we're not getting the l, p, r associated with the last entry.

hanxiao: or is it simply appending, based on prev. `self._start`?

hanxiao: but then why isn't `self._start` updated though?

anyway, a unit test on visiting all entries after a couple of updates would be clear enough to reveal the problem

cristianmtr: Can you check the new version?

jina/executors/indexers/keyvalue.py
```diff
@@ -37,7 +37,8 @@ def close(self):
 
     def get_add_handler(self):
         # keep _start position as in pickle serialization
-        return self.WriteHandler(self.index_abspath, 'ab')
+        handler = self.WriteHandler(self.index_abspath, 'ab')
+        return handler
```
hanxiao, line 41: dont see the need of change here

jina/executors/indexers/keyvalue.py
```diff
@@ -69,12 +70,58 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
             # print(f'l: {l} p: {p} r: {r} r+l: {r + l} size: {self._size}')
 
     def query(self, key: int) -> Optional[bytes]:
+        # FIXME this shouldn't be required
+        if self.query_handler is None:
+            self.query_handler = self.get_query_handler()
```
hanxiao, line 75:
```suggestion

```

jina/executors/indexers/keyvalue.py
```diff
@@ -69,12 +70,58 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
...
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        new_values = []
+        new_keys = []
+        for key, new_value in zip(keys, values):
+            pos_info = self.query_handler.header.get(key, None)
+            if pos_info is not None:
+                p, r, l = pos_info
+                with mmap.mmap(self.query_handler.body, offset=p, length=l) as m:
+                    old_value = m[r:]
```
hanxiao, line 90: This is simply
```suggestion
            old_value = self.query(key)
```

jina/executors/indexers/keyvalue.py
```diff
@@ -69,12 +70,58 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
...
+                p = int(_start / self._page_size) * self._page_size  #: offset of the page
+                r = _start % self._page_size  #: the remainder, i.e. the start position given the offset
+                adder.header.write(
+                    np.array(
+                        (key, p, r, r + l),
+                        dtype=np.int64
+                    ).tobytes()
+                )
+                _start += l
+                adder.body.write(new_value)
```
hanxiao, line 109: isn't this self.add()?

cristianmtr: No, because we need the `adder` handler, and we need a custom `_start` position

jina/executors/indexers/keyvalue.py
cristianmtr: Yes, but it fails sometimes. Need to debug it more and figure out why

jina/executors/indexers/keyvalue.py
cristianmtr: Yep, was debugging something.

jina/executors/indexers/keyvalue.py
```diff
@@ -28,7 +28,7 @@ class ReadHandler:
         def __init__(self, path):
             with open(path + '.head', 'rb') as fp:
                 tmp = np.frombuffer(fp.read(), dtype=np.int64).reshape([-1, 4])
-                self.header = {r[0]: r[1:] for r in tmp}
+                self.header = {r[0]: None if np.array_equal(tmp[0][1:], (-1, -1, -1)) else r[1:] for r in tmp}
```
cristianmtr, line 31: @florian-hoenicke Not sure I understand what this is doing

cristianmtr: Checks if elements from `[1:]` in the first row are `(-1, ...)` ? What would that mean?

cristianmtr: Oh I see those are the markers of having been deleted. Can't you just delete the key from the header completely and then rewrite it? It's much smaller than the body so it shouldn't be a performance hindrance.

Or at least make that `(-1, ...)` a constant

cristianmtr, line 31: My comments got deleted somehow, writing again:

Cant you just rewrite the entire header? Much smaller than body, shouldnt be a perf hindrance.

Or at least make `(-1, ...)` a constant

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +31,48 @@ def validate(req):
...
+
+
+def test_binarypb_delete(tmpdir):
+    # FIXME how to open a specific file in a dir. os.path.join(tmpdir, file) doesn't work
+    with BinaryPbIndexer('pbdix2.bin') as idxer:
+        idxer.add([1, 2, 3], [b'oldvalue', b'same', b'random'])
+        idxer.save()
+        assert idxer.size == 3
+        assert idxer.query(1) == b'oldvalue'
+        idxer.delete(iter([1, 2]))
```
cristianmtr, line 73: Shouldn't `.save(` be required here, as after `.add(`?

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +31,48 @@ def validate(req):
...
+        assert idxer.query(1) == None
```
cristianmtr, line 74: `        assert idxer.size == 1` fails

jina/executors/indexers/keyvalue.py
```diff
@@ -67,14 +67,46 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
...
+        if self.query_handler is None:
+            self.query_handler = self.get_query_handler()
         pos_info = self.query_handler.header.get(key, None)
         if pos_info is not None:
             p, r, l = pos_info
             with mmap.mmap(self.query_handler.body, offset=p, length=l) as m:
                 return m[r:]
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        # check: hard fail (raises) on key not found
```
florian-hoenicke, line 82: @cristianmtr what about the sharding case, where we want to have an update on all shards having the key. I think, if  a shard don't have a key, it should silently ignore an update or delete request. I don't think there is a way of knowing which key is persistent in which shard from outside.

JoanFM:
Hey guys, fix the `commit-lint` problem asap (then rebasing becomes a mess).

tests/unit/executors/indexers/test_binary_indexer.py
cristianmtr: Yes, it passes, but I see your point

jina/executors/indexers/keyvalue.py
florian-hoenicke: Now, I created a constant for `(-1, -1, -1)`.

jina/executors/indexers/keyvalue.py
cristianmtr: I see what you mean. But isn't that something the tail pea should handle? It should catch these errors, right?

cristianmtr:
> Hey guys, fix the `commit-lint` problem asap (otherwise rebasing becomes a mess).

Done

jina/executors/indexers/keyvalue.py
florian-hoenicke: okay, then let's leave it like this for now.

jina/executors/indexers/keyvalue.py
```diff
@@ -67,14 +69,45 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
             self.write_handler.body.write(value)
             self._size += 1
             # print(f'l: {l} p: {p} r: {r} r+l: {r + l} size: {self._size}')
+        self.write_handler.flush()
 
     def query(self, key: int) -> Optional[bytes]:
+        if self.query_handler is None:
+            self.query_handler = self.get_query_handler()
```
nan-wang, line 76: Is this part necessary?

JoanFM: I think this is not needed, but the problem is that the `test` is doing something that we are not doing. in the same context  managed indexer we are indexing, saving and querying

JoanFM: It is not necessary and dangerous for now. Query handler and Write handler I think should not be opened at the same time.

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +29,78 @@ def validate(req):
...
+        'workspace': f'{tmpdir}',
+        'name': None,
+        'on_gpu': False,
+        'warn_unnamed': False,
+        'max_snapshot': 0,
+        'py_modules': None,
+        'pea_id': '{root.metas.pea_id}',
+        'separated_workspace': '{root.metas.separated_workspace}',
+    }
+    with BinaryPbIndexer('pbidx', metas=metas) as idxer:
```
JoanFM, line 49: this test should be changed:

1. Using `test_metas` fixture will give what u expect
2. What is 'pbidx'
3. For now, I think we cannot `add` and `query` in the same `context managed` indexer. You should close the context after the addition, and then open a new indexer for querying, and then yet another for updating, and yet another for later query.  (I think this holds true specially for VectorIndexers and I think also for BinaryPbIndexer)

JoanFM (changes requested):

jina/executors/indexers/keyvalue.py
JoanFM: I think the reason refers to the comments I added in the test

cristianmtr: It was removed, but it a symptom of another problem we discussed over zoom. basically we need to flip the mutex and close handlers between ops. This is now done in the `.update`

florian-hoenicke:
@JoanFM @hanxiao Is there anything left to do for this pull request?

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +30,111 @@ def validate(req):
         d.ClearField('embedding')
     with f:
         f.search(docs_no_embedding, output_fn=validate, override_doc_id=False)
+
+
+def test_binarypb_update(tmpdir):
```
JoanFM, line 35: use test_metas fixture in our conftest.py

JoanFM: This test is too long. Could you divide it in different ones? I see comments that suggest that different behaviors are being tested and I think is better if they are reflected in a separate test

JoanFM (changes requested):

jina/executors/indexers/keyvalue.py
```diff
@@ -75,6 +78,35 @@ def query(self, key: int) -> Optional[bytes]:
...
                 return m[r:]
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        # check: hard fail (raises) on key not found
+        missed = []
+        for key in keys:
+            if self.query_handler.header.get(key) is None:
+                missed.append(key)
+        if missed:
+            # FIXME get indexer name
```
bhavsarpratik, line 88: Anything left to fix here?

jina/executors/indexers/keyvalue.py
```diff
@@ -67,6 +69,7 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
             self.write_handler.body.write(value)
             self._size += 1
             # print(f'l: {l} p: {p} r: {r} r+l: {r + l} size: {self._size}')
```
bhavsarpratik, line 71: Remove the comment

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +30,111 @@ def validate(req):
...
+    # required for creating workspace
+    metas = {
```
bhavsarpratik, line 37: This can be a pytest fixture used for below tests broken into many tests.

bhavsarpratik (changes requested):

jina/executors/indexers/keyvalue.py
cristianmtr: Wasn't sure if there is a way to get the name of the indexer. What i'm logging here is the path to the file instead

jina/executors/indexers/keyvalue.py
cristianmtr: done

tests/unit/executors/indexers/test_binary_indexer.py
cristianmtr: sure

tests/unit/executors/indexers/test_binary_indexer.py
cristianmtr: yes

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -31,76 +32,85 @@ def validate(req):
...
-    }
-    with BinaryPbIndexer('pbidx', metas=metas) as idxer:
+def test_binarypb_add_and_update_not_working(tmpdir, test_metas):
+    with BinaryPbIndexer(metas=test_metas) as idxer:
+        idxer.add([11, 12], [b'eleven', b'twelve'])
+        idxer.save()
+        # FIXME `add` and `update` won't work in the same context
+        # since `.save` calls `.flush` on a closed handler
+        # and the handler needs to have been
+        # closed for us to allow querying in the `.update`
```
bhavsarpratik, line 81: Remove this if not needed

bhavsarpratik (approved):

jina/executors/indexers/keyvalue.py
```diff
@@ -75,6 +77,34 @@ def query(self, key: int) -> Optional[bytes]:
...
                 return m[r:]
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        # check: hard fail (raises) on key not found
+        missed = []
+        for key in keys:
+            if self.query_handler.header.get(key) is None:
+                missed.append(key)
+        if missed:
+            raise KeyError(f'Key(s) {missed} were not found in {self.save_abspath}')
```
bwanglzu, line 87: I guess you don't need to loop over all the keys and raise the error.
```python
for key in keys:
    if not self.query_handler.header.get(key):
        raise KeyError('...')
```

jina/executors/indexers/__init__.py
```diff
@@ -215,6 +215,12 @@ def query(self, key: Any) -> Optional[Any]:
         """
         raise NotImplementedError
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        raise NotImplementedError
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        raise NotImplementedError
```
bwanglzu, line 222: curious, why not `__setitem__` and `__del__`?

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +30,87 @@ def validate(req):
...
+        assert idxer.size == 3
+
+    with BaseIndexer.load(save_abspath) as idxer:
+        assert idxer.query(1) == b'newvalue'
+        assert idxer.query(2) == b'same'
+        assert idxer.query(3) == b'random'
+        assert idxer.query(99) is None
+
+
+def test_binarypb_add_and_update_not_working(tmpdir, test_metas):
```
bwanglzu, line 74: in `test_metas` we already set `tmpdir`, so you don't have to use the `tmpdir` fixture anymore

bwanglzu (changes requested):

jina/executors/indexers/__init__.py
florian-hoenicke: Nice catch! Actually, I don't see any argument for not using them. Changing this could be done in another ticket.
@maximilianwerk Do you know why we don't use dunder methods here?

jina/executors/indexers/keyvalue.py
florian-hoenicke: Nice point. True, all missed update-keys are collected in memory. As a tradeoff we get have a better error handling. 
Actually, we have all keys in memory already (ReadHandler.header). Since keys are expected to be much smaller than values, we decided to accept the tradeoff.

tests/unit/executors/indexers/test_binary_indexer.py
florian-hoenicke: thanks, I removed the tmpdir parameter

cristianmtr (approved):
LGTM now. Any other feedback? We should merge this

jina/executors/indexers/keyvalue.py
cristianmtr: We want to log them all in one go

JoanFM (changes requested):
We cannot merge this, you have done some updates to the `hub` submoudle that should not be there.

cristianmtr:
> We cannot merge this, you have done some updates to the `hub` submoudle that should not be there.

Not sure how that happened, but removed. Should I squash the commits before merging?

JoanFM (approved):

jina/executors/indexers/__init__.py
maximilianwerk: Currently, `__getitem__` is just calling `query` in the `BaseKVIndexer`. Following this approach, we would need to implement `__setitem__` with either calling `add` or `update`. Since we bind the CRUD operations even to dedicated request objects, it would be weird to first make an explicit update call, and then convert it to a `__setitem__` which needs to figure out again, whether it is a `add` or `update`. Anyhow, for convenience, we should add the `__del__` and `__setitem__` later on.

jina/executors/indexers/keyvalue.py
```diff
@@ -75,6 +77,34 @@ def query(self, key: int) -> Optional[bytes]:
             with mmap.mmap(self.query_handler.body, offset=p, length=l) as m:
                 return m[r:]
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        # check: hard fail (raises) on key not found
+        missed = []
+        for key in keys:
+            if self.query_handler.header.get(key) is None:
```
maximilianwerk, line 84: Can you please add an integration test for two updates calls in a row on the same `idxer` object? I'd expect this line to fail then. If that is intended, then write a test, which expects to fail in the second `update` call in order to document, that double update on the very same `idxer` object is not possible.

tests/unit/executors/indexers/test_binary_indexer.py
bwanglzu: @florian-hoenicke great! while I see another two , can you also remove them?

bwanglzu (dismissed):
lgtm if you can remove the unused tmpdirs.

jina/executors/indexers/keyvalue.py
cristianmtr: You mean like this https://github.com/jina-ai/jina/blob/e318088a30897ba3ca538de8e0f5f471af92688c/tests/unit/executors/indexers/test_binary_indexer.py#L119 ? Seems to pass

cristianmtr:
> lgtm if you can remove the unused tmpdirs.

Done, removed

bwanglzu (approved):
LGTM!

jina/executors/indexers/keyvalue.py
maximilianwerk: Yes exactly. Even better.

jina/executors/indexers/keyvalue.py
```diff
@@ -66,7 +68,7 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
             self._start += l
             self.write_handler.body.write(value)
             self._size += 1
-            # print(f'l: {l} p: {p} r: {r} r+l: {r + l} size: {self._size}')
+        self.write_handler.flush()
```
JoanFM, line 71: shoul we maybe have a lazy flush? and flush only if we need the guarantee? or it feels like early optimization?

tests/unit/executors/indexers/test_binary_indexer.py
```diff
@@ -30,3 +30,105 @@ def validate(req):
...
+        assert idxer.query(2) == b'same'
+        assert idxer.query(3) == b'random'
+        assert idxer.query(99) is None
+
+
+def test_binarypb_add_and_update_not_working(test_metas):
+    with BinaryPbIndexer(metas=test_metas) as idxer:
+        idxer.add([11, 12], [b'eleven', b'twelve'])
+        idxer.save()
+        # FIXME `add` and `update` won't work in the same context
```
JoanFM, line 78: this may be fixed with the lazy flush?

JoanFM: I think add and update should be supported; if we merge this please create a ticket to keep track of this

cristianmtr: Yes, I will

cristianmtr: https://github.com/jina-ai/jina/issues/1413

jina/executors/indexers/keyvalue.py
cristianmtr: Opened an issue https://github.com/jina-ai/jina/issues/1413

PR #1385, opened 2020-12-02
feat: add multimodal set

JoanFM:


bwanglzu (approved):
lgtm, I'll figure out a better way to ensure a callback is called (if exist)

jina/types/sets/document_set.py
```diff
@@ -88,3 +88,21 @@ def build(self):
 
     def sort(self, *args, **kwargs):
         self._docs_proto.sort(*args, **kwargs)
+
+
+class MultiModalDocumentSet(DocumentSet):
```
nan-wang, line 93: @JoanFM We need a consistent naming here. `MultiModalDocumentSet` and `MultimodalDocument` is not a good choice.

JoanFM: Oh yes I will change it, what is preferred?

nan-wang: Personally,  `MultimodalDocumentSet ` is prefered because `multimodal` is one word.

maximilianwerk: In case, we might have a `CrossModalDocumentSet` in the future, I'd prefer `MultiModalDocumentSet`.

JoanFM: I think there will not be a CrossModalDocumentSet

JoanFM: Opened a PR to fix this #1390

tests/integration/multimodal/test_multimodal_parallel.py
```diff
@@ -80,8 +80,9 @@ def multimodal_all_types_documents():
     return docs
 
 
-def test_multimodal_all_types_parallel(multimodal_all_types_documents):
+def test_multimodal_all_types_parallel(multimodal_all_types_documents, is_callback_called):
     def validate_response(resp):
+        is_callback_called._callback_called = True
```
maximilianwerk, line 85: I don't see, why this should be better than `global`. I believe `global` itself is better, since it is explicit and not implicit.

maximilianwerk, line 85: This is overkill and not the right way. We should use the build in and standardized way of doing it. There are two ways. Either we use `unittest.mock.Mock` like:
```
response_mock = Mock(wraps=validate_response)
index_gt_flow.index(input_fn=multimodal_all_types_documents, output_fn=response_mock)
```
Or we add `pytest-mock` as a dependency and do:
```
def test_multimodal_all_types_parallel(multimodal_all_types_documents, mocker):
response_mock = mocker.Mock(wraps=validate_response)
index_gt_flow.index(input_fn=multimodal_all_types_documents, output_fn=response_mock)
```
I'd prefer using unittest.mock.Mock.

JoanFM: I am working on it in general in another PR will try to implement ur suggestions, is there a way to have it easily included in the tests?

maximilianwerk: It is exactly as easy as I wrote. just mock the callback function and test with `mock.assert_called()` if the callback was called.

JoanFM: yes, yes. I mean something more automatic without much effort, But I think there isnot. About to change in #1391

PR #1391, opened 2020-12-03
test: first batch of ensuring callback called

JoanFM:


JoanFM:
@maximilianwerk @bwanglzu I am not sure why a test is breaking.

@maximilianwerk wouldn't global pose problems when validate is a function shared by many tests?

JoanFM:
It works on local

JoanFM:
I think this might be caused because the folder may not be reachable from the github action?

extra-requirements.txt
```diff
@@ -55,6 +55,7 @@ cookiecutter:               hub, devel, cicd
 pytest:                     test
 pytest-xdist==1.34.0:       test
 pytest-timeout:             test
+pytest-mock:                test
```
maximilianwerk, line 58: Since `unittest` is natively installed with python, I'd rather do the 
```
from unittest.mock import Mock
...
Mock(wrap=...)
```
instead of using pytest-mock. Apart from that, I like the PR.

JoanFM: I chose the `pytest` version just because we are using all the `pytest` ecosystem, but no strong opinion about this.

@bwanglzu any consideration about this?

tests/integration/mime/test_mime.py
```diff
@@ -28,48 +30,62 @@ def input_fn3():
...
     with f:
-        f.index(input_fn=input_fn)
+        f.index(input_fn=input_fn, output_fn=response_mock)
 
+    response_mock.assert_called()
+    response_mock_2 = mocker.Mock()
     f = Flow().add(uses='- !Buffer2URI | {mimetype: png, base64: true}')
     with f:
-        f.index(input_fn=input_fn)
+        f.index(input_fn=input_fn, output_fn=response_mock_2)
```
rutujasurve94, line 43: Can we `parametrise` this test by declaring pairwise [`inputfn`, `responsemock`] list at the top of the function?

extra-requirements.txt
rutujasurve94: `pytest-mock` is a good choice as we simply inject `mocker` in the argument and inside the test body we can explicitly define return values rather than having to patch on top of the function call (if we use Mock) and explicitly injecting the mocked patches - increases the number of arguments making the test look clumsy.

tests/integration/mime/test_mime.py
JoanFM: It is a sequential test, so is not either or, but one after the other.

bwanglzu (commented):
it looks good to me, while I didn't propose to use mock because I don't know if it's good to mock in integration test.

JoanFM:
> it looks good to me, while I didn't propose to use mock because I don't know if it's good to mock in integration test.

You made me think and I tested, and u are right, it makes the error message very wordy if a different assertion fails!

JoanFM:
@maximilianwerk with this solution, when a test fails, the output message is:

```
self = <Mock id='139638584934544'>, name = '__name__'

    def __getattr__(self, name):
        if name in {'_mock_methods', '_mock_unsafe'}:
            raise AttributeError(name)
        elif self._mock_methods is not None:
            if name not in self._mock_methods or name in _all_magics:
                raise AttributeError("Mock object has no attribute %r" % name)
        elif _is_magic(name):
>           raise AttributeError(name)
E           AttributeError: __name__

/usr/lib/python3.7/unittest/mock.py:600: 
```
Wouldn't we prefer the ugly `fixture`? The global variable will be hard to use when the `validate` callback function is shared by more than one test case which is quite a common thing in our codebase.

maximilianwerk:
> it looks good to me, while I didn't propose to use mock because I don't know if it's good to mock in integration test.

I think it is, since we use this mock only exactly as the decorator Joan implemented before. There nothing actually hidden away, but just tracked, whether the function is called. And the verbose Exception is OK in my opinion. If something goes wrong in a test, you anyhow need a moment to check what happens.

JoanFM:
> > it looks good to me, while I didn't propose to use mock because I don't know if it's good to mock in integration test.
> 
> I think it is, since we use this mock only exactly as the decorator Joan implemented before. There nothing actually hidden away, but just tracked, whether the function is called. And the verbose Exception is OK in my opinion. If something goes wrong in a test, you anyhow need a moment to check what happens.

Okey, as long as we are awar of the tradeoff, good for me!

maximilianwerk:
> @maximilianwerk with this solution, when a test fails, the output message is:
> 
> ```
> self = <Mock id='139638584934544'>, name = '__name__'
> 
>     def __getattr__(self, name):
>         if name in {'_mock_methods', '_mock_unsafe'}:
>             raise AttributeError(name)
>         elif self._mock_methods is not None:
>             if name not in self._mock_methods or name in _all_magics:
>                 raise AttributeError("Mock object has no attribute %r" % name)
>         elif _is_magic(name):
> >           raise AttributeError(name)
> E           AttributeError: __name__
> 
> /usr/lib/python3.7/unittest/mock.py:600: 
> ```
> 
> Wouldn't we prefer the ugly `fixture`? The global variable will be hard to use when the `validate` callback function is shared by more than one test case which is quite a common thing in our codebase.

I agree with not using the `global` variable. Anyhow, I prefer using the `mock` approach, since it is the pytest standard one. And for me the output is like
```
        mock = mocker.Mock()
        my_func(mock, False)
>       mock.assert_called()
E       AssertionError: Expected 'mock' to have been called.

tests/unit/test_callback.py:18: AssertionError
```

JoanFM:
> > @maximilianwerk with this solution, when a test fails, the output message is:
> > ```
> > self = <Mock id='139638584934544'>, name = '__name__'
> > 
> >     def __getattr__(self, name):
> >         if name in {'_mock_methods', '_mock_unsafe'}:
> >             raise AttributeError(name)
> >         elif self._mock_methods is not None:
> >             if name not in self._mock_methods or name in _all_magics:
> >                 raise AttributeError("Mock object has no attribute %r" % name)
> >         elif _is_magic(name):
> > >           raise AttributeError(name)
> > E           AttributeError: __name__
> > 
> > /usr/lib/python3.7/unittest/mock.py:600: 
> > ```
> > 
> > 
> > Wouldn't we prefer the ugly `fixture`? The global variable will be hard to use when the `validate` callback function is shared by more than one test case which is quite a common thing in our codebase.
> 
> I agree with not using the `global` variable. Anyhow, I prefer using the `mock` approach, since it is the pytest standard one. And for me the output is like
> 
> ```
>         mock = mocker.Mock()
>         my_func(mock, False)
> >       mock.assert_called()
> E       AssertionError: Expected 'mock' to have been called.
> 
> tests/unit/test_callback.py:18: AssertionError
> ```

my point is when an inner assertion fails, not when it is not called

maximilianwerk:
> my point is when an inner assertion fails, not when it is not called

While that is true, you cannot circumvent the stack-trace of the flow when assertion deep inside. So the output will be equally ugly with the decorator and I'd then default to mock.

nan-wang (approved):
LGTM👍

tests/integration/level_depth/test_search_different_depths.py
```diff
@@ -45,13 +45,13 @@ def validate_granularity_1(resp):
...
+    with Flow.load_config('flow-query.yml') as search_flow:
         search_flow.search(
             input_fn=search_data,
-            output_fn=validate_granularity_1,
+            output_fn=response_mock,
             callback_on='body',
         )
 
     del os.environ['JINA_TEST_LEVEL_DEPTH_WORKSPACE']
+    response_mock.assert_called()
```
maximilianwerk, line 57: I have doubled checked and think we are doing it wrong. It should be:
```
response_mock = mocker.Mock()
...
    output_fn=response_mock

for arguments in response_mock.call_args_list:
    validate_granularity_1(*arguments)
```
With this we get rid of the assertion stack trace, since the assert happens outside of the flow.

JoanFM: What if the `callback` is called more than once? When you provide more than one requests in input?

PR #1392, opened 2020-12-03
Feat add random

theUnkownName:


JoanFM (commented):
Just to make sure, the idea is to chanhe how document ID is computed and used but not to add a new field in the document class.

hanxiao:
sorry i'm going to reject this PR. it's not in the correct direction. @JoanFM is right

maximilianwerk:
> sorry i'm going to reject this PR to save everybody time, it's not in the correct direction. @JoanFM is right
> 
> change protobuf for new fields/modify old fields, not the Document type, `Document` type is a "View" to Protobuf object. For detailed designs see my blog post

I don't get this comment.

The point of this PR is to remove the hashing as a document's id. So you can either provide a `custom_id` to the `Document` or if non is provided, generate a random id. In either case, this is consecutively forwarded to `self._document.id` via `self.id` via a `setter` method.

Furthermore, there is no point in having the `Document` context-managed, since there is no point in calculating the id as a hash at the end.

Anyhow, the `self.random_id` is a fragment of previous thoughts, which needs to be completely removed again.

JoanFM:
I agree with Max, I do not think we need to close this, also since it is just a `Draft`. My comment was just a warning in case we were heading in the wrong direction, but it was easily handled since it was in a very early stage

hanxiao:
- protobuf level change required:
   - add `str identity`
   - redoc the comment of `id` in protobuf to as "the hash value of the document in hexstring"
   - or, reuse `id` for your `identity`, add a new `str hash_val`
- add/modify interface to `Document` accordingly

Principle: `Document` as well as all other Jina Data Type do not store any extra data. They are stateless, they rely on the Protobuf object as the "storage layer", they are building a "View" to that storage layer.

maximilianwerk:
>     * protobuf level change required:
>       
>       * add `str identity`
>       * redoc the comment of `id` in protobuf to as "the hash value of the document in hexstring"
>       * or, reuse `id` for your `identity`, add a new `str hash_val`
> 
>     * add/modify interface to `Document` accordingly
> 
> 
> Principle: `Document` as well as all other Jina Data Type do not store any extra data. They are stateless, they rely on the Protobuf object as the "storage layer", they are building a "View" to that storage layer.

The hash will mostly be used for checking, if we saw a document with the same content earlier. Thus, it's use is (at least currently) limited to a single use-case. What about the following suggestion:

* redoc the comment of `id` in protobuf to "the unique id of a document" and reuse it for the identity.
* calculate the hash inside the document with a `get_hash` function always on the fly and don't store it. This also makes it really explicit to the user of `get_hash` that all latest document changes are incorporated into the `hash`.

One question that remains from my side: Currently the `UniqueId` supports all the types of `str`, `bytes` and `int`. This needs to stay for the different use-cases in different drivers/executors, right?

theUnkownName:
I removed all `random_id` and set the `custom_id` as discussed. But there are some questions left:

Should we keep the context manager of the Document or remove it? We thought about removing it in last discussions we had because now the id is set either manually, or if none provided, a random one is generated. If we remove it I need to refactor some tests that use the context manager, but wanted to check first with you if this makes sense or not.

Second, if the id is set manually, what you think the type should be? int/string?

maximilianwerk:
> I removed all `random_id` and set the `custom_id` as discussed. But there are some questions left:
> 
> Should we keep the context manager of the Document or remove it? We thought about removing it in last discussions we had because now the id is set either manually, or if none provided, a random one is generated. If we remove it I need to refactor some tests that use the context manager, but wanted to check first with you if this makes sense or not.
> 
> Second, if the id is set manually, what you think the type should be? int/string?

For the second: I think it doesn't matter, as long as we use `UniqueId` to convert one into the other.

tests/unit/types/test_documentset.py
```diff
@@ -1,11 +1,13 @@
 from jina import Document
 from jina.types.sets import DocumentSet
 from tests import random_docs
+from tests import random_docs_new_api
```
bwanglzu, line 4: please merge master branch to your branch, `random_docs_new_api ` has been merged back to `random_docs` in #1356

maximilianwerk: Just remark: I prefer rebasing, since it makes a cleaner commit history. Anyhow, since we mostly "squash and merge" the difference is minor.

maximilianwerk:
We need to still be able to create `Document`s via the `with` statement. Otherwise, we will break unit tests in the hub. So let's keep this for now and let's now refactor not-yet refactored unit tests.

hanxiao:
@theUnkownName @JoanFM @maximilianwerk I think this branch suffers from a wrong merge or git operations, it suddenly becomes too large:
image

hanxiao (changes requested):
need a clean rebase before continue. This PR is too large and very likely due to some bad git operations.

maximilianwerk:
Seems like it. I'll fix it tomorrow.

maximilianwerk:
I'll close this PR, create a new branch from master and cherry-pick the right commits.

PR #1393, opened 2020-12-03
fix: change numpy metric cosine

bhavsarpratik:
Changed default `NumpyIndexer` metric to `cosine` for better `default`


nan-wang (approved):
LGTM👍

PR #1396, opened 2020-12-04
fix: minor thing in test shards

cristianmtr:


tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -181,4 +181,4 @@ def test_numpy_indexer_empty_data(batch_size, compress_level, test_metas):
         assert isinstance(indexer, NumpyIndexer)
         idx, dist = indexer.query(query, top_k=4)
         assert idx == None
```
JoanFM, line 183: assert not idx is more elegant?

cristianmtr: That can also mean `False`, no?

JoanFM: fair enough!

JoanFM (approved):

PR #1398, opened 2020-12-04
refactor: clean env variable when closing pea

JoanFM:
**Changes introduced**
I have seen weird behavior where it seemed that some process may be recycled (does that make sense)?


bhavsarpratik (approved):

hanxiao (changes requested):
let's put this on hold as we clarify the reason. Recycle process is not possible

JoanFM:
image

This is what raised my eyebrow

JoanFM:
> let's put this on hold as we clarify the reason. Recycle process is not possible

I understood it now, We run some tests with backend `thread` and so the `JINA_POD_NAME` gets in the main process, and I guess next test starts flow in the same process and from then on, the logs look quite nasty

hanxiao:
so close this PR?

JoanFM:
i think it shows better logs when testing

PR #1403, opened 2020-12-07
fix: check jina version in hub modules when pushing

cristianmtr:
+ Add `--no-overwrite` option to not allow pushing image with same module version and jina version

### DIFF
--- a/jina/docker/hubio.py
+++ b/jina/docker/hubio.py
@@ -16,7 +16,7 @@ from .helper import credentials_file
 from .hubapi import _list, _register_to_mongodb, _list_local
 from ..clients.python import ProgressBar
 from ..enums import BuildTestLevel
-from ..excepts import DockerLoginFailed, HubBuilderError, HubBuilderBuildError, HubBuilderTestError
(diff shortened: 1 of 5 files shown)


jina/docker/hubio.py
```diff
@@ -172,37 +172,34 @@ def push(self, name: str = None, readme_path: str = None, build_result: Dict = N
...
+                                  f'please use "jina hub build" to build the image first '
+                                  f'before pushing.')
+
+        if build_result:
+            if build_result.get('is_build_success', False):
+                _register_to_mongodb(logger=self.logger, summary=build_result)
+            if build_result.get('details', None) and build_result.get('build_history', None):
+                self._write_slack_message(build_result, build_result['details'], build_result['build_history'])
 
-        except Exception as ex:
```
deepankarm, line 204: why did we remove the `try..except` here?

cristianmtr: because the `raise` wasn't propagated. It needs to be propagated in order to be caught and dealt with by the handler. For now it's only used in the test, with `pytest.raises ...`

deepankarm: We need to find a different way to propagate the raise without removing the overall `Exception` as this is one of the errors that signifies an error in hub-builder.

deepankarm (approved):

Issue #1337, opened 2020-11-24
🚨 BREAKING CHANGES 0.8.0 -> 0.9.0


nan-wang:
This issue is created to keep track of breaking changes between `0.8.0` and `0.9.0` (scheduled release on Dec. 21, 2020)

## Template

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1343 | Yes|  |

<table stype="margin-left:auto;margin-right:auto;">
<tr>
<td> </td>
<td> v0.8.0 </td>
<td> v0.9.0 </td>
</tr>
<tr>
<td> </td>
<td> <sub> 

```python
from jina import Document()
with Document() as d:
    c = Document(id=f'1:0>16')
    d.chunks.append(c)

with Document() as d:
     c = d.chunks.append()
     c.id = f'1:0>16'    
```
</sub> </td>
<td> <sub> 

```python
from jina import Document()
with Document() as d:
    c = d.chunks.add_chunks()
    c.id = f'1:0>16'  
```

</sub> </td>
</tr>
</table>

----
| PR |  Rolled Out? | Override |
| --- | --- | --- |
|#1379 | Yes |  |

- `exit-no-dump` arg removed from `CLI` arg. 
-  If you had `exit-no-dump` to `True` now you will need to have `read_only` to `True`. 

----
| PR |  Rolled Out? | Override |
| --- | --- | --- |
|#1406  | Yes |  #1418 |
| #1418 | Yes | |

- `merge_all` executor alias removed, use `_merge_chunks` or `_merge_matches` to decide which dimension to merge from.
- If you were using `merge_all` inside `uses_after` of an indexer **Pod**, switch to `_merge_matches`. If you were using it inside `uses_before` of a **Pod** collecting `chunks` from different **Flow** branches, use `_merge_chunks`

---
| PR |  Rolled Out? | Override |
| --- | --- | --- |
|#1494 | Yes |  |

- `CollectMatches2DocRankDriver` has been renamed to `AggregateMatches2DocRankDriver`.

---
| PR |  Rolled Out? | Override |
| --- | --- | --- |
|#1495 | Yes  | |

- Refactor the YAML file parsing backend from `ruamel.yaml` to `pyyaml`

---

## Template
| PR |  Rolled Out? | Override |
| --- | --- | --- |
| |  | |

<table stype="margin-left:auto;margin-right:auto;">
<tr>
<td> </td>
<td> v0.8.0 </td>
<td> v0.9.0 </td>
[...]

PR #1406, opened 2020-12-07
fix: add _merge_matches and _merge_chunks syntax sugar

### LINKED ISSUE #1404: top_k is not working on multiple shards
**Describe the bug**
Problem is on query time. when i've multiple shards (>1) and if i set top_k == (total len of docs) on quey time than it's not returning all docs result. i also passing `polling='all'` and `uses_reducing='_merge_all'` on query flow still not working.

**Environment**
```
jina                          0.8.3
jina-proto                    0.0.76
jina-vcs-tag                  (unset)
libzmq                        4.3.3
pyzmq                         1.19.4
protobuf                      3.14.0
proto-backend                 cpp
grpcio                        1.33.2
ruamel.yaml                   0.16.12
python                        3.7.9
platform                      Linux
platform-release              5.4.0-56-generic
platform-version              #62~18.04.1-Ubuntu SMP Tue Nov 24 10:07:50 UTC 2020
architecture                  x86_64
processor                     x86_64
jina-resources                /media/kavan/WORK/Job/jina_search_engine/venv/lib/python3.7/site-packages/jina/resources
JINA_ARRAY_QUANT              (unset)
JINA_BINARY_DELIMITER         (unset)
JINA_CONTRIB_MODULE           (unset)
JINA_CONTRIB_MODULE_IS_LOADING(unset)
JINA_CONTROL_PORT             (unset)
JINA_DB_COLLECTION            (unset)
JINA_DB_HOSTNAME              (unset)
JINA_DB_NAME                  (unset)
JINA_DB_PASSWORD              (unset)
JINA_DB_USERNAME              (unset)
JINA_DEFAULT_HOST             (unset)
JINA_DISABLE_UVLOOP           (unset)
JINA_EXECUTOR_WORKDIR         (unset)
JINA_FULL_CLI                 (unset)
JINA_IPC_SOCK_TMP             (unset)
JINA_LOG_CONFIG               (unset)
JINA_LOG_NO_COLOR             (unset)
JINA_POD_NAME                 (unset)
JINA_PROFILING                (unset)
JINA_RANDOM_PORTS             (unset)
JINA_SOCKET_HWM               (unset)
JINA_TEST_GPU                 (unset)
JINA_TEST_PRETRAINED          (unset)
JINA_VCS_VERSION              (unset)
```

JoanFM:
Fixes #1404


jina/drivers/reduce.py
```diff
@@ -20,7 +20,7 @@ class ReduceAllDriver(BaseRecursiveDriver):
         It uses the last request as a reference.
     """
 
-    def __init__(self, traversal_paths: Tuple[str] = ('c',), *args, **kwargs):
+    def __init__(self, traversal_paths: Tuple[str] = ('m',), *args, **kwargs):
```
cristianmtr, line 23: Would this affect other examples? I see some failing unit tests and integration tests (and running them on master works)

cristianmtr (changes requested):

jina/drivers/reduce.py
JoanFM: It will affect whenever a `ReduceAllDriver` is used that is not used to accumulate `matches`. So either we change this or we change the `_merge_all` in the resources, which I will do now.

JoanFM: now u can see the alternative

JoanFM: We could consider in the future adding different `aliases` depending on the reduction to be done, but now this would be the best option

jina/resources/executors._merge_all.yml
```diff
@@ -5,6 +5,8 @@ metas:
 requests:
   on:
     [SearchRequest, TrainRequest, IndexRequest]:
-      - !ReduceAllDriver {}
+      - !ReduceAllDriver
+        with:
+          traversal_paths: ['m']
```
cristianmtr, line 10: So for me ideally I'd like to minimize the number of times we have exceptions / configuration changes at the yaml file. These should be available to be made in module/examples etc, but not be part of the core resources, I'd say. These should be as easy to read as possible, with changes pushed towards the code itself

jina/drivers/reduce.py
cristianmtr: I agree

jina/resources/executors._merge_all.yml
JoanFM: the point is flexibility, if we want to merge `matches` it will be `matches` (but sometimes one may want to merge other things). This is some work we need to make to ease things to the user, but at this point, we need to make a decision

JoanFM: let me do something here

cristianmtr: So what is the *sensible* default here? I'd say matches.

jina/drivers/reduce.py
JoanFM: what about this new alias?

JoanFM: will need to add a test

cristianmtr (approved):
Would a quick test be needed?

JoanFM:
> Would a quick test be needed?

Added some changes in some tests to validate

jina/resources/executors._merge_chunks.yml
```diff
@@ -0,0 +1,10 @@
+!BaseExecutor
+with: {}
+metas:
+  name: merge
+requests:
+  on:
+    [SearchRequest, TrainRequest, IndexRequest]:
+      - !ReduceAllDriver {}
```
maximilianwerk, line 8: For the sake of correctness, this should be
```
- !ReduceAllDriver
  with:
    traversal_paths: ['c']
```
and not `['c', 'r']`.

JoanFM: it is only c by definition of ReduceAllDriver

maximilianwerk (approved):
LGTM

JoanFM:
@nan-wang does this deserve a place in #1337 ?

nan-wang:
> @nan-wang does this deserve a place in #1337 ?

@JoanFM I saw `_merge_all` is still there. Do we plan to remove it?

JoanFM:
> > @nan-wang does this deserve a place in #1337 ?
> 
> @JoanFM I saw `_merge_all` is still there. Do we plan to remove it?

I would eventually remove

nan-wang:
Gotcha, then it is not a breaking change for the time being

PR #1407, opened 2020-12-07
ci: add test jinad checks for pr on master

JoanFM:


.github/workflows/test-jinad.yml
```diff
@@ -2,10 +2,8 @@ name: Test jinad
 
 on:
   pull_request:
-    #branches:
-    #  - master
-    branches-ignore:
-      - '**'
+    branches:
```
cristianmtr, line 5: Is this related to the problem we were having with jinad tests failing? Are they ok now?

JoanFM: they are almost ready

hanxiao (approved):

PR #1408, opened 2020-12-07
fix(gateway): event loop for rest server

deepankarm:


JoanFM (commented):
No tests can be added to validate this?

deepankarm:
> No tests can be added to validate this?

I need to think about tests for this, as this is complex. will work on this, while solving issues with pytest.asyncio

JoanFM (approved):

PR #1410, opened 2020-12-07
refactor(client): refactor async client

hanxiao:


jina/clients/__init__.py
```diff
@@ -43,111 +42,48 @@ def py_client(**kwargs) -> 'PyClient':
         to perform `index`, `search` or `train`, py_client needs to be awaited, as it is a coroutine
 
     """
+
+    warnings.warn('this method is depreciated after 0.8.3, use "py_client" instead', DeprecationWarning)
```
deepankarm, line 46: why can't we remove this?

hanxiao: unit tests still use them, new tests are missing

PR #1414, opened 2020-12-08
feat: add optimizer

bhavsarpratik:
Optimiser and discoverer for hyperparameters


jina/optimize/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+from jina.flow import Flow
+from jina.helper import colored
+from jina.logging import default_logger as logger
+from type import IntegerParameter, load_optimization_parameters
+
+
+class FlowRunner:
+    def __init__(
+        self,
+        index_document_generator,
```
JoanFM, line 17: Why would I need to pass query_document_generator if I do not want to evaluate a query flow? It feels that this class is very specialized in a single way of evaluating.

jina/optimize/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+        query_document_generator,
+        index_batch_size,
+        query_batch_size,
+        env_yaml=None,
+        workspace_env="JINA_WORKSPACE",
```
JoanFM, line 22: why do not you get the workspace directly instead of the name of an environment variable?

jina/optimize/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+        overwrite_workspace=False,
+    ):
+
+        self.index_document_generator = index_document_generator
```
JoanFM, line 26: why do we have generators and batch_sizes as members of the class and not passed to the `run` methods. Remember generators can not be reused for iteration again.

jina/optimize/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+        self.index_document_generator, index_document_generator = tee(
+            self.index_document_generator
+        )
+
+        with Flow.load_config(index_yaml) as f:
+            f.index(index_document_generator, batch_size=self.index_batch_size)
+
+    def run_querying(self, query_yaml, callback):
+        self._load_env()
+        self.query_document_generator, query_document_generator = tee(
```
JoanFM, line 80: what is tee doing?

jina/optimize/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+        all_params = {**self.flow_runner.env_parameters, **study.best_trial.params}
+        yaml.dump(all_params, open(self.best_config_filepath, "w"))
+        logger.info(colored(f"Number of finished trials: {len(study.trials)}", "green"))
+        logger.info(colored(f"Best trial: {study.best_trial.params}", "green"))
+        logger.info(colored(f"Time to finish: {study.best_trial.duration}", "green"))
+
+    def optimize_flow(self, n_trials, direction="maximize", seed=42):
+        import optuna
+
+        sampler = optuna.samplers.TPESampler(seed=seed)
```
JoanFM, line 228: what other samplers are available?

JoanFM (changes requested):
Some minor comments.

I would also call the module `optimizer` and not `optimize`. And would have more split of classes in different files. In _init_ only the main interface that the user should use should be seen

hanxiao:
Always remember to add tests. No test no merge

jina/optimize/__init__.py
bhavsarpratik: I will keep them `None`

jina/optimize/__init__.py
bhavsarpratik: This allows us to use ENV variable which is the current general practise. We don't need it for optimizer as new yamls are generated for every trial with value of workspace instead of ENV variable. I was thinking of making flow runner a general class to be used for any flows and hence kept this ENV support.

jina/optimize/__init__.py
bhavsarpratik: `tee` is used to make a generator copy and rerun generators

jina/optimize/__init__.py
JoanFM: The fact that is general practice does not mean in my opinion it should be used inside the code. One thing is what the client may provide and the other is how we use it. If we need to facilitate things to the user I think it should be in a different layer not in this core object.

jina/optimize/__init__.py
bhavsarpratik: There are many. I will keep an arg for this

jina/optimize/__init__.py
JoanFM: I still do not think we should own a generator.

- Is it a generator or a generator function? if it is a generator it will only be available once, so owning something that will not be useful during the instance lifetime is not good

- Is the generator something that the user needs to know when instantiating the FlowRunner?

jina/optimize/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+from .parameters import IntegerParameter, load_optimization_parameters
+
+
+class FlowRunner:
+    def __init__(
+        self,
+        index_document_generator=None,
+        query_document_generator=None,
+        index_batch_size=None,
+        query_batch_size=None,
```
maximilianwerk, line 20: Perhaps we can abstract a flow run with kind of a `FlowCompositions`  which we define as `namedtuple('FlowComposition', ['yaml_file', 'document_generator', 'batch_size', 'callback', 'target_endpoint', 'overwrite_workspace'])`, where `target_endpoint` defines, whether we call the `search` or `index` function of the flow. Then you could define your flow runs in a consecutive order and let the FlowRunner just run whichever list of flows you hand over.

maximilianwerk: Ah without the `yaml_file` most probably.

jina/optimize/__init__.py
bhavsarpratik: It is a generator. `tee` is used to make a copy of the generator so that we can rerun generators. We will think of moving it to run methods

jina/optimize/__init__.py
bhavsarpratik: Yes we can think on this

bwanglzu:
one suggestion based on what max mentioned during show and tell, if the yaml logic is pretty much same, maybe we can factor it out as a mixin.

jina/optimize/parameters.py
```diff
@@ -0,0 +1,181 @@
...
+
+    @classmethod
+    def _get_instance_from_yaml(cls, constructor, node):
+        data = ruamel.yaml.constructor.SafeConstructor.construct_mapping(
+            constructor, node, deep=True
+        )
+        return cls(**data)
+
+
+class IntegerParameter(OptimizationParameter):
```
JoanFM, line 54: I think we are here mixing `HyperParamSpace` concepts with `Sampling` concepts, it may be interesting to separate these 2 concepts. 

And so, this `Sampling` is the one responsible to translate `HyperParamSpace` to `optuna` parameters.

Like this you may not suffer this `hierarchy explosion` (See how often you have to give `low` and `high` parameters) to adapt to different `sampling` methods

bhavsarpratik:
@ALL refactoring is still under process.

jina/optimize/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+        self.n_docs = 0
+
+    def get_mean_evaluation(self):
+        if self.op_name:
+            return self.evaluation_values[self.op_name] / self.n_docs
+        return {
+            metric: val / self.n_docs for metric, val in self.evaluation_values.items()
+        }
+
+    def process_result(self, response):
```
JoanFM, line 151: are we surr this evaluation will work on all documents? Isnt it only considering mean_value for only the first Request?

JoanFM (changes requested):

jina/optimize/__init__.py
bhavsarpratik: The callback allows passing the eval name. The user can also write its own callback and pass to Optimizer.

JoanFM: my point is that the callback only considers one response. but an optimizer cycle may consist of many Requests. 

So if I send 100 queries in batches of 10 it seems it will individually consider 10 at a time.

bhavsarpratik: It's giving mean evaluation of the first metric in the below dict (for all docs)
{metric: val / self.n_docs for metric, val in self.evaluation_values.items()}

JoanFM: Yes, sorry the `=` led me to think it was not being updated

bhavsarpratik:
@ALL check the draft of new FlowRunner and MultiFlowRunner

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,105 @@
+from pathlib import Path
+from ruamel.yaml import YAML
+
+from jina.helper import colored
+from jina.logging import default_logger as logger
+from .parameters import load_optimization_parameters
+
+
+class EvaluationCallback:
+    def __init__(self, op_name=None):  # Todo: better name than op_name
```
JoanFM, line 10: please add type hinting

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        trial_parameters[self.workspace_env] = str(trial_workspace)
+
+        trial.workspace = trial_workspace
+        return trial_parameters
+
+    def _objective(self, trial):
+        trial_parameters = self._trial_parameter_sampler(trial)
+
+        self.flow_runner.run_indexing(
```
JoanFM, line 72: what if I just want to optimize a `Flow` without `querying`?

jina/optimizers/discovery.py
```diff
@@ -0,0 +1,128 @@
...
+    with open(filename, "r") as file:
+        return file.readlines()
+
+
+def _extract_executor_files(flows):
+    executor_files = set()
+    for flow_definition in flows.values():
+        for line in flow_definition:
+            stripped = line.strip()
+            if stripped.startswith("uses"):
```
JoanFM, line 20: what happens if I have a `docker image`?

jina/optimizers/discovery.py
```diff
@@ -0,0 +1,128 @@
...
+def config_global_environment():
+    os.environ.setdefault("JINA_WORKSPACE", "workspace_discovery")
+    os.environ.setdefault("JINA_MYENCODER_TARGET_DIMENSION", "64")
+
+
+def main():
+    config_global_environment()
+
+    run_parameter_discovery(
+        flow_files=[Path("flows/index.yml"), Path("flows/evaluate.yml")],
```
JoanFM, line 121: should this belong to `jina` main objects? like another CLI application?

JoanFM (commented):
I like much better the new `FlowRunner`.

Another comment is, can the `integration test` be minimalised? We do not need to have so much code for an integration test in my opinion, it seems like having a complete new project.

jina/optimizers/discovery.py
maximilianwerk: Yes, that will definitely happen.

jina/optimizers/discovery.py
bhavsarpratik: Good catch! Any suggestions for changing the logic?

jina/optimizers/discovery.py
bhavsarpratik: This file is under construction. Max will push the changes

jina/optimizers/__init__.py
bhavsarpratik: I will add typehint and docstring once we freeze the code

jina/optimizers/new_flow_runner.py
```diff
@@ -0,0 +1,167 @@
...
+        for file_path in self.pod_dir.glob("*.yml"):
+            parameters = yaml.load(file_path)
+            if "components" in parameters:
+                for i, component in enumerate(parameters["components"]):
+                    parameters["components"][i] = self._replace_param(
+                        component, trial_parameters
+                    )
+            parameters = self._replace_param(parameters, trial_parameters)
+            new_pod_file_path = self.pod_workspace / file_path.name
+            yaml.dump(parameters, open(new_pod_file_path, "w"))
```
maximilianwerk, line 64: That is also the way, we should handle yaml files in the discoverer. Thanks for showcasing it.

maximilianwerk (changes requested):
Some code style remarks.

It is deemed bad practice, to declare some `self.X` parameters outside of the `__init__`. Once exception: When you use `@property`.

Then there is the trade-off between a property function and a get function. Rule of thumb:
If the result calculation is cheap (as it is here) use property.
If the result calculation is not cheap use a get method.

jina/optimizers/__init__.py
maximilianwerk: The behavior of the Optimizer is not yet adapted to the new `FlowRunner` since @bhavsarpratik first wanted to get feedback on the changes on `FlowRunner` itself before doing premature changes on the Optimizer. Fair enough from my side.

jina/optimizers/new_flow_runner.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class MultiFlowRunner:
+    def __init__(self, flows, workspace="workspace"):
+        self.flows = flows
+        self.workspace = workspace
+
+    def run(self, trial_parameters):
+        for flow_def in self.flows:
+            flow = FlowRunner(**flow_def)
```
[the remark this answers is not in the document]
bhavsarpratik: Oops yes that was the intention. `FlowRunner` will not be called inside `MultiFlowRunner`.

jina/optimizers/discovery.py
bhavsarpratik: We can check for stripped.endswith('.yml')

JoanFM (commented):
Please do proper rebasing

PR #1415, opened 2020-12-08
feat: numpy indexer update/delete

cristianmtr:
Adding update & delete ops to NumpyIndexer

- [x] implement basic approach: mask filtering at `query_by_id`
- [x] add mask filtering to querying by vectors
- Faiss and Annoy use masking process --> separate PRs once this is merged https://github.com/jina-ai/jina-hub/pull/1280
- more testing?

### DIFF
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -5,7 +5,7 @@ import gzip
 import os
 from functools import lru_cache
 from os import path
-from typing import Optional, List, Union, Tuple, Dict
+from typing import Optional, List, Union, Tuple, Dict, Iterator
 
 import numpy as np
 
@@ -48,6 +48,7 @@ class BaseNumpyIndexer(BaseVectorIndexer):
         self.dtype = None
         self.compress_level = compress_level
         self.key_bytes = b''
+        self.valid_indices = np.array([], dtype=bool)
         self.key_dtype = None
         self._ref_index_abspath = None
 
@@ -57,6 +58,7 @@ class BaseNumpyIndexer(BaseVectorIndexer):
             self.dtype = ref_indexer.dtype
             self.compress_level = ref_indexer.compress_level
             self.key_bytes = ref_indexer.key_bytes
+            self.valid_indices = ref_indexer.valid_indices
             self.key_dtype = ref_indexer.key_dtype
             self._size = ref_indexer._size
             # point to the ref_indexer.index_filename
@@ -115,6 +117,7 @@ class BaseNumpyIndexer(BaseVectorIndexer):
         self._validate_key_vector_shapes(keys, vectors)
         self.write_handler.write(vectors.tobytes())
         self.key_bytes += keys.tobytes()
+        self.valid_indices = np.concatenate((self.valid_indices, np.full(len(keys), True)))
         self.key_dtype = keys.dtype.name
         self._size += keys.shape[0]
 
@@ -123,7 +126,7 @@ class BaseNumpyIndexer(BaseVectorIndexer):
 
         :return: a numpy ndarray of vectors
         """
-        vecs = self.raw_ndarray
+        vecs = self.raw_ndarray[self.valid_indices]
         if vecs is not None:
             return self.build_advanced_index(vecs)
 
@@ -154,18 +157,29 @@ class BaseNumpyIndexer(BaseVectorIndexer):
             return self._load_gzip(self.index_abspath)
         elif self.size is not None and os.stat(self.index_abspath).st_size:
             self.logger.success(f'memmap is enabled for {self.index_abspath}')
-            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r', shape=(self.size, self.num_dim))
+            # `==` is required. `is False` does not work in np
+            deleted_keys = len(self.valid_indices[self.valid_indices == False])  # noqa
+            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r',
+                             shape=(self.size + deleted_keys, self.num_dim))
 
     def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
-        int_ids = [self.ext2int_id[j] for j in ids]
-        return self.raw_ndarray[int_ids]
+        """
+        Search the index by the external key (passed during `.add(`)
+
+        :param ids: The list of keys to be queried
+        """
+        self.check_keys(ids)
+        indices = [self.ext2int_id[key] for key in ids]
+        return self.raw_ndarray[indices]
 
     @cached_property
     def int2ext_id(self) -> Optional['np.ndarray']:
         """Convert internal ids (0,1,2,3,4,...) to external ids (random index) """
         if self.key_bytes and self.key_dtype:
             r = np.frombuffer(self.key_bytes, dtype=self.key_dtype)
-            if r.shape[0] == self.size == self.raw_ndarray.shape[0]:
+            # `==` is required. `is False` does not work in np
+            deleted_keys = len(self.valid_indices[self.valid_indices == False])  # noqa
+            if r.shape[0] == self.size + deleted_keys == self.raw_ndarray.shape[0]:
                 return r
             else:
                 self.logger.error(
@@ -178,6 +192,29 @@ class BaseNumpyIndexer(BaseVectorIndexer):
         """Convert external ids (random index) to internal ids (0,1,2,3,4,...) """
         if self.int2ext_id is not None:
             return {k: idx for idx, k in enumerate(self.int2ext_id)}
+
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        self.delete(keys)
+        self.add(np.array(keys), np.array(values))
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        self.check_keys(keys)
+        for key in keys:
+            # mark as `False` in mask
+            self.valid_indices[self.ext2int_id[key]] = False
+        self._size = self.size - len(list(keys))
+        self.touch()
+
+    def check_keys(self, keys):
+        """checks if keys are present in the indexer. raises if any of them are not"""
+        missed = []
+        for key in keys:
+            # if it never existed or if it's been marked as deleted in the current index
+            # using `is False` doesn't work
+            if key not in self.ext2int_id.keys() or self.valid_indices[self.ext2int_id[key]] == False: # noqa
+                missed.append(key)
+        if missed:
+            raise KeyError(f'Key(s) {missed} were not found in {self.save_abspath}')
 
 
 @lru_cache(maxsize=3)
@@ -286,7 +323,8 @@ class NumpyIndexer(BaseNumpyIndexer):
             raise NotImplementedError(f'{self.metric} is not implemented')
 
         idx, dist = self._get_sorted_top_k(dist, top_k)
-        return self.int2ext_id[idx], dist
+        indices = self.int2ext_id[self.valid_indices][idx]
+        return indices, dist
 
     def build_advanced_index(self, vecs: 'np.ndarray'):
         return vecs
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -5,7 +5,7 @@ import gzip
 import os
 from functools import lru_cache
 from os import path
-from typing import Optional, List, Union, Tuple, Dict, Iterator
+from typing import Optional, List, Union, Tuple, Dict, Sequence
 
 import numpy as np
 
@@ -179,7 +179,7 @@ class BaseNumpyIndexer(BaseVectorIndexer):
             r = np.frombuffer(self.key_bytes, dtype=self.key_dtype)
             # `==` is required. `is False` does not work in np
             deleted_keys = len(self.valid_indices[self.valid_indices == False])  # noqa
-            if r.shape[0] == self.size + deleted_keys == self.raw_ndarray.shape[0]:
+            if r.shape[0] == (self.size + deleted_keys) == self.raw_ndarray.shape[0]:
                 return r
             else:
                 self.logger.error(
@@ -193,11 +193,11 @@ class BaseNumpyIndexer(BaseVectorIndexer):
         if self.int2ext_id is not None:
             return {k: idx for idx, k in enumerate(self.int2ext_id)}
 
-    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+    def update(self, keys: Sequence[int], values: Sequence[bytes], *args, **kwargs) -> None:
         self.delete(keys)
         self.add(np.array(keys), np.array(values))
 
-    def delete(self, keys: Iterator[int], *args, **kwargs):
+    def delete(self, keys: Sequence[int], *args, **kwargs) -> None:
         self.check_keys(keys)
         for key in keys:
             # mark as `False` in mask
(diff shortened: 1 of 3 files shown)


jina/executors/indexers/vector.py
```diff
@@ -115,6 +117,7 @@ def add(self, keys: 'np.ndarray', vectors: 'np.ndarray', *args, **kwargs) -> Non
         self._validate_key_vector_shapes(keys, vectors)
         self.write_handler.write(vectors.tobytes())
         self.key_bytes += keys.tobytes()
+        self.existing_keys = np.concatenate((self.existing_keys, np.repeat(True, len(keys))))
```
maximilianwerk, line 120: You can use `np.full()` instead of `np.repeat`. That is the recommended way, as far as I understand here https://stackoverflow.com/questions/21174961/how-to-create-a-numpy-array-of-all-true-or-all-false

cristianmtr: Oh yeah, I was actually using that in other places. Should make it consistent

JoanFM:
@cristianmtr @maximilianwerk @florian-hoenicke 

I do not think we should handle `Faiss and Annoy` (and any other `advanced` vector indexers) in this PR.

They are using an underlying `NumpyIndexer` structure and since they will have to reload the Annoy or Faiss index at query time, it won't be necessary.

I think this belongs to another story.

cristianmtr:
> @cristianmtr @maximilianwerk @florian-hoenicke
> 
> I do not think we should handle `Faiss and Annoy` (and any other `advanced` vector indexers) in this PR.
> 
> They are using an underlying `NumpyIndexer` structure and since they will have to reload the Annoy or Faiss index at query time, it won't be necessary.
> 
> I think this belongs to another story.

Oh it won't be in this PR. I'll create separate PRs.

But from what I understood looking at those two, we should add a way of filtering out the masked rows, right? Otherwise when they re-build the index it will have old (masked) data in there.

jina/executors/indexers/vector.py
```diff
@@ -154,18 +157,27 @@ def raw_ndarray(self) -> Optional['np.ndarray']:
...
-            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r', shape=(self.size, self.num_dim))
+            # `==` is required. `is False` does not work in np
+            deleted_keys = len(self.existing_keys[self.existing_keys == False]) # noqa
+            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r',
+                             shape=(self.size + deleted_keys, self.num_dim))
 
     def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
         int_ids = [self.ext2int_id[j] for j in ids]
-        return self.raw_ndarray[int_ids]
+        # TODO discussion on what we want to return on key that has been deleted
```
maximilianwerk, line 167: It should have the same behavior, as if we would query for a key that is not present.

jina/executors/indexers/vector.py
```diff
@@ -179,6 +191,19 @@ def ext2int_id(self) -> Optional[Dict]:
         if self.int2ext_id is not None:
             return {k: idx for idx, k in enumerate(self.int2ext_id)}
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        self.delete(keys)
+        self.add(np.array(keys), np.array(values))
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        # TODO check if they exist
```
maximilianwerk, line 199: Should this be part of this PR already?

maximilianwerk (changes requested):
Good job. Some remarks. That might be an edge case, but can you check what happens, when you delete all existing vectors?

jina/executors/indexers/vector.py
cristianmtr: Atm that is not handled. On master it simply raises

```
        int_ids = [self.ext2int_id[j] for j in ids]
```

Is that what we want?

jina/executors/indexers/vector.py
cristianmtr: Oh right. I'll add that and a test for it.

jina/executors/indexers/vector.py
maximilianwerk: OK, to be fair, I don't see the use-case of that function so I cannot give you a good answer. But it seems like we should raise an `AttributeError` until the use-case is understood. Please check, if the exception is catched in the calling function and if you could find a pattern, that fits. Thanks!

cristianmtr: Shouldn't it be `KeyError`? I guess it is similar in logic to how in the `PbIndexer` we handle missing keys on update/delete https://github.com/jina-ai/jina/blob/4ff025588fcfdb46a41bb011f260b1a53217c3c5/jina/executors/indexers/keyvalue.py#L82

maximilianwerk: Yes you are absolutely right. :)

cristianmtr:
Added the checking of keys, with an example test.

jina/executors/indexers/vector.py
```diff
@@ -260,7 +292,8 @@ def _get_sorted_top_k(dist: 'np.array', top_k: int) -> Tuple['np.ndarray', 'np.n
 
         return idx, dist
 
-    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple[Optional['np.ndarray'], Optional['np.ndarray']]:
+    def query(self, queries: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple[
```
JoanFM, line 295: Do not change the signature, it is inherited from another Base class, and we will need to change in every inherited class, so if we decide to do this, we will do it in a different PR

JoanFM (changes requested):

jina/executors/indexers/vector.py
cristianmtr: Yes, reverted. I did add a overwrite inside the function. Is that OK? It was a bit confusing before, I thought. Especially since `keys` comes into play when figuring out the indices after masking

jina/executors/indexers/vector.py
```diff
@@ -154,18 +157,24 @@ def raw_ndarray(self) -> Optional['np.ndarray']:
...
-            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r', shape=(self.size, self.num_dim))
+            # `==` is required. `is False` does not work in np
+            deleted_keys = len(self.valid_indices[self.valid_indices == False])  # noqa
+            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r',
+                             shape=(self.size + deleted_keys, self.num_dim))
 
-    def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
-        int_ids = [self.ext2int_id[j] for j in ids]
-        return self.raw_ndarray[int_ids]
+    def query_by_id(self, keys: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
```
JoanFM, line 165: also here avoid changing signature in this PR

JoanFM (dismissed):

jina/executors/indexers/vector.py
```diff
@@ -154,18 +157,31 @@ def raw_ndarray(self) -> Optional['np.ndarray']:
             return self._load_gzip(self.index_abspath)
         elif self.size is not None and os.stat(self.index_abspath).st_size:
             self.logger.success(f'memmap is enabled for {self.index_abspath}')
-            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r', shape=(self.size, self.num_dim))
+            # `==` is required. `is False` does not work in np
```
JoanFM, line 160: I guess it is because some lambda function may work inside for each index?

tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -35,13 +35,76 @@ def test_numpy_indexer(batch_size, compress_level, test_metas):
...
+        with pytest.raises(KeyError):
+            _ = indexer.query_by_id(vec_idx)
+        query_results = indexer.query_by_id(vec_idx[keys_to_delete:])
+        expected = vec[keys_to_delete:]
+        np.testing.assert_allclose(query_results, expected, equal_nan=True)
+
+
 @pytest.mark.parametrize('batch_size, compress_level', [(None, 0), (None, 1), (16, 0), (16, 1)])
-def test_numpy_indexer_known(batch_size, compress_level, test_metas):
+def test_numpy_indexer_known_and_delete(batch_size, compress_level, test_metas):
```
JoanFM, line 103: Is it possible to duplicate this test instead of changing it? This test is done for every `indexer` in the hub and I would prefer not to break this.

JoanFM (commented):
Just a minor thing

jina/executors/indexers/vector.py
```diff
@@ -154,18 +157,31 @@ def raw_ndarray(self) -> Optional['np.ndarray']:
...
     def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
-        int_ids = [self.ext2int_id[j] for j in ids]
-        return self.raw_ndarray[int_ids]
+        """
+        Search the index by the external key (passed during `.add(`)
+
+        :param ids: The list of keys to be queried
+        """
+        # better naming. cannot change function signature for now
+        keys = ids
```
maximilianwerk, line 172: I believe `deprecated_alias` is the decorator, that we have for exactly that use-case.

jina/executors/indexers/vector.py
```diff
@@ -272,21 +311,24 @@ def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple[Option
             Distance (the smaller the better) is returned, not the score.
 
         """
+        # better name. cannot change function signature for now
+        vector_queries = keys
```
maximilianwerk, line 315: I am not convinced by the renaming ;) It verschlimmbessert (german word for "worse but better") things.

maximilianwerk (approved):
I am OK with the renaming, even thought I am not 100% convinced that it is necessary. The rest LGTM.

jina/executors/indexers/vector.py
cristianmtr: You mean this? https://stackoverflow.com/a/49802489

jina/executors/indexers/vector.py
cristianmtr: Do you mean the approach, or the names themselves? I found the original ones didn't really match

tests/unit/executors/indexers/test_numpyindexer.py
cristianmtr: Sure, done

jina/executors/indexers/vector.py
maximilianwerk: Did not know this, just saw the decorator earlier. Interesting read.

jina/executors/indexers/vector.py
maximilianwerk: Both. But in this case also the variable name. I see where it comes from, but in the search context, they are `keys`.

cristianmtr: Ok, reverted

maximilianwerk: The conversation in the related issue will eventually bring a solution for renaming.

maximilianwerk (approved):
LGTM

florian-hoenicke (dismissed):

jina/executors/indexers/vector.py
```diff
@@ -179,6 +193,29 @@ def ext2int_id(self) -> Optional[Dict]:
...
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        self.check_keys(keys)
+        for key in keys:
+            # mark as `False` in mask
+            self.valid_indices[self.ext2int_id[key]] = False
+        self._size = self.size - len(list(keys))
+        self.touch()
+
+    def check_keys(self, keys):
```
hanxiao, line 208:
```suggestion
    def check_keys(self, keys: Iterator[int]):
```

jina/executors/indexers/vector.py
```diff
@@ -179,6 +193,29 @@ def ext2int_id(self) -> Optional[Dict]:
         if self.int2ext_id is not None:
             return {k: idx for idx, k in enumerate(self.int2ext_id)}
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        self.delete(keys)
+        self.add(np.array(keys), np.array(values))
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
```
hanxiao, line 200:
```suggestion
    def delete(self, keys: Iterator[int], *args, **kwargs) -> None:
```

jina/executors/indexers/vector.py
```diff
@@ -179,6 +193,29 @@ def ext2int_id(self) -> Optional[Dict]:
         if self.int2ext_id is not None:
             return {k: idx for idx, k in enumerate(self.int2ext_id)}
 
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
```
hanxiao, line 196:
```suggestion
    def update(self, keys: Sequence[int], values: Sequence[bytes], *args, **kwargs) -> None:
```

jina/executors/indexers/vector.py
```diff
@@ -5,7 +5,7 @@
 import os
 from functools import lru_cache
 from os import path
-from typing import Optional, List, Union, Tuple, Dict
+from typing import Optional, List, Union, Tuple, Dict, Iterator
```
hanxiao, line 8:
```suggestion
from typing import Optional, List, Union, Tuple, Dict, Sequence
```

jina/executors/indexers/vector.py
```diff
@@ -154,18 +157,29 @@ def raw_ndarray(self) -> Optional['np.ndarray']:
             return self._load_gzip(self.index_abspath)
         elif self.size is not None and os.stat(self.index_abspath).st_size:
             self.logger.success(f'memmap is enabled for {self.index_abspath}')
-            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r', shape=(self.size, self.num_dim))
+            # `==` is required. `is False` does not work in np
+            deleted_keys = len(self.valid_indices[self.valid_indices == False])  # noqa
+            return np.memmap(self.index_abspath, dtype=self.dtype, mode='r',
+                             shape=(self.size + deleted_keys, self.num_dim))
 
     def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
```
hanxiao, line 165:
```suggestion
    def query_by_id(self, ids: Union[Sequence[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
```

jina/executors/indexers/vector.py
```diff
@@ -154,18 +157,29 @@ def raw_ndarray(self) -> Optional['np.ndarray']:
...
 
     @cached_property
     def int2ext_id(self) -> Optional['np.ndarray']:
         """Convert internal ids (0,1,2,3,4,...) to external ids (random index) """
         if self.key_bytes and self.key_dtype:
             r = np.frombuffer(self.key_bytes, dtype=self.key_dtype)
-            if r.shape[0] == self.size == self.raw_ndarray.shape[0]:
+            # `==` is required. `is False` does not work in np
+            deleted_keys = len(self.valid_indices[self.valid_indices == False])  # noqa
+            if r.shape[0] == self.size + deleted_keys == self.raw_ndarray.shape[0]:
```
hanxiao, line 182:
```suggestion
            if r.shape[0] == (self.size + deleted_keys) == self.raw_ndarray.shape[0]:
```

hanxiao (changes requested):

hanxiao (approved):

hanxiao (approved):

PR #1427, opened 2020-12-10
[WIP] refactor(flow): copy_flow => inplace

lsgrep:


jina/flow/__init__.py
```diff
@@ -267,7 +267,7 @@ def needs_all(self, name: str = 'joiner', *args, **kwargs) -> 'Flow':
 
     def add(self,
             needs: Union[str, Tuple[str], List[str]] = None,
-            copy_flow: bool = True,
+            inplace: bool = True,
```
JoanFM, line 270: it totally feels like `inplace` and `copy_flow` are exactly opposite concepts?

JoanFM (changes requested):

jina/flow/__init__.py
lsgrep: I am still working on the PR.  I will ping you once I've done with it.  Sorry, I think I should have finished it before creating the PR.

JoanFM: no problem, you can convert it to `Draft` to communicate that is still in progress

lsgrep: @JoanFM, Before I make changes to the 20+ places,  I would like to discuss my reasoning for this PR.  I think repeating things, in general, is an anti-pattern and I've seen `inplace` used in popular frameworks such as Numpy,  TensorFlow for similar use cases.

JoanFM: yes, but my feeling is that `inplace` and `copy_flow` have exactly the opposite logic right?

meaning, when `copy_flow` is `True`, `inplace` should be `False` and viceversa?

lsgrep: Yes. I understand the differences and I will update the code accordingly. I am just making sure the change is recommended.

lsgrep:
In `Frow().plot` method,   `copy_flow` parameter is not used in the function body, and conceptually `plot` should be idempotent?  Should I remove it?

PR #1435, opened 2020-12-10
test: refactor primitive type test

bwanglzu:
This PR:

1. Added `DocGroundtruthPair` pair tests.
2. Structured primitive types with the same code structure as we have in core

image


tests/unit/types/document/test_helper.py
```diff
@@ -0,0 +1,77 @@
+import types
+
+import pytest
+
+from jina import Document
+from jina.types.document.helper import DocGroundtruthPair
+
+@pytest.fixture(scope='function')
```
JoanFM, line 8: scope is function by default right?

bwanglzu: yes..i like to be explicit so added function here

JoanFM (approved):

tests/unit/types/document/test_helper.py
JoanFM: like a lot the folder structure matching

PR #1440, opened 2020-12-11
feat: document random_id

theUnkownName:
UPDATE [2020-21-16 21:21 CET by @maximilianwerk]
This PR now consistently renames `hash` back to `id` and also changes the semantic back to a simple `id`. In case no `id` is provided for a `Document`, a random one will be generated.

Furthermore, some artifacts of the old `hash` are still existing, since they are still used in the Hub. Once I refactored all code in the Hub to the new `id`, I will remove the missing old `hash` artifacts.

Finally, a new `content_hash` is "introduced". Alternative formulation: The hash-semantic of the old `id/hash` is now inside the `content_hash`


theUnkownName:
I made the first PR refactoring the places where the `update_id` was called. 
Now the id will either be set by the client or if none was set, it will be a random one. 

And to still be able to have the hash of the document there is a new `get_document_hash` function. If this is ok I will then refactor all the tests that were using `update_id`

jina/clients/python/request.py
```diff
@@ -31,8 +31,8 @@ def _build_doc_from_content():
             return data, DataInputType.DOCUMENT
         try:
             d = Document(data, **kwargs)
-            if override_doc_id:
-                d.update_id()
+            #if override_doc_id:
```
JoanFM, line 34: here if it may be good to provide some logic to assign random documents if client did not take care of it?

jina/types/document/__init__.py
```diff
@@ -135,6 +135,12 @@ def __init__(self, document: Optional[DocumentSourceType] = None,
                              f'if you are trying to set the content '
                              f'you may use "Document(content=your_content)"') from ex
 
+        if custom_id is None:
```
JoanFM, line 138: not sure if we should, but would it make sense here to check if `self.id` has already been set? in case this is already created from an existing `documentProto`?

jina/types/document/uid.py
```diff
@@ -37,6 +37,19 @@ def new_doc_hash(doc: 'DocumentProto') -> int:
...
 
 
+def get_document_hash(doc: 'DocumentProto') -> str:
+    """ Generate a new hexdigest based on the content of the document.
+
+    :param doc: a non-empty document
+    :return: the hexdigest based on :meth:`blake2b`
+    """
+    d = doc
+    if _doc_field_mask:
```
JoanFM, line 47: does this _doc_field_mask exist?

JoanFM (commented):
Looks like a good direction to go for me.

jina/types/document/uid.py
theUnkownName: Is set as None by default. I'm not 100% how the doc_id was generated, but I'm taking this function exactly as the previous one `new_doc_id` that was doing the same, just a different name makes more sense since it's not setting the id anymore

jina/clients/python/request.py
maximilianwerk: If a `Document` is initiated in the first place, you either provide an `id` or a random one is generated.

jina/types/document/__init__.py
maximilianwerk: Very good point. :+1:

theUnkownName:
Hey people, as an update, there's a `content_hash` on the `DocumentProto` now, Max told me that this needs another step to transform the `jina.proto` into python file, anyone knows how to proceed on this?

maximilianwerk:
> Hey people, as an update, there's a `content_hash` on the `DocumentProto` now, Max told me that this needs another step to transform the `jina.proto` into python file, anyone knows how to proceed on this?

I think that file explains how it works: https://github.com/jina-ai/jina/blob/master/jina/proto/build-proto.sh

So it should be `docker run -v $(pwd)/jina/proto:/jina/proto jinaai/protogen`

theUnkownName:
> > Hey people, as an update, there's a `content_hash` on the `DocumentProto` now, Max told me that this needs another step to transform the `jina.proto` into python file, anyone knows how to proceed on this?
> 
> I think that file explains how it works: https://github.com/jina-ai/jina/blob/master/jina/proto/build-proto.sh
> 
> So it should be `docker run -v $(pwd)/jina/proto:/jina/proto jinaai/protogen` (perhaps you have to build the docker image beforehand)

Yeah it was like that, I didn't have to do anything else

jina/types/document/__init__.py
```diff
@@ -402,7 +403,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.update_id()
+        pass
```
cristianmtr, line 406: So is the Document supposed to be used as a context manager now or not?

jina/types/document/uid.py
cristianmtr: I can't find any reference of this `_doc_field_mask` being modified, only read.

jina/types/document/__init__.py
maximilianwerk: In my opinion, you shouldn't for now. But since it is advertised in the Native Data Types blog article and is also already used all over the place, we choose to keep the possibility.

jina/types/document/__init__.py
```diff
@@ -135,6 +137,12 @@ def __init__(self, document: Optional[DocumentSourceType] = None,
                              f'if you are trying to set the content '
                              f'you may use "Document(content=your_content)"') from ex
 
+        if custom_id is None:
+            import random
+            custom_id = random.randint(0, 9)
```
cristianmtr, line 142: So is the id a string (from the function signature) or an int?

jina/types/sets/chunk.py
```diff
@@ -30,7 +30,8 @@ def append(self, document: 'Document', **kwargs) -> 'Document':
 
         if not chunk.mime_type:
             chunk.mime_type = self._ref_doc.mime_type
-        chunk.update_id()
+        #chunk.update_id()
```
cristianmtr, line 33: What's the reason behind leaving these comments around?

tests/unit/types/document/test_document.py
```diff
@@ -105,7 +105,7 @@ def test_bad_good_doc_id():
 
 def test_id_context():
     with Document() as d:
-        assert not d.id
+        #assert not d.id
```
cristianmtr, line 108: Same here

jina/types/sets/chunk.py
maximilianwerk: reminder to make sure, that one is completely refactored. Will be removed before merging. Good that you point it out.

jina/types/document/__init__.py
maximilianwerk: It does not matter. Good opportunity to dig a little deeper: Try to find out why is does not matter by looking at the implementation of `self.id`.

jina/types/document/__init__.py
maximilianwerk: Now, that I understood the whole concept better, I think that should be exactly the way of setting `id`. There will be no `custom_id` after my next commit.

maximilianwerk:
Just rebased. Please review once again @JoanFM @theUnkownName

jina/clients/request.py
```diff
@@ -18,12 +18,10 @@
                                      Tuple[DocumentSourceType, DocumentSourceType]]]
 
 
-def _build_doc(data, data_type: DataInputType, override_doc_id, **kwargs) -> Tuple['Document', 'DataInputType']:
+def _build_doc(data, data_type: DataInputType, **kwargs) -> Tuple['Document', 'DataInputType']:
```
JoanFM, line 21: why don't we allow override_doc_id so that the user can choose that we give for them a `random id? Have we checked if some example uses this?

jina/types/document/__init__.py
```diff
@@ -135,6 +137,10 @@ def __init__(self, document: Optional[DocumentSourceType] = None,
                              f'if you are trying to set the content '
                              f'you may use "Document(content=your_content)"') from ex
 
+        if self._document.id is None or self._document.id == '':
```
JoanFM, line 140: self._document.id.is_empty() may work better? not sure if works in python

jina/types/document/__init__.py
```diff
@@ -135,6 +137,10 @@ def __init__(self, document: Optional[DocumentSourceType] = None,
...
+            import random
+            self.id = random.randint(0, 9223372036854775807)
```
JoanFM, line 142: isn't there some Python built-in Integer.Max or something?

jina/types/document/__init__.py
```diff
@@ -402,7 +401,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
```
JoanFM, line 403: eventually we will remove the `context manager` concept from `document` right? or should we at least do some check like wether `id` is not None or not empty?

tests/unit/test_helper.py
```diff
@@ -64,21 +64,6 @@ def test_np_int():
     assert hash2bytes(np.int64(a)) == hash2bytes(a)
 
 
-def test_hash():
```
JoanFM, line 67: have these functions been removed? I have not seen the removal of these functions in this PR right?

jina/clients/request.py
maximilianwerk: I don't get the question. Either, the user gives and `id` in the data or the `Document` gets a random `id`. In either case, the user can decide to overwrite the `id` later via `doc.id = 1337`.

jina/types/document/__init__.py
maximilianwerk: This is `2**64/2`, which is what should well fit into 64 bit. Apparently, we could directly use `2**64`. Haven't thought too much about it. I'll spend another thought tomorrow.

jina/types/document/__init__.py
maximilianwerk: It cannot be None or empty at the point. It will always be populated. This is just for backward compatibility and for "niceness".

jina/types/document/__init__.py
JoanFM: yes but python must have some default value for this.

tests/unit/test_helper.py
maximilianwerk: Haha, so predictable. When I removed this test, I knew you would write that.

`new_doc_id` was renamed to `get_content_hash` and `new_doc_hash` was removed. So there is no point in testing forward-backward converting. I thought about a test like
```
hash_ = get_content_hash(d)
assert hash_ == hash2id(id2hash(hash_))
```
But this already shows, that the semantic of `id` and `hash` are not precise anymore (Beware, it is NOT `id2hash(hash2id(hash_))`!). I will have another thought about this tomorrow with a fresh mind.

jina/types/document/__init__.py
maximilianwerk: Nope, python does not. `int` is unbounded in python. Anyhow, `numpy` has: `np.iinfo(np.int64).max`, which is `9223372036854775807`. Since we do `int(value).to_bytes(_digest_size, sys.byteorder, signed=True)` to convert it to bytes, this should also be the right number. If I understand correctly, we could also have negative `id`s, but that feels weird. So I restrict for the `numpy` max.

JoanFM: yes I am not saying number is wrong, at least have a CONSTANT well documented rather than a magic number.

jina/types/document/__init__.py
maximilianwerk: Well apparently mine is the best practice in python as far as I know.

JoanFM: create a constant on top. MAX_DOCUMENT_ID

jina/executors/rankers/__init__.py
```diff
@@ -32,7 +32,9 @@ class Chunk2DocRanker(BaseRanker):
 
     required_keys = {'text'}  #: a set of ``str``, key-values to extracted from the chunk-level protobuf message
     COL_MATCH_PARENT_HASH = 'match_parent_hash'
+    COL_MATCH_PARENT_ID = 'match_parent_hash'
```
maximilianwerk, line 35: In case someone wonders, what this is about: This enables us to first change everything in the hub from `COL_MATCH_PARENT_HASH` to `COL_MATCH_PARENT_ID` without breaking and then remove the `COL_MATCH_PARENT_ID`, since it is an `id` again.

maximilianwerk:
I updated the initial message of @theUnkownName

jina/types/sets/document.py
```diff
@@ -30,6 +31,8 @@ def insert(self, index: int, doc: 'Document') -> None:
     def __setitem__(self, key, value: 'Document'):
         if isinstance(key, int):
             self._docs_proto[key].CopyFrom(value)
+        elif isinstance(key, UniqueId):
+            self._docs_map[str(key)].CopyFrom(value)
```
maximilianwerk, line 35: I am not sure, if that is the right way to go. Perhaps we should also always convert into `UniqueId`, since it is the jina native `id`. Opinion?

jina/clients/request.py
JoanFM: ok good, using this `_build_doc` will generate random ids, if the user wants custom ids, they have to use their own generator. Fair enough

jina/types/sets/document.py
maximilianwerk: Thought about this a bit more. This can go away in the moment we remove the `__hash__` function from `UniqueId`, because afterwards `str` and `UniqueId` will get the same hash again.

jina/proto/jina.proto
```diff
@@ -74,6 +74,9 @@ message DocumentProto {
     // A hexdigest that represents a unique document ID
     string id = 1;
 
+    // A hexdigest that represents the hash of the content of the document
+    string content_hash = 24;
```
JoanFM, line 78: has the `proto` been compiled?

JoanFM:
I am surprised to see so many tests passing, how are the `Incremental Indexing` tests passing?

jina/proto/jina.proto
theUnkownName: Yes, Max pointed me on how to do this

jina/types/document/__init__.py
maximilianwerk: And out of a sudden, it is not `pass` anymore. Reasoning: It makes sense to update the `content_hash` after `Document` is completed. Otherwise, the user would have to call the `update_content_hash` manually.

maximilianwerk:
> I am surprised to see so many tests passing, how are the `Incremental Indexing` tests passing?

While I see your point, the `get_duplicate_docs` function, which is used for all the incremental indexing tests is written in a way, that it tests for duplicate `ids` and not duplicate `content`. :D If I change to the following, the tests fail as expected (`d.text` is the change):

```
def get_duplicate_docs(num_docs=10):
    ...
    num_uniques = len(set(d.text for d in result))
    return result, num_uniques
```

Point is: that will only be repaired, after the changes to the `DocIdCache`. So I'd leave it for now.

JoanFM (dismissed):

jina/types/document/uid.py
```diff
@@ -28,13 +28,21 @@
...
+def get_content_hash(doc: 'DocumentProto') -> str:
+    """ Generate a new hexdigest based on the content of the document.
 
-def new_doc_hash(doc: 'DocumentProto') -> int:
-    return id2hash(new_doc_id(doc))
+    :param doc: a non-empty document
+    :return: the hexdigest based on :meth:`blake2b`
+    """
+    # TODO: once `new_doc_id` is removed, the content of this function can directly move to the `Document`.
+    return blake2b(doc.SerializeToString(), digest_size=_digest_size).hexdigest()
```
cristianmtr, line 45: Shouldn't the content hash not depend on the id?

cristianmtr (changes requested):
Shouldn't the content hash not depend on the id?

jina/types/document/uid.py
maximilianwerk: Fair enough. That's where the `_doc_field_mask` comes into play. I'll add that.

theUnkownName: It should depend on the document content only. The ID needs to be totally independent.

theUnkownName: Ah it was the other way around...

maximilianwerk:
> Shouldn't the content hash not depend on the id?

I will fix this in the cleanup PR tomorrow after I did the Hub changes. Thus I merge the current state for now.

maximilianwerk:
> Shouldn't the content hash not depend on the id?

I added the hash calculation, which omits the `id`. This should pretty much be the same as the old hash calculation.

jina/drivers/index.py
```diff
@@ -27,14 +27,14 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
...
+            self.exec_fn(np.array([int(doc.id) for doc in docs_pts]), np.stack(embed_vecs))
 
 
 class KVIndexDriver(BaseIndexDriver):
     """Serialize the documents/chunks in the request to key-value JSON pairs and write it using the executor
     """
 
     def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
-        keys = [hash(doc.id) for doc in docs]
+        keys = [int(doc.id) for doc in docs]
```
florian-hoenicke, line 38: does this have to be done in the VectorIndexDriver as well?

maximilianwerk: It is. See line 30.

maximilianwerk (approved):
Was approved before. Tests now succeed.

jina/drivers/index.py
florian-hoenicke: ah okay, I searched for VectorIndexDriver and did not find a hit. 😳

PR #1443, opened 2020-12-11
fix: redundant translation in readme Chinese

smy0428:
fix: redundant translation in readme Chinese


bwanglzu (approved):
LGTM, @smy0428 thanks!

Issue #975, opened 2020-09-21
HubIO not building images with base of local image


JoanFM:
**Describe the bug**
When running unit tests for PR #968, since communication layer has been changed, running even a dry run with a docker image running some old version of jina gives errors.

Then is where it is shown the need to change the image of `jinaai/jina:devel` to `jinaai/jina:test-pip` (Another question is if what we really have to test is the ability to obtain `jinaai/jina:devel` (if this is the case we have a little problem))

`jinaai/jina:test-pip` is created by our CI to have an image with the codebase that is being tested. 
When this changes are applied, the tests of `hubio` still fail with a different message:

`FAILED tests/unit/docker/test_hub_build.py::test_hub_build_pull - docker.errors.BuildError: manifest for jinaai/jina:test-pip not found: manifest unknown: manifest unknown`

(Note that I have changed some names and paths for the tests)
But when I locally run :
`docker build -f Dockerfile -t jinahub/pod.encoder.dummy_mwu_encoder:0.0.6 .` inside the `hub-mwu` folder, the image is successfully built.

This last comment leads me to the conclusion that there is some discrepancy between `hubio` and the docker client.

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACORE-240) by [Unito](https://www.unito.io/learn-more)

PR #1452, opened 2020-12-13
test: refactor driver test

bwanglzu:
minor test refactoring on the `querylang` and `driver`


tests/unit/drivers/test_matches2doc_rank_drivers.py
```diff
@@ -41,17 +41,15 @@ def create_document_to_score():
     # |- matches: (id: 3, parent_id: 1, score.value: 3),
     # |- matches: (id: 4, parent_id: 1, score.value: 4),
     # |- matches: (id: 5, parent_id: 1, score.value: 5),
-
-    doc = jina_pb2.DocumentProto()
+    doc = Document()
```
JoanFM, line 44: the idea herr is to see if we can refactor the logic of the test itself to be less dependant on id? I am not sure it is possible though

JoanFM: One problem is that  this test is so dependant on document logic that any document refactor involves a lot of effort to regfactor these tests. is there a way to avoid this?

tests/unit/drivers/querylang/queryset/test_lookup.py
```diff
@@ -173,3 +175,30 @@ def test_lookup_leaf_None():
...
+def test_docs_filter():
+    s = random_docs(10)
+    ss = QuerySet(s).filter(tags__id__lt=5, tags__id__gt=3)
+    ssr = list(ss)
+    assert len(ssr) == 1
+    for d in ssr:
+        assert (3 < d.tags['id'] < 5)
+
+
+def test_docs_filter_equal():
```
florian-hoenicke, line 188: There is some redundant code. What about some kind of get_data Funktion?

florian-hoenicke (changes requested):

tests/unit/drivers/querylang/queryset/test_lookup.py
bwanglzu: these codes were copy-pasted from another file, I'll refactor this, thanks @florian-hoenicke !

tests/unit/drivers/test_matches2doc_rank_drivers.py
bwanglzu: the option I can see is to maintain fixtures in a separate file so that we can reuse/modify it anywhere

JoanFM: butis there so many shared code?

bwanglzu: there could be, if we "composite" small fixtures into bigger ones, a lot of code are actually shared

tests/integration/hub_usage/dummyhub_abs/__init__.py
```diff
@@ -1,4 +1,4 @@
-from helper import foo
+from .helper import foo
```
JoanFM, line 1: I think the point of this file is to have this and fail for that reason

tests/unit/drivers/test_matches2doc_rank_drivers.py
JoanFM: okey that'd be great.

tests/integration/hub_usage/dummyhub_abs/__init__.py
bwanglzu: okay I'll revert this change

tests/unit/drivers/querylang/queryset/test_lookup.py
bwanglzu: created a fixture for this

florian-hoenicke (approved):

bwanglzu:
related to #975 , hubio test both locally and in the ci.

bwanglzu:
local test passed after running `docker build -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .`

nan-wang (changes requested):
minor changes

nan-wang (approved):

bwanglzu:
@nan-wang sorry I was off yesterday, just saw your message. The test was actually copied from another file `test_queryset.py`, I merged `test_queryset` into `test_lookup` since they're in the same file in our codebase. I think the objective of the test is just to double check number of chunks.

PR #1450, opened 2020-12-12
refactor(client): refactor code, style of pyclient

hanxiao:
- Simplify the implementation of `PyClient`, `GatewayPea`.
- Keep async, streaming, prefetch, jupyter-notebook features support.

image

PR #1455, opened 2020-12-14
[not to be merged] test: stuck on error

deepankarm:
- One more reason for adding a `runtime` was to fix this.
- Not sure why this test-case was removed in #1450

### DIFF
--- a/tests/unit/clients/python/test_async_client.py
+++ b/tests/unit/clients/python/test_async_client.py
@@ -1,0 +1,24 @@
+import pytest
+import numpy as np
+
+from jina.flow import Flow
+from jina.excepts import BadClientCallback
+
+
+def validate(x):
+    raise NotImplementedError
+
+
+def test_client_on_error():
+    # This will fail and get stuck
+    # Not sure why this test case was removed
+    f = Flow().add()
+    with pytest.raises(BadClientCallback):
+        with f:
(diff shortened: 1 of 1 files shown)


tests/unit/clients/python/test_async_client.py
```diff
@@ -0,0 +1,24 @@
...
+def test_client_on_error():
+    # This will fail and get stuck
+    # Not sure why this test case was removed
+    f = Flow().add()
+    with pytest.raises(BadClientCallback):
+        with f:
+            f.index_ndarray(np.random.random([5, 4]), output_fn=validate, continue_on_error=False)
+
+    # Recreating the flow instance will make it work.
+    # But reusing the flow upon an error is a good test case
```
hanxiao, line 21: > # But reusing the flow upon an error is a good test case

But this describes another scenario, which is different than the discussion in Slack.

In this particular test, when you write two tests in a row, you are testing the following case:
- You are testing exception in `callback`, not error in request generator
- The exception breaks the `async stub.Call` 
- Server probably has something hold in the stream
- Restart the client, keep server untouched.
- Now, server stucks

hanxiao: https://github.com/grpc/grpc/issues/10974

PR #1459, opened 2020-12-14
feat: optimiser with new flow runner and flow v1 parser

bhavsarpratik:
Optimiser with new flow runner, v1 parser with env support and test


jina/optimizers/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        trial_workspace = Path(
+            "JINA_WORKSPACE_" + "_".join([str(v) for v in trial_parameters.values()])
+        )
+        trial_parameters[self.workspace_env] = str(trial_workspace)
+
+        trial.workspace = trial_workspace
+        return trial_parameters
+
+    def _objective(self, trial):
+        self.eval_flow_runner.callback.flush()
```
JoanFM, line 71: can't just a new callback be created? That would be more `functional-like`

jina/optimizers/flow_runner.py
```diff
@@ -0,0 +1,87 @@
...
+        flow_yaml = self._create_trial_flow(workspace, trial_parameters)
+        with Flow.load_config(flow_yaml) as f:
+            getattr(f, self.task)(
+                self.documents,
+                batch_size=self.batch_size,
+                output_fn=self.callback,
+            )
+
+
+class MultiFlowRunner:
```
JoanFM, line 80: To have an easier interface, this could behave exactly as a `FlowRunner` if you could inherit from `FlowRunner`. This design pattern would be the same as `CompoundExecutor` for `Executors`?

JoanFM: Like this you would have the same exact `run` interface and for the user it would be exactly the same running a `FlowRunner` and a `MultiFlowRunner`

jina/optimizers/parameters.py
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+yaml.register_class(IntegerParameter)
+
+class FloatParameter(OptimizationParameter):
+    def __init__(
+        self,
+        low: float,
+        high: float,
+        step_size: Optional[float] = None,
```
JoanFM, line 87: why is step_size for Float Optional and not for Integer?

I think that for these classes we will soon have an explosion. 

I would separate the concerns of defining the `HyperParamSpace (low, high...)` and the `Sampling` which defines `step_size, log ...` (like this you could have just one FloatParameter work with `UniformSampler` or with `StepSampler`)

jina/optimizers/parameters.py
```diff
@@ -0,0 +1,181 @@
...
+            "name": self.env_var,
+            "low": self.low,
+            "high": self.high,
+            "step": self.step_size,
+            "log": self.log,
+        }
+
+yaml.register_class(FloatParameter)
+
+class UniformParameter(OptimizationParameter):
```
JoanFM, line 110: In relation to previous comment, why is this not considered a `FloatParameter` if it is an hyperparam space formed by Floats?

jina/optimizers/parameters.py
```diff
@@ -0,0 +1,181 @@
...
+    def to_optuna_args(self):
+        return {
+            "name": self.env_var,
+            "low": self.low,
+            "high": self.high,
+        }
+
+yaml.register_class(UniformParameter)
+
+class LogUniformParameter(OptimizationParameter):
```
JoanFM, line 126: and here

tests/integration/optimizers/fashion/data.py
```diff
@@ -0,0 +1,60 @@
+import gzip
```
JoanFM, line 1: for integration tests, isn't there a more minimal test that do not require these amount of code?

tests/integration/optimizers/optimize.py
```diff
@@ -0,0 +1,42 @@
+from fashion.data import get_data
+from fashion.evaluation import index_document_generator, evaluation_document_generator
+
+from jina.optimizers import OptunaOptimizer, EvaluationCallback
+from jina.optimizers.flow_runner import FlowRunner
+
+DATA_DIRECTORY = "data"
+
+
+def main():
```
JoanFM, line 10: feels weird to have a `main` under `tests`

tests/integration/optimizers/test_optimizer.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_optimizer():
+    """This will run complete optimisation.
+    Todo: fix number of index and query doc so that test doesnt run long
+    Todo: assert to check eval key and value for values generated in config/best_config.yml
+    Todo: after optimisation is complete, assert to check trial pods and flows do not have any evironment variable.
+          This can be checked by some of our current yaml related functions in flow runner
+          and changing its functionality to assert values are not starting with $.
+    """
+    main()
```
JoanFM, line 12: put the main directly here

JoanFM (changes requested):

jina/optimizers/parameters.py
maximilianwerk: This mimics exactly the optuna behavior. When we have another sampler, which we can supply to optuna, we can think about abstracting this. I believe you think way to complicated here. This is a optimization, where you won't have thousands of thousands of executions but rather in the hundreds with a huge parameter set. Fine-grained control over the sampling will not make your life much easier.

jina/optimizers/__init__.py
bhavsarpratik: Passing the callback without init in the optimizer to recreate every time will require to pass the eval_name to Optimizer which I didn't like.

jina/optimizers/flow_runner.py
bhavsarpratik: Sure, I will check on this

jina/optimizers/__init__.py
JoanFM: I think is best practica not to reset the object

JoanFM: but its okey

JoanFM (commented):
Just a comment. Will you add specific unitary tests?

tests/integration/optimizers/fashion/data.py
bhavsarpratik: I am not able to find anything which we can trim

tests/integration/optimizers/optimize.py
bhavsarpratik: This is temporary

tests/integration/optimizers/test_optimizer.py
bhavsarpratik: Yes that's the plan once everything is ready

jina/optimizers/__init__.py
maximilianwerk: I disagree. Resetting objects like this is a root for ugly and hard-to-track bugs. Please don't do that. There are two possibilities to solve that in general: 
1. Just give the eval_name to the `OptunaOptimizer`
2. The evaluator must implement a `empty_copy` function, which will return an empty copy of the evaluator. Guarantee, that custom evaluators implement it by defining an interface (named `EvaluationCallback`?), that evaluators must implement and inherit from this in the `MeanEvaluationCallback`

The first option is bad, as @bhavsarpratik already discovered, since it will not work with custom evaluator. Anyhow, the second is totally valid.

bhavsarpratik:
We will add unit tests for discovery and maybe parameters module

JoanFM:
> We will add unit tests for discovery and maybe parameters module

unittests need to exist for every module and class

jina/optimizers/__init__.py
JoanFM: I agree this shoukd be avoided but do not understand the evaluators copy. optimizer does not see the evaluator, just Flows containing Evaluators.

tests/integration/optimizers/fashion/data.py
```diff
@@ -0,0 +1,60 @@
...
+import numpy as np
+import os
+
+import urllib.request
+from pathlib import Path
+
+from jina.clients.python import ProgressBar
```
florian-hoenicke, line 8: ProgressBar fails to import for me

tests/integration/optimizers/pods/components.py
```diff
@@ -0,0 +1,37 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+import numpy as np
+
+from jina.executors.encoders import BaseImageEncoder
+from jina.optimizer.parameters import IntegerParameter
```
florian-hoenicke, line 7: from jina.optimizers.parameters import IntegerParameter

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        yaml = YAML(typ="rt")
+        yaml.dump(study.best_trial.params, open(self.best_config_filepath, "w"))
+        logger.info(colored(f"Number of finished trials: {len(study.trials)}", "green"))
+        logger.info(colored(f"Best trial: {study.best_trial.params}", "green"))
+        logger.info(colored(f"Time to finish: {study.best_trial.duration}", "green"))
+
+    def optimize_flow(
+        self, n_trials, sampler="TPESampler", direction="maximize", seed=42
+    ):
+        import optuna
```
florian-hoenicke, line 95: will optuna be added to the core requirements?

bhavsarpratik: I don't think so but maybe we can have dependencies installed with jina[optimizer]

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,100 @@
+from pathlib import Path
+from ruamel.yaml import YAML
+
+from .new_flow_runner import MultiFlowRunner
+from jina.helper import colored
+from jina.logging import default_logger as logger
+from .parameters import load_optimization_parameters
+
+
+class EvaluationCallback:
```
florian-hoenicke, line 10: I know, it is just a draft but before I forget to mention it - We need a documentation on class level and public method level.

JoanFM:
> > We will add unit tests for discovery and maybe parameters module
> 
> unittests need to exist for every module and class

Please make sure this is done

jina/optimizers/__init__.py
nan-wang: Yes, the docstring is a must-have.

jina/optimizers/discovery.py
```diff
@@ -0,0 +1,128 @@
...
+from jina.logging import default_logger as logger
+from jina.helper import yaml
+
+
+def _read_file(filename):
+    with open(filename, "r") as file:
+        return file.readlines()
+
+
+def _extract_executor_files(flows):
```
JoanFM, line 15: Not to be changed now, but this will make impossible to optimize executors user with ContainerPeas. 

In another iteration this must be addressed.

JoanFM: In ContainerPeas the only way to parametrize executors is through the argument "uses_internal"

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+            trial_parameters[param.env_var] = getattr(trial, param.method)(
+                **param.to_optuna_args()
+            )
+
+        trial_workspace = Path(
+            "JINA_WORKSPACE_" + "_".join([str(v) for v in trial_parameters.values()])
+        )
+        trial_parameters[self.workspace_env] = str(trial_workspace)
+
+        trial.workspace = trial_workspace
```
JoanFM, line 67: Why is the input value changed. Input parameters should remain unchanged

JoanFM (changes requested):

jina/optimizers/discovery.py
```diff
@@ -0,0 +1,128 @@
...
+
+    _write_new_executors(executor_configurations)
+    _write_optimization_parameter(
+        executor_configurations, target_file, overwrite_parameter_file
+    )
+
+
+def config_global_environment():
+    os.environ.setdefault("JINA_WORKSPACE", "workspace_discovery")
+    os.environ.setdefault("JINA_MYENCODER_TARGET_DIMENSION", "64")
```
JoanFM, line 114: why is this part of the default params?

jina/optimizers/flow_runner.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class FlowRunner:
+    def __init__(
+        self,
+        flow_yaml,
+        documents,
+        batch_size,
+        pod_dir,
+        task,  # this can be only index or search as it is used to call the flow API
```
JoanFM, line 18: task should be a parameter of run. a Flow should be described by the Pods and its parameters. what Request is sent to query is independent from it

jina/optimizers/parameters.py
```diff
@@ -0,0 +1,181 @@
...
+import ruamel.yaml
+
+from jina.helper import yaml
+
+
+class OptimizationParameter:
+    def __init__(
+        self,
+        parameter_name: str,
+        executor_name: str = None,
```
JoanFM, line 13: if they can be None please use Optional

jina/optimizers/parameters.py
JoanFM: If the use is not ready okey, but I think when it may come to the point of trying different strategies with the same ParameterSpace, this abstraction can make a significant difference.

tests/integration/optimizers/fashion/data.py
JoanFM: you can have a single executor with a dummy set of parameters that one can easily understand what is the best parameter set. 

You can have a DummyExecutor with 2 parameters with 3 options each for which there is only one convination that provides good results (for instance u could evaluate embedding distance) . And assert that the best set of parameters is the only possible one. That would be a good test documenting how the optimizer works and what it achieves.

JoanFM (changes requested):

jina/optimizers/discovery.py
bhavsarpratik: Max is going to push the final changes for this file

jina/optimizers/discovery.py
bhavsarpratik: This will be removed by Max

jina/optimizers/flow_runner.py
bhavsarpratik: This will cause problem with `MultiFlowRunner`'s run as we will have to pass the `task` to it now.

JoanFM: if we have MultiFlowRunner inherit itself from FlowRunner (or a common BaseClass), this would not be a problem because they would have the same interface.

jina/optimizers/__init__.py
```diff
@@ -9,17 +9,16 @@
...
     def __init__(self, eval_name=None):
-        self.op_name = eval_name
+        self.eval_name = eval_name
         self.evaluation_values = {}
         self.n_docs = 0
 
-    def flush(self):
-        self.evaluation_values = {}
-        self.n_docs = 0
+    def get_fresh_callback(self):
```
JoanFM, line 16: I would have this logic in the caller of this method and not as a method here.

jina/optimizers/__init__.py
```diff
@@ -38,17 +37,17 @@ def __call__(self, response):
 class OptunaOptimizer:
     def __init__(
         self,
-        index_flow_runner,
-        eval_flow_runner,
+        multi_flow,
```
JoanFM, line 40: is it necessarily a multiFlow?

jina/optimizers/__init__.py
bhavsarpratik: Isn't this a general pattern of callback class we would want the user and our developers to follow?

jina/optimizers/__init__.py
bhavsarpratik: It can also be just one flow but has to be wrapped in `MultiFlowRunner`

JoanFM: that's why I think it would be good to have `MultiFlowRunner`  inherit from `FlowRunner`. But no problem

JoanFM:
@bhavsarpratik make sure u fix the conflicts

tests/integration/optimizers/fashion/data.py
```diff
@@ -0,0 +1,58 @@
...
+            'filename': os.path.join(data_directory, 'index-labels'),
+        },
+        'query-labels': {
+            'url': 'http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/t10k-labels-idx1-ubyte.gz',
+            'filename': os.path.join(data_directory, 'query-labels'),
+        },
+        'index': {
+            'url': 'http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/train-images-idx3-ubyte.gz',
+            'filename': os.path.join(data_directory, 'index'),
+        },
```
hanxiao, line 49: all these functions are redundant, why not `import` them from `jina` directly?

tests/integration/optimizers/test_optimizer.py
```diff
@@ -0,0 +1,12 @@
+from .optimize import main
+
+
+def test_optimizer():
+    """This will run complete optimisation.
+    Todo: fix number of index and query doc so that test doesnt run long
```
hanxiao, line 6: correct syntax is `# TODO: xxx (Reminder, Date)`

tests/integration/optimizers/fashion/data.py
JoanFM: this test will disappear and be much more simplified

jina/optimizers/__init__.py
```diff
@@ -59,36 +89,50 @@ def _trial_parameter_sampler(self, trial):
...
+            'JINA_WORKSPACE_' + '_'.join([str(v) for v in trial_parameters.values()])
         )
         trial_parameters[self.workspace_env] = str(trial_workspace)
 
         trial.workspace = trial_workspace
         return trial_parameters
 
     def _objective(self, trial):
-        self.multi_flow.flows[self.eval_flow_index].callback = self.multi_flow.flows[self.eval_flow_index].callback.get_fresh_callback()
+        self.multi_flow.flows[self.eval_flow_index].callback = self.multi_flow.flows[
```
JoanFM, line 100:
```suggestion
       eval_flow = self.multi_flow.flows[self.eval_flow_index]
       eval_flow.callback = eval_flow.callback.get_fresh_callback()
```

jina/optimizers/__init__.py
```diff
@@ -59,36 +89,50 @@ def _trial_parameter_sampler(self, trial):
...
+            self.eval_flow_index
+        ].callback.get_fresh_callback()
         trial_parameters = self._trial_parameter_sampler(trial)
         self.multi_flow.run(trial_parameters, workspace=trial.workspace)
-        evaluation_values = self.multi_flow.flows[self.eval_flow_index].callback.get_mean_evaluation()
+        evaluation_values = self.multi_flow.flows[
```
JoanFM, line 105:
```suggestion
        evaluation_values = eval_flow.callback.get_mean_evaluation()
```

jina/optimizers/__init__.py
```diff
@@ -1,22 +1,35 @@
...
+    def __init__(self, eval_name: Optional[str] = None):
+        """
+        :param eval_name: evaluation name as required by evaluator
+        """
         self.eval_name = eval_name
-        self.evaluation_values = {}
+        self.evaluation_values = defaultdict(float)
         self.n_docs = 0
 
     def get_fresh_callback(self):
```
JoanFM, line 27: this should be a classmethod.

JoanFM (changes requested):

jina/optimizers/flow_runner.py
```diff
@@ -24,13 +24,17 @@ def __init__(
         :param documents: iterator with list or generator for getting the documents
         :param batch_size: batch size used in the flow
         :param task: task of the flow which can be `index` or `search`
```
JoanFM, line 26: task is better passed as argument of `run` I think, like this the `FlowRunner` instance is just dependant of the `Flow`. `task` is just syntax sugar for querying in jina

JoanFM (changes requested):
Minor changes plus missing unittests for `Optimizer` and `EvaluationCallback`

jina/optimizers/flow_runner.py
bhavsarpratik: I am not sure of a situation where I will use the same flow for both `index` and `query`. If a flow is always linked to a task, let's go with the current design.

tests/unit/optimizers/test_parameters.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.mark.parametrize('paramter_class, inputs, outputs', arg_dict)
+def test_parameters(paramter_class, inputs, outputs):
+    param = paramter_class(**inputs)
+    assert param.to_optuna_args() == outputs
+
+
+def test_parameter_file_loading():
+    load_optimization_parameters(cur_dir / 'parameters.yml')
```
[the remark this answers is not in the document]
bhavsarpratik: This test is to check all parameter classes are registered in yaml otherwise it will throw error like `!xxxx` not found

jina/optimizers/flow_runner.py
JoanFM: a Flow is by design not linked to a task

JoanFM: but it is okey leave it like this

tests/unit/optimizers/test_parameters.py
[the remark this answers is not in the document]
JoanFM: but it has a return value and it needs to be asserted, also add type hinting to that function

tests/unit/optimizers/test_flow_runner.py
```diff
@@ -0,0 +1,51 @@
+from pathlib import Path
+
+from tests import random_docs
+from jina.optimizers.flow_runner import FlowRunner
+
+import pytest
+
+cur_dir = Path(__file__).parent
+
+def test_flow_runner(tmpdir, mocker):
```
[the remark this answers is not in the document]
JoanFM: This conversation is resolved because there is a commit coming with these changes?

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+        **kwargs
+    ):
+        """
+        :param n_trials: evaluation trials to be run
+        :param sampler: optuna sampler
+        :param direction: direction of the optimization from either of `maximize` or `minimize`
+        :param seed: random seed for reproducibility
+        :param kwargs: extra parameters for optuna sampler
+        """
+        import optuna
```
JoanFM, line 143: Use `ImportExtensions` to guide the user when this is not installed

JoanFM (changes requested):

tests/unit/optimizers/test_flow_runner.py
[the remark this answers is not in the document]
bhavsarpratik: Yes

JoanFM:
optuna needs to be added to `devel` tag so that is installed for integration tests

jina/optimizers/__init__.py
bhavsarpratik: For this to be a `staticmethod`, it needs to have the `args` needed to create it which we don't have in `OptunaOptmizer`

JoanFM: u can extract it from the last `evaluationCallback`.

u would do `self.callback = EvaluationCallback.get_fresh_callback(self.callback.eval_name)`

bhavsarpratik: If we are using from the `self` it doesn't change Optimizer's behaviour since earlier it also uses from self. The intention was not to get a new callback with new op_name but the same op_name. Please confirm the change

JoanFM: sure, no problem

bhavsarpratik:
@JoanFM I added `devel` to optuna but still it failed to import in the test

JoanFM:
> @JoanFM I added `devel` to optuna but still it failed to import in the test

sorry, the tag is `test`

jina/optimizers/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+        logger.info(colored(f'Avg {op_name}: {eval_score}', 'green'))
+        return eval_score
+
+    def optimize_flow(
+        self,
+        n_trials: int,
+        sampler: str = 'TPESampler',
+        direction: str = 'maximize',
+        seed: int = 42,
+        result_processor = OptunaResultProcessor,
```
JoanFM, line 131: remove this and create direction OptunaResultProcessor in return

JoanFM (changes requested):

JoanFM (approved):
Approving as first iteration

jina/optimizers/flow_runner.py
```diff
@@ -0,0 +1,102 @@
...
+class FlowRunner:
+    """Module to define and run a flow."""
+
+    def __init__(
+        self,
+        flow_yaml: str,
+        documents: Iterator,
+        batch_size: int,
+        task: str,  # this can be only index or search as it is used to call the flow API
+        callback: Optional = None,
```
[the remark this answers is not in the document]
nan-wang: Maybe this is better? It will be a `callable` taking any number of arguments and returning `None`.
```
Optional[Callable[..., None]]
```

nan-wang (dismissed):
LGTM👍

JoanFM (approved):

PR #1463, opened 2020-12-15
fix: jupyter exception handling

hanxiao:


jina/helper.py
```diff
@@ -714,9 +729,18 @@ def run(self):
...
+            thread = RunThread()
+            thread.start()
+            thread.join()
+            try:
+                return thread.result
+            except AttributeError:
+                from .excepts import BadClient
+                raise BadClient('something wrong when running the eventloop, result can not be retrieved')
+        else:
+            raise RuntimeError('you have an eventloop running but not using jupyter/ipython, '
```
deepankarm, line 742: `jinad` runs inside `fastapi+uvicorn` which has a default event loop (again based on uvloop). Why do we want to raise a `RuntimeError`?

jina/flow/asyncio.py
```diff
@@ -0,0 +1,299 @@
...
+
+from . import Flow
+from ..clients import InputFnType
+from ..clients.asyncio import AsyncClient
+
+if False:
+    import numpy as np
+
+
+class AsyncFlow(Flow):
```
deepankarm, line 12: Should we implement an async context manager for AsyncFlow? 
```
async with AsyncFlow().add() as f:
   await f.index(['abc'])
```

hanxiao: no. `async with` is only implemented when one wants to waive the control during `__enter__` `__exit__`. We don't have such needs.

deepankarm: This is because we have taken special care not to propagate the `async...await` to the caller. e.g. - `Client` and `AsyncClient` should be context managers, so that they can be used outside a Flow with ease. Similarly `GatewayPea`  & `RESTGatewayPea` can be an async context manager, even though I don't see heavy usage for the same.

PR #1465, opened 2020-12-15
feat: helloworld evaluate html

JoanFM:
**Changes introduced**
Add Evaluation value to helllo world html

### DIFF
--- a/jina/helloworld/helper.py
+++ b/jina/helloworld/helper.py
@@ -5,6 +5,7 @@ import gzip
 import os
 import urllib.request
 import webbrowser
+import random
 
 import numpy as np
 from pkg_resources import resource_filename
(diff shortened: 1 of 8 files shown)


JoanFM:
image (10)

jina/resources/helloworld.html
```diff
@@ -23,6 +23,7 @@ <h2 style="color: #1E6E73">What just happened?</h2>
         </div>
         
         <hr style="border: 1px dashed #009999;">
+        <th align="left"><h3 style="color: #1E6E73">Label Evaluation {% EVALUATION %}</h3></th>
```
bhavsarpratik, line 26: We can add here x% matched true label in top 5 result

bhavsarpratik (changes requested):

jina/resources/helloworld.html
JoanFM: I prefer to be like this in case we ever change eval_at and do not remember

bhavsarpratik: I think the % number itself doesn't give clarity on the evaluation done and needs some definition. For auto-update, you can load the yaml and pass its eval_at to this string

maximilianwerk:
Can we have the evaluator disablelable (however you write, that you are able to disable it)? For the speed test it would be nice to not run the evaluation, since that would be a quite a change since the very beginning in the measured functionality. Or we say, we don't care and it should be part of the hello-world. Would also be fine for me. Just wanted to raise awareness.

JoanFM:
> Can we have the evaluator disablelable (however you write, that you are able to disable it)? For the speed test it would be nice to not run the evaluation, since that would be a quite a change since the very beginning in the measured functionality. Or we say, we don't care and it should be part of the hello-world. Would also be fine for me. Just wanted to raise awareness.



> Can we have the evaluator disablelable (however you write, that you are able to disable it)? For the speed test it would be nice to not run the evaluation, since that would be a quite a change since the very beginning in the measured functionality. Or we say, we don't care and it should be part of the hello-world. Would also be fine for me. Just wanted to raise awareness.

latecny tracking works on its own way not running hello world directly (which I find odd actually). but safe for now

jina/helloworld/helper.py
```diff
@@ -5,15 +5,59 @@
...
+        d = Document(content=target['index']['data'][j])
+        d.update_id()
+        label_int = target['index-labels']['data'][j][0]
+        d.tags['label_id'] = str(label_int)
+        yield d
+
+
+def query_generator(num_docs: int, target: dict):
+    for n in range(num_docs):
+        n = random.randint(0, 1000)  # there are 10000 query examples, so that's the limit
```
nan-wang, line 55:
```suggestion
        num_data = len(target['query-labels']['data'])
        n = random.randint(0, num_data)

nan-wang (changes requested):
minor changes

bhavsarpratik (dismissed):

jina/helloworld/helper.py
JoanFM: Done!

nan-wang (approved):
LGTM👍

PR #1466, opened 2020-12-15
feat(driver): add docidcache for content

cristianmtr:
- Make DocIDCache capable of detecting collisions on content-level.
- NOTE: Tests are only on the `Indexer` level, since the Drivers for Update/Del. haven't been implemented yet. Integration tests will be required when that is done.

**Discussion required:** do we set the `field` (`id`/`content_hash`) at the Driver level or at the Indexer level? Atm it's done at the Driver level.

### DIFF
--- a/jina/drivers/cache.py
+++ b/jina/drivers/cache.py
@@ -1,31 +1,42 @@
 from typing import Any, Dict
 
 from .index import BaseIndexDriver
+from ..executors.indexers.cache import DATA_FIELD
 
 if False:
     from .. import Document
     from ..types.sets import DocumentSet
 
+ID_KEY = 'id'
+CONTENT_HASH_KEY = 'content_hash'
+
 
 class BaseCacheDriver(BaseIndexDriver):
     """
     The driver related to :class:`BaseCache`
-
     """
+
+    supported_fields = [ID_KEY, CONTENT_HASH_KEY]
+    default_field = ID_KEY
 
     def __init__(self, with_serialization: bool = False, *args, **kwargs):
         """
-
         :param with_serialization: feed serialized doc to the CacheIndexer
         :param args:
         :param kwargs:
         """
         self.with_serialization = with_serialization
         super().__init__(*args, **kwargs)
+        self.field = kwargs.get('field', self.default_field)
+        if self.field not in self.supported_fields:
+            raise ValueError(f"Field '{self.field}' not in supported list of {self.supported_fields}")
 
     def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
         for d in docs:
-            result = self.exec[d.id]
+            if self.field == ID_KEY:
+                result = self.exec[d.id]
+            elif self.field == CONTENT_HASH_KEY:
+                result = self.exec[d.content_hash]
             if result is None:
                 self.on_miss(d)
             else:
@@ -33,17 +44,19 @@ class BaseCacheDriver(BaseIndexDriver):
 
     def on_miss(self, doc: 'Document') -> None:
         """Function to call when doc is missing, the default behavior is add to cache when miss
-
         :param doc: the document in the request but missed in the cache
         """
+        data = doc.id
+        if self.field == CONTENT_HASH_KEY:
+            data = doc.content_hash
+
         if self.with_serialization:
-            self.exec_fn(doc.id, doc.SerializeToString())
+            self.exec_fn(doc.id, doc.SerializeToString(), **{DATA_FIELD: data})
         else:
-            self.exec_fn(doc.id)
+            self.exec_fn(doc.id, **{DATA_FIELD: data})
 
     def on_hit(self, req_doc: 'Document', hit_result: Any) -> None:
         """ Function to call when doc is hit
-
         :param req_doc: the document in the request and hitted in the cache
         :param hit_result: the hit result returned by the cache
         :return:
--- a/jina/executors/indexers/cache.py
+++ b/jina/executors/indexers/cache.py
@@ -1,11 +1,11 @@
 import tempfile
-from typing import Optional
-
-import numpy as np
+from typing import Optional, Iterator
 
 from . import BaseKVIndexer
-from ...helper import cached_property
+from ...helper import deprecated_alias, check_keys_exist
 from ...types.document.uid import UniqueId
+
+DATA_FIELD = 'data'
 
 
 class BaseCache(BaseKVIndexer):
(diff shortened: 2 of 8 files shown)


jina/drivers/cache.py
```diff
@@ -22,10 +27,21 @@ def __init__(self, with_serialization: bool = False, *args, **kwargs):
...
+            raise ValueError(f"Field '{self.field}' not in supported list of {self.supported_fields}")
 
     def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
         for d in docs:
-            result = self.exec[d.id]
+            # TODO waiting for new API from docId refactor PR
+            if self.field == 'id':
+                result = self.exec[d.id]
+            elif self.field == 'content':
+                # TODO what about blob?
```
JoanFM, line 40: blob is one of the potential contents

cristianmtr: yes but it's a different field, it's `.blob` So should that be a different field for the driver parameter? Because `.text` is the same as `.content`.

jina/resources/executors._unique.yml
```diff
@@ -1,6 +1,9 @@
 !DocIDCache
 with:
   index_path: cache.tmp
+  # TODO Should we rename _unique to _unique_id ?
```
maximilianwerk, line 4: No, it should be `_unique_content_hash` or so. Since the old `id` is what the current `content_hash` is.

tests/unit/drivers/test_cache_content_driver.py
```diff
@@ -0,0 +1,42 @@
...
+from jina import DocumentSet
+from jina.drivers.cache import BaseCacheDriver
+from jina.executors.indexers.cache import DocIDCache
+from jina.proto import jina_pb2
+from jina.types.document import uid, Document, bytes2id
+from tests import random_docs
+from tests.unit.drivers.test_cache_id_driver import MockCacheDriver
+
+
+def test_cache_content_driver(tmp_path):
```
maximilianwerk, line 16: Yesterday, we had the example with for different documents, which collides on either `id` or `content`. Before continuing implementing, I'd put in the four use cases to have a complete view on what should be solved.

maximilianwerk (commented):
Good approach to first find all the construction places.

jina/resources/executors._unique.yml
cristianmtr: Yes, but now we will have caches on the two possible fields:

-  `id` (which is not a hash, but passed by the user or assigned randomly)
- `content_hash` (which was the old `id`).

So we will need two `_unique_something.yml`, right?

The old one did you use `content_hash`, but it was actually using the field `id`. So the question is, which perspective do we preserve? The caller (which only cares it's `id`), or internal (which knows the old `id` is the new `content_hash`)?

tests/unit/drivers/test_cache_content_driver.py
cristianmtr: Oh yeah, missed those. Will add them

jina/drivers/__init__.py
```diff
@@ -189,6 +189,7 @@ def expect_parts(self) -> int:
     @property
     def msg(self) -> 'Message':
         """Get the current request, shortcut to ``self.pea.message``"""
+        # FIXME pea is None when running `test_cache_content_driver_same_content`
```
JoanFM, line 192: the drivers tests are done without peas to allow them being `unittesty`

jina/resources/executors._unique.yml
JoanFM: What I understood this will be `_unique` and will ensure that the document is not reindexed and so. I think `unique_something` would expose too much complexity to the user. What this does is guarantee that no duplicated is indexed while keeping it in sync when deletions or updates happen

tests/unit/drivers/test_cache_content_driver.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+import pytest
+
+from jina import DocumentSet
+from jina.drivers.cache import BaseCacheDriver, CONTENT_HASH_KEY
+from jina.executors.indexers.cache import DocIDCache
+from jina.proto import jina_pb2
+from jina.types.document import uid, Document, bytes2id
+from tests import random_docs
+from tests.unit.drivers.test_cache_id_driver import MockCacheDriver
```
JoanFM, line 13: avoid importing from different tests. I actually do not understand why are we having a different test file if the class is the same?

jina/resources/executors._unique.yml
cristianmtr: But "duplicated" based on what field? `id` should be the default. Even though `id` now means the random id assigned by user (or us), and not `content_hash` (which is what `id` used to be/mean)

jina/executors/indexers/cache.py
```diff
@@ -24,33 +24,58 @@ def post_init(self):
...
+    # TODO do we really need a dict-like object?
+    store = {}
+
     def __init__(self, index_filename: str = None, *args, **kwargs):
         if not index_filename:
             # create a new temp file if not exist
             index_filename = tempfile.NamedTemporaryFile(delete=False).name
         super().__init__(index_filename, *args, **kwargs)
 
     def add(self, doc_id: UniqueId, *args, **kwargs):
```
JoanFM, line 37: how can I add something to the `hash -> doc_id` ?

cristianmtr: What do you mean?

Atm the API for `add` is 

(from the Driver)

```
            self.exec_fn(doc.id, **{DATA_FIELD: key})
```

where the `key` can be either the `doc.id` or the `content_hash`. I think this could be simplified somehow, but not sure yet, as I'm still unclear on the expected use cases. 

So atm the data structure is either

`docid --> docid` (redundant, could be simplified somehow)
or
`docid --> content_hash`.

Why would you need to add `content_hash` -> `doc_id`?

JoanFM: For Incremental Indexing, one may want to guarantee that the document with that `content` is not there and so u would check if that `content_hash` has been seen.

cristianmtr: But that's a query, right?

you'd first need to create a DocIdCache with a Driver that use the field `content_hash` for checking.

Then you add docs.

Then you do `query([content_hash])`

cristianmtr: Is this what you mean? https://github.com/jina-ai/jina/blob/958a29d84a7e1939c0e603fd65dc145536e67c22/tests/unit/drivers/test_cache_driver.py#L148

cristianmtr:
Closing, because of `commitlint` CI issue. New PR here: https://github.com/jina-ai/jina/pull/1510

Issue #1382, opened 2020-12-02
Provide batching decorator ready to batch `Ranker` input data (query_meta, doc_meta, prev_scores)


JoanFM:
**Describe the feature**
I am not sure if it is ready, but we need to make sure that `Rankers` can leverage the batching decorator capability.

**Motivation**
We can have `Neural Network` based rankers that could highly benefit from batching their inputs

PR #1467, opened 2020-12-15
feat: add batching support for rankers

deepampatel:
Previously i had tried to merge all batching decorators into one, after your comments below i have reverted the commit back and added a new `@batching_ranker_input` decorator  only for ranker executors. The `@batching` and `@batching_mutli_input` are untouched.


jina/executors/decorators.py
```diff
@@ -213,77 +220,78 @@ def batching(func: Callable[[Any], np.ndarray] = None,
...
-
+                    
                 @batching(batch_size = 64)
                 def train(self, batch: 'numpy.ndarray', *args, **kwargs):
                     gpu_train(batch)
-    """
 
+            class MultiModalExecutor:
+                
+                @batching(batch_size = 64, slice_on=[1,2])
```
JoanFM, line 244: Just as a comment, there is a special batching decorator for `MultiModalExecutor` itself.

JoanFM (commented):
Hey @deepampatel ,

Thank you very much for your contribution.

As a comment, I would just recommend do not try to fit every type of batching pattern in a single function since it can become really unreadable and unmantainable. So if you feel like you need a special function for some special executor or type of input you are welcome to add it!

Great work!

jina/executors/decorators.py
deepampatel: > Hey @deepampatel ,
> 
> Thank you very much for your contribution.
> 
> As a comment, I would just recommend do not try to fit every type of batching pattern in a single function since it can become really unreadable and unmantainable. So if you feel like you need a special function for some special executor or type of input you are welcome to add it!
> 
> Great work!

@JoanFM  Thanks for the quick feedback. I agree with you on the special function for special executor class.  i will add a separate class for dictionary input type.
WDYT about having `slice_on` parameter as a `Union[int, List[int]]` instead of just `int` value in `@batching_multi_input` . Dont have any particular example in mind rn where this can be used but this thought came up when i was trying to merge both decorators.

jina/executors/decorators.py
```diff
@@ -375,3 +382,107 @@ def arg_wrapper(*args, **kwargs):
...
+                ) -> 'np.ndarray': 
+                ...
+
+                @batching(batch_size = 64, slice_on=2, num_data = 1)
+                def score(
+                    self, query_meta: Dict, old_match_scores: Dict, match_meta: Dict
+                ) -> 'np.ndarray': 
+                ...
+
+                @batching(batch_size = 64, slice_on=[2])
```
JoanFM, line 426: wrong docstring right?

deepampatel: yes

JoanFM:
> This commit is a feat but also a refactor commit .
> 
> 1. feat: add support for batching dictionary data, so it can be used to add batching support for rankers.
> 2. wip: refactor `@batching` and  `@batching_multi_input` into a single decorator
> 
> ##### TODO : add support for label_on and ordinal_idx_arg

Now I see the description, please unless it looks very very clean, avoid having batching and barching_multi_input inside a single decorator. it was split for better readability

deepampatel:
> > This commit is a feat but also a refactor commit .
> > 
> > 1. feat: add support for batching dictionary data, so it can be used to add batching support for rankers.
> > 2. wip: refactor `@batching` and  `@batching_multi_input` into a single decorator
> > 
> > ##### TODO : add support for label_on and ordinal_idx_arg
> 
> Now I see the description, please unless it looks very very clean, avoid having batching and barching_multi_input inside a single decorator. it was split for better readability


Previously i had tried to merge all batching decorators into one, after your comments above i have reverted the commit back and added a new `@batching_ranker_input` decorator only for ranker executors. The `@batching`and `@batching_multi_input` are untouched. @JoanFM

jina/executors/decorators.py
```diff
@@ -365,3 +372,107 @@ def arg_wrapper(*args, **kwargs):
         return _batching(func)
     else:
         return _batching
+
+def batching_ranker_input(func: Callable[[Any], np.ndarray] = None,
+                         batch_size: Union[int, Callable] = None,
+                         num_batch: Optional[int] = None,
+                         split_over_axis: int = 0,
+                         merge_over_axis: int = 0,
```
JoanFM, line 380: since it is only for ranker, this will not be different than 0, so let's remove this. Also the `split_over_axis` (will it be different than 0?

JoanFM: the same for `slice_on`?

jina/executors/decorators.py
```diff
@@ -365,3 +372,107 @@ def arg_wrapper(*args, **kwargs):
...
+            _num_data = num_data
+            if _num_data is not None:
+                if isinstance(_slice_on, List):
+                    raise ValueError(f'When using num_data in @batching_ranker, an integer value '
+                'for slice_on (default=1) is expected.')
+                _slice_on = [_slice_on + i for i in range(0, _num_data)]
+            else:
+                if isinstance(_slice_on, int):
+                    _slice_on = [_slice_on]
+                    _num_data = 1
```
JoanFM, line 436: _num_data is not used after here

JoanFM (changes requested):

jina/executors/decorators.py
deepampatel: `slice_on` i feel can be a configurable parameter.

JoanFM: at least `slice_on` should be set to the default that the `rankers` would use

jina/executors/decorators.py
```diff
@@ -365,3 +372,95 @@ def arg_wrapper(*args, **kwargs):
...
+                    
+    """
+    def _batching(func):
+        @wraps(func)
+        def arg_wrapper(*args, **kwargs):
+            _slice_on = slice_on
+            _num_data = num_data
+            if _num_data is not None:
+                if isinstance(_slice_on, List):
+                    raise ValueError(f'When using num_data in @batching_ranker_input, an integer value '
```
JoanFM, line 420: then please fix the `slice_on` type hint, since it suggests a `List` is accepted

JoanFM: I would just adapt the `type hint` and remove this check

jina/executors/decorators.py
```diff
@@ -365,3 +372,95 @@ def arg_wrapper(*args, **kwargs):
         return _batching(func)
     else:
         return _batching
+
+def batching_ranker_input(func: Callable[[Any], np.ndarray] = None,
+                         batch_size: Union[int, Callable] = None,
+                         num_batch: Optional[int] = None,
+                         slice_on: Union[int, List[int]] = 2,
+                         num_data: int = None) -> Any:
```
JoanFM, line 380: num_data default is 3 I think for this case no?

jina/executors/decorators.py
JoanFM: Avoid this noisy part, consider num_data is never None (type hint suggests so)

JoanFM (changes requested):
There is a lot of code that is common (or can be made common) between `batching_multi_input` and `batching_input_ranker`. I am sure some helper functions can be made to unify a little bit of both

jina/executors/decorators.py
deepampatel: I was thinking, `slice_on` and `num_data` when used together, allow you to batch consecutive parameters. There might be a case where you want to batch non - consecutive parameters, lets say [1,3], in those cases we can have `slice_on` as list of indices. So whenever  `slice_on` is passed as a list, `num_data` (type hint needs to be updated to optional) cannot be used. wdyt?

jina/executors/decorators.py
deepampatel: out of `self, query_meta, old_match_scores, match_meta` we only want to batch `old_match_scores` . other two are meta which dont need to batched.

jina/executors/decorators.py
JoanFM: Since there is not yet a case where we require this, I'd prefer to keep it simple

jina/executors/decorators.py
```diff
@@ -437,19 +421,19 @@ def arg_wrapper(*args, **kwargs):
...
-            data_iterators = [batch_iterator(_get_slice(args[i],total_size), b_size) for i in _slice_on]
+            yield_dict = [isinstance(args[slice_on + i], Dict) for i in range(0,num_data)]
+            data_iterators = [batch_iterator(_get_slice(args[slice_on + i],total_size), b_size) for i in range(0, num_data)]
 
             for batch in data_iterators[0]:
-                args[_slice_on[0]] = dict(batch) if yield_dict[0] else batch
-                for idx,slice_idx in enumerate(_slice_on[1:]):
-                    batch_idx = next(data_iterators[idx+1])
-                    if yield_dict[idx+1]:
+                args[slice_on] = dict(batch) if yield_dict[0] else batch
```
JoanFM, line 431: why is `yield_dict` needed? what is batch returning?

deepampatel: batch returns a list of tuples of (key, value)

JoanFM: ok, I see. Maybe it is better to consider in `batch_iterator` the case where a `dict` is provided?

JoanFM: It feels that until obtaining the `iterators` the code with `batching_multi_input` may be merged?

jina/helper.py
```diff
@@ -107,7 +107,10 @@ def batch_iterator(data: Iterable[Any], batch_size: int, axis: int = 0,
         data = iter(data)
         # as iterator, there is no way to know the length of it
         while True:
-            chunk = tuple(islice(data, batch_size))
+            if yield_dict:
```
JoanFM, line 110: better to add before `elif(data, Iterable)` an `elif(data, dict)`, like this it will work without the need of the `yield_dict` parameter.

JoanFM (changes requested):

jina/helper.py
deepampatel: `data` needs to be sliced before passing it to `batch_iterator` because we have `num_batch` that inturn sets the  `total_size` . There are two options i think : 
1. Pass `total_size` to `batch_iterator` and slice the data inside the `batch_iterator`
2. Slice it before passing it to `batch_iterator` and pass 'yield_dict`. In this case dict when sliced is converted to a iterable .
wdyt?

JoanFM: OK I see, let's keep passing `yield_dict` as argument so that we do not affect current behavior (the code is a little verbose already)

jina/executors/decorators.py
```diff
@@ -365,3 +372,77 @@ def arg_wrapper(*args, **kwargs):
...
+            default_logger.debug(
+                f'batching enabled for {func.__qualname__} batch_size={b_size} '
+                f'num_batch={num_batch}')
+
+            # assume all datas have the same length
+            full_data_size = _get_size(args[slice_on])
+            total_size = _get_total_size(full_data_size, b_size, num_batch)
+            final_result = []
+            yield_dict = [isinstance(args[slice_on + i], Dict) for i in range(0,num_data)]
+            data_iterators = [batch_iterator(_get_slice(args[slice_on + i],total_size), b_size , yield_dict=yield_dict[i]) for i in range(0, num_data)]
```
JoanFM, line 428: let's adapt this part together with `batching_multi_input` if possible. `get_slice` can be used also there right?. Also separating comma before `total_size`

deepampatel: Since both `batch_multi_input` and `batch_ranker_input` are almost similar, we can completely remove `batch_ranker_input` and update `batching` and `batch_multi_input` with `get_slice` and `yield_dict` to handle dictionary data. or  keep 'batch_ranker_input' separate?

JoanFM: yes lets try, but yield_dict can be avoided in regular batching. it can be set to default false in the batch_decorator

deepampatel: if we are removing `batching_ranker_input` maybe we should add dictionary support to `batching` no?

JoanFM: maybe in the future, but now I would prefer to keep it conservative and just touch where it may be used

deepampatel: For rankers we actually need only one parameter `old_match_scores` to be batched. So we still need to use `batch_multi_input`  ?

JoanFM: not true, you may want to batch also match meta

JoanFM (approved):

deepampatel:
#1382

JoanFM:
recheckcla

JoanFM (changes requested):
Hey @deepampatel ,

thanks for your great contribution, now we just need you to sign the CLA before merging the PR

deepampatel:

JoanFM (approved):

Issue #1468, opened 2020-12-15
Driver handling Reduce and Route logic together


JoanFM:
**Describe the bug**
When investigating #1302 I found that the issue was not that IDLE was not handled by ControlReqDriver but that we were not using `RouteDriver` in the `HeadPea`. So then, how can I use `HeadPea` to `Reduce`  incoming messages and at the same time to `Route` inside the `Pod`?

JoanFM:
this could be handled by allowing Composition of ControlDrivers. In comparison with ExecutableDrivers, sequential application does not seem to make much sense

hanxiao:
can you add a minimum test case to show this problem?

JoanFM:
I did @hanxiao added in #1469

JoanFM:
Test added in #1469

PR #1469, opened 2020-12-16
test: add test for issue 1468

JoanFM:
**Changes introduced**
Test showing problem with #1468 

It hangs unless wither of the 2 things happen:
- Remote _merge_chunks
- Reduce parallel to 1


tests/integration/issues/github_1468/flow.yml
```diff
@@ -0,0 +1,17 @@
...
+    uses: pods/filter_chunk_1.yml
+    read_only: true
+  filter_chunk_2:
+    uses: pods/filter_chunk_1.yml
+    read_only: true
+    needs: [gateway]
+  merge_route_process:
+    uses: _forward
+    parallel: 2
+    uses_before: _merge_chunks
```
hanxiao, line 16: is it intentional, you could write `uses: _merge_chunks`

is it suggesting the problem only exists when you set both?

JoanFM: `uses` could be anything else. 
I want `_merge_chunks` to accumulate all the `chunks` from the 2 incoming paths in the HeadPea, but at the same time HeadPea must handle Routing.

hanxiao: but why a `needs` here does not fit your need?

```yaml
  merge_route_process:
    uses: _merge_chunks
    parallel: 2
#    uses_before: _merge_chunks
    needs: [filter_chunk_1, filter_chunk_2]

```

JoanFM: ah will this merge everything? then yes I will test tomorrow. But then the load balancing could fail to send every pair of chunks to the same shard right?

hanxiao: no i mean you can always separate it into two pods, e.g.
```yaml
!Flow
version: '1.0'
pods:
  - name: filter_chunk_1
    uses: pods/filter_chunk_1.yml
    read_only: true
  - name: filter_chunk_2
    uses: pods/filter_chunk_1.yml
    read_only: true
    needs: [gateway]
  - needs: [filter_chunk_1, filter_chunk_2]
  - name: merge_route_process
    parallel: 2

```

JoanFM: ok yes good. However there is big way for improvement. The big issue that I see is that then just changing the amount of parallelization to the same Pod is not enough which adds complexity to the user

JoanFM:
I still think @hanxiao it can make sense to merge it and look at it and see if we can improve this

hanxiao (approved):

PR #1470, opened 2020-12-16
feat: add mapping executor, test drivers

JoanFM:
@nan-wang this is more or less what I had in mind in big words, but it does indeed open the door to some unmantainable complexity

### DIFF
--- a/jina/docker/hub_test_generators.py
+++ b/jina/docker/hub_test_generators.py
@@ -1,0 +1,29 @@
+import random
+import string
+
+import numpy as np
+
+from jina import Document
+
+
+def text_document_generator(num_docs):
+    for i in range(num_docs):
+        length = random.randint()
+        document = Document(text=''.join(random.choices(string.ascii_uppercase + string.digits, k=length)))
+        yield document
+
+
+def image_document_generator(num_docs):
+    length = random.randint()
+    height = random.randint()
+    for i in range(num_docs):
+        document = Document(blob=np.random.rand(height, length, 3))
+        yield document
+
+
+def multimodal_document_generator(num_docs):
+    length = random.randint()
+    height = random.randint()
+    for i in range(num_docs):
+        document = Document(blob=np.random.rand(height, length, 3))
+        yield document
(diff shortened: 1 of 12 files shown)


jina/executors/__init__.py
```diff
@@ -424,6 +424,19 @@ def load_config(cls: Type[AnyExecutor], source: Union[str, TextIO], separated_wo
...
+        """Build an executor from a YAML file.
+
+        :param source: the file path of the YAML file or a ``TextIO`` stream to be loaded from
+        :param separated_workspace: the dump and data files associated to this executor will be stored separately for
+                each parallel pea, which will be indexed by the ``pea_id``
+        :param pea_id: the id of the storage of this parallel pea, only effective when ``separated_workspace=True``
+        :return: an executor object
+        """
+        return cls.load_config(source, separated_workspace, pea_id, read_only)
+
```
nan-wang, line 439: Where this function is used?

nan-wang (commented):
This looks good. We can stick to this strategy for the time being.

jina/executors/__init__.py
JoanFM: this is not ready yet. it wouls be uses bu hubio to load the executor class (not the instance) to see which test_drivers has

jina/docker/hub_test_generators.py
```diff
@@ -0,0 +1,29 @@
...
+def text_document_generator(num_docs):
+    for i in range(num_docs):
+        length = random.randint()
+        document = Document(text=''.join(random.choices(string.ascii_uppercase + string.digits, k=length)))
+        yield document
+
+
+def image_document_generator(num_docs):
+    length = random.randint()
+    height = random.randint()
```
maximilianwerk, line 18: They should be parameters and only random if not set. Imagine someone does
```
index_docs = image_document_generator(10)
index index_docs...
query_docs = image_document_generator(5)
query query_docs
```
Then they would not have equal size which might crash out of the box.

JoanFM: Do not look at the details now. this is very very early stage and far from being ready. this will not be used by users outside though

JoanFM:
This is just a Proof of Concept for @bwanglzu to keep in mind, but will close to avoid having noise in the list of Open PRs

Issue #1444, opened 2020-12-11
Custom Driver


janandreschweiger:
**Describe the feature**
The ability to add custom drivers or a new driver that fits our needs.

**Why?**
When using the EncodeDriver, the encode method of our executor is called for every document of our IndexRequest. (Unfortunately, all available executable drivers inherit from the BaseRecursiveDriver.) This is a problem, because we want to encode all sentences at once. Otherwise, the index-runtime is way too long. We use the sentence-transformers library which offers a much faster encoding of multiple sentences at one shot.

So in a nutshell, we need a Driver that executes the encoder with all the documents of the request only once.

Our custom encode-executor:
```
from jina.executors.encoders import BaseEncoder
from jina.helper import cached_property

class SentenceEncoder(BaseEncoder):
    def __init__(self,
            model_name: str = 'xlm-r-distilroberta-base-paraphrase-v1',
            max_words: int = 128,
            device: str = None, # None (auto), cpu, cuda, mkldnn, opengl, opencl, ideep, hip, msnpu, xla
            *args,
            **kwargs):
        super().__init__(*args, **kwargs)
        self.model_name = model_name
        self.max_words = max_words
        self.device = device

    @cached_property
    def model(self):
        from sentence_transformers import SentenceTransformer
        return SentenceTransformer(self.model_name, device=self.device)

    def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':  # IS CALLED FOR EVERY DOCUMENT OF THE REQUEST
            return self.model.encode(data)
```

It would be great if someone could help us. We really need this in order to apply jina.ai in our firm.
Thank you!

JoanFM:
Hey @janandreschweiger;

You can add custom drivers the same way you add custom executors using the py_modules parameter in the metas of the pod yaml definition.

JoanFM:
As per the need of such a driver, you can control with batch_size the amount of documents that are encoded at once.

Is your document of the semantic unit that you want to embed to?

JoanFM:
you can use the @batching decorator that controls how many documents of the request are considered at the same time for the function execution and which is specifically targeted for encoding in batches.

janandreschweiger:
Hey @JoanFM, thanks for helping me out again.

**Batching**
Unfortunately, using the batching decorator doesn't change anything:
```
# other code
    @batching(batch_size = 64)
    def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
[...]

PR #1483, opened 2020-12-18
feat: add logic to batch at encode driver level

JoanFM:
**Changes introduced**

Related to feature requested in #1444 .

Right now, `EncodeDriver` when working with `documents` at `granularity>0` can only batch documents to the executor that have a common parent_id. 

With this PR I try to add the capacity to "accumulated" or "batch" at driver level, so that `self.apply_all` is called with more documents in the `DocumentSet` and user can better leverage the benefits of `batching` at `executor` level.


JoanFM:
I am thinking it may be better to specialize it only at `Driver level` so that the `interfaces` properly match

jina/types/sets/cache.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class CacheDocumentSet:
+
+    def __init__(self,
+                 capacity: Optional[int] = None,
+                 *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.capacity = capacity
+        self._doc_set = DocumentSet()
```
rutujasurve94, line 12: We could also have an `id` for the `CacheDocumentSet` so that when a new set is created because `left capacity` is smaller than number of documents, the new `CacheSet` for the overflow has a subsequent identifier than the previous `CacheSet`.

jina/types/sets/cache.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @property
+    def left_capacity(self):
```
rutujasurve94, line 15: nit: `left_capacity` could be renamed to `available_capacity` as left indicates subtraction

rutujasurve94:
> I am thinking it may be better to specialize it only at `Driver level` so that the `interfaces` properly match

This is a great idea.

jina/types/sets/cache.py
JoanFM: I will have it private for `EncodeDriver` because it is very specific to help in this functionality, do not want to provide identifiers or have it be a primary type for Jina

JoanFM:
Enables #1493

jina/drivers/encode.py
```diff
@@ -1,23 +1,117 @@
...
+        docs = args[0]
+        if not force_flush and \
+                self.cache_set is not None and \
+                self.cache_set.available_capacity > 0:
+            left_docs = self.cache_set.cache(docs)
+            while len(left_docs) > 0:
+                self._apply_cache()
+                left_docs = self.cache_set.cache(left_docs)
+            if self.cache_set.available_capacity == 0:
+                self._apply_cache()
```
nan-wang, line 30: `func` is not used in this part, which makes it not a *real* decorator.

nan-wang (changes requested):

jina/drivers/encode.py
```diff
@@ -1,23 +1,117 @@
...
+        super().__init__(*args, **kwargs)
+        self.batch_size = batch_size
+        if self.batch_size:
+            self.cache_set = EncodeDriver.CacheDocumentSet(capacity=self.batch_size)
+        else:
+            self.cache_set = None
+
+    def __call__(self, *args, **kwargs):
+        self._traverse_apply(self.docs, *args, **kwargs)
+        self._apply_cache(force_flush=True, **kwargs)
```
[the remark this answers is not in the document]
JoanFM: Because there may be some left-overs from the `request` (Imagine request batch size is 100 documents and driver_batch_size is 9) There will be 9 apply_all of batch_size 9 and then there will be 1 document leftover from the request that needs to be applied (1 that will remain in the cache). And it would not be safe to assume that another request would come

jina/drivers/encode.py
JoanFM: yes it is used, when cache_set is full the recursivity will end up here again and call `f`

JoanFM: But yes, I know itr is not a common recylcable decorator

JoanFM: I will adapt it in the `_apply_all`

jina/drivers/encode.py
```diff
@@ -1,25 +1,84 @@
...
+    def __init__(self,
+                 batch_size: Optional[int] = None,
+                 *args, **kwargs):
+        """
+        Extract from different documents.
+        :param batch_size: number of documents to be used simultaniously in the encoder :meth:apply_all.
+        It is specially useful when the same EncoderExecutor can be used for documents of different granularities
+         (chunks, chunks of chunks ...)
+
+    .. warning::
```
hanxiao, line 63: wrong indent

jina/drivers/encode.py
```diff
@@ -1,25 +1,84 @@
...
+            self._doc_set = DocumentSet(docs_proto=[])
+
+        @property
+        def available_capacity(self):
+            return self.capacity - len(self._doc_set)
+
+        def cache(self, docs: DocumentSet):
+            docs_to_append = min(len(docs), self.available_capacity)
+            for doc in docs[: docs_to_append]:
+                self._doc_set.append(doc)
```
hanxiao, line 45:
```suggestion
            self._doc_set.extend(docs[: docs_to_append])
```
DocSet has `extend` method

jina/drivers/encode.py
```diff
@@ -1,25 +1,84 @@
...
     """
 
-    def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
-        contents, docs_pts, bad_docs = docs.all_contents
+    class CacheDocumentSet:
+        """Helper class to accumulate documents from differents document Set in a single DocumentSet
+         to help guarantee that the encoder driver can consume documents in fixed batch sizes to allow
+         the EncoderExecutors to leverage its batching abilities.
+
+         It is useful to have batching even when chunks are involved"""
```
hanxiao, line 29: i have a feeling this should be in types? or even the `__add__` of `DocumentSet`? feel that this should be more native and general. Will look at this later

jina/drivers/encode.py
JoanFM: Done

jina/drivers/encode.py
JoanFM: I also had the feeling, but I was being conservative as per how universal this may be needed, and that's why I kept it inside `EncodeDriver`

hanxiao (approved):

hanxiao:
able to merge atm, with potential refactoring https://github.com/jina-ai/jina/issues/1532

PR #1488, opened 2020-12-18
refactor: update helloworld evaluate

JoanFM:


jina/helloworld/helper.py
```diff
@@ -77,8 +82,11 @@ def write_html(html_path):
             open(html_path, 'w') as fw:
         t = fp.read()
         t = t.replace('{% RESULT %}', '\n'.join(result_html))
-        evaluation_percentage = evaluation_value/num_docs_evaluated * 100.0
-        t = t.replace('{% EVALUATION %}', '{:.2f}%'.format(evaluation_percentage))
+        precision_evaluation_percentage = evaluation_value['evaluate-Precision@5'] / num_docs_evaluated * 100.0
```
hanxiao, line 85: where is `@5` hardcoded?

hanxiao: should be `@{args.num_query}`

JoanFM: @5 is the eval_at

JoanFM: I can do the trick to select any key containing `Precision`

hanxiao: but why is it 5, shouldn't this be `top_k`?

hanxiao: i see so the `eval_at` is an attribute in Executor, making  it impossible to read from querylang/driver
```
class BaseRankingEvaluator(BaseEvaluator):
    """A :class:`BaseRankingEvaluator` evaluates the content of matches against the expected GroundTruth.
    It is used to evaluate messages coming out from Indexers and Rankers and compares matches with groundtruths
    """

    def __init__(self, eval_at: int, *args, **kwargs):
        """"
        :param eval_at: k at which evaluation is performed
        """
        self.eval_at = eval_at
```

hanxiao (approved):
approved, refactor on `eval_at` in another PR

PR #1492, opened 2020-12-19
fix: access token cred file for hub push 

rutujasurve94:
- Fetches `GITHUB_TOKEN` from the `env` and makes it available during reading access token from `credentials_file` during `ci` run


tests/unit/docker/test_hub_build.py
```diff
@@ -59,6 +66,7 @@ def test_hub_build_push():
         '--keywords', summary['manifest_info']['keywords'][0],
         '--type', summary['manifest_info']['type']
     ])
+    monkeypatch.setattr(yaml, 'load', access_token_github)
```
deepankarm, line 69: `JAML.load()` will get executed only if `credentials_file().is_file()` is True. Please patch that too

deepankarm, line 69: `monkeypatch` needs to be done before `build & push` (L#48)

tests/unit/docker/test_hub_build.py
```diff
@@ -1,14 +1,20 @@
...
 from jina.excepts import ImageAlreadyExists
 from jina.helper import yaml
+from jina.docker.helper import credentials_file
 from jina.parser import set_hub_build_parser, set_hub_list_parser, set_hub_pushpull_parser
 
 cur_dir = Path(__file__).parent
 
+@pytest.fixture
+def access_token_github():
+    token = os.environ['GITHUB_TOKEN']
```
deepankarm, line 16:
```suggestion
    token = os.environ,get('GITHUB_TOKEN', None)
```
Please add a way to error out, if env is not set

deepankarm (changes requested):

tests/unit/docker/test_hub_build.py
```diff
@@ -67,12 +77,15 @@ def test_hub_build_push():
 
 
 @pytest.mark.skip(reason='Non reproducible error and super flaky in github. Please repair it.')
```
deepankarm, line 79: Please remove the `skip` here to make sure the tests are run

tests/unit/docker/test_hub_build.py
```diff
@@ -37,13 +45,15 @@ def test_hub_build_uses():
     # build again it shall not fail
     HubIO(args).build()
 
-
-def test_hub_build_push():
+def test_hub_build_push(monkeypatch, access_token_github):
```
JoanFM, line 48: please add an skip if GITHUB TOKEN is not found so that does not fail in local

tests/unit/docker/test_hub_build.py
```diff
@@ -66,13 +76,15 @@ def test_hub_build_push():
     assert manifests[0]['name'] == summary['manifest_info']['name']
 
 
-@pytest.mark.skip(reason='Non reproducible error and super flaky in github. Please repair it.')
-def test_hub_build_push_push_again():
+def test_hub_build_push_push_again(monkeypatch, access_token_github):
```
JoanFM, line 79: this should be moved to an integration test and not a unittest

JoanFM (changes requested):

tests/unit/docker/test_hub_build.py
JoanFM: also the one pushing

JoanFM (changes requested):

tests/unit/docker/test_hub_build.py
rutujasurve94: moved under `test_hub_usage` integration test

tests/unit/docker/test_hub_build.py
JoanFM: can u move this one also under integration?

tests/unit/docker/test_hub_build.py
```diff
@@ -1,14 +1,22 @@
...
+from jina.docker import hubapi
 from jina.excepts import ImageAlreadyExists
 from jina.jaml import JAML
+from jina.helper import expand_dict
+from jina.docker.helper import credentials_file
 from jina.parser import set_hub_build_parser, set_hub_list_parser, set_hub_pushpull_parser
 
 cur_dir = Path(__file__).parent
 
+@pytest.fixture
```
deepankarm, line 16: This can be removed here, as tests are moved to integration.

tests/integration/hub_usage/test_hub_usage.py
```diff
@@ -76,3 +86,71 @@ def test_build_timeout_ready():
...
+        '--keywords', summary['manifest_info']['keywords'][0],
+        '--type', summary['manifest_info']['type']
+    ])
+    response = HubIO(args).list()
+    manifests = response
+
+    assert len(manifests) >= 1
+    assert manifests[0]['name'] == summary['manifest_info']['name']
+
+
```
deepankarm, line 122: `@pytest.mark.skipif(condition='os.environ.get(\'GITHUB_TOKEN\')==None', reason='Token not found')` here too.

.github/workflows/ci.yml
```diff
@@ -108,6 +108,7 @@ jobs:
         env:
           JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
           JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
deepankarm, line 111: Since `GITHUB_TOKEN` is not used in unit-tests anymore, please remove this.

deepankarm (changes requested):
Few more minor changes

deepankarm (approved):

hanxiao (approved):

PR #1499, opened 2020-12-20
docs(devices): improve TorchDevice docstring

tadej-redstone:
I improved the `TorchDevice` docs, namely:
- set `torch.set_grad_enabled(False)` as the example does not include any training/gradients - this is good practice as it saves memory and increases performance
- `_input` tensor created using `as_tensor` with the `device` parameter - no need to explicitly check for GPU, as if only CPU is available no in-memory copy will be made
-  `_output` tensor now calls the `cpu()` method without checking for GPU - as before, if there is no GPU, then this operation is a noop, i.e. no in-memory copy will be made.

### DIFF
--- a/jina/executors/devices.py
+++ b/jina/executors/devices.py
@@ -48,12 +48,11 @@ class TorchDevice(BaseDevice):
             def encode(self, data, *args, **kwargs):
                 # use your awesome model to encode/craft/score
                 import torch
-                _input = torch.from_numpy(data)
-                if self.on_gpu:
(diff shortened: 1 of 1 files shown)


JoanFM:
recheckcla

JoanFM (approved):

JoanFM:
Hey @tadej-redstone,

Could you please sign the `CLA` before we can merge this PR?

tadej-redstone:

Issue #1330, opened 2020-11-23
top_k not working from Python client


bhavsarpratik:
**Describe the bug**
top_k set via python client is failing in current southpark [test](https://github.com/jina-ai/examples/pull/289/checks?check_run_id=1441373067)

JoanFM:
I think it may be related to the new `jina primitive types`

bhavsarpratik:
Han has made new `QueryLang` changes in the queryset in `flow.search`. Our test are also passing.

bwanglzu:
duplicate to #1316

hanxiao:
this issue is not solved.

Issue #1316, opened 2020-11-20
setting top_k from the client is not working


nan-wang:
**Describe the bug**
The integration test of running Southpark example fails. 

https://github.com/jina-ai/examples/pull/259/checks?check_run_id=1422319675

**Describe how you solve it**

---

**Environment**

**Screenshots**

hanxiao:
this issue is not solved

JoanFM:
> this issue is not solved

This issue was thought to be solved at syntax sugar level. it is impossible to adapt to the exact Flow the user will be using.

So no solution can be universal

PR #1501, opened 2020-12-20
refactor: hello-world, evaluator, top_k, eval_at

hanxiao:
see linked issues


tests/integration/issues/github_976/test_topk_rest_api.py
```diff
@@ -24,6 +24,6 @@ def test_top_k_with_rest_api(query_dict):
         req = request.Request(f'http://0.0.0.0:{port}/api/search', data=query,
                               headers={'content-type': 'application/json'})
         resp = request.urlopen(req).read().decode('utf8')
-        assert json.loads(resp)['queryset'][0]['name'] == 'VectorSearchDriver'
-        assert json.loads(resp)['queryset'][0]['parameters']['top_k'] == TOP_K
+        assert json.loads(resp)['queryset'][0]['name'] == 'SliceQL'
```
JoanFM, line 27: the user then cant use a SliceQL in the middle of the Flow. that can make for a very hard to debug problem

hanxiao: the current `VectorSearchDriver` does not garantee `top_k`, it is wrongly implemented. jina hello-world says `topk=50` but the result only shows 20.

hanxiao: I'm very surprised how these two past issues #1330 #1316 are considered as done.

JoanFM: yes, but because under our current implementation is very hard to guarantee this. We cannot target a specific Pod to send top_k. in my opinion, current solution is the best tradeoff.

But anyway, It is important to know that no solution right now will be perfect

hanxiao: if the user does not use `SliceQL` then there is no way to select top_k.

JoanFM: when we first implemented top_k we were aware that it was just syntax sugar to help the first user, but that if someone wanted to have full control of the parameters they should send their own QueryLang with the Queries.

Thats why the issues were considered done, because the issues came from a degradation of that original thought.

PR #1508, opened 2020-12-20
fix(helloworld): hello-world evaluator use running avg directly

hanxiao:


JoanFM:
I think is not a good idea to encourage to use this feature. it is impossible to reset and leads to stateful computations.

JoanFM:
Plus this feature goes against the idea of having the ability of having a contunious Flow instance running. We have no way to reset the running average (and I think we should not have one for now)

PR #1510, opened 2020-12-21
feat(driver): add docidcache for content

cristianmtr:
- Make DocIDCache capable of detecting collisions on content-level.
- NOTE: Tests are only on the `Indexer` level, since the Drivers for Update/Del. haven't been implemented yet. Integration tests will be required when that is done.

**Discussion required:** do we set the `field` (`id`/`content_hash`) at the Driver level or at the Indexer level? 

Atm it's done at the Driver level. Currently it is set at the `DocIDCache` level but read and used by the Driver.

**Tests** Added an integration test in incremental indexing, where we are caching by content_hash

NOTE: This is a replacement of #1466 because that one was failing the commitlint


jina/executors/indexers/cache.py
```diff
@@ -24,33 +24,58 @@ def post_init(self):
...
+        self.store[doc_id] = kwargs.get(DATA_FIELD)
+
+    @deprecated_alias(doc_id='data')
+    def query(self, data: UniqueId, *args, **kwargs) -> Optional[bool]:
+        """
+        @param data: either the `id` or the `content_hash` of a `Document`
+        """
+        # print(f'query = {data}, self.store = {self.store}')
+        status = (data in self.store.values()) or None
+        print(f'query = {data}. status = {status}')
```
JoanFM, line 48: remove prints

jina/executors/indexers/cache.py
```diff
@@ -24,33 +24,58 @@ def post_init(self):
...
+        if missed:
+            raise KeyError(f'Keys {missed} were not found. No operation performed...')
+
+        for key in keys:
+            del self.store[key]
+            self._size -= 1
 
     def get_add_handler(self):
-        return open(self.index_abspath, 'ab')
+        pass
```
JoanFM, line 78: who will implement these?

tests/unit/drivers/test_cache_driver.py
```diff
@@ -26,11 +28,11 @@ def docs(self):
         return DocumentSet(list(random_docs(10)))
 
 
-def test_cache_driver_twice(tmp_path):
-    filename = tmp_path / 'test-tmp.bin'
+def test_cache_driver_twice(tmpdir):
     docs = DocumentSet(list(random_docs(10)))
     driver = MockCacheDriver()
-    with DocIDCache(filename) as executor:
+    # FIXME DocIdCache doesn't use tmpdir, it saves in curdir
```
JoanFM, line 34: you need to use `test_metas`

tests/unit/drivers/test_cache_driver.py
```diff
@@ -40,41 +42,48 @@ def test_cache_driver_twice(tmp_path):
             driver._traverse_apply(docs)
 
         # new docs
-        docs = list(random_docs(10))
+        docs = list(random_docs(10, start_id=100))
         driver._traverse_apply(docs)
+        filename = executor.save_abspath
 
         # check persistence
+        executor.save()
```
JoanFM, line 50: no need, executor as `context manager` will save itself if it has been updated

tests/unit/drivers/test_cache_driver.py
```diff
@@ -84,8 +93,86 @@ def test_cache_driver_from_file(tmp_path):
...
+
+    with BaseExecutor.load(filename) as executor:
+        assert executor.query(doc1.content_hash) is True
+        assert executor.query(old_hash) is None
+
+    # delete
+    with BaseExecutor.load(filename) as executor:
+        executor.delete([UniqueId(doc1.id)])
+        executor.save()
```
JoanFM, line 154: no need to save explicitly here

JoanFM (changes requested):

tests/unit/drivers/test_cache_driver.py
cristianmtr: Tried, same problem

jina/executors/indexers/cache.py
cristianmtr: That's one of the things to discuss still. What should the internal data structure be? Can I have a `BinaryPbIndexer` internally? How do I assign it a filename based on the filename that the `DocIDCache` is assigned?

JoanFM: I do not think a `BinaryPbIndexer` is the right one, since you do not store `Protobuf` documents. It is a KVIndexer and you can just write to a file as it was done previously right? The only difference is what will be written there or read

cristianmtr: Previously it was only one array. Now I need a KV indexer, yes. But do I implement another one just for this?

cristianmtr: Also I'm still not sure whether the field should be set in the Driver or in the DocIDCache. If we don't set it in DocIDCache, we duplicate data. However, if we don't set it in Driver we break the API of the CRUD methods (because we would need to pass the entire document to a query (not just a field)

JoanFM: you could have it in the executor and consulted from the Driver too right?

JoanFM: > Previously it was only one array. Now I need a KV indexer, yes. But do I implement another one just for this?

The point is, by inheritance, DocIDCache is a `KVIndexer` and it was implemented as this, so yes we have are indeed implementing a new one

tests/unit/drivers/test_cache_driver.py
JoanFM: Is not that u do `DocIDCache(metas=test_metas)` , metas can only be passed by `BaseExecutor.load_config` I think.

What you can do is to set `executor.workspace = tmpdir`

cristianmtr: But here it works https://github.com/jina-ai/jina/blob/f143a073857bd547e4af12caf127e526217d40d6/tests/unit/executors/indexers/test_numpyindexer.py#L23 What am I missing?

jina/executors/indexers/cache.py
cristianmtr: But the internal data structure should also be completely new? Just using a `dict` is probably not efficient enough, right? Before we were only storing one np array of the ids. Now we need kv of `docid -> doc.content_hash`

cristianmtr: It could be two arrays, mapped from one to the other by positional index

JoanFM: seems a good idea

jina/executors/indexers/cache.py
```diff
@@ -22,35 +26,104 @@ def post_init(self):
...
+    class CacheHandler:
+        def __init__(self, path, logger):
+            self.path = path
+            try:
+                # TODO maybe mmap?
+                self.ids = pickle.load(open(path + '.ids', 'rb'))
+                self.cache = pickle.load(open(path + '.cache', 'rb'))
+            except FileNotFoundError as e:
+                logger.warning(f'File path did not exist : {path}.ids or {path}.cache: {repr(e)}. Creating new CacheHandler...')
+                self.ids = []
```
JoanFM, line 42: why one is named ids and the other cache? If this is itself a 'cache'?

cristianmtr: It's sort of a KV store, with one being the keys (`.ids`), and the other being the values (`.cache`). The first one contains the `ids` of the documents, the other contains the `.content_hash`-es. It is what I discussed here https://github.com/jina-ai/jina/pull/1510#discussion_r546762157

cristianmtr: Two possible improvements:

- using `np.mmap`. What do you think?
- if the field we cache by is `id` we don't need to use the `.cache` as a separate store. This one I'll do for sure now.

JoanFM: I would then call one ids and the other hashes better?

JoanFM: I would not do early optimization. I think the complete CRUD is complex enough, I would have the complete thing properly tested, and later think of potential improvements, what do you think?

cristianmtr: Sure, then apart from the workspace issue, what else needs to be done?

The tests in `test_cache_driver` are unit tests.

One integration test is the incremental indexing. They are all passing.

jina/executors/indexers/cache.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, index_filename: str = None, *args, **kwargs):
             raise ValueError(f"Field '{self.field}' not in supported list of {self.supported_fields}")
 
     @deprecated_alias(doc_id='doc')
-    def add(self, doc: 'Document', *args, **kwargs):
+    def add(self, doc, *args, **kwargs):
```
JoanFM, line 65: u need to have the trick 

if False:
from jina import Document

JoanFM (changes requested):

JoanFM:
> * Make DocIDCache capable of detecting collisions on content-level.
> * NOTE: Tests are only on the `Indexer` level, since the Drivers for Update/Del. haven't been implemented yet. Integration tests will be required when that is done.
> 
> **Discussion required:** do we set the `field` (`id`/`content_hash`) at the Driver level or at the Indexer level? Atm it's done at the Driver level.
> 
> NOTE: This is a replacement of #1466 because that one was failing the commitlint

I think is better to have it at Driver level. It is better to keep the Executor unaware of any Jina data type.

jina/executors/indexers/cache.py
cristianmtr: That fails 

> > * Make DocIDCache capable of detecting collisions on content-level.
> > * NOTE: Tests are only on the `Indexer` level, since the Drivers for Update/Del. haven't been implemented yet. Integration tests will be required when that is done.
> > 
> > **Discussion required:** do we set the `field` (`id`/`content_hash`) at the Driver level or at the Indexer level? Atm it's done at the Driver level.
> > NOTE: This is a replacement of #1466 because that one was failing the commitlint
> 
> I think is better to have it at Driver level. It is better to keep the Executor unaware of any Jina data type.

That was the original approach. Two problems:

1. Since it doesn't know which field it's caching, we can't optimize memory/space usage to avoid duplicating the ids (when we are caching by id)
2. It requires that whenever a user wants to use `DocIDCache` with `content_hash` (no the default option, `id`) to have to replicate the yml to specify it at the Driver level as parameter:


```yaml
!DocIDCache
with:
  index_path: cache.tmp
metas:
  name: cache
requests:
  on:
    [SearchRequest, TrainRequest, IndexRequest, ControlRequest]:
      - !RouteDriver {}
    IndexRequest:
      - !TaggingCacheDriver
        with:
          field: content_hash
          tags:
            is_indexed: true
      - !FilterQL
        with:
          lookups: {tags__is_indexed__neq: true}
```

vs

```yaml
!DocIDCache
with:
  index_path: cache.tmp
  field: content_hash
metas:
  name: cache
```

cristianmtr:
> > * Make DocIDCache capable of detecting collisions on content-level.
> > * NOTE: Tests are only on the `Indexer` level, since the Drivers for Update/Del. haven't been implemented yet. Integration tests will be required when that is done.
> > 
> > **Discussion required:** do we set the `field` (`id`/`content_hash`) at the Driver level or at the Indexer level? Atm it's done at the Driver level.
> > NOTE: This is a replacement of #1466 because that one was failing the commitlint
> 
> I think is better to have it at Driver level. It is better to keep the Executor unaware of any Jina data type.

That was the original approach. Two problems:

1. Since it doesn't know which field it's caching, we can't optimize memory/space usage to avoid duplicating the ids (when we are caching by id)
2. It requires that whenever a user wants to use `DocIDCache` with `content_hash` (no the default option, `id`) to have to replicate the yml to specify it at the Driver level as parameter:


```yaml
!DocIDCache
with:
  index_path: cache.tmp
metas:
  name: cache
requests:
  on:
    [SearchRequest, TrainRequest, IndexRequest, ControlRequest]:
      - !RouteDriver {}
    IndexRequest:
      - !TaggingCacheDriver
        with:
          field: content_hash
          tags:
            is_indexed: true
      - !FilterQL
        with:
          lookups: {tags__is_indexed__neq: true}
```

vs

```yaml
!DocIDCache
with:
  index_path: cache.tmp
  field: content_hash
metas:
  name: cache
```

jina/executors/indexers/cache.py
cristianmtr: That failed before. I rebased on master again, and it seems to have been fixed. Yep, will do

JoanFM:
> > > * Make DocIDCache capable of detecting collisions on content-level.
> > > * NOTE: Tests are only on the `Indexer` level, since the Drivers for Update/Del. haven't been implemented yet. Integration tests will be required when that is done.
> > > 
> > > **Discussion required:** do we set the `field` (`id`/`content_hash`) at the Driver level or at the Indexer level? Atm it's done at the Driver level.
> > > NOTE: This is a replacement of #1466 because that one was failing the commitlint
> > 
> > 
> > I think is better to have it at Driver level. It is better to keep the Executor unaware of any Jina data type.
> 
> That was the original approach. Two problems:
> 
> 1. Since it doesn't know which field it's caching, we can't optimize memory/space usage to avoid duplicating the ids (when we are caching by id)
> 2. It requires that whenever a user wants to use `DocIDCache` with `content_hash` (no the default option, `id`) to have to replicate the yml to specify it at the Driver level as parameter:
> 
> ```yaml
> !DocIDCache
> with:
>   index_path: cache.tmp
> metas:
>   name: cache
> requests:
>   on:
>     [SearchRequest, TrainRequest, IndexRequest, ControlRequest]:
>       - !RouteDriver {}
>     IndexRequest:
>       - !TaggingCacheDriver
>         with:
>           field: content_hash
>           tags:
>             is_indexed: true
>       - !FilterQL
>         with:
>           lookups: {tags__is_indexed__neq: true}
> ```
> 
> vs
> 
> ```yaml
> !DocIDCache
> with:
>   index_path: cache.tmp
>   field: content_hash
> metas:
>   name: cache
> ```

ok, but can the driver extract it from the executor and extract the field before calling the executor?

cristianmtr:
@JoanFM I guess so. It could be done here https://github.com/jina-ai/jina/blob/19d9f4fd1880ca8ba12307547fd21999a49a474a/jina/drivers/cache.py#L27

So what would be the approach?

- we set it from the Executor config level
- but the Driver processes the Document and passes the correct params to the Executor CRUD methods
- the Executor still knows whether it's gonna duplicate and possibly optimize by not saving ids twice

JoanFM:
> @JoanFM I guess so. It could be done here
> 
> https://github.com/jina-ai/jina/blob/19d9f4fd1880ca8ba12307547fd21999a49a474a/jina/drivers/cache.py#L27
> 
> So what would be the approach?
> 
> * we set it from the Executor config level
> * but the Driver processes the Document and passes the correct params to the Executor CRUD methods
> * the Executor still knows whether it's gonna duplicate and possibly optimize by not saving ids twice

Yes, I think is a good approach

jina/executors/indexers/cache.py
JoanFM: Good, I think is really good job, I am sure we may find some nuances when integrating all together, but besides the comment on the `executor-driver-logic` I think is good

jina/executors/indexers/cache.py
```diff
@@ -22,35 +24,102 @@ def post_init(self):
...
+            self._size -= 1
 
     def get_add_handler(self):
-        return open(self.index_abspath, 'ab')
+        pass
+
+    def get_query_handler(self) -> CacheHandler:
+        return self.CacheHandler(self.index_abspath, self.logger)
 
     def get_create_handler(self):
```
JoanFM, line 124: how is then the data created?

cristianmtr: we don't need a separate handler

cristianmtr: https://github.com/jina-ai/jina/blob/bbc73f514ee6000e3a7cc1509c2db3294241e53d/jina/executors/indexers/cache.py#L76

JoanFM: ok, may u add small comment about this?

cristianmtr: good point, did

JoanFM (approved):

hanxiao (approved):

PR #1516, opened 2020-12-21
fix(flow): fix flow cls register

hanxiao:


deepankarm:
Should we allow to load a Flow from a string based yaml? This can be useful in `jinad`.

image

hanxiao:
@deepankarm it supports already. see https://docs.jina.ai/chapters/cli/jina-pea.html?highlight=uses

PR #1539, opened 2020-12-24
refactor: redesign of BasePea and BaseRuntime

hanxiao:
Big important ticket, during the holidays so I will start simple. 

## TL;DR

The current implementation of `Pea`, `Runtime`, and `Pod` introduces the following structure.

```
Pod 
 | 
 -  LocalRuntime
     |
     - Pea
 - ContainerRuntime (No pea)
 - AsyncRuntime (No Pea)
 - RemoteRuntime (No Pea)
```

Roughly speaking, this PR will **"switch" the semantics** of current `Pea` & `Runtime`, and change it into:

```
Pod 
 | 
 -  Pea
     |
     - ZEDRuntime (ZMQlet+Executor+Driver)
     - ContainerRuntime
     - AsyncRuntime (for gRPC)
     - RemoteRuntime
```

## Current Problems

The philosophy of separating Runtime Manager and Runtime is a **good direction**. However, the **realization** of this philosophy has the following problems:
- It weakens the design of `Pea`, breaking the semantic relation of `Pod`<> `Pea`. It injects a `Runtime` in-between: now the `Pod` is maintaining multiple `Runtime` instead of `Pea`s. Such big change with little notice & adaption on other modules/blog/docs/Jina 101 is not acceptable. 
- When looking at the `runtimes` module implementation, it is half-done. 
   - Many of the methods are legacy methods & docstrings from `BasePea`, making people question about the design phiolophy (which has nothing wrong).
   - All `Runtime` classes do not leverage inherit methods and each seems to have its own way of managing the lifecycle.

All in all, the current implemenation in `runtimes` module gives me a strong impression that it is a not complete work. It does not exhaustively abstract the philosophy of separating Runtime Manager and Runtime. Therefore, changes must be done before 1.0, asap.

## This PR

1. I will change the semantic back to `Pod` manages `Pea`, and `Pea` manages `Runtime`.
2. `Runtime` is a member of `Pea`, managed by the `Pea`. It is defined as "a procedure that blocks the main process once running, therefore must be put into a separated thread/process. Any program/library/package/module that blocks the main
    process, can be formulated into a :class:`BaseRuntime` class and then be used in :class:`BasePea`." [See `jina/peapods/runtimes/base.py` for details](https://github.com/jina-ai/jina/blob/9d1e06a92c1a0f5910454657c38d45e0f3d6e1a1/jina/peapods/runtimes/base.py) and [how now `BasePea` is managing `Runtime`](https://github.com/jina-ai/jina/blob/9d1e06a92c1a0f5910454657c38d45e0f3d6e1a1/jina/peapods/peas/base.py).
3. For the sake of clarity, I'm removing all `pea` & `runtime` modules in this PR atm. These modules have to be heavily refactored anyway. 
4. There will be only one class of `Pea`, but `Runtime` has many subclasses, e.g. for remote, async, container, etc.

## Progress

- [x] Restructure the semantics of `Pea` and `Runtime`
- [x] Design `BaseRuntime` and four essential interfaces.
- [x] Design `Pea` to use `BaseRuntime`
- [x] Refactor `Pod` to use `Pea`
- [x] Fix & refactor remote, container, async features in old `Pea`/`Runtime`
- [x] Refactor `parser` to separate `Pea` & `Runtime` argument
- [x] All tests

## ETA

Extremely important refactoring. By the end of this year.


JoanFM:
I disagree that it confuses the user in the definition of Pods and Peas.

I think current design clarifies better what IS a Pea and what is NOT a Pea. It is more important to have a Pea properly context managed by a Runtime than the fact that a Pod manages a Pea (which is False in the direction of this PR).

Inheriting from Pea things that do not have to know anything about Executors or ZmqLets confuses even more the concepts of Pods and Peas.

Current design has a better separation of concerns clarifying better the different ways of parallelizing Peas (local,  remote and in docker) and the actual work of a Pea.  

Runtime has very different work because they do very different things. Inheriting should be made when possible but it is not a benefit to try to fit very different logics into the same structure using inheritance as a way trying to recycle code at all costs even when readbility is compromised.

hanxiao:
>  It is more important to have a Pea properly context managed by a Runtime than the fact that a Pod manages a Pea

To understand this PR, the old `Runtime` is the new `Pea` in this PR.

JoanFM:
I would suggest tob create a new class to encapsulate the (ZMQLet + Executor) in a context manager in the way BasePea is designed right now such that ZEDRuntime can ensure its proper context management.

JoanFM:
> > It is more important to have a Pea properly context managed by a Runtime than the fact that a Pod manages a Pea
> 
> To understand this PR, the old `Runtime` is the new `Pea` in this PR.

and the old Pea is the new Runtime?

jina/peapods/runtimes/base.py
```diff
@@ -0,0 +1,90 @@
...
+     Note, another way to jump out from :meth:`serve_forever` is raise exceptions from it. This will immediately move to
+     :meth:`teardown`.
+
+     .. seealso::
+
+        :class:`BasePea` for managing a :class:`Runtime` object's lifecycle.
+
+     """
+
+    def serve_forever(self):
```
hanxiao, line 45: change this to `run_forever`

hanxiao:
> > > It is more important to have a Pea properly context managed by a Runtime than the fact that a Pod manages a Pea
> > 
> > 
> > To understand this PR, the old `Runtime` is the new `Pea` in this PR.
> 
> and the old Pea is the new Runtime?

yes

jina/peapods/peas/base.py
```diff
@@ -0,0 +1,150 @@
...
+
+        def _finally():
+            self.is_ready.clear()
+            self.is_shutdown.set()
+            self._unset_envs()
+
+        self._set_envs()
+
+        try:
+            self.runtime.setup()
```
JoanFM, line 68: if possible, using Runtime as context manager will be cleaner and more robust than a complicated set of Exceptions. 

I also think that the different Runtimes are so diferent that I think is better to have different run methods for each. It is easier to read specialized methods than fitting 3 things in the same template.

JoanFM: For instance, RuntimeTerminated is only thrown by a local Runtime. Each of them will have their own nuances and I think is better to have it prepared to be of different nature, so that they can evolve in different directions.

hanxiao: > Each of them will have their own nuances

The `BaseRuntime` will capture this pattern for all.

Note throwing exception `RuntimeTerminated ` is not defined as *the way* to end from `run_forever`, see `BaseRuntime` docstring: calling `cancel` will can end a forever loop peacefully. The Runtime designer must implement it.

jina/peapods/runtimes/zed.py
```diff
@@ -0,0 +1,226 @@
...
+        self._executor.close()
+
+    #: Private methods required by :meth:`setup`
+
+    def _load_executor(self):
+        """Load the executor to this BasePea, specified by ``uses`` CLI argument.
+        """
+        try:
+            try:
+                self._executor = BaseExecutor.load_config(self.args.uses,
```
JoanFM, line 57: a big benefit of the latest refactor is to have executor context managed by the Pea which is a very robust design. Can we keep it this way?

JoanFM (changes requested):

jina/peapods/runtimes/zed.py
JoanFM: is there any plan to benefit of this? that would give a lot of robustness in ensuring the proper closure of Executors. Instead of trusting that the developer will safely capture every potential problem

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -0,0 +1,222 @@
...
+from .... import Request
+from ....enums import SkipOnErrorType
+from ....excepts import NoExplicitMessage, ExecutorFailToLoad, MemoryOverHighWatermark, ChainedPodException, \
+    BadConfigSource, RuntimeTerminated
+from ....executors import BaseExecutor
+from ....logging.profile import used_memory, TimeDict
+from ....proto import jina_pb2
+
+
+class ZEDRuntime(ZMQRuntime):
```
JoanFM, line 20: This should be a `ContextManager` for `Executor`.

jina/peapods/peas/__init__.py
```diff
@@ -1,338 +1,174 @@
...
-        """Save the contained executor according to the `dump_interval` parameter
-        """
-        if (time.perf_counter() - self.last_dump_time) > self.args.dump_interval > 0:
-            self.executor.save()
-            self.last_dump_time = time.perf_counter()
-            if hasattr(self, 'zmqlet'):
-                self.zmqlet.print_stats()
+            self.logger.error(f'{ex!r} during {self.runtime.setup!r}', exc_info=self.args.show_exc_info)
+        else:
+            self.is_ready.set()
```
JoanFM, line 65: this else construction isnt it a bit weird?

jina/peapods/peas/__init__.py
```diff
@@ -1,338 +1,174 @@
...
-    @property
-    def expect_parts(self) -> int:
-        """The expected number of partial messages before trigger :meth:`handle` """
-        return self.args.num_part if self.message.is_data_request else 1
+            try:
+                self.runtime.teardown()
+            except Exception as ex:
+                self.logger.error(f'{ex!r} during {self.runtime.teardown!r}', exc_info=self.args.show_exc_info)
+        finally:
+            _finally()
```
JoanFM, line 80: finally is not a good method name

jina/peapods/peas/__init__.py
```diff
@@ -1,338 +1,174 @@
...
+            self.runtime = self._get_runtime_cls()(self.args)  # type: 'BaseRuntime'
+        except Exception as ex:
+            self.logger.error(f'{ex!r} during {self.runtime_cls.__init__!r}', exc_info=self.args.show_exc_info)
+            raise RuntimeFailToStart from ex
 
-    @property
-    def request(self) -> 'Request':
-        """Get the current request body inside the protobuf message"""
-        return self._request
+    def run(self):
```
JoanFM, line 45: having a specializrd Pea to handle ZEQRuntime makes the Runtime of the main part of the logic much clearer and robust. instead of adapting so many different runtimes to be handled in the same exact way.

JoanFM (changes requested):

jina/peapods/peas/__init__.py
hanxiao: one of the big headache is the explosive polymorphism on `Pod`, `Pea` and `Runtime`. In this PR I killed all subclasses of `Pod` and `Pea`, only `Runtime` allows polymorphism.

jina/peapods/runtimes/zed.py
hanxiao: >  executor context managed , ... , is there any plan to benefit of this?

There is no magic on context manager. It relies on handwritten code such as `.start()`, `.close`. Previously one of the big headache is you never now when the context manager is ending its context, in main process/subprocess, in side forever loop or outside forever loop. It is unrealistic to just believe `with ...` will magically handle everything.

So it's not part of the PR, as I see it is a sugary syntax which only weakly benefits people who work on `Runtime` but no others. Further improvement in separate PR is welcome but please first understand all corner cases described in the `BasePea:start()`

jina/peapods/peas/__init__.py
hanxiao: The use of the else clause is better than adding additional code to the try clause because it avoids accidentally catching an exception that wasn’t raised by the code being protected by the try ... except statement.

JoanFM: good point

hanxiao: one of the big benefits of the four-function-design of `BaseRuntime` is its decoupling the progress-status of the Runtime: for example, you never have to inject `is_ready` from `Pea` into some `Runtime.func()` just to let `Pea` to know `Runtime` is ready. 

Therefore, try except at `BasePea` must catch exception in a precise way. It must see the four functions from `Runtime` as the atomic function and catch each of them precisely.

PR #1541, opened 2020-12-25
feat: return result from flow and client

hanxiao:
add `return_results` (by default = `False`) in client CLI, so that when turning on, results are returned as a list. Can be used with `Flow` and `AsyncFlow`, e.g.

`AsyncFlow`:

```python
with AsyncFlow(return_results=True).add() as f:
    r = await f.index_ndarray(np.random.random([10, 2]))
    assert isinstance(r, list)
    assert isinstance(r[0], Response)
```

SyncFlow
```python
with Flow(return_results=True).add() as f:
    r = f.index_ndarray(np.random.random([10, 2]))
    assert isinstance(r, list)
    assert isinstance(r[0], Response)
```

Design decision:
- dropping asyncio generator. Complicated, can not unify the Sync & AsyncFlow interface in short time.
- people who wants to use this function is more likely to fetch all result at once, otherwise, `on_done` callback is sufficient.
- complicated usage e.g. fully delegating async generator to users is possible, but sounds overengeering to me atm, can be future work. Out of scope of this PR.

### DIFF
--- a/jina/clients/base.py
+++ b/jina/clients/base.py
@@ -114,6 +114,7 @@ class BaseClient:
                            on_done: Callable,
                            on_error: Callable = None,
                            on_always: Callable = None, **kwargs):
+        result = []
         try:
             self.input_fn = input_fn
             req_iter, tname = self._get_requests(**kwargs)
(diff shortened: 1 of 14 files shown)


jina/clients/base.py
```diff
@@ -151,6 +154,8 @@ async def _get_results(self, input_fn: Callable,
                                      'please double check your input iterator') from rpc_ex
             else:
                 raise BadClient(msg) from rpc_ex
+        if self.args.return_results:
+            return result
```
JoanFM, line 158: would it make sense to return a generator to make it more general?

hanxiao: tried. too complicated., async generator is depreciated in python asyncio

hanxiao: also, if async generator is desired, why not directly use `on_done`?

hanxiao: see design decision in PR body

JoanFM: that is true. Just a comment that felt more natural, but no big deal

Issue #1545, opened 2020-12-27
Index stored under metas['workspace']/None-{root.metas.pea_id}/ even when no `parallelism` is used


JoanFM:
**Describe the bug**
FaissIndexer could not be built. ( I think the reason why many Hub executors failed to be built with latest jina version is also related to this).

 The reason is that some tests expect that the index files to be stored under `metas['workspace']/index_filename`. However, the path is stored under ` metas['workspace']/None-{root.metas.pea_id}/index_filename`.

I think the problem comes when the JAML module was added.

JoanFM:
Related to self reference in metas:

```pea_id: ${{root.metas.pea_id}}  # this may result in self-referred
separated_workspace: ${{root.metas.separated_workspace}}  # this may result in self-referred```

PR #1547, opened 2020-12-28
test: add test for 1545

JoanFM:
**Changes introduced**
Add test proving the problem with #1545


hanxiao:
let me fix this real quick

JoanFM:
> let me fix this real quick, taking this branch now

perfect thanks!

PR #1550, opened 2020-12-28
fix: add crud numpy indexer

JoanFM:


hanxiao (approved):
i see some of redudancy in CRUDBase and Crud, but given this is a temp solution, good to go

Issue #1502, opened 2020-12-20
Questions regarding MongoDBIndexer


ace-kay-law-neo:
He Jina team, I had a look at the MongoDBIndexer. Some things however remained unclear.

**MongoDB-Id vs Jina-Id**
MongoDBs id is a 24byte hex number. Jina's id is only of the size 16. How does this work together? Is there some internal translation?

**MongoDB only?**
Does the indexer store all information in the database or has it some data (e.g. the Id translation) stored locally like the BinaryPbIndexer indexer from which it inherits.

**Switch between Indexer**
Is it possible to switch between MongoDBIndexer and BinaryPdIndexer based on the document. Or does one need to write a custom executor therefor.

Thanks!!!

janandreschweiger:
I'm particularly interested in the ID-question. I'd would be cool to use the mongo _id as the jina document id. Unfortunately, there jina restricts us to 16 characters. Can someone have a look on this one?

cristianmtr:
> I'm particularly interested in the ID-question. I'd would be cool to use the mongo _id as the jina document id. Unfortunately, jina restricts us to 16 characters. Can someone have a look on this one?

I am not an expert in Mongo, but the test suite wrote for the indexer works. What is the issue that you're experiencing with the id being only 16 characters?

Unfortunately at the moment Jina IDs are calculated in a specific way.

In the next release we are going to allow custom ids that can be of any length (as they are no longer hashes of the content):

```
In [8]: a = Document(id='a'*24)
In [9]: a.id
Out[9]: 'aaaaaaaaaaaaaaaaaaaaaaaa'
```

> Switch between Indexer
> Is it possible to switch between MongoDBIndexer and BinaryPdIndexer based on the document. Or does one need to write a custom executor therefor.

No, not at the moment. And I'm not sure what the use case would be here

janandreschweiger:
Thank you @cristianmtr! Custom ids would be great. Will they be available in 0.8.x or 0.9.0

cristianmtr:
> Thank you @cristianmtr! Custom ids would be great. Will they be available in 0.8.x or 0.9.0?

Seems like it's there in `0.8.11`

deepankarm:
Hello @janandreschweiger, glad to see your interest in Jina.

[...]

PR #1554, opened 2020-12-28
fix: custom long ids for rankers

### LINKED ISSUE #1553: Rankers cannot handle long custom ids
Hi Jina team,

it's awesome that you now support custom ids. We use MongoDB in combination with jina, so our id-length is 24 instead of 16.
This is a problem, because the Chunk2DocRanker and some RankDrivers assume that the id is of type np.int64.

Please have a look:
- https://github.com/jina-ai/jina/blob/bf97c36ba65c08f620d504ed556ca508fdf2368f/jina/executors/rankers/__init__.py#L92
- https://github.com/jina-ai/jina/blob/bf97c36ba65c08f620d504ed556ca508fdf2368f/jina/drivers/rank/aggregate/__init__.py#L201-L203

An update that fixes our error would be awesome.
```
OverflowError: Python int too large to convert to C long
```

janandreschweiger:
fixes #1553


janandreschweiger:
@JoanFM could you please have a look on the failed integration-test? I don't think that my changes caused this error:
```
failed to stop. 'ZmqStreamlet' object has no attribute 'io_loop'
```

janandreschweiger:
Hey @JoanFM, my changes passed all checks. I've added a test case for longer ids.
It would be cool if you could review my branch and merge it if my changes are fine.

JoanFM:
We will review them. However, I am a little concerned about the implications of the change. our Document ID is built from an int64 (8 bytes) and from the MongoDB documentation, they use 12 bytes, 4 of which are dedicated to store a timestamp.

Could you detail how do you interact with a MongoDB Index? Is it the external source from which jina Documents are generated? or are you using our MongoDBIndexer?

janandreschweiger:
As @cristianmtr mentioned [here](https://github.com/jina-ai/jina/issues/1502#issuecomment-748830502), Jina now supports custom ids with a length of 24 characters.

**Our application:**
We already have a MongoDB database which contains our documents. Now we want to use Jina for search. MongoDB uses ids of length 24. The documents are stored on MongoDB and the parent-ids of our chunks are therefore of length 24.

ace-kay-law-neo:
Thanks for opening this PR @janandreschweiger. After I was told that Jina supports 24 character ids #1502, I was very happy and connected Jina with our Mongo database. Unfortunately, I got this error:
```OverflowError: Python int too large to convert to C long```

I hope that your changes will fix this issue, because we cannot alter our mongo-ids.

File: jina/types/document/uid.py [hunk 1]
```diff
@@ -31,7 +31,7 @@
 from jina.logging import default_logger
 
 _digest_size = 8
-_id_regex = re.compile(r'[0-9a-fA-F]{16}')
+_id_regex = re.compile(r'[0-9a-fA-F]+')
```
maximilianwerk, line 34: If we do that, the number of characters must be a multiple of 2, since we hexlify them at some point. Could you adapt the regex please?

maximilianwerk (changes requested):
Apart from the change in the regex, we must touch the `BinaryPBIndexer` since it expects a fixed length id which is mapable to `np.int64` for performance reasons. We have to double check, what the consequences of a variable length id would be. Sorry for not having a solution right away.

jina/drivers/rank/__init__.py
```diff
@@ -52,8 +52,8 @@ def _apply_all(self, docs: 'DocumentSet', context_doc: 'Document', *args,
         # if at the top-level already, no need to aggregate further
         query_meta = context_doc.get_attrs(*self.exec.required_keys)
 
-        old_match_scores = {int(match.id): match.score.value for match in docs}
-        match_meta = {int(match.id): match.get_attrs(*self.exec.required_keys) for match in docs}
+        old_match_scores = {str(match.id): match.score.value for match in docs}
```
cristianmtr, line 55: Why do you convert to string if you then convert back to `UniqueId`?

cristianmtr:
> Apart from the change in the regex, we must touch the `BinaryPBIndexer` since it expects a fixed length id which is mapable to `np.int64` for performance reasons. We have to double check, what the consequences of a variable length id would be. Sorry for not having a solution right away.

Yes, I just tested adding to `BinaryPBIndexer` with a key `str` of 24 characters and it fails. And we cannot have `dtype=np.object` for the ReadHandler

https://github.com/jina-ai/jina/blob/4ff025588fcfdb46a41bb011f260b1a53217c3c5/jina/executors/indexers/keyvalue.py#L32

Not sure what the solution would be here

JoanFM:
> > Apart from the change in the regex, we must touch the `BinaryPBIndexer` since it expects a fixed length id which is mapable to `np.int64` for performance reasons. We have to double check, what the consequences of a variable length id would be. Sorry for not having a solution right away.
> 
> Yes, I just tested adding to `BinaryPBIndexer` with a key `str` of 24 characters and it fails. And we cannot have `dtype=np.object` for the ReadHandler
> 
> https://github.com/jina-ai/jina/blob/4ff025588fcfdb46a41bb011f260b1a53217c3c5/jina/executors/indexers/keyvalue.py#L32
> 
> Not sure what the solution would be here

maybe we will need them to work with our new JinaID type?

nero-nazok:
Hi @cristianmtr!
Custom ids are useful when storing the documents outside of Jina (e.g. in MongoDB). So the BinaryPBIndexer won't actually be used with long ids.
Thanks for opening this PR. We really look forward to this feature.

jina/types/document/uid.py
janandreschweiger: Sure, no problem

cristianmtr:
> Hi @cristianmtr!
> Custom ids are useful when storing the documents outside of Jina (e.g. in MongoDB). So the BinaryPBIndexer won't actually be used with long ids.
> Thanks for opening this PR. We really look forward to this feature.

Yes, but since you are pushing this change to Jina core, it is basically saying that by default Jina will support these sort of IDs seamlessly. https://github.com/jina-ai/jina/pull/1554/files#diff-30f4f5f24cb0e08e122a5e704ff612265c1f37e4c5f9296feed3dd2bb8001ea3R125

We need to make sure they are handled properly there too

maximilianwerk:
> > Apart from the change in the regex, we must touch the `BinaryPBIndexer` since it expects a fixed length id which is mapable to `np.int64` for performance reasons. We have to double check, what the consequences of a variable length id would be. Sorry for not having a solution right away.
> 
> Yes, I just tested adding to `BinaryPBIndexer` with a key `str` of 24 characters and it fails. And we cannot have `dtype=np.object` for the ReadHandler
> 
> https://github.com/jina-ai/jina/blob/4ff025588fcfdb46a41bb011f260b1a53217c3c5/jina/executors/indexers/keyvalue.py#L32
> 
> Not sure what the solution would be here

Can you do the following: 

- replace `dtype=np.int64` by `dtype=[np.object, np.int64, np.int64, np.int64]` in the `jina/executors/indexers/keyvalue.py`
- add unit tests to `tests/unit/executors/indexers/test_binary_indexer.py` and `tests/unit/executors/indexers/test_numpyindexer.py` with custom `id`s of different sizes.

It it works, we should be good.

jina/drivers/rank/__init__.py
janandreschweiger: Hey @cristianmtr,
I now removed the string-uniqueid-conversion in (e204475c658d19)[https://github.com/janandreschweiger/jina/commit/e204475c658d19b937d3a18c3cb175ac55533d49]. Unfortunately, the jinad-test (ubuntu-latest, 3.7) now fails. I tried several things, but I cannot fix this error. Could you maybe have a look on it?

JoanFM: Hey @janandreschweiger,

have u rebased on master? these tests should work if you realign, since jinad has recently been updated

janandreschweiger: Thanks @JoanFM, I will try that after adding some unit tests.

janandreschweiger: Hey @JoanFM the intergration test still fails. Should I revert my changes and use strings?

JoanFM: Hey @janandreschweiger,

No, do not revert it. We will check what can be the reason, but one colleague is doing some refactoring on that. We will investigate what can be the problem and see how it can be fixed. For now let's focus on having unit and integration tests working

JoanFM (commented):
why did u revert the change?

janandreschweiger:
@JoanFM I wanted to check whether the integration tests failed due to altering the BinaryPBIndexer:
`tmp = np.frombuffer(fp.read(), dtype=np.int64).reshape([-1, 4])` ->
`tmp = np.frombuffer(fp.read(), dtype=[np.object, np.int64, np.int64, np.int64).reshape([-1, 4])`

JoanFM:
Hey @janandreschweiger ,

Apparently the test is failing currently on master, so it is very unlikely to be caused by your changes, we will try to fix this test asap so that we can move on with this PR if the assumptions from @maximilianwerk are met

janandreschweiger:
Nevertheless I think enabling long-ids for the BinaryPbIndexer isn't that easy. The type-hints and numpy-dtypes should be changed for these methods:
- add
- query
- update
- delete

maximilianwerk:
> Nevertheless I think enabling long-ids for the BinaryPbIndexer isn't that easy. The type-hints and numpy-dtypes should be changed for these methods:
> 
>     * add
> 
>     * query
> 
>     * update
> 
>     * delete

That is only partly true. The type hints are fine. They are python `int`, which should be fine also for longer ids, since whatever you give is implicitly converted to an `int`. If you replace the `np.int64` with what I described above, it should work out of the box. Have you tested it or did it crash?

Also the `VectorIndexer` should be fine, since we do a conversion to an internal `id` and there is no conversion to some `numpy` type.

janandreschweiger:
Hey @JoanFM, could you please have a look on this error. It just states "The action timed out." Do think this is caused by my changes? How can I fix it?

cristianmtr:
> Hey @JoanFM, could you please have a look on this error. It just states "The action timed out." Do think this is caused by my changes? How can I fix it?

Where is this error? The checks are currently being re-run

JoanFM:
> > Hey @JoanFM, could you please have a look on this error. It just states "The action timed out." Do think this is caused by my changes? How can I fix it?
> 
> Where is this error? The checks are currently being re-run

I did trigger the rerun

jina/types/document/uid.py
```diff
@@ -1,13 +1,9 @@
 """
-
 Remarks on the ``id``, we have three views for it
-
 - ``id``: ``str`` is a hex string, for non-binary environment such as HTTP, CLI, HTML and also human-readable. it will be used as the major view.
 - ``bytes``: ``bytes`` is the binary format of str, it has 8 bytes fixed length, so it can be used in the dense file storage, e.g. BinaryPbIndexer, as it requires the key has to be fixed length.
 - ``int``:``int64`` (formerly names ``hash``) is the integer form of bytes, as 8 bytes map to int64 . This is useful when sometimes you want to use key along with other numeric values together in one ndarray, such as ranker and Numpyindexer
```
maximilianwerk, line 5: This docstring must be adjusted to state.

jina/executors/indexers/keyvalue.py
```diff
@@ -29,7 +29,7 @@ def flush(self):
     class ReadHandler:
         def __init__(self, path):
             with open(path + '.head', 'rb') as fp:
-                tmp = np.frombuffer(fp.read(), dtype=np.int64).reshape([-1, 4])
+                tmp = np.frombuffer(fp.read(), dtype=[np.object, np.int64, np.int64, np.int64]).reshape([-1, 4])
```
maximilianwerk, line 32: I am pretty sure, that replacement must also happen in line 65 and 101 of this file.

maximilianwerk (changes requested):
Please update the docstring of `uid.py`. Furthermore, the change in `keyvalue.py` must be done consistently. Thanks a lot for all your efforts. Great commitment from your side!

jina/types/document/uid.py
janandreschweiger: I think I already did that. Please have a look at the note in docstring. Or do you mean something else?

maximilianwerk: `int` now is not an `int64` anymore, but a regular python `int` which an be of arbitrary length. Also all the "fixed length" documentation does not hold true anymore, right?

jina/executors/indexers/keyvalue.py
```diff
@@ -62,7 +62,7 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
             self.write_handler.header.write(
                 np.array(
                     (key, p, r, r + l),
-                    dtype=np.int64
+                    dtype=np.object
```
maximilianwerk, line 65: Sorry for not being completely concrete in the first place. I meant:
```suggestion
                    dtype=[np.object, np.int64, np.int64, np.int64]
```

jina/executors/indexers/keyvalue.py
```diff
@@ -98,7 +98,7 @@ def delete(self, keys: Iterator[int], *args, **kwargs):
             self.write_handler.header.write(
                 np.array(
                     np.concatenate([[key], HEADER_NONE_ENTRY]),
-                    dtype=np.int64
+                    dtype=np.object
```
maximilianwerk, line 101: Same here.
```suggestion
                    dtype=[np.object, np.int64, np.int64, np.int64]
```

jina/types/document/uid.py
janandreschweiger: Right, sorry I missed that.

maximilianwerk:
Hey @janandreschweiger since our unit tests timed out regularly, we have changed our CI/CD pipeline. The branch must be rebased on our current master. Do you like to do it or would you like me to take over the branch and fix the issues?

Cheers
Maximilian

janandreschweiger:
Hey @maximilianwerk I tried it, but it somehow didn't work. So yes, it would be very nice if you could give it a try.

maximilianwerk:
Reopened with freshly merged master branch into it in #1666

PR #1555, opened 2020-12-28
feat: document set dundermethod add and iadd

florian-hoenicke:


jina/types/sets/document.py
```diff
@@ -63,6 +63,19 @@ def __getitem__(self, item):
         else:
             raise IndexError(f'do not support this index {item}')
 
+    def __add__(self, other: 'DocumentSet'):
+        sum = DocumentSet([])
```
JoanFM, line 67: sum seems like a weird name?

hanxiao: shadowing built-in, better use other, e.g. `_tmp`, or simply `v`

florian-hoenicke: thanks changed

JoanFM (approved):

PR #1565, opened 2020-12-30
refactor(peas): avoid nested-try-exception to improve readability

nan-wang:
refactor(peas): avoid nested-try-exception to improve readability

### DIFF
--- a/jina/peapods/peas/__init__.py
+++ b/jina/peapods/peas/__init__.py
@@ -121,21 +121,29 @@ class BasePea(ExitStack):
 
         """
         try:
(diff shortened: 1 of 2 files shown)


JoanFM (approved):

hanxiao:
close as it is overridden by #1539 

also:
- i don't see the point of changing nest try to duplicate try-catch.
- also, this nested try-catch is temp solution, as https://jinaai.slack.com/archives/C018QC78L84/p1608653613005700?thread_ts=1608653438.004800&cid=C018QC78L84

PR #1566, opened 2020-12-30
refactor(tests): clean up codes in testing `hubapi`

nan-wang:
refactor(tests): clean up codes in testing `hubapi`


JoanFM (dismissed):

florian-hoenicke:
I think it is failing because the access.yml is not defined right?
We could create it before running `_fetch_access_token`

florian-hoenicke (changes requested):
I think it is failing because the access.yml is not defined right?
We could create it before running _fetch_access_token

tests/unit/docker/test_hubapi.py
```diff
@@ -52,9 +49,7 @@ def test_hubapi_list(mocker):
...
 
-@mock.patch('jina.docker.hubapi.JAML.load')
 def test_fetch_access_token(mocker):
-    token_string = json.dumps({'access_token':'dummy_token'})
-    token_json = json.loads(token_string)
-    mocker.return_value = token_json
-    _fetch_access_token(logger=getLogger())
\ No newline at end of file
+    target_val = 'dummy_token'
+    mocker.patch('jina.docker.hubapi.JAML.load', return_value={'access_token': target_val})
```
florian-hoenicke, line 54: we could insert to make sure the file exists credentials_file().touch()

florian-hoenicke (changes requested):

tests/unit/docker/test_hubapi.py
nan-wang: Do you mean to insert here? We've ensured the file exists. Anyway, touching an empty one doesn't make sense in this case. 

https://github.com/jina-ai/jina/blob/master/jina/docker/hubapi.py#L250

nan-wang: ok, I see your point.

PR #1570, opened 2020-12-30
refactor: change metric evaluator name adapt to eval_at

JoanFM:


florian-hoenicke (changes requested):
line 87 in test_ranking_evaluate_driver  has to be changed to   
`assert doc.evaluations[0].op_name == 'Precision@2'`

Also for some reason the function `test_ranking_evaluate_driver` is there twice. We can remove one of them.

JoanFM:
> line 87 in test_ranking_evaluate_driver has to be changed to
> `assert doc.evaluations[0].op_name == 'Precision@2'`
> 
> Also for some reason the function `test_ranking_evaluate_driver` is there twice. We can remove one of them.

Is it exactly the same? if it is exactly the same yes we can remove it, otherwise we rename one.

I saw, I want to change to 

`assert doc.evaluations[0].op_name == executor.metric`

Would u have time to do these couple of changes on this branch?

florian-hoenicke:
sure - I do it

tests/unit/drivers/test_rankingevaluation_driver.py
```diff
@@ -65,26 +65,26 @@ def add_matches(doc: jina_pb2.DocumentProto, num_matches):
...
-def test_ranking_evaluate_driver(ruuningavg_rank_evaluate_driver,
+def test_ranking_evaluate_ruuningavg_driver(ruuningavg_rank_evaluate_driver,
                                  ground_truth_pairs):
     ruuningavg_rank_evaluate_driver.attach(executor=PrecisionEvaluator(eval_at=2), pea=None)
     ruuningavg_rank_evaluate_driver._apply_all(ground_truth_pairs)
     for pair in ground_truth_pairs:
         doc = pair.doc
         assert len(doc.evaluations) == 1
-        assert doc.evaluations[0].op_name == 'Precision@N'
+        assert doc.evaluations[0].op_name == 'Precision@2'
```
JoanFM, line 87: can u change this to have 
``` executor = PrecisionEvaluator(eval_at=2) 
driver.attach(executor=executor, pea=None)

assert driver.op_name == executor.metric
```
Naming might be wrong, but like this the responsibility of testing the actual name is tested in the executor while the driver tests that the name is taken from executor

JoanFM (commented):
I like it

JoanFM:
@florian-hoenicke if u can, can u try running jina hello-world and see if the evaluation results are properly shown in the html resulting?

florian-hoenicke:
As a result I get:
`Precision@50: 65.47% Recall@50: 0.55%`
The precision looks really low to me.
From what I see in the results, it should be much higher.

JoanFM:
> As a result I get:
> `Precision@50: 65.47% Recall@50: 0.55%`
> The precision looks really low to me.
> From what I see in the results, it should be much higher.

thats okey I just want to check the name

hanxiao (changes requested):
I removed this intentionally. To me it is redundant as we already have `self.name` for all executor. Can we just use `self.name` instead of creating a new property.

I suggest remove redundant `metric` property and use:
```python
op.name = f'{self.exec.name}@{self.eval_at}'
```

JoanFM:
> I removed this intentionally. To me it is redundant as we already have `self.name` for all executor. Can we just use `self.name` instead of creating a new property.
> 
> I suggest remove redundant `metric` property and use:
> 
> ```python
> op.name = f'{self.exec.name}@{self.eval_at}'
> ```

it was to mantain alignment with hub but yes we can do this. Just wanted to align all

JoanFM:
Will do as @hanxiao suggests

florian-hoenicke:
I created a new pr for the two tests having the same method name.
https://github.com/jina-ai/jina/pull/1573/files

nan-wang (dismissed):
LGTM👍

nan-wang:
The unit test failed
https://github.com/jina-ai/jina/pull/1570/checks?check_run_id=1645486983#step:6:2024

JoanFM:
> The unit test failed
> https://github.com/jina-ai/jina/pull/1570/checks?check_run_id=1645486983#step:6:2024

Thanks for pointing it out

nan-wang (approved):
LGTM👍

PR #1577, opened 2021-01-02
fix(remote): remote type check and pod invocation

deepankarm:
This needs to be fixed to make https://github.com/jina-ai/jinad/pull/78 work. `jinad` checks will still fail with this PR.

### DIFF
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -16,7 +16,7 @@ class BasePod(ExitStack):
     Internally, the peas can run with the process/thread backend. They can be also run in their own containers
     """
 
-    def __init__(self, args: 'argparse.Namespace', needs: Set[str] = None):
+    def __init__(self, args: Union['argparse.Namespace', Dict], needs: Set[str] = None):
         """
 
         :param args: arguments parsed from the CLI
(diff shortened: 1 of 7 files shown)


deepankarm:
https://github.com/jina-ai/jinad/pull/78/checks?check_run_id=1643055891
`jinad@fix-runtime-refactotring` works with this branch.

jina/peapods/pods/__init__.py
```diff
@@ -31,7 +31,12 @@ def __init__(self, args: 'argparse.Namespace', needs: Set[str] = None):
         self.deducted_head = None
         self.deducted_tail = None
 
-        self.peas_args = self._parse_args(args)
+        if isinstance(args, Dict):
+            # This is used when a Pod is created in a remote context, where peas & their connections are already given.
+            self.peas_args = args
+        else:
+            self.peas_args = self._parse_args(args)
```
hanxiao, line 38: when is dict used? inside `JinadRuntime` sending this to `jinad`?

hanxiao (approved):

jina/peapods/pods/__init__.py
deepankarm: https://github.com/jina-ai/jina/blob/master/jina/peapods/runtimes/jinad/__init__.py#L28

before sending these args, they're converted to dict & on the receiving end, jinad converts dict to `argparse`

PR #1580, opened 2021-01-03
refactor(logs): remove legacy SSE logserver from core

hanxiao:
`jinad` and `fluentd` logger have been there for quite a while, yet there is no clue they work with dashboard. 

I'm leaving this PR here to push the progress a bit and clarify the scope. This PR removes the legacy SSE logger completely from the core, **breaking** the connection between the core and dashboard.


JoanFM (changes requested):
tests/integration/logging should be also removed

tests/integration/encode_driver_batching/test_encode_driver_batching.py
```diff
@@ -82,10 +82,10 @@ def document_generator(num_docs, num_chunks, num_chunks_chunks):
         yield doc
 
 
-@pytest.mark.parametrize('request_batch_size', [5, 100, 500])
+@pytest.mark.parametrize('request_batch_size', [8, 16, 32])
 @pytest.mark.parametrize('driver_batch_size', [8, 16, 64])
```
JoanFM, line 86: reduce also driver batch sizes so that most of the tests the driver batch size is smaller than the request size

JoanFM (changes requested):

tests/integration/encode_driver_batching/test_encode_driver_batching.py
JoanFM, line 86:
```suggestion
@pytest.mark.parametrize('driver_batch_size', [3, 4, 13])
```

tests/integration/encode_driver_batching/test_encode_driver_batching.py
```diff
@@ -82,10 +82,10 @@ def document_generator(num_docs, num_chunks, num_chunks_chunks):
...
 def test_encode_driver_batching(request_batch_size, driver_batch_size, tmpdir):
-    num_docs = 1315
+    num_docs = 128
```
JoanFM, line 88:
```suggestion
    num_docs = 137
```

JoanFM (changes requested):

PR #1583, opened 2021-01-03
test: add test with local flow and remote pods

JoanFM:
**Changes introduced**
- Test added that tests having a local Flow with remote Pods
- Uses complete_path to ensure files are found before sending to remote instances


jina/peapods/runtimes/jinad/api.py
```diff
@@ -9,12 +9,22 @@
...
 from ....jaml import JAML
 from ....logging import JinaLogger
+from ....jaml.helper import parse_config_source
 
 
 def _add_file_to_list(_file: str, _file_list: Set, logger: 'JinaLogger'):
     if _file and _file.endswith(('yml', 'yaml', 'py')):
-        if Path(_file).is_file():
-            _file_list.add(_file)
+        _, real_file = parse_config_source(_file,
```
florian-hoenicke, line 17: what do you think about using `_complete_path(path)`  and making it public `complete_path(path)`?
It would be easier to read.

florian-hoenicke (changes requested):

jina/peapods/runtimes/jinad/api.py
JoanFM: not strong opinion

florian-hoenicke: it would shrink 9 lines into 1

JoanFM: It may be better, but the number of lines are fake

nan-wang (approved):
LGTM👍

tests/integration/daemon/distributed/test_simple_local_remote/test_integration.py
```diff
@@ -0,0 +1,32 @@
...
+    text = 'cats rules'
+    m = mocker.Mock()
+
+    def validate_output(resp):
+        m()
+        assert len(resp.index.docs) == 1
+        assert resp.index.docs[0].text == text
+
+    os.environ['JINA_ENCODER_HOST'] = '172.28.1.1'
+    os.environ['JINA_INDEXER_HOST'] = '172.28.1.2'
```
nan-wang, line 24: Is it possible to put these into the `.env`? This can make maintenance easier.

nan-wang (changes requested):
LGTM👍

tests/integration/daemon/distributed/test_simple_local_remote/test_integration.py
JoanFM: no, because these env live locally with the Flow. so they are fixed by the docker compose definition

nan-wang (approved):
LGTM👍

PR #1588, opened 2021-01-04
test: add querylang set test

bwanglzu:
Add unit tests for QueryLangSet and minor bug fixing.

**What has been fixed**:
1. `__setitem__` expect set, not return
2. `self._docs_map` is not a valid variable in `QueryLangSet`.
3. `QueryLangProto` does not have an id field, while calling `build`.

Question: do we really need `build` method in `QueryLangSet`? the docstring was copied from `Document`

### DIFF
--- a/jina/types/sets/querylang.py
+++ b/jina/types/sets/querylang.py
@@ -31,7 +31,7 @@ class QueryLangSet(MutableSequence):
         if isinstance(key, int):
             self._querylangs_proto[key].CopyFrom(value.as_pb_object)
         elif isinstance(key, str):
-            return self._querylangs_map[key].CopyFrom(value.as_pb_object)
+            self._querylangs_map[key].CopyFrom(value.as_pb_object)
         else:
             raise IndexError(f'do not support this index {key}')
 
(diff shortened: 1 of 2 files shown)


jina/types/sets/querylang.py
```diff
@@ -85,4 +85,5 @@ def build(self):
         """Build a doc_id to doc mapping so one can later index a Document using
         doc_id as string key
         """
-        self._docs_map = {d.id: d for d in self._querylangs_proto}
+        # TODO This is a temp fix, QueryLangProto do not have an id field.
+        self._querylangs_map = {d.name: d for d in self._querylangs_proto}
```
nan-wang, line 89: Please fix the docstrings as well
```suggestion
        # TODO This is a temp fix, QueryLangProto do not have an id field.
        self._querylangs_map = {q.name: q for q in self._querylangs_proto}
```

nan-wang (changes requested):

jina/types/sets/querylang.py
```diff
@@ -85,4 +85,5 @@ def build(self):
         """Build a doc_id to doc mapping so one can later index a Document using
         doc_id as string key
```
hanxiao, line 86:
```suggestion
        """Build a name to QueryLang mapping so one can later index a QueryLang using
        name as string key
```

PR #1596, opened 2021-01-05
feat: volumes can be mapped to arbitrary path in container

### LINKED ISSUE #1438: Problem updating `faiss-search-example` (General problems with `indexers` , `ref_indexer` and  `ContainerPea`)
**Describe the bug**
I have been trying to update `faiss-search-example` (https://github.com/jina-ai/examples/pull/318) and I am struggling to get it right.

So the problem is related to `workspace` `ref_indexers` and `ContainerPea`

Let me try to explain in detail:
- At index time `NumpyIndexer` is working with `index_workspace` as workspace, there it serializes itself to `indexer_workspace/wrapidx.bin`

- At query time, the docker image is started mapping volume `indexer_workspace` (but the problem is that inside `docker` this path is not proper with respect to `workspace` inside the docker image (from where `jina pod` entrypoint is run). Then there is a missmatch of the `workspace` stored in `wrapidx.bin` and the one that `FaissIndexer` should work on.

- At query time `NumpyIndexer` is passed as `ref_indexer` to the `FaissIndexer`, and the `index_abspath` is extracted from the `ref_indexer` which is not considering where the volume is mapped.

As my experience, this problem has 2 sides of the story:
- We should think of providing more freedom to the user on what folder is mapping and where it will be mapped inside the container.

- Have more clarity in from where the `ContainerPea` is working (absolute, relative, etc...)

- Is the fact that we are using `workspace` as our base for hub image could conflict with the `workspace` with indexer files?

**EXTRA PROPOSAL:**
Create an extra examples test for `faiss-search-example`

JoanFM:
**Changes introduced**
Iteration to fix #1438 

It allows to have volumes mapped to arbitrary path in the container. This combined with #1595  can allow using `ref_indexer` inside Docker

### DIFF
--- a/jina/parsers/peapods/runtimes/container.py
+++ b/jina/parsers/peapods/runtimes/container.py
@@ -16,8 +16,8 @@ def mixin_container_runtime_parser(parser):
                     help='pull the latest image before running')
     gp.add_argument('--volumes', type=str, nargs='*', metavar='DIR',
                     help='the path on the host to be mounted inside the container. '
-                         'if separated by ":" the first part will be considered as the local path and the second '
-                         'part, the path in the container system. '
-                         'If no split provided, they will be mounted to the root path, i.e. /user/test/my-workspace '
-                         'will be mounted to /my-workspace inside the container. all volumes are mounted with '
+                         'if separated by ":" the first part will be considered as the local host path and the second '
+                         'part is the path in the container system. '
+                         'If no split provided, then the basename of that directory will be mounted into container's root path, e.g. --volumes="/user/test/my-workspace" '
+                         'will be mounted into "/my-workspace" inside the container. all volumes are mounted with '
                          'read-write mode.')
--- a/jina/parsers/peapods/runtimes/container.py
+++ b/jina/parsers/peapods/runtimes/container.py
@@ -16,5 +16,8 @@ def mixin_container_runtime_parser(parser):
                     help='pull the latest image before running')
     gp.add_argument('--volumes', type=str, nargs='*', metavar='DIR',
                     help='the path on the host to be mounted inside the container. '
-                         'they will be mounted to the root path, i.e. /user/test/my-workspace will be mounted to '
-                         '/my-workspace inside the container. all volumes are mounted with read-write mode.')
+                         'if separated by ":" the first part will be considered as the local path and the second '
+                         'part, the path in the container system. '
+                         'If no split provided, they will be mounted to the root path, i.e. /user/test/my-workspace '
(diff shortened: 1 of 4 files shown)


jina/parsers/peapods/runtimes/container.py
```diff
@@ -16,5 +16,8 @@ def mixin_container_runtime_parser(parser):
                     help='pull the latest image before running')
     gp.add_argument('--volumes', type=str, nargs='*', metavar='DIR',
                     help='the path on the host to be mounted inside the container. '
-                         'they will be mounted to the root path, i.e. /user/test/my-workspace will be mounted to '
-                         '/my-workspace inside the container. all volumes are mounted with read-write mode.')
+                         'if separated by ":" the first part will be considered as the local path and the second '
+                         'part, the path in the container system. '
+                         'If no split provided, they will be mounted to the root path, i.e. /user/test/my-workspace '
+                         'will be mounted to /my-workspace inside the container. all volumes are mounted with '
+                         'read-write mode.')
```
hanxiao, line 23:
```suggestion
                         'if separated by ":" the first part will be considered as the local host path and the second '
                         'part is the path in the container system. '
                         'If no split provided, then the basename of that directory will be mounted into container's root path, e.g. --volumes="/user/test/my-workspace" '
                         'will be mounted into "/my-workspace" inside the container. all volumes are mounted with '
                         'read-write mode.')
```

jina/parsers/peapods/runtimes/container.py
```diff
@@ -16,5 +16,8 @@ def mixin_container_runtime_parser(parser):
                     help='pull the latest image before running')
     gp.add_argument('--volumes', type=str, nargs='*', metavar='DIR',
                     help='the path on the host to be mounted inside the container. '
-                         'they will be mounted to the root path, i.e. /user/test/my-workspace will be mounted to '
-                         '/my-workspace inside the container. all volumes are mounted with read-write mode.')
+                         'if separated by ":" the first part will be considered as the local host path and the second '
+                         'part is the path in the container system. '
+                         'If no split provided, then the basename of that directory will be mounted into container's root path, e.g. --volumes="/user/test/my-workspace" '
```
hanxiao, line 21:
```suggestion
                         'If no split provided, then the basename of that directory will be mounted into container\'s root path, e.g. --volumes="/user/test/my-workspace" '
```

JoanFM:
@hanxiao could you fix the commit-lint error, it is a common trap when commiting suggestions

JoanFM:
> @hanxiao could you fix the commit-lint error, it is a common trap when commiting suggestions

Fixed this , no prob

hanxiao (approved):

PR #1597, opened 2021-01-05
fix: clarify py_modules usage in 1546

hanxiao:
https://github.com/jina-ai/jina-hub/blob/master/README.md#use-py_modules-to-import-multiple-files


hanxiao:
> What is the learnings from here? Which patterns can be imported and which can't?

in test docstring

jina/importer.py
```diff
@@ -277,7 +277,12 @@ def _path_import(absolute_path: str) -> Optional[ModuleType]:
             spec = importlib.util.spec_from_file_location('jinahub', absolute_path)
             module = importlib.util.module_from_spec(spec)
             user_module_name = os.path.splitext(os.path.basename(absolute_path))[0]
-            sys.modules[f'{spec.name}.{user_module_name}'] = module
+            if user_module_name == '__init__':
```
JoanFM, line 280: this may put some problems in many hub executors right?

hanxiao: no, irrelevant, magic is in `spec.loader.exec_module(module)`

PR #1551, opened 2020-12-28
feat(gateway): websocket based streaming and client

deepankarm:
[NOT TO BE MERGED]

PR #1608, opened 2021-01-06
feat(gateway): websocket based streaming and client

deepankarm:
Recreated duplicate of https://github.com/jina-ai/jina/pull/1551

#### Gateway
Added a `StreamingEndpoint` (**WS /stream**) to `RESTGateway`

#### Client
Added `WebSocketClient` & `AsyncWebSocketClient` to stream requests from `StreamingEndpoint` for `Flow` & `AsyncFlow` respectively

#### Flow
Enabled websocket based clients for REST gateway

image

#### Notes & Issues:
- ~~With the current implementation, gateway is blocking, as we wait for each response to come back on the websocket~~ **(FIXED)**
- Indexing/searching for flows with `needs` fail -  `Flow().add().add(needs='gateway')` **(TO BE FIXED)**
- Unlike gRPC we cannot send an arbitrary python iterator via websockets, hence every request/response is converted to `bytes` or `json` at the gateway & the client. Hence the performance might be poor compared to gRPC.
- Gateway expects the input to be in `bytes` only for now. Need to add support for `json`
- The reason for having a different way of `send` & `recv` in websockets when compared to `grpc.async_call` - [websockets-comment](https://github.com/aaugustin/websockets/issues/807#issuecomment-752636210)
- The reason for selecting `wsproto` library over `websockets` with UvicornServer, which adds a new dependency to jina - [This PR](https://github.com/encode/uvicorn/pull/538) was never merged in uvicorn.


jina/peapods/runtimes/asyncio/rest/app.py
```diff
@@ -0,0 +1,164 @@
...
+                        await websocket.send_bytes(response.SerializeToString())
+                    else:
+                        await websocket.send_json(response.to_json())
+                    self.num_responses += 1
+            except Exception as e:
+                logger.error(f'Got an exception in handle_send: {repr(e)}')
+
+        async def decode(self, websocket: WebSocket, message: Message) -> Any:
+            if 'text' in message or 'json' in message:
+                self.client_encoding = 'text'
```
florian-hoenicke, line 153: this could be a good test case to cover

hanxiao:
im taking this branch

Issue #1575, opened 2020-12-31
PR#1361 test is misleading, reverting


hanxiao:
**Describe the bug**

This test is misleading as `HubIO._test_build` does not accept image as `str`

```python
image, log = self._client.images.build(...)
```

Please read Docker SDK carefully and please notice that:

> The first item is the :py:class:`Image` object ...

Mentioned here: https://github.com/jina-ai/jina/pull/1361#pullrequestreview-560556769

Based on this questionable test, I worry about the effectiveness of PR#1361

**Describe how you solve it**

---

**Environment**

**Screenshots**

hanxiao:
In #1539 I removed the test introduced by #1361 as it didn't test properly - rthe test itself is buggy. Please clarify the purpose of #1361  and redesign the test. @rutujasurve94

rutujasurve94:
Redesigning this test, updating on this issue thread.

PR #1609, opened 2021-01-06
fix: test for hubio test build

rutujasurve94:
Fix for #1575 using `docker image` object instead of `string` argument for `test_build` in `hubio`

### DIFF
--- a/tests/unit/docker/test_hub_build_level.py
+++ b/tests/unit/docker/test_hub_build_level.py
@@ -1,0 +1,40 @@
+import os
+
+import docker
+import pytest
+import mock
+from mock import Mock, patch
+
+from jina.docker.hubio import HubIO
+from jina.enums import BuildTestLevel
+from jina.peapods import Pod
+from jina.executors import BaseExecutor
+from jina.executors import AnyExecutor
+from jina.parsers.hub import set_hub_build_parser
+
+@pytest.fixture
+def mock_load_config():
+    return BaseExecutor
+
+cli = docker.APIClient(base_url='unix://var/run/docker.sock')
+
+def test_hub_build_level_pass(monkeypatch, mock_load_config):
+
+    monkeypatch.setattr(BaseExecutor, "load_config", mock_load_config)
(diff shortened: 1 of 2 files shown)


tests/unit/docker/test_hub_build_level.py
```diff
@@ -0,0 +1,40 @@
...
+
+@pytest.fixture
+def mock_load_config():
+    return BaseExecutor
+
+cli = docker.APIClient(base_url='unix://var/run/docker.sock')
+
+def test_hub_build_level_pass(monkeypatch, mock_load_config):
+
+    monkeypatch.setattr(BaseExecutor, "load_config", mock_load_config)
```
deepankarm, line 23: What does this mock do?

tests/unit/docker/test_hub_build_level.py
```diff
@@ -0,0 +1,40 @@
...
+    args = set_hub_build_parser().parse_args(['path/hub-mwu', '--push', '--host-info', '--test-level', 'EXECUTOR'])
+
+    docker_image = cli.get_image('jinahub/pod.dummy_mwu_encoder')
+    p_names, failed_levels = HubIO._test_build(docker_image, BuildTestLevel.EXECUTOR, "sample/yaml.yaml", 60, True)
```
deepankarm, line 27: In what case, does a user send an `image` object? 
Should this remain an `str` and inside the `_test_build` we do the pull/build? @JoanFM

nan-wang (changes requested):

tests/unit/docker/test_hub_build_level.py
rutujasurve94: this is an `image` object as per our recent discussion

tests/unit/docker/test_hub_build_level.py
rutujasurve94: Removed the mock to use an actual `test-yaml` file for the `BaseExecutor` as discussed.

deepankarm (approved):

nan-wang (approved):
LGTM👍

PR #1610, opened 2021-01-06
refactor: moving jinad into daemon module of the core

hanxiao:
- [x] code 
- [x] test
- [x] Dockerfile
- [x] setup 
- [x] README (entrypoint changed, will rewrite in the next PR)
- [x] API docs generation
- [x] GH workflows


daemon/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+                    log_level='error')
+    server = Server(config=config)
+    server.run()
+    daemon_logger.info('bye!')
+
+
+def _start_fluentd():
+    daemon_logger.info('starting fluentd')
+    cfg = pkg_resources.resource_filename('jina', 'resources/fluent.conf')
+    subprocess.Popen(['fluentd', '-c', cfg])
```
deepankarm, line 75: Fluentd also ships as a daemon (td-agent), which means we can have systemctl manage it along with jinad. We can avoid having `fluentd` invocation inside python code, as this makes `jinad` unusable without `fluentd` command. 
https://github.com/jina-ai/jinad/blob/main/scripts/deb-systemd.sh#L49

hanxiao: let's handle this as another PR. right now it is too ahead to put `systemctl` together, the goal of this PR is to make the jinad usable in CICD.

hanxiao: btw,
```bash
curl -L https://raw.githubusercontent.com/jina-ai/jinad/main/scripts/deb-systemd.sh | bash
```
this is good, but later will make it shorter, something like `curl -L https://daemon-sh.jina.ai | bash`

PR #1611, opened 2021-01-06
fix: content hash should not depend on chunks

cristianmtr:
Test added was failing before the change introduced in the hashing method

### DIFF
--- a/jina/types/document/uid.py
+++ b/jina/types/document/uid.py
@@ -45,6 +45,8 @@ def get_content_hash(doc: 'DocumentProto') -> str:
     doc_without_id = DocumentProto()
     doc_without_id.CopyFrom(doc)
     doc_without_id.id = ""
+    del doc_without_id.chunks[:]
+    del doc_without_id.matches[:]
     return blake2b(doc_without_id.SerializeToString(), digest_size=_digest_size).hexdigest()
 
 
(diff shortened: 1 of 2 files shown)


jina/types/document/uid.py
```diff
@@ -45,6 +45,7 @@ def get_content_hash(doc: 'DocumentProto') -> str:
     doc_without_id = DocumentProto()
     doc_without_id.CopyFrom(doc)
     doc_without_id.id = ""
+    del doc_without_id.chunks[:]
```
JoanFM, line 48: delete matches as well then

JoanFM (changes requested):

jina/types/document/uid.py
JoanFM, line 48: This method should be overriden for MultiModalDocument where the chunks contents are considered for id. it can be done later

cristianmtr: done

theUnkownName: I thought the id should be only from the content itself not the chunks?

cristianmtr: I'll open an issue as a reminder. I'll also mark it as 'good first issue', maybe someone will pick it up before we get to it ;)

cristianmtr:
Seems some async test is failing, `tests/unit/flow/test_asyncflow.py::test_run_async_flow_other_task_concurrent`. https://github.com/jina-ai/jina/pull/1611/checks?check_run_id=1657292065#step:6:2069

Doesn't seem to be related, does it?

hanxiao (approved):

JoanFM (commented):
This content hash logic should be based on a mask or a set of fields to be added (whitelisted) more than a set of fields to be removed. 

It will be otherwise unmantainable as the structure of the document grows.

PR #1613, opened 2021-01-06
crud intergration tests: docidcache-centric

cristianmtr:
Main integration tests for CRUD, focusing on using the new DocIDCache functionality

- [x] Document content_hash incorrectly includes child chunks
- [x] docs generator and testing it
- [x] basic structure of tests & parameters
- [x] `UniquePbIndexer` and `UniqueVectorIndexer` do not support `delete` and `update`
- [x] issue with handling missing keys
- [x] DocIDCache uses savepath / filename differently than most Indexers
- [x] adding Delete and Update to CacheDriver


JoanFM:
Now that I think, DocIDCache does not seem to be a proper name for a class that can cache based on content hash. I think this needs to be renamed DocCache.

cristianmtr:
> Now that I think, DocIDCache does not seem to be a proper name for a class that can cache based on content hash. I think this needs to be renamed DocCache.

I'll open a PR to refactor it everywhere and then will rebase

jina/executors/indexers/cache.py
```diff
@@ -127,11 +127,13 @@ def delete(self, keys: Iterator['UniqueId'], *args, **kwargs):
 
     def get_add_handler(self):
         # not needed, as we use the queryhandler
-        pass
+        # FIXME better way to silence warnings
+        return 1
```
JoanFM, line 131: if it is not used, raise a NotImplementedError

jina/types/document/uid.py
```diff
@@ -45,6 +45,8 @@ def get_content_hash(doc: 'DocumentProto') -> str:
     doc_without_id = DocumentProto()
     doc_without_id.CopyFrom(doc)
     doc_without_id.id = ""
+    del doc_without_id.chunks[:]
```
JoanFM, line 48: realign with master to have these changes

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,226 @@
...
+    vector_pod = dict(
+        uses=cur_dir / 'yml' / 'cp_cache_pb.yml',
+        name='inc_vec'
+    )
+    pb_pod = dict(
+        uses=cur_dir / 'yml' / 'cp_cache_vector.yml',
+        name='inc_doc'
+    )
+
+    if shards == 3:
```
JoanFM, line 34: You can have this config in the `pb_pod` and have `shards` parametrized. This construction is too noisy.

`separated_workspace`, `polling` and `uses_after` will not affect if shards=1

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+        vector_pod.update(
+            sharding_config
+        )
+
+    f: Flow = (Flow()
+               .add(**pb_pod)
+               .add(**vector_pod))
+
+    if indexers == 'parallel':
```
JoanFM, line 53: if-else better construction than changing f.

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,226 @@
...
+            timeout_ready='-1'
+        )
+        pb_pod.update(
+            sharding_config
+        )
+        vector_pod.update(
+            sharding_config
+        )
+
+    f: Flow = (Flow()
```
JoanFM, line 49: no type hint needed inside the code

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,226 @@
...
+        vector_pod.update(
+            dict(
+                needs=['gateway']
+            )
+        )
+        f = (Flow()
+             .add(**pb_pod)
+             .add(**vector_pod)
+             .add(needs=['inc_vec', 'inc_doc']))
+    flow_plot = os.path.join(os.environ.get('JINA_TEST_CACHE_CRUD_WORKSPACE'), f'flow-{indexers}-{shards}.svg')
```
JoanFM, line 63: why this?

tests/integration/docidcache/test_crud_cache.py
JoanFM: I do prefer to have the `ymls` parametrized with `environment variables` (or moustache ones) so that just reading the yaml files, it is clear what is being tested.

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+
+    f: Flow = (Flow()
+               .add(**pb_pod)
+               .add(**vector_pod))
+
+    if indexers == 'parallel':
+        vector_pod.update(
+            dict(
+                needs=['gateway']
```
JoanFM, line 56: needs can also be parametrized with ENV variables.

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,226 @@
...
+                          (3, True, 0),
+                          (3, True, 10),
+                          (3, True, 100),
+                          (5, False, 0),
+                          (5, False, 10),
+                          (5, False, 100),
+                          (5, True, 0),
+                          (5, True, 10),
+                          (5, True, 100)])
+def test_docs_generator(chunks, content_same, nr):
```
JoanFM, line 119: we should not be adding a test for a generator added for the test

JoanFM (changes requested):

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,257 @@
...
+
+import numpy as np
+import pytest
+
+from jina import Flow, Document
+from jina.executors.indexers import BaseIndexer
+from jina.executors.indexers.cache import DocIDCache
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
+from jina.executors.indexers.vector import NumpyIndexer
+from tests.integration.incremental_indexing import random_workspace
```
JoanFM, line 11: not a fan of adding dependencies between tests, better duplicate code

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,257 @@
...
+                    else:
+                        c.text = f'hello world from chunk {j}'
+                        c.embedding = np.random.random([9])
+
+                next_chunk_id += 1
+                d.chunks.append(c)
+        yield d
+
+
+@pytest.mark.parametrize('chunks, content_same, nr',
```
JoanFM, line 105: Cartesian product is achieved this way

```suggestion
@pytest.mark.parametrize('chunks', [0, 3, 5])
@pytest.mark.parametrize('content_same', [False, True])
@pytest.mark.parametrize('nr', [0, 10, 100])
```

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,257 @@
...
+            if content_same:
+                assert c.text == 'hello world from chunk'
+                np.testing.assert_almost_equal(c.embedding, c_embedding)
+            else:
+                assert c.text == f'hello world from chunk {j}'
+                assert c.embedding.shape == c_embedding.shape
+    print(f'{len(ids_used)=}')
+
+
+@pytest.mark.parametrize('indexers, field, shards, chunks, same_content',
```
JoanFM, line 154: same here

JoanFM (changes requested):

jina/executors/indexers/cache.py
cristianmtr: This crashes the execution

tests/integration/docidcache/test_crud_cache.py
cristianmtr: I would still need to keep it as code for the parallel (`needs`) part, right?

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Was just for debugging

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Was needed for sanity testing some of the issues. Wanted to clarify what was wrong: the expectation of the test, or the document generation itself.

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Hm not sure how to do that. Is there an example somewhere?

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Was only temporary. I pruned them a lot, so as not to have too many (unnecessary) tests

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,257 @@
...
+import numpy as np
+import pytest
+
+from jina import Flow, Document
+from jina.executors.indexers import BaseIndexer
+from jina.executors.indexers.cache import DocIDCache
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
+from jina.executors.indexers.vector import NumpyIndexer
+from tests.integration.incremental_indexing import random_workspace
+from pathlib import Path
```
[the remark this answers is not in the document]
cristianmtr: Yes. Not sure why PyCharm didn't automatically sort them.

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Removed. Didn't notice it was defined there when I did auto-import

tests/integration/docidcache/test_crud_cache.py
JoanFM: needs and parallel are independent things, but u can also paramrtrize a list with env variable. I think is more readable to have flows in yaml files

jina/executors/indexers/cache.py
JoanFM: so why is this used? why does this even exist? I do prefer to have pass than return `1`

JoanFM (commented):
Is it possible to have the proper way of building cartesian products in `pytest`,

and to parametrize Flows with variables instead of with complex `if ... else ...` logic

cristianmtr:
> Is it possible to have the proper way of building cartesian products in `pytest`,
> 
> and to parametrize Flows with variables instead of with complex `if ... else ...` logic

I don't want to do a full cartesian product. I only created them initially to see the full spread of possible combinations. They were too many. I have pruned them manually to only certain combination that I thought would capture the corner cases.

I am working on refactoring the get flow to use yml but it's harder than I expected.

I am not sure how to add the `needs=...` with env. vars.

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,271 @@
...
+                    else:
+                        c.text = f'hello world from chunk {j}'
+                        c.embedding = np.random.random([9])
+
+                next_chunk_id += 1
+                d.chunks.append(c)
+        yield d
+
+
+@pytest.mark.parametrize('chunks, same_content, nr',
```
JoanFM, line 99: here, which possibilites are pruned?

I see every combination of [0, 3, 5] and [False, True] and [0, 10, 100]

tests/integration/docidcache/test_crud_cache.py
JoanFM: here yes there is pruning.

Does it save so much time to prune the cartesian product? because if it is not so big I think readability can be improved a lot

JoanFM:
> > Is it possible to have the proper way of building cartesian products in `pytest`,
> > and to parametrize Flows with variables instead of with complex `if ... else ...` logic
> 
> I don't want to do a full cartesian product. I only created them initially to see the full spread of possible combinations. They were too many. I have pruned them manually to only certain combination that I thought would capture the corner cases.
> 
> I am working on refactoring the get flow to use yml but it's harder than I expected.
> 
> I am not sure how to add the `needs=...` with env. vars.

This can work:

```  indexer:
    needs: $JINA_INDEXER_NEEDS
```

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Not pushed yet. Working on it

JoanFM:
> > Is it possible to have the proper way of building cartesian products in `pytest`,
> > and to parametrize Flows with variables instead of with complex `if ... else ...` logic
> 
> I don't want to do a full cartesian product. I only created them initially to see the full spread of possible combinations. They were too many. I have pruned them manually to only certain combination that I thought would capture the corner cases.
> 
> I am working on refactoring the get flow to use yml but it's harder than I expected.
> 
> I am not sure how to add the `needs=...` with env. vars.

You can set `os.environ['JINA_NEEDS'] = '[pod1, pod2]'` being `[pod1, pod2]` a string and not a list

jina/executors/indexers/cache.py
```diff
@@ -122,11 +122,13 @@ def delete(self, keys: Iterator['UniqueId'], *args, **kwargs):
...
 
     def get_query_handler(self) -> CacheHandler:
-        return self.CacheHandler(self.index_abspath, self.logger)
+        return self.CacheHandler(self.save_abspath, self.logger)
```
JoanFM, line 129: are we sure about that? index_abspath makes more sense since it should contain the actual indexed data?

tests/integration/docidcache/test_crud_cache.py
JoanFM: This is not the way to do `cartesian product` in pytest.

It should be :

```
@pytest.mark.parametrize('chunks', [0, 3, 5])
@pytest.mark.parametrize('docs_same_content', [False, True])
@pytest.mark.parametrize('docs_nr', [0,10,100])
```

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,215 @@
...
+def config(field, tmp_workspace, shards, indexers):
+    os.environ['JINA_SHARDS'] = str(shards)
+    os.environ['JINA_CACHE_FIELD'] = field
+    os.environ['JINA_TEST_CACHE_CRUD_WORKSPACE'] = str(tmp_workspace)
+    os.environ['JINA_KV_IDX_NAME'] = KV_IDX_FILENAME.split('.bin')[0]
+    os.environ['JINA_VEC_IDX_NAME'] = VEC_IDX_FILENAME.split('.bin')[0]
+    if indexers == 'parallel':
+        # the second indexer will be directly connected to entry gateway
+        os.environ['JINA_INDEXER_NEEDS'] = 'gateway'
+        os.environ['JINA_MERGER_NEEDS'] = "[inc_vec, inc_doc]"
```
JoanFM, line 29: consistency with single quote if possible

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,215 @@
...
+
+    docs.extend(new_docs)
+    # delete
+    with f:
+        f.delete(docs)
+
+    check_indexers_size(chunks, 0, field, tmp_path, same_content, shards, 'delete')
+
+
+def check_indexers_size(chunks, nr_docs, field, tmp_path, same_content, shards, post_op):
```
JoanFM, line 178: This part is really too complex, and the code seems to assume a lot of features from a single test, which may make it hard to mantain. Every small change to things in the `indexers` can make this example fail.

JoanFM: I think if we load the Indexer from the `indexer` yaml with the same workspace it will be able to load it from the `.bin` without this hardcode tricks

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,215 @@
...
+
+from jina import Flow, Document
+from jina.executors.indexers import BaseIndexer
+from jina.executors.indexers.cache import DocIDCache
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
+from jina.executors.indexers.vector import NumpyIndexer
+
+cur_dir = Path(os.path.dirname(os.path.abspath(__file__)))
+
+KV_IDX_FILENAME = 'kv_idx.bin'
```
JoanFM, line 16: related to other message above, seems weird to depend so heavily on having these file names

JoanFM (changes requested):

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,215 @@
...
+        tmp_path,
+        indexers,
+        field,
+        shards,
+        chunks,
+        same_content
+):
+    config(field, tmp_path, shards, indexers)
+    print(f'{tmp_path=}')
+    f = Flow.load_config('yml/flow.yml')
```
JoanFM, line 148: can u use absolute path with reference filename dir so that test can be executed from different paths

jina/executors/indexers/cache.py
cristianmtr: Nope, that provides it a weird path, smth like `/tmp/whatever`. Using `save_abspath` provides the path to the workspace + shard dir + `name`

tests/integration/docidcache/test_crud_cache.py
cristianmtr: This test should be self-contained. Any other special corner-cases should become their own tests.

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Well, we need to get the right file

tests/integration/docidcache/test_crud_cache.py
JoanFM: If you use `BaseExecutor.load_config` from the `indexer` uses you will be able to check the `size`. The test code would look much simpler

tests/integration/docidcache/test_crud_cache.py
JoanFM: No if u use `BaseExecutor.load_config` from the `indexer` uses config, it will be able to load from the `bin` automatically

JoanFM: I know it is not ideal to trust other features, but in this case I do believe that it simplifies the code drastically

JoanFM:
Problem seems to be that `update` is not clearing properly somehow `blob shape`. I think it may be more of a primitive types fault than the Update method itself, but it may have exposed the issue and it needs to be solved to continue

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Done

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,314 @@
...
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
+from jina.executors.indexers.vector import NumpyIndexer
+
+cur_dir = Path(os.path.dirname(os.path.abspath(__file__)))
+
+KV_IDX_FILENAME = 'kv_idx.bin'
+VEC_IDX_FILENAME = 'vec_idx.bin'
+DOCS_TO_SEARCH = 1
+TOP_K = 5
+BATCH_SIZE = 4
```
JoanFM, line 19: To adapt to latest changes, change `BATCH_SIZE` to `REQUEST_SIZE` and also the `kwarg` in the .index()

JoanFM (changes requested):
Just a minor thing,

However, I still think we should think how to better test for the index size without having such a complex code for testing

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,314 @@
...
+DOCS_TO_INDEX = 10
+
+pytest.register_assert_rewrite('check_indexers_size')
```
JoanFM, line 22: what does this do? just curious?

tests/integration/docidcache/test_crud_cache.py
```diff
@@ -0,0 +1,314 @@
...
+                    else:
+                        c.text = f'hello world from chunk {j}'
+                        c.embedding = np.random.random(d_embedding.shape)
+
+                next_chunk_id += 1
+                d.chunks.append(c)
+        yield d
+
+
+docs_chunks = [0, 3, 5]
```
JoanFM, line 74: this is not used anymore ??

JoanFM (changes requested):

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Right, will remove. Thanks. They're declared in-line

tests/integration/docidcache/test_crud_cache.py
cristianmtr: I was trying to fix a problem with pytest not showing me the full assert location in PyCharm. Thought this would help. Didn't. Will remove

tests/integration/docidcache/test_crud_cache.py
cristianmtr: Done

cristianmtr:
> Just a minor thing,
> 
> However, I still think we should think how to better test for the index size without having such a complex code for testing

I agree that the engineering of the tests shouldn't be so complex. 

But on the other hand, as an advantage, we can see more clearly how changing some parameters in the cache would affect the size / output of the indexers. Having them as separate tests, with hardcoded apparently _magic_ numbers would have made them more opaque.

JoanFM (approved):

PR #1616, opened 2021-01-07
refactor: doc content hash

hanxiao:
refactor on #1611

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -2,17 +2,18 @@ import base64
 import os
 import urllib.parse
 import urllib.request
-import warnings
-from typing import Union, Dict, Optional, TypeVar, Any, Callable, Sequence
+from hashlib import blake2b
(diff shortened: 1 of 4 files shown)


jina/types/document/__init__.py
```diff
@@ -186,9 +187,16 @@ def modality(self, value: str):
     def content_hash(self):
         return self._document.content_hash
 
-    def update_content_hash(self):
-        """Update the document hash according to its content."""
-        self._document.content_hash = get_content_hash(self._document)
+    def update_content_hash(self, mask: Tuple[str] = ('id', 'chunks', 'matches', 'content_hash')) -> None:
```
JoanFM, line 190: I opened the issue to have exactly the opposite logic, I want the mask to be the fields that we want to include and not the ones we want to exclude.

cristianmtr: > I opened the issue to have exactly the opposite logic, I want the mask to be the fields that we want to include and not the ones we want to exclude.

I also think this is more resilient to any future adding of new fields

PR #1619, opened 2021-01-07
fix: remove warning jina version missing from manifest

JoanFM:


jina/docker/hubio.py
```diff
@@ -31,7 +31,7 @@
 
 _allowed = {'name', 'description', 'author', 'url',
             'documentation', 'version', 'vendor', 'license', 'avatar',
-            'jina_version', 'platform', 'update', 'keywords'}
+            'platform', 'update', 'keywords'}
```
cristianmtr, line 34: If it's not there will it be *allowed* to be added?

JoanFM: Well, we did accept to remove this right? I just want to remove the warning because it seems missleading

cristianmtr: Ok. Then maybe `_allowed` is the wrong name for this variable. It should be `_expected` or something like that

JoanFM: I think we do not need all of them

cristianmtr: But then it's not about being `allowed`, it's about being `required`, perhaps

JoanFM: I do not know what's the best logic or if there are required fields @deepankarm ?

deepankarm: We have a list of `required` & a list of `_allowed`. `ValueError` is raised if something is missing from `required` & a warning is given if something is missing from `_allowed`. 
https://github.com/jina-ai/jina/blob/master/jina/docker/hubio.py#L590

nan-wang (approved):
LGTM👍

PR #1622, opened 2021-01-07
fix: example requires numpy to be imported

FionnD:
A user copying and pasting the existing sample will hit the following error. 
```
NameError: name 'numpy' is not defined
```

This is avoidable by including import numpy in the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Let's create some random data and index it:
 
 ```python
 from jina import Document
+import numpy
 
 with Flow().add() as f:
     f.index((Document() for _ in range(10)))  # index raw Jina Documents
--- a/README.md
+++ b/README.md
@@ -133,8 +133,8 @@ To visualize the Flow, simply chain it with `.plot('my-flow.svg')`. If you are u
 Let's create some random data and index it:
 
 ```python
+import numpy 
 from jina import Document
-import numpy
 
 with Flow().add() as f:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,6 +134,7 @@ Let's create some random data and index it:
 
 ```python
 from jina import Document
+import numpy
```
JoanFM, line 137: please move numpy import before jina to respect standard import order

JoanFM (changes requested):

README.md
FionnD: good spot, reordered

cristianmtr (approved):

JoanFM (commented):
@alexcg1 do we want to add this or we assume it can be incomplete?

theUnkownName (approved):
Looks good

hanxiao (changes requested):
it was intentionally removed for saving one-line, example code snippets are for showing capability in a quick & lean way. It is optimized for size, not for reproducibility. For reproducibility, I will use [repl.it](https://repl.it/) for all example snippet.

hanxiao:
will merge for now as it is just one line code change. For other example snippet, please see my comment above

alexcg1:
Very much agreed with @hanxiao . README should be for showing /how/ code works in general, not showing /exact/ code to pull it off

PR #1623, opened 2021-01-07
fix: fix eval name from driver

JoanFM:


hanxiao (approved):

jina/drivers/evaluate.py
```diff
@@ -55,7 +55,7 @@ def _apply_all(
             if self._running_avg:
                 evaluation.value = self.exec.mean
 
-            if hasattr(self.exec, 'eval_at'):
+            if hasattr(self.exec, 'eval_at') and self.exec.eval_at is not None:
```
hanxiao, line 58:
```suggestion
            if getattr(self.exec, 'eval_at', None):
```

jina/drivers/evaluate.py
```diff
@@ -55,7 +55,7 @@ def _apply_all(
             if self._running_avg:
                 evaluation.value = self.exec.mean
 
-            if hasattr(self.exec, 'eval_at'):
+            if getattr(self.exec, 'eval_at', None):
                 evaluation.op_name = f'{self.exec.__class__.__name__}@{self.exec.eval_at}'
             else:
                 evaluation.op_name = self.exec.__class__.__name__
```
hanxiao, line 61:
```suggestion
                evaluation.op_name = f'{self.exec.name}@{self.exec.eval_at}'
            else:
                evaluation.op_name = self.exec.name
```

I think we can simply use `.name` attribute of the executor - one of the `metas` attribute of any `Executor`. Also, since `eval_at` is kind of `BaseRankingEvaluator` specific, the `BaseRankingEvaluator` could just have the following somewhere in `__init__` of `post_init`:

```
if self.eval_at:
   self.name += f`@{self.eval_at}`
```

JoanFM: We could, but now changing again in Hub is too much?

hanxiao: Hub follows core.

JoanFM: ah yes makes sense

nan-wang: Why this will lead to many changes in Jina Hub?

JoanFM: no, that is my bad.

The name is very noisy for what I have seen, though

JoanFM:
```    .. confval:: name

        the name of the executor.

        :type: str
        :default: class name plus a random string
```

JoanFM: I think the naming is better with the current changes

nan-wang (approved):
LGTM👍

PR #1628, opened 2021-01-08
fix: rewrite handling of missing keys for indexers

cristianmtr:
This will be required for the CRUD integration tests suite

- handling of missing keys should not `Raise`
- missing keys are filtered out of the request
- we do log a `warning` - up to the user to interpret if that's expected or not

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,7 +133,3 @@ shell/jina-wizard.sh
 # IntelliJ IDEA
 *.iml
 .idea
-
-# tests binaries
-/tests/**/*.bin
-*.bin
--- a/jina/executors/indexers/__init__.py
+++ b/jina/executors/indexers/__init__.py
@@ -162,6 +162,18 @@ class BaseIndexer(BaseExecutor):
         except:
             pass
 
+    def _filter_nonexistent_keys(self, keys: Iterator, existent_keys: Iterator, check_path: str):
+        indices_to_drop = []
+        keys = list(keys)
+        for key_index, key in enumerate(keys):
+            if key not in existent_keys:
+                indices_to_drop.append(key_index)
+        if indices_to_drop:
+            self.logger.warning(
+                f'Key(s) {[keys[i] for i in indices_to_drop]} were not found in {check_path}. Continuing anyway...')
+            keys = [keys[i] for i in range(len(keys)) if i not in indices_to_drop]
+        return keys
+
 
 class BaseVectorIndexer(BaseIndexer):
     """An abstract class for vector indexer. It is equipped with drivers in ``requests.on``
--- a/jina/executors/indexers/cache.py
+++ b/jina/executors/indexers/cache.py
@@ -3,7 +3,6 @@ import tempfile
 from typing import Optional, Iterator
 
 from jina.executors.indexers import BaseKVIndexer
-from jina.helper import check_keys_exist
 
 DATA_FIELD = 'data'
 ID_KEY = 'id'
@@ -99,9 +98,7 @@ class DocIDCache(BaseCache):
         """
         :param keys: list of Document.id
         :param values: list of either `id` or `content_hash` of :class:`Document"""
-        missed = check_keys_exist(keys, self.query_handler.ids)
-        if missed:
-            raise KeyError(f'Keys {missed} were not found in {self.index_abspath}. No operation performed...')
+        keys = self._filter_nonexistent_keys(keys, self.query_handler.ids, self.save_abspath)
 
         for key, cached_field in zip(keys, values):
             key_idx = self.query_handler.ids.index(key)
@@ -113,9 +110,7 @@ class DocIDCache(BaseCache):
         """
         :param keys: list of Document.id
         """
-        missed = check_keys_exist(keys, self.query_handler.ids)
-        if missed:
-            raise KeyError(f'Keys {missed} were not found in {self.index_abspath}. No operation performed...')
+        keys = self._filter_nonexistent_keys(keys, self.query_handler.ids, self.save_abspath)
 
         for key in keys:
             key_idx = self.query_handler.ids.index(key)
--- a/jina/executors/indexers/keyvalue.py
+++ b/jina/executors/indexers/keyvalue.py
@@ -78,22 +78,14 @@ class BinaryPbIndexer(BaseKVIndexer):
                 return m[r:]
 
     def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
-        # check: hard fail (raises) on key not found
-        missed = []
-        for key in keys:
-            if self.query_handler.header.get(key) is None:
-                missed.append(key)
-        if missed:
-            raise KeyError(f'Key(s) {missed} were not found in {self.save_abspath}')
-
-        # hack
-        self.query_handler.close()
-        self.handler_mutex = False
-        self.delete(keys)
+        keys = self._filter_nonexistent_keys(keys, self.query_handler.header.keys(), self.save_abspath)
+        self._delete(keys)
         self.add(keys, values)
         return
 
-    def delete(self, keys: Iterator[int], *args, **kwargs):
+    def _delete(self, keys: Iterator[int]):
+        self.query_handler.close()
+        self.handler_mutex = False
         for key in keys:
             self.write_handler.header.write(
                 np.array(
@@ -102,8 +94,12 @@ class BinaryPbIndexer(BaseKVIndexer):
                 ).tobytes()
             )
             if self.query_handler:
-                self.query_handler.header[key] = None
+                del self.query_handler.header[key]
             self._size -= 1
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        keys = self._filter_nonexistent_keys(keys, self.query_handler.header.keys(), self.save_abspath)
+        self._delete(keys)
 
 
 class DataURIPbIndexer(BinaryPbIndexer):
--- a/jina/executors/indexers/vector.py
+++ b/jina/executors/indexers/vector.py
@@ -126,26 +126,29 @@ class BaseNumpyIndexer(BaseVectorIndexer):
         self.key_dtype = keys.dtype.name
         self._size += keys.shape[0]
 
-    def _check_keys(self, keys: Sequence[int]) -> None:
-        missed = []
-        for key in keys:
-            # if it never existed or if it's been marked as deleted in the current index
-            # using `is False` doesn't work
-            if key not in self.ext2int_id.keys() or self.valid_indices[self.ext2int_id[key]] == False:  # noqa
-                missed.append(key)
-        if missed:
-            raise KeyError(f'Key(s) {missed} were not found in {self.save_abspath}')
-
     def update(self, keys: Sequence[int], values: Sequence[bytes], *args, **kwargs) -> None:
-        self.delete(keys)
-        self.add(np.array(keys), np.array(values))
+        keys = self._filter_nonexistent_keys(keys, self.ext2int_id.keys(), self.save_abspath)
+        # could be empty
+        if keys:
+            # expects np array for computing shapes
+            keys = np.array(list(keys))
+            self._delete(keys)
+            self.add(np.array(keys), np.array(values))
+
+    def _delete(self, keys):
+        # could be empty
+        if keys:
+            # expects np array for computing shapes
+            keys = np.array(list(keys))
+            for key in keys:
+                # mark as `False` in mask
+                self.valid_indices[self.ext2int_id[key]] = False
+                self._size -= 1
 
     def delete(self, keys: Sequence[int], *args, **kwargs) -> None:
-        self._check_keys(keys)
-        for key in keys:
-            # mark as `False` in mask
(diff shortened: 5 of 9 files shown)


jina/executors/indexers/keyvalue.py
```diff
@@ -78,22 +79,17 @@ def query(self, key: int) -> Optional[bytes]:
...
-        self.query_handler.close()
-        self.handler_mutex = False
-        self.delete(keys)
+        keys = check_keys_exist(keys, self.query_handler.header.keys(), self.save_abspath, self.logger)
+        self.delete(keys, **{'keys_precomputed': True})
         self.add(keys, values)
         return
 
     def delete(self, keys: Iterator[int], *args, **kwargs):
+        if kwargs.get('keys_precomputed') is not True:
```
JoanFM, line 88: why do we pass extra arguments? not expected in such key interface I would say

jina/helper.py
```diff
@@ -693,12 +693,16 @@ def run(self):
         return asyncio.run(func(*args, **kwargs))
 
 
-def check_keys_exist(keys_to_check, existing_keys):
-    missed = []
-    for k in keys_to_check:
-        if k not in existing_keys:
-            missed.append(k)
-    return missed
+def check_keys_exist(keys: Iterator, keys_existing: Iterator, save_abspath: str, logger: 'JinaLogger'):
```
JoanFM, line 699: the name save abspath of this argument links it heavily with an executor, I would call it check_path

JoanFM (changes requested):

jina/helper.py
cristianmtr: Done

jina/executors/indexers/keyvalue.py
JoanFM: the interfaces should never be changed, pasa executor parameter if needed but do not change interface or assume things will be passed

cristianmtr: It's a small optimization. 

When we call an `.update`, the keys will be computed for the update anyway. The update will then call the `delete` with only the filtered out keys. 

If we don't do it this way, the `.delete` method will have to recalculate and filter out the missing keys.

I can add a comment

jina/executors/indexers/keyvalue.py
```diff
@@ -78,22 +79,17 @@ def query(self, key: int) -> Optional[bytes]:
...
-            if self.query_handler.header.get(key) is None:
-                missed.append(key)
-        if missed:
-            raise KeyError(f'Key(s) {missed} were not found in {self.save_abspath}')
-
-        # hack
-        self.query_handler.close()
-        self.handler_mutex = False
-        self.delete(keys)
+        keys = check_keys_exist(keys, self.query_handler.header.keys(), self.save_abspath, self.logger)
```
JoanFM, line 82: if check does return something the name seems wrong. Also try to keep variables inmutable as much as possible

jina/executors/indexers/keyvalue.py
cristianmtr: > the interfaces should never be changed, pasa executor parameter if needed but do not change interface or assume things will be passed

I do not EXPECT it to be there. That's why it's an optional `kwargs`. Again, just a small optimization

JoanFM: in which contrxt is this optimixzation and why does it only works sometimes?

cristianmtr:
Not sure why I can't reply to this comment: https://github.com/jina-ai/jina/pull/1628#discussion_r553921838 
but @JoanFM here's the reply

When we call an .update, the keys will be computed for the update anyway. The update will then call the delete with only the filtered out keys.

If we don't do it this way, the .delete method will have to recalculate and filter out the missing keys.

When `.delete` is called on its own (not part of an update), then the optimization will not happen.

jina/executors/indexers/keyvalue.py
cristianmtr: When we call an .update, the keys will be computed for the update anyway. The update will then call the delete with only the filtered out keys.

If we don't do it this way, the .delete method will have to recalculate and filter out the missing keys

jina/executors/indexers/keyvalue.py
cristianmtr: Yes, good point, renamed

jina/helper.py
```diff
@@ -693,12 +696,16 @@ def run(self):
...
+def filter_keys_nonexistent(keys: Iterator, keys_existing: Iterator, check_path: str, logger: 'JinaLogger'):
+    keys = set(keys)
+    missed = set()
+    for key in keys:
+        if key not in keys_existing:
+            missed.add(key)
+    if missed:
+        logger.warning(f'Key(s) {missed} were not found in {check_path}. Continuing anyway...')
+        keys -= missed
+    return keys
```
florian-hoenicke, line 708: `keys` could be a generator instead. Therefore we wouldn't have to load all keys in memory if used somewhere else. In our case we have the keys in memory anyways.

jina/helper.py
```diff
@@ -693,12 +696,16 @@ def run(self):
         return asyncio.run(func(*args, **kwargs))
 
 
-def check_keys_exist(keys_to_check, existing_keys):
-    missed = []
-    for k in keys_to_check:
-        if k not in existing_keys:
-            missed.append(k)
-    return missed
+def filter_keys_nonexistent(keys: Iterator, keys_existing: Iterator, check_path: str, logger: 'JinaLogger'):
```
florian-hoenicke, line 699: `keys_existing` should not be type of Iterator but maybe Set.

jina/helper.py
cristianmtr: Well they are already loaded in memory, for performing the calculation above

jina/helper.py
cristianmtr: It depends on how the indexer stores the keys. I will cast them to `set` to be sure during the calculation

JoanFM: if this is to be used from different points then have a private _delete with an extra argument and have it called from update and delete method with the proper value.

JoanFM: Sorry is not in this discussion but innthe kwargs one

jina/executors/indexers/keyvalue.py
cristianmtr: Response to your suggestion about `_delete`

Like so ? https://github.com/jina-ai/jina/blob/3d8d678a5591861aae1608589f4a08cec7306109/jina/executors/indexers/keyvalue.py#L82

JoanFM: yes more or less

jina/executors/indexers/cache.py
```diff
@@ -99,9 +99,7 @@ def update(self, keys: Iterator['UniqueId'], values: Iterator[any], *args, **kwa
         """
         :param keys: list of Document.id
         :param values: list of either `id` or `content_hash` of :class:`Document"""
-        missed = check_keys_exist(keys, self.query_handler.ids)
-        if missed:
-            raise KeyError(f'Keys {missed} were not found in {self.index_abspath}. No operation performed...')
+        keys = filter_keys_nonexistent(set(keys), set(self.query_handler.ids), self.save_abspath, self.logger)
```
JoanFM, line 102: can we call it `existent_keys`?

jina/executors/indexers/cache.py
```diff
@@ -113,9 +111,7 @@ def delete(self, keys: Iterator['UniqueId'], *args, **kwargs):
         """
         :param keys: list of Document.id
         """
-        missed = check_keys_exist(keys, self.query_handler.ids)
-        if missed:
-            raise KeyError(f'Keys {missed} were not found in {self.index_abspath}. No operation performed...')
+        keys = filter_keys_nonexistent(set(keys), set(self.query_handler.ids), self.save_abspath, self.logger)
```
JoanFM, line 114: same here?

jina/executors/indexers/vector.py
```diff
@@ -137,15 +137,26 @@ def _check_keys(self, keys: Sequence[int]) -> None:
             raise KeyError(f'Key(s) {missed} were not found in {self.save_abspath}')
 
     def update(self, keys: Sequence[int], values: Sequence[bytes], *args, **kwargs) -> None:
-        self.delete(keys)
-        self.add(np.array(keys), np.array(values))
+        keys = filter_keys_nonexistent(set(keys), set(self.ext2int_id.keys()), self.save_abspath, self.logger)
```
JoanFM, line 140: same here?

jina/executors/indexers/vector.py
```diff
@@ -137,15 +137,26 @@ def _check_keys(self, keys: Sequence[int]) -> None:
...
+            self.delete(keys, **{'keys_precomputed': True})
+            self.add(np.array(keys), np.array(values))
 
     def delete(self, keys: Sequence[int], *args, **kwargs) -> None:
-        self._check_keys(keys)
-        for key in keys:
-            # mark as `False` in mask
-            self.valid_indices[self.ext2int_id[key]] = False
-            self._size -= 1
+        if kwargs.get('keys_precomputed') is not True:
```
JoanFM, line 149: the idea would be to have in this class a 

```
def _delete(self, keys, keys_precomputed: bool = False)
```
that is called from different parts

JoanFM (changes requested):

jina/executors/indexers/vector.py
cristianmtr: We don't even need the `precomputed_keys` arg if we opt for a private `_delete`. The public methods would simply handle the keys, since they are always called first. THis is the way I've implemented it now

.gitignore
```diff
@@ -135,5 +135,6 @@ shell/jina-wizard.sh
 .idea
 
 # tests binaries
-/tests/**/*.bin
-*.bin
+**/**/*.bin.ids
```
JoanFM, line 138: this being needed is a smell that we are not properly using `tmpdir` in tests, where does this happen?

cristianmtr: the DocIDCache doesn't store the bin files properly when initiated on its own (not part of a flow)

JoanFM: why not? are u passing `test_metas`?

jina/executors/indexers/vector.py
cristianmtr: Also done for the numpyindexer now

tests/unit/executors/indexers/test_numpyindexer.py
```diff
@@ -271,8 +271,7 @@ def test_numpy_update_delete(compress_level, test_metas):
     with BaseIndexer.load(save_abspath) as indexer:
         assert isinstance(indexer, NumpyIndexer)
         # this will fail cause the key doesn't exist
```
maximilianwerk, line 273: Remove/update comment

jina/helper.py
```diff
@@ -693,12 +697,16 @@ def run(self):
         return asyncio.run(func(*args, **kwargs))
 
 
-def check_keys_exist(keys_to_check, existing_keys):
-    missed = []
-    for k in keys_to_check:
-        if k not in existing_keys:
-            missed.append(k)
-    return missed
+def filter_nonexistent_keys(keys: Iterator, existent_keys: Iterator, check_path: str, logger: 'JinaLogger'):
```
maximilianwerk, line 700: I would prefer having this method as a part of a base class of all the using classes. Might be `BaseIndexer`? Having the logger as an argument is a smell in my opinion.

maximilianwerk (commented):
Just organizational/cleanup comments.

.gitignore
cristianmtr: Hm I don't really remember. I will have a look at it in the upcoming PR where I'm adding the drivers for delete and update to DocCache

tests/unit/executors/indexers/test_numpyindexer.py
cristianmtr: Yep, done

jina/helper.py
JoanFM: I agree

cristianmtr: Ah good point. Done

.gitignore
cristianmtr: Nvm, found the problem and fixed it

jina/executors/indexers/__init__.py
```diff
@@ -162,6 +162,18 @@ def flush(self):
         except:
             pass
 
+    def filter_nonexistent_keys(self, keys: Iterator, existent_keys: Iterator, check_path: str):
```
JoanFM, line 165: can u make it "private"?

jina/helper.py
```diff
@@ -21,6 +21,9 @@
 
 import numpy as np
 
+if False:
```
JoanFM, line 24: why is this needed'

JoanFM (changes requested):

jina/helper.py
cristianmtr: Ah it was from when the `filter_keys` was here. Removing the import left a `pass` behind it seems

jina/executors/indexers/__init__.py
cristianmtr: done

JoanFM (approved):

PR #1633, opened 2021-01-08
refactor: parallelize cicd

hanxiao:


.github/workflows/cd.yml
```diff
@@ -125,7 +125,19 @@ jobs:
...
+        # But what if someone create a new test folder but forget to update
+        # here, isn't this newly added folder missed?
+        # yes, it will be missed. However, the solution is quite simple:
+        # e.g.
+        # 1. write a small script to read this YAML and validate the "test-path" below, if inconsistent, then exit(1)
+        # 2. add step1 script as a job before "core-test" job.
+        #
+        # I will leave this task to whoever see this message and want to learn Github Action & DevOps.
+        # welcome to submit your PR here: https://github.com/jina-ai/jina/pull/
+        test-path: [unit/*.py, integration/*.py, integration/crud,integration/encode_driver_batching,integration/eval_flow,integration/evaluation,integration/gateway_non_blocking,integration/high_order_matches,integration/hub_usage,integration/incremental_indexing,integration/issues,integration/level_depth,integration/logging,integration/memmap,integration/mime,integration/multimodal,integration/ref_indexer,unit/clients,unit/docker,unit/drivers,unit/executors,unit/flow,unit/logging,unit/mwu-encoder,unit/peapods,unit/proto,unit/types,unit/yaml]
```
JoanFM, line 140: this is very very hard to mantain, the benefit wont compensate

JoanFM (changes requested):
I would not go in this direction

.github/workflows/cd.yml
hanxiao: why? isn't this just a script thing? i can hack it in an hour

JoanFM: because anytime a new folder is added we need to remember to add this, and ci cd is something people are not so used to touch

hanxiao: the value of `strategy.matrix` can be set dynamically, so this should solve the problem

hanxiao: it's doc-ed here https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#fromjson

JoanFM: yes but being doced does not  guarantee at all rhat it will be done. People do not check docs when doing a PR.

JoanFM: and it will be even harder as the team grows

hanxiao: i mean: since it is dynamic, I will write a code to autofill in this `matrix`, the team won't have to know about how it works.

hanxiao: i will add autofill in this PR, not leaving it to others.

PR #1639, opened 2021-01-10
refactor(cli): move show-exc-info as PPR common args

hanxiao:


hanxiao:
this PR skips some Hub IO related test due to the bad gateway problem of the hub. #1641 , please revert the skip after #1641 is fixed

PR #1642, opened 2021-01-10
fix(daemon): fix some helper text for redocs

hanxiao:


daemon/api/endpoints/flow.py
```diff
@@ -208,6 +205,7 @@ async def _ping(
     _, args, _ = ArgNamespace.get_parsed_args(kwargs, set_client_cli_parser())
     client = Client(args)
     try:
+        # TODO: this introduces side-effect, need to be refactored. (2020.01.10)
```
deepankarm, line 208: `Client` & `WebsocketClient` need to be refactored to work as a context manager. This interface can tell the daemon client if the gateway server is alive.

PR #1657, opened 2021-01-12
fix: sharding parallel one

florian-hoenicke:
The following case leads to `pea_id == -1`:
```
 with Flow().add(
      uses=yaml_file,
      shards=1
      separated_workspace=True,
  ) as f:
       f.index(input_fn=docs)
```

```
#yaml_file
!BinaryPbIndexer
with:
  index_filename: doc.gzip
metas:
  name: docIndexer
  workspace: $JINA_TOPK_DIR

requests:
  on:
    IndexRequest:
      - !KVIndexDriver
        with:
          executor: docIndexer
          traversal_paths: ['r']
```

### DIFF
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -105,7 +105,7 @@ class BasePod(ExitStack):
             peas_args['head'] = _copy_to_head_args(args, args.polling.is_push)
             peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args['head'], peas_args['tail'])
-        elif getattr(args, 'uses_before', None) or getattr(args, 'uses_after', None):
+        else:
             args.scheduling = SchedulerType.ROUND_ROBIN
             if getattr(args, 'uses_before', None):
                 self.is_head_router = True
@@ -114,10 +114,6 @@ class BasePod(ExitStack):
                 self.is_tail_router = True
                 peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args.get('head', None), peas_args.get('tail', None))
-        else:
-            self.is_head_router = False
-            self.is_tail_router = False
-            peas_args['peas'] = [args]
 
         # note that peas_args['peas'][0] exist either way and carries the original property
         return peas_args
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -105,7 +105,7 @@ class BasePod(ExitStack):
             peas_args['head'] = _copy_to_head_args(args, args.polling.is_push)
             peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args['head'], peas_args['tail'])
-        else:
+        elif getattr(args, 'uses_before', None) or getattr(args, 'uses_after', None):
             args.scheduling = SchedulerType.ROUND_ROBIN
             if getattr(args, 'uses_before', None):
                 self.is_head_router = True
@@ -114,6 +114,10 @@ class BasePod(ExitStack):
                 self.is_tail_router = True
                 peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args.get('head', None), peas_args.get('tail', None))
+        else:
+            self.is_head_router = False
+            self.is_tail_router = False
+            peas_args['peas'] = [args]
 
         # note that peas_args['peas'][0] exist either way and carries the original property
         return peas_args
--- a/jina/peapods/pods/helper.py
+++ b/jina/peapods/pods/helper.py
@@ -4,13 +4,13 @@ from typing import List
 
 from ... import __default_host__
 from ...enums import SchedulerType, SocketType, PeaRoleType
-from ...helper import random_port, get_random_identity, get_public_ip, get_internal_ip, is_valid_local_config_source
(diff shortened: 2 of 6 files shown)


jina/peapods/pods/helper.py
```diff
@@ -4,13 +4,13 @@
...
 from ...enums import SchedulerType, SocketType, PeaRoleType
-from ...helper import random_port, get_random_identity, get_public_ip, get_internal_ip, is_valid_local_config_source
+from ...helper import random_port, get_random_identity, get_public_ip, get_internal_ip
 
 
 def _set_peas_args(args: Namespace, head_args: Namespace = None, tail_args: Namespace = None) -> List[Namespace]:
     result = []
 
-    for idx in range(args.parallel):
+    for idx in range(getattr(args, 'parallel', 1)):
```
JoanFM, line 13: this is a smell, parallel should be well defined here

JoanFM (changes requested):

JoanFM:
I think this can be expected, pea_id is only used to define separate workspaces, and I think in executors the pea_id is checked against -1

jina/peapods/pods/__init__.py
```diff
@@ -105,7 +105,7 @@ def _parse_args(self, args: Namespace) -> Dict[str, Optional[Union[List[Namespac
             peas_args['head'] = _copy_to_head_args(args, args.polling.is_push)
             peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args['head'], peas_args['tail'])
-        elif getattr(args, 'uses_before', None) or getattr(args, 'uses_after', None):
+        else:
```
JoanFM, line 108: I think you should be able to safely do :+1: 
```python
elif args.uses_before or args.uses_after
```

jina/peapods/pods/__init__.py
```diff
@@ -105,7 +105,7 @@ def _parse_args(self, args: Namespace) -> Dict[str, Optional[Union[List[Namespac
...
             args.scheduling = SchedulerType.ROUND_ROBIN
             if getattr(args, 'uses_before', None):
```
JoanFM, line 110: maybe here u can also assume the attribute is there since last refactoring on parsing

jina/peapods/pods/__init__.py
```diff
@@ -114,10 +114,6 @@ def _parse_args(self, args: Namespace) -> Dict[str, Optional[Union[List[Namespac
                 self.is_tail_router = True
                 peas_args['tail'] = _copy_to_tail_args(args)
             peas_args['peas'] = _set_peas_args(args, peas_args.get('head', None), peas_args.get('tail', None))
-        else:
-            self.is_head_router = False
```
JoanFM, line 118: I think this logic needs to remain there, are you sure it needs removal?

JoanFM:
@florian-hoenicke it would be good to describe what is the expected value of `pea_id` in this case? 

would it be 1 or 0?

JoanFM:
Does this `pea_id` value has any further impact?

florian-hoenicke:
> Does this `pea_id` value has any further impact?

Yes, the `pea_id` is used to generate the path the index is stored at. If it is -1, then the path becomes `.../docIndexer--1/doc.gzip`.

If we have 2 shards, the indexes are stored at `.../docIndexer-1` and `.../docIndexer-2`.
In case of 1 shard it is messed up. It stores at `.../docIndexer--1` and searches at `.../docIndexer-0`.
From my understanding, it should store and search at .../docIndexer-1` instead

JoanFM:
> > Does this `pea_id` value has any further impact?
> 
> Yes, the `pea_id` is used to generate the path the index is stored at. If it is -1, then the path becomes `.../docIndexer--1/doc.gzip`.
> 
> If we have 2 shards, the indexes are stored at `.../docIndexer-1` and `.../docIndexer-2`.
> In case of 1 shard it is messed up. It stores at `.../docIndexer--1` and searches at `.../docIndexer-0`.
> From my understanding, it should store and search at .../docIndexer-1` instead

I think the correct way is to only consider the `pea_id` to create the file only if `separated_workspace` is `True`

JoanFM (commented):
add unittest with to check `current_workspace` does not include any reference to `pea_id`?

jina/peapods/pods/helper.py
florian-hoenicke: thanks. I solved it in a different way now

jina/peapods/pods/__init__.py
florian-hoenicke: thanks, I'm not touching this code anymore

jina/peapods/pods/__init__.py
florian-hoenicke: thanks. It was just a draft. I solved it in a different way now

jina/peapods/pods/__init__.py
florian-hoenicke: thanks, I don't think it is used, but now I keep it there because it is out of scope for this pr

florian-hoenicke:
> add unittest with to check `current_workspace` does not include any reference to `pea_id`?

I don't update the pea_id anymore, but change the way the workspace path is created. The test validates that the index file  is created in the right folder:
```
    with BaseExecutor.load((random_workspace / 'doc_idx.bin')) as doc_indexer:
        assert isinstance(doc_indexer, BinaryPbIndexer)
        assert doc_indexer.size == num_uniq_docs
```
There exists another test validating that the workspace is set up correctly in case we have more than one shard.`tests/unit/test_index.py:test_index`

JoanFM (approved):

PR #1658, opened 2021-01-12
refactor: shadowing built ins

florian-hoenicke:


JoanFM (approved):

theUnkownName (dismissed):

jina/docker/helper.py
```diff
@@ -63,5 +63,5 @@ def sleep(self, seconds=5):
         self.logger.debug(f'sleeping for {seconds} seconds')
         time.sleep(seconds)
 
-    def __exit__(self, type, value, traceback):
```
hanxiao, line 66: this is inherited from the exit stack builtin, i don't think we need to change that

hanxiao: by inherited, i mean if you type `__exit__` in Pycharm, it auto-complete with this signature

hanxiao (changes requested):
keep `__exit__` signature as `object.__exit__`, others look good

jina/docker/helper.py
hanxiao: okay my bad, somehow this one is not correctly auto-filled by Pycharm. The correct signature should be 
`object.__exit__(self, exc_type, exc_value, traceback)`

let's change to that
https://docs.python.org/3.7/reference/datamodel.html#object.__exit__

florian-hoenicke: thanks, I changed it

theUnkownName (approved):

hanxiao (approved):

PR #1660, opened 2021-01-12
refactor: unused definitions

florian-hoenicke:


tests/unit/test_echostream.py
```diff
@@ -20,19 +20,8 @@ def test_simple_zmqlet():
...
-        '--port-out', '12346',
-        '--socket-in', 'PULL_BIND',
-        '--socket-out', 'PUSH_BIND',
-        '--uses', '_logforward',
-        '--timeout-ctrl', '-1'
-    ])
-
     logger = logging.getLogger('zmq-test')
-    with BasePea(args2) as z1, Zmqlet(args, logger) as z:
+    with Zmqlet(args, logger) as z:
```
JoanFM, line 24: I think this may be on putpose to test ifnthey can work in parallel

JoanFM (changes requested):

tests/unit/test_echostream.py
florian-hoenicke: okay, makes sense. I put it back and removed just the z1 parameter since it is not used any further

JoanFM (approved):

PR #1666, opened 2021-01-12
Custom long ids

maximilianwerk:
This is the branch from PR #1554 where I merged current jina master into the branch. This is the starting point for unit/integration test fixing and should finally add the wanted feature.

@janandreschweiger I did it in a way, that your commits are still preserved and you get the commits correctly attributed :) I'll close your PR and will take it from here. Again, thanks a lot for your efforts. This is much appreciated.


tests/unit/test_workspace.py
```diff
@@ -50,29 +51,46 @@ def test_compound_indexer(tmpdir, pea_id):
...
-            assert not a.is_updated
-            a[0].add([j, j * 2, j * 3], [bytes(j), bytes(j * 2), bytes(j * 3)])
-            a[0].add([j, j * 2, j * 3], [bytes(j), bytes(j * 2), bytes(j * 3)])
-            assert a[0].is_updated
-            assert a.is_updated
-            assert not a[1].is_updated
-            a[1].add(np.array([j * 2, j * 2 + 1]), all_vecs[(j * 2, j * 2 + 1), :])
-            assert a[1].is_updated
-            a.save()
+        with BaseExecutor.load_config(executor_yml, separated_workspace=True, pea_id=j) as indexer:
```
maximilianwerk, line 72: @bwanglzu @JoanFM How do you like to put the yml string directly into the test. Advantages: much better readable without file jumping and you can adapt the workspace without first writing to env variable and then read again. This is only a good approach for yaml with <20 lines, but here I really prefer it much.

hanxiao: make sense to me

JoanFM: is this even possible to have a basic `yml` and pass extra parameters? even without the metas?

JoanFM: For me having this is kind of perfect, I kind of got used to have `yml` files independently but it is okey for me to have it embedded in the code, but when it comes to `flow` yamls with several `pods` I do believe is better to have it as independent file

florian-hoenicke: I like this a lot since it is much more explicit what is gone tested.
The only down site I see is that we don't have syntax highlighting and auto format anymore.
But since the files are very small, this won't hurt to0 much.

maximilianwerk:
I looked further into longer `ids` and for the `BinaryPbIndexer` this results in the following problem: We are currently using `ids` of type `np.int64` which is exactly 8 bytes. When we move to `np.object` we cannot load the data from disk as before, since numpy does not support `np.frombuffer` for `np.object` fields. I see three possible solutions:

1. Use a `int2ext_id` and `ext2int_id` construct as we use in the `BaseNumpyIndexer` which is a lookup table that maps any external id to an internal one, which can be of type `np.int64` again.
2. Change the way we do serialization/deserialization in the `BinaryPbIndexer` for the `ReadHandler._header` object. We could do it as done [here](https://stackoverflow.com/a/30699208), but this would mean, we have to write the whole file either on each flush or on the close.
3. Don't allow arbitrary key length for the `BinaryPbIndexer` and if you want to use keys with that are longer than `int64`, you have to use one of the production-ready indexers from the Hub. However, that would destroy the drop-in-replacement style in Jina and I would discourage this approach.

@JoanFM @florian-hoenicke @cristianmtr please comment on this, since you have worked more with the Indexers than I did.

JoanFM:
> I looked further into longer `ids` and for the `BinaryPbIndexer` this results in the following problem: We are currently using `ids` of type `np.int64` which is exactly 8 bytes. When we move to `np.object` we cannot load the data from disk as before, since numpy does not support `np.frombuffer` for `np.object` fields. I see three possible solutions:
> 
> 1. Use a `int2ext_id` and `ext2int_id` construct as we use in the `BaseNumpyIndexer` which is a lookup table that maps any external id to an internal one, which can be of type `np.int64` again.
> 2. Change the way we do serialization/deserialization in the `BinaryPbIndexer` for the `ReadHandler._header` object. We could do it as done [here](https://stackoverflow.com/a/30699208), but this would mean, we have to write the whole file either on each flush or on the close.
> 3. Don't allow arbitrary key length for the `BinaryPbIndexer` and if you want to use keys with that are longer than `int64`, you have to use one of the production-ready indexers from the Hub. However, that would destroy the drop-in-replacement style in Jina and I would discourage this approach.
> 
> @JoanFM @florian-hoenicke @cristianmtr please comment on this, since you have worked more with the Indexers than I did.

I think the best approach for now would be the 1st one. The benefit of dealing in the same way with Vector and KVIndexers is good since the learnings and optimizations taken from one, cane be used for the other

florian-hoenicke:
Maybe it is out of scope for this ticket, but should be done from my perspective:
There are a lot of `d.id = f'{idx:0>16}'` in the tests of the core. Since we don't have to use 16 chars, all the tests can be refactored.

maximilianwerk:
> 
> I think the best approach for now would be the 1st one. The benefit of dealing in the same way with Vector and KVIndexers is good since the learnings and optimizations taken from one, cane be used for the other

Beware, that only the `BinaryPbIndexer` would do that. `RedisIndexer` and other advanced `KVIndexer` don't need this approach, since they anyhow accept longer ids. As a consequence we should need to split out the `int2ext` and `ext2int` code into another class which we mix into `BaseNumpyIndexer` and `BinaryPbIndexer`. Sounds good for me, but is a bit more effort than expected.

JoanFM:
> > I think the best approach for now would be the 1st one. The benefit of dealing in the same way with Vector and KVIndexers is good since the learnings and optimizations taken from one, cane be used for the other
> 
> Beware, that only the `BinaryPbIndexer` would do that. `RedisIndexer` and other advanced `KVIndexer` don't need this approach, since they anyhow accept longer ids. As a consequence we should need to split out the `int2ext` and `ext2int` code into another class which we mix into `BaseNumpyIndexer` and `BinaryPbIndexer`. Sounds good for me, but is a bit more effort than expected.

Okey, but is the right way to do this, and having mixins is a solution I like

maximilianwerk:
@janandreschweiger This refactoring of the `BinaryPbIndexer` will take a bit longer and more involved than expected. Anyhow, we decided it will be a feature of Jina 1.0. So we will work on it in the next week and I'd ask you to stay tuned a bit longer :)

maximilianwerk:
This will not be fixed now. For a more detailed information see #1553

Good2G (dismissed):

maximilianwerk (commented):
I believe this refactoring now makes the code clearer. Not only are long custom `ids` possible, but also the interface between index driver and executor got much better.

I would like to get some input about the idea to move the numpy conversion to the executor in order to hide the `key_length` from the driver.

hanxiao (changes requested):
the `tests/integration/evaluation/` keeps failing, not due to flaky though. must be some side-effect

Issue #1672, opened 2021-01-12
_check_on_gpu method of BaseExecutor does not properly check for GPU.


tadejsv:
Here's the method in question:

https://github.com/jina-ai/jina/blob/d40aee59615b34e39a8dbad538f50343d2bed46e/jina/executors/__init__.py#L129-L137

This method checks for `nvcc`, which is the CUDA compiler. However, in a lot of cases, only prebuilt CUDA binaries are installed, without the compiler. In fact, this is the case for the recommended way of installing pytorch. Say you have this image

``` dockerfile
FROM continuumio/miniconda3 AS conda

RUN conda install pytorch cudatoolkit=11.0 -c pytorch

FROM nvidia/cuda:11.0-base-ubuntu20.04

# Prepare shell and file system
ENV LANG=C.UTF-8 LC_ALL=C.UTF-8 SHELL=/bin/bash
ENV PATH /opt/conda/bin:$PATH
SHELL ["/bin/bash", "-c"]

# Copy over conda and bashrc, install environment
COPY --from=conda /opt/ /opt/
RUN ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
    echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
    echo "conda activate base" >> ~/.bashrc
```

Here, executing `nvcc --version` will return nothing. However, this image has a perfectly working CUDA installation, as you can see
``` python
>>> import torch
>>> torch.cuda.is_available()
True
```

## Solution

I propose one of two things:
- Replace `['nvcc', '--version']` with `['nvidia-smi']`. This can cause "false positives" - i.e. user uses base CUDA image and forgets to install CUDA. But that's his mistake, and he will spot it easily when torch/tf complains no CUDA is available ;) Better than the current situation when the user is prevented from using a working CUDA installation
- Remove this check entirely. More or less the same reason as above - I mean if user wants to use a GPU it's his responsibility to ensure things work, and error from DL packages will make it clear that CUDA is missing

JoanFM:
> Here's the method in question:
> 
> https://github.com/jina-ai/jina/blob/d40aee59615b34e39a8dbad538f50343d2bed46e/jina/executors/__init__.py#L129-L137
> 
[...]

PR #1674, opened 2021-01-12
refactor: remove GPU check in Executor

tadejsv:
As discussed in #1672 , currently `_check_on_gpu` method will disable GPU even in some cases where there is a perfectly fine CUDA installation on the machine.

It is better to leave it up to the users to ensure that CUDA/GPU is working - the errors from various deep learning frameworks should quickly tell them if something is missing anyways.


JoanFM (commented):
For back compatibility, have u checked if there is any executor in the hub calling this method? 

If they are, I would just have it pass, remove them from those hub executors and later remove it totally from here

tadejsv:
> For back compatibility, have u checked if there is any executor in the hub calling this method?
> 
> If they are, I would just have it pass, remove them from those hub executors and later remove it totally from here

Ok, I went over jina and jina hub codebase, looking for all instances of "gpu" (case insensitive). Nothing calls this method directly, and no hub pod is testing on GPU (they all either not set `on_gpu`, or set it to true if `JINA_TEST_GPU` is in environment, which it never is).

I have not found any other potential use of this method. But you can double check to be sure. I am not sure why the CI is failing

nan-wang (approved):
LGTM👍

PR #1679, opened 2021-01-13
ci: tag docker image with github run id

JoanFM:


florian-hoenicke (approved):
makes sense. With this we handle build options differently.

hanxiao (changes requested):
1. we dont have `${{ matrix.build_option }}`, it is an empty value. This PR takes no effect.
2. use {{env.GITHUB_RUN_ID}}, Unique ID, Safe String, no escaping required
```
ci.yaml: jina:test-pip${{env.GITHUB_RUN_ID}}
python: uses=f'docker://jina:test-pip{os.environ[GITHUB_RUN_ID]}'
```
ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables

hanxiao (approved):

PR #1681, opened 2021-01-13
feat: draft hub integration test

bwanglzu:
**So the ticket can be handled as follows:**

- Attach default_drivers class method to all BaseExecutor classes. (maybe rename this to test_drivers)

**Then the problem comes to:**

- When I build a hub image, and I use --test-uses, I receive a docker image file with the hubpod.

**The challenge is:**

- Find out which Executor has to be tested.

- Once I know which Executor has to be tested:

- Loop for all the test_drivers and build the config.yml (or any other interface):

- This new config.yml pass to uses_internal argument in testLevel.BUILD_FLOW

- Once we have the Flow properly built knowing that the Pod we test is equipped with the Executor-Driver pair that we want to test.

- Then we query the Flow with a suitable query_generator and just have a small check that the Flow run successfully.

**How to find a suitable query_generator?**

- Every driver or executor or some labeling may have a kind of input_generator to run against.

- For instance: NumericEncoders will receive documents with Blob. TextEncoders will receive documents with text … Some logic like this must be implemented


bwanglzu:
Had a discussion with @JoanFM , we'll find other better solutions for this issue.

PR #1682, opened 2021-01-13
fix: conditional args

florian-hoenicke:
In case we have a separated workspace and set shards to 1, search does not work.
We get rid of this case by forcing to have `separated_workspace=False` if `parallel==1`

### DIFF
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -227,6 +227,8 @@ class BasePod(ExitStack):
                 args.runtime_cls = 'RESTRuntime'
             else:
                 args.runtime_cls = 'GRPCRuntime'
+        if 'parallel' in args and args.parallel == 1:
+            args.separated_workspace = False
 
     def connect_to_tail_of(self, pod: 'BasePod'):
         """Eliminate the head node by connecting prev_args node directly to peas """


JoanFM (commented):
I think this should be completed with a robust chanhe also on executor/indexer size. I remember the logic there was robust to this condition.

jina/peapods/pods/__init__.py
```diff
@@ -227,6 +227,8 @@ def _set_conditional_args(args):
                 args.runtime_cls = 'RESTRuntime'
             else:
                 args.runtime_cls = 'GRPCRuntime'
+        if 'parallel' in args and args.parallel == 1:
```
JoanFM, line 230: parallel is now guaranteed to be in args

JoanFM (changes requested):

jina/peapods/pods/__init__.py
JoanFM: and it would be good to add a warning

florian-hoenicke: > parallel is now guaranteed to be in args
not in case the gateway is added

JoanFM: ok, is a smell but good point

florian-hoenicke: > and it would be good to add a warning

Thanks, good idea!
However, this is actually expected behaviour:
The documentation for `--separated-workspace` in `mixin_zed_runtime_parser` says: `the data and config files are separated for each pea in this pod, only effective when BasePod\'s parallel > 1`.
In addition there is no logging used in `jina/peapods/pods/__init__.py` for all similar cases.

JoanFM (approved):

PR #1685, opened 2021-01-14
fix: kv default traversal

cristianmtr:


JoanFM (commented):
@maximilianwerk I think this is the value we should put as default, do you agree?

maximilianwerk:
> @maximilianwerk I think this is the value we should put as default, do you agree?

Yes.

maximilianwerk (dismissed):
LGTM

cristianmtr:
Some test was failing

florian-hoenicke (dismissed):
Nice feature. For most of the real world cases it will make the yaml configuration smaller.
The test fails because there is a `traversal_paths: ['r']` missing on the `SearchRequest` in `tests/integration/crud/yaml/index_kv.yml`.
The test `tests/integration/crud/test_crud.pytest_delete_kv:test_crud.py` fails because it is just using the `kv indexer` and tries to get matches.

cristianmtr:
> Nice feature. For most of the real world cases it will make the yaml configuration smaller.
> The test fails because there is a `traversal_paths: ['r']` missing on the `SearchRequest` in `tests/integration/crud/yaml/index_kv.yml`.
> The test `tests/integration/crud/test_crud.pytest_delete_kv:test_crud.py` fails because it is just using the `kv indexer` and tries to get matches.

Strange, I cannot reproduce the second one. It passes for me. I fixed the first one and started a new run of the CI. Let's see...

cristianmtr:
The test `tests/integration/test_helloworld.py::test_helloworld_py` keeps failing in CI, but cannot reproduce locally

tests/unit/drivers/test_kv_search_driver.py
```diff
@@ -42,10 +42,11 @@ def __init__(self, *args, **kwargs):
             self.db[int(doc.id)] = doc.SerializeToString()
 
 
-
 class SimpleKVSearchDriver(KVSearchDriver):
 
     def __init__(self, *args, **kwargs):
+        if kwargs.get('traversal_paths') is None:
```
JoanFM, line 48: add `traversal_paths` to set of arguments

JoanFM (changes requested):

tests/unit/drivers/test_kv_search_driver.py
```diff
@@ -42,11 +42,10 @@ def __init__(self, *args, **kwargs):
             self.db[int(doc.id)] = doc.SerializeToString()
 
 
-
 class SimpleKVSearchDriver(KVSearchDriver):
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, traversal_paths=['r'], *args, **kwargs):
```
cristianmtr, line 47: @JoanFM like so?

JoanFM: yes

JoanFM (dismissed):

theUnkownName (approved):

tests/integration/test_helloworld.py
```diff
@@ -91,7 +91,7 @@ def test_helloworld_flow(tmpdir):
 
     # run it!
     with f:
-        f.index(_input_ndarray(targets['index']['data']), batch_size=args.index_batch_size)
+        f.index(_input_ndarray(targets['index']['data']), batch_size=args.index_request_size)
```
JoanFM, line 94: can u call it `request_size=`?

JoanFM (changes requested):

tests/integration/test_helloworld.py
cristianmtr: done

JoanFM (approved):

PR #1686, opened 2021-01-14
feat: merge matches topk

florian-hoenicke:


jina/resources/executors._merge_matches_topk.yml
```diff
@@ -0,0 +1,17 @@
+!BaseExecutor
+with: {}
+metas:
+  name: merge_matches
```
cristianmtr, line 4: DIfferent name, right?

florian-hoenicke: true thanks

jina/resources/executors._merge_matches_topk.yml
```diff
@@ -0,0 +1,17 @@
...
+requests:
+  on:
+    [SearchRequest, TrainRequest, IndexRequest, DeleteRequest, UpdateRequest]:
+      - !ReduceAllDriver
+        with:
+          traversal_paths: ['m']
+      - !SliceQL
+        with:
+          start: 0
+          end: $JINA_TOPK
```
JoanFM, line 14: this has to be set arbitrarily and overriden by QueryLangDriver, but setting an env variable is bad since we do not know it will exist. you can write 10

JoanFM (changes requested):

jina/resources/executors._merge_matches_topk.yml
hanxiao: the new syntax for env var substituion  is https://docs.jina.ai/chapters/yaml/substitute.html#confval-${{ENV.var}}

JoanFM: but we should not consider this is there right?

JoanFM (changes requested):
It could be good to add this in one test? maybe take an integration test and change the name and see it works. Plus remove the env var

cristianmtr (approved):

JoanFM (approved):

Issue #1671, opened 2021-01-12
Enable using GPU for docker containers


tadejsv:
I would like to be able to start GPU-enabled docker containers with Jina.

The following arguments need to be passed to Docker Python SDK run command for this to happen:
- `runtime`: for GPU to work, users should set this to `"nvidia"`
- `environment`: If working on a multi-GPU machine, users need to pass env variables like `CUDA_VISIBLE_DEVICES` to select which GPU the container should use
- `ipc_mode` or `shm_size`: This one is not necessary, but would be nice to have. This increases the amount of memory (RAM) available to the container - for larger models it might be needed.

Adding this would require changes in `/jina/peapods/runtimes/container/__init__.py`. If you wish, I can prepare a PR for this myself

JoanFM:
Hey @tadej-redstone ,

This is definitely an interesting point to look at. We will take a look at it, but maybe we need a way to pass arbitraty extra parameters or extra "kwargs" to the docker sdk to be able to respond to custom needs.

JoanFM:
> Hey @tadej-redstone ,
> 
> This is definitely an interesting point to look at. We will take a look at it, but maybe we need a way to pass arbitraty extra parameters or extra "kwargs" to the docker sdk to be able to respond to custom needs.

This would also involve changes in the parser. You are of course more than welcome to propose a PR for this solution

tadejsv:
> > Hey @tadej-redstone ,
> > This is definitely an interesting point to look at. We will take a look at it, but maybe we need a way to pass arbitraty extra parameters or extra "kwargs" to the docker sdk to be able to respond to custom needs.
> 
> This would also involve changes in the parser. You are of course more than welcome to propose a PR for this issue

I agree, probably being able to pass custom kwargs is the best option. If this is not very hard to do, I might make a shot at a PR for it

JoanFM:
> > > Hey @tadej-redstone ,
> > > This is definitely an interesting point to look at. We will take a look at it, but maybe we need a way to pass arbitraty extra parameters or extra "kwargs" to the docker sdk to be able to respond to custom needs.
> > 
> > 
> > This would also involve changes in the parser. You are of course more than welcome to propose a PR for this issue
> 
> I agree, probably being able to pass custom kwargs is the best option. If this is not very hard to do, I might make a shot at a PR for it

I guess we will need to check what the docker sdk allows to.

tadejsv:
[...]

PR #1690, opened 2021-01-14
feat: pass arbitrary kwargs to docker sdk

JoanFM:
**Changes introduced**
As requested in #1671 , adding arbitrary parameters can be useful to get custom behavior or options for the docker images supporting Peas living in Docker images

### DIFF
--- a/daemon/models/custom.py
+++ b/daemon/models/custom.py
@@ -55,13 +55,13 @@ def get_pydantic_fields(config: Union[dict, argparse.ArgumentParser]):
                                   description=arg['help'])
             all_options[arg_key] = (arg_type, current_field)
 
-    # Issue: For all args that have a default value which comes from a function call (get_random_identity() or random_port()),
-    # 1st pydantic model invocation sets these default values, means build_pydantic_model(...) sets the args, not SinglePodModel()
-    # In case of multiple Pods, port conflict happens because of same port set as default in both.
-    # TODO(Deepankar): Add support for `default_factory` for default args that are functions
+    # Issue: For all args that have a default value which comes from a function call (get_random_identity() or
+    # random_port()), 1st pydantic model invocation sets these default values, means build_pydantic_model(...) sets
+    # the args, not SinglePodModel() In case of multiple Pods, port conflict happens because of same port set as
+    # default in both. TODO(Deepankar): Add support for `default_factory` for default args that are functions
     if isinstance(config, argparse.ArgumentParser):
         # Ignoring first 3 as they're generic args
-        from jina.parsers.helper import KVAppendAction
+        from jina.parsers.helper import KVAppendAction, DockerKwargsAppendAction
         for arg in config._actions[3:]:
             arg_key = arg.dest
             arg_type = arg.type
@@ -72,7 +72,7 @@ def get_pydantic_fields(config: Union[dict, argparse.ArgumentParser]):
             if hasattr(arg_type, '__self__'):
                 arg_type = type(arg.default) if arg.default else int
             arg_type = str if isinstance(arg_type, argparse.FileType) else arg_type
-            arg_type = dict if type(arg) == KVAppendAction else arg_type
+            arg_type = dict if isinstance(arg, (KVAppendAction, DockerKwargsAppendAction)) else arg_type
 
             current_field = Field(default=arg.default,
                                   example=arg.default,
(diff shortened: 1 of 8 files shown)


tadejsv:
Oh it was this easy 😂

JoanFM:
> Oh it was this easy

Have to test still, so not sure it was this easy, just a draft

tadejsv:
Would you be able to add a test for using this in a yaml config as well?

JoanFM:
> Would you be able to add a test for using this in a yaml config as well?

I am thinking it won't look very nice, I am trying to see how can I pass direct dictionaries to argparser

JoanFM:
> > > Would you be able to add a test for using this in a yaml config as well?
> > 
> > 
> > I am thinking it won't look very nice, I am trying to see how can I pass direct dictionaries to argparser
> 
> I also came across this problem when thinking how to implement this yesterday. I think the way you did it is probably the best way currently - argparser is unfortunately very limited - you could use something like `eval`, but it's use is discouraged.
> 
> A more fundamental solution would require moving away from Argparser, but that would take a long time

I found a quite good solution with a helper we are already using.

To use it in the yaml files, the syntax would be similar to the usage of `env` in `Flow.yml` as it can be seen in:

`tests/integration/optimizers/flow.yml`

tadejsv:
> > > > Would you be able to add a test for using this in a yaml config as well?
> > > 
> > > 
> > > I am thinking it won't look very nice, I am trying to see how can I pass direct dictionaries to argparser
> > 
> > 
> > I also came across this problem when thinking how to implement this yesterday. I think the way you did it is probably the best way currently - argparser is unfortunately very limited - you could use something like `eval`, but it's use is discouraged.
> > A more fundamental solution would require moving away from Argparser, but that would take a long time
> 
> I found a quite good solution with a helper we are already using.
> 
> To use it in the yaml files, the syntax would be similar to the usage of `env` in `Flow.yml` as it can be seen in:
> 
> `tests/integration/optimizers/flow.yml`

Pretty cool :) I guess for now it is not possible to pass list/dict arguments, right? I mean the following
``` python
docker_kwargs = {'environment': ['VAR1=BAR', 'VAR2=FOO']}

**Comment by JoanFM** [comment]
> > > > > Would you be able to add a test for using this in a yaml config as well?
> > > > 
> > > > 
> > > > I am thinking it won't look very nice, I am trying to see how can I pass direct dictionaries to argparser
> > > 
> > > 
> > > I also came across this problem when thinking how to implement this yesterday. I think the way you did it is probably the best way currently - argparser is unfortunately very limited - you could use something like `eval`, but it's use is discouraged.
> > > A more fundamental solution would require moving away from Argparser, but that would take a long time
> > 
> > 
> > I found a quite good solution with a helper we are already using.
> > To use it in the yaml files, the syntax would be similar to the usage of `env` in `Flow.yml` as it can be seen in:
> > `tests/integration/optimizers/flow.yml`
> 
> Pretty cool :) I guess for now it is not possible to pass list/dict arguments, right? I mean the following [but through command line/yaml]
> 
> ```python
> docker_kwargs = {'environment': ['VAR1=BAR', 'VAR2=FOO']}
> ```

Not sure, should be tested  If u want u can add a tesr and do a PR to this branch or I will try to test tmr

**Comment by tadejsv** [comment]
I can try testing it a bit later yes

**Comment by JoanFM** [comment]
Hey @tadejsv , I did some changes to allow passing strings and arbitrary values, plus I added a test to showcase and document how it can be passed into a `yaml`

**Comment by tadejsv** [comment]
Great work @JoanFM !!

File: daemon/models/custom.py [hunk 1]
```diff
@@ -72,7 +72,7 @@ def get_pydantic_fields(config: Union[dict, argparse.ArgumentParser]):
             if hasattr(arg_type, '__self__'):
                 arg_type = type(arg.default) if arg.default else int
             arg_type = str if isinstance(arg_type, argparse.FileType) else arg_type
-            arg_type = dict if type(arg) == KVAppendAction else arg_type
+            arg_type = dict if (isinstance(arg, KVAppendAction) or isinstance(arg, DockerKwargsAppendAction)) else arg_type
```
tadejsv, line 75: You can also do `isinstance(arg, (KVAppendAction, DockerKwargsAppendAction))` :)

rutujasurve94 (approved):

PR #1693, opened 2021-01-14
feat: support json print in loginfo

cristianmtr:
I found this more useful when debugging. I can just copy-paste the output into a JSON tree viewer, and collapse/expand as needed during inspection.

### DIFF
--- a/jina/drivers/control.py
+++ b/jina/drivers/control.py
@@ -3,6 +3,8 @@ __license__ = "Apache-2.0"
 
 import time
 
+from google.protobuf.json_format import MessageToJson
+
 from . import BaseDriver
 from .querylang.queryset.dunderkey import dunder_get
 from ..excepts import UnknownControlCommand, RuntimeTerminated, NoExplicitMessage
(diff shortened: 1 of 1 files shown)


jina/drivers/control.py
```diff
@@ -22,17 +24,24 @@ def envelope(self) -> 'jina_pb2.EnvelopeProto':
 class LogInfoDriver(BaseControlDriver):
     """Log output the request info"""
 
-    def __init__(self, key: str = 'request', *args, **kwargs):
+    def __init__(self, key: str = 'request', json=False, *args, **kwargs):
```
JoanFM, line 27: should we do `json` by default? I think it could be better to have Json directly

JoanFM: in case we mantain, please add type hint

cristianmtr: added type hint. Not sure about making it default or not. What do others think?

cristianmtr: I made it default to `True`

JoanFM (approved):

PR #1701, opened 2021-01-15
fix: KVSearchDriver `is_merge` should be `False` by default

cristianmtr:
This is to prevent duplicating the embedding and raising the reshaping error in Document (`ValueError: cannot reshape array of size 10 into shape (10,10)`)

A more robust solution should be discussed in the future


JoanFM (approved):

cristianmtr:
Fixed a test

JoanFM (approved):

Issue #1670, opened 2021-01-12
save_config not working for `TransformerTorchEncoder`


JoanFM:
**Describe the bug**
Tests fail for https://github.com/jina-ai/jina-hub/pull/2158 due to some problems when dumping this executor with `JAML`

JoanFM:
This is very weird, because it seems to appear in hub builds but not to be consistent

tadejsv:
@JoanFM, I came to the bottom of this. The problem appears here

https://github.com/jina-ai/jina/blob/321a890ba85cb8c0266b239bc0e7ea34bccf7ae0/jina/executors/__init__.py#L58-L73

See the line 65 there? Well, what happens is the following: by the time the **local** `TransformerTorchEncoder` is imported (and thus registered), then **jina package** `TransformerTorchEncoder` has already been imported, and thus registered. This means that `'TransformerTorchEncoder'` in `reg_cls_set`, and this the whole if statement will be skipped - leading to `store_init_kwargs` args not getting saved, and this not being part of the yaml dump.

Can you take it from here? One way to fix it would be to remove the if statement, but I am not sure this has any wider consequences, so I would prefer someone other than me do this PR.

PS: The statement is also not getting executed on the following two executors: `TransformEncoder` and `ChromaPitchEncoder`. I have checked and there are in fact two different executors with the same class name in the code.

tadejsv:
PS: you can consistently reproduce the problem with the code from #1527 , just replace `TransformerTFEncoder` with `TransformerTorchEncoder`

tadejsv:
Also, in local testing I can confirm that replacing the line 65 with `if True:` fixes the problem.

JoanFM:
> @JoanFM, I came to the bottom of this. The problem appears here
> 
> https://github.com/jina-ai/jina/blob/321a890ba85cb8c0266b239bc0e7ea34bccf7ae0/jina/executors/__init__.py#L58-L73
> 
> See the line 65 there? Well, what happens is the following: by the time the **local** `TransformerTorchEncoder` is imported (and thus registered), the **jina package** `TransformerTorchEncoder` has already been imported, and thus registered. This means that `'TransformerTorchEncoder'` is already in `reg_cls_set`, and this the whole if statement will be skipped - leading to `store_init_kwargs` args not getting saved, and this not being part of the yaml dump.
> 
[...]

Issue #1527, opened 2020-12-22
Params not saved in config


tadejsv:
There is a bug that when importing some encoders directly from the source file, all config params will not be saved. This is causing the failing tests in jina-ai/jina-hub#1349, although for some mysterious reason no failure occurs in jina-ai/jina-hub#1350.

To replicate, first cd to the proper folder (starting at the root of `jina-hub`):

```
cd encoders/nlp
```
Then, open a python terminal and execute
``` python
from TransformerTFEncoder import TransformerTFEncoder
encoder = TransformerTFEncoder('distilbert-base-uncased')
encoder.save_config()
```

Now examine the resulting file, it will look something like this
``` yaml
!TransformerTFEncoder
metas:
  name: TransformerTFEncoder.TransformerTFEncoder-1f530892
  pea_id: 0
  pea_workspace: .//None-{root.metas.pea_id}
  separated_workspace: false
requests:
  on:
    ControlRequest:
    - !ControlReqDriver
      with:
        priority: 0
    IndexRequest:
    - &id001 !EncodeDriver
      with:
        executor: null
        method: encode
        priority: 0
        traversal_paths:
        - c
        - r
    SearchRequest:
    - *id001
    TrainRequest:
    - *id001
```

Note that the custom parameters are not saved! I also get the same problem with `TransformerTorchEncoder`

If you replace the encoder import statement with 
``` python
from jina.hub.encoders.nlp.TransformerTFEncoder import TransformerTFEncoder
```

the whole thing will work, and the yaml file will have this at the end
``` yaml

with:
  layer_index: -1
  max_length: null
  model_save_path: null
  pooling_strategy: mean
  pretrained_model_name_or_path: distilbert-base-uncased
  truncation_strategy: longest_first
```

PR #1702, opened 2021-01-15
fix: register class should not only depend on name

JoanFM:
Linked to #1670  and #1527

### DIFF
--- a/tests/unit/test_exectype.py
+++ b/tests/unit/test_exectype.py
@@ -7,9 +7,20 @@ from jina.executors import BaseExecutor
 from jina.jaml import JAML
 
 
-def test_exec_type(tmpdir):
+@pytest.fixture
+def unregister():
     from jina.executors.indexers import BaseIndexer
-    assert 'BaseIndexer' in BaseExecutor._registered_class
+    if 'tests.unit.test_exectype.BaseIndexer' in BaseIndexer._registered_class:
+        BaseIndexer._registered_class.remove('tests.unit.test_exectype.BaseIndexer')
+    yield
+    if 'tests.unit.test_exectype.BaseIndexer' in BaseIndexer._registered_class:
+        BaseIndexer._registered_class.remove('tests.unit.test_exectype.BaseIndexer')
+
+
+@pytest.mark.parametrize('f_register', [True, False])
+def test_exec_type(tmpdir, f_register, unregister):
+    from jina.executors.indexers import BaseIndexer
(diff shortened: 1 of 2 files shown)


tadejsv:
One question here, not really related to the PR but more of a fundamental python question: how is it possible that `jina.hub.encoders.nlp.TransformerTorchEncoder` (and litreally every other encoder) was registered? I checked and it is never referenced/imported anywhere (the tests will only import the local one).

JoanFM:
> One question here, not really related to the PR but more of a fundamental python question: how is it possible that `jina.hub.encoders.nlp.TransformerTorchEncoder` (and litreally every other encoder) was registered? I checked and it is never referenced/imported anywhere (the tests will only import the local one).

I believe is some prework done as u can see when u do jina check.

tadejsv:
> > One question here, not really related to the PR but more of a fundamental python question: how is it possible that `jina.hub.encoders.nlp.TransformerTorchEncoder` (and litreally every other encoder) was registered? I checked and it is never referenced/imported anywhere (the tests will only import the local one).
> 
> I believe is some prework done as u can see when u do jina check.

You're right, I found that `__init__.py` file executes this: 
``` python
_ji.import_classes('jina.hub', show_import_table=False, import_once=True)
```

tadejsv:
@JoanFM I think this problem (of registring a class with only the name) also appears in some other places - definitely for drivers. I found those instances by searching for `__name__`. Probably would be best to solve it there as well?

JoanFM:
> @JoanFM I think this problem (of registring a class with only the name) also appears in some other places - definitely for drivers. I found those instances by searching for `__name__`. Probably would be best to solve it there as well?

these MetaClass is only used for Executors, so at least for storing init kwargs should be okey.

tests/unit/test_exectype.py
```diff
@@ -28,18 +40,21 @@ def assert_bi():
     # we override BaseIndexer now, without force it shall not store all init values
```
nan-wang, line 40: Would you please change the comments here as well? By this point, `BaseIndexer` has not registered in `reg_cls_set` yet and `store_init_kwargs` will be executed

nan-wang (changes requested):
LGTM👍

tests/unit/test_exectype.py
JoanFM: Done!

nan-wang (approved):
LGTM👍

PR #1711, opened 2021-01-16
refactor: optional disable uses_b/a if parallel is 1

JoanFM:
**Changes introduced**
Right now if we use `uses_before` or `uses_after` it is added even when parallel is 1. Although it can be desired to have a `uses_after` or `uses_before` it is more common to only want it when there are multiple Peas inside a Pod.

Add an option to disable them if parallel is 1. Like this one can keep the same `Flow` structure for `sharding` and no `sharding` and just parametrize the number of parallels or other options while not wasting CPU and network hops by these extra Peas

### DIFF
--- a/jina/parsers/peapods/pods/base.py
+++ b/jina/parsers/peapods/pods/base.py
@@ -9,12 +9,15 @@ def mixin_base_pod_parser(parser):
     gp = add_arg_group(parser, title='Pod')
 
     gp.add_argument('--uses-before', type=str,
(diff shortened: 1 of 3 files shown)


florian-hoenicke (changes requested):
Can we add a small test here?

JoanFM:
> Can we add a small test here?

sure, wilk

JoanFM:
> Can we add a small test here?

Done!

nan-wang (approved):
LGTM👍

PR #1716, opened 2021-01-17
test: crud sharding

florian-hoenicke:


tests/integration/sharding/test_sharding.py
```diff
@@ -0,0 +1,225 @@
...
+from jina.executors.indexers import BaseIndexer
+from jina.flow import Flow
+
+random.seed(0)
+np.random.seed(0)
+
+
+def get_index_flow(yaml_file, num_shards):
+    f = Flow().add(
+        uses='yaml/' + yaml_file,
```
JoanFM, line 19: lets use cur_dir as in other tests to make this test runnable from different folders so that paths are absolute

tests/integration/sharding/test_sharding.py
```diff
@@ -0,0 +1,225 @@
...
+        shards=num_shards,
+        separated_workspace=True,
+    )
+    return f
+
+
+def get_delete_flow(yaml_file, num_shards):
+    f = Flow().add(
+        uses='yaml/' + yaml_file,
```
JoanFM, line 28: same here

JoanFM (commented):
Does this belong better as subfolder of integration/crud maybe?

JoanFM (commented):
If the logic is counting on the fact that the default `request_size` at index time is set to 100, I would set it explicitly in these tests.

tests/integration/sharding/test_sharding.py
florian-hoenicke: thanks, changed

tests/integration/sharding/test_sharding.py
florian-hoenicke: thanks, changed

florian-hoenicke:
> If the logic is counting on the fact that the default `request_size` at index time is set to 100, I would set it explicitly in these tests.

Good point. Therefore, the tests will not break in case we change the default value. Thanks.

florian-hoenicke:
Since https://github.com/jina-ai/jina/pull/1719/files is merged, the tests of this pr can run through.

> If the logic is counting on the fact that the default request_size at index time is set to 100, I would set it explicitly in these tests.

I think if we would move this to crud testing, we would have to move other features which work with crud as well (e.g. docidcache)
Delete and update is required in many features - like index and search.
Therefore, whenever we develop a new feature, we need to test it not only with index and search but also with update and delete.
Putting all these test cases into one folder would not be an option for us.

jina/resources/executors._merge_matches_topk.yml
```diff
@@ -8,10 +8,6 @@ requests:
       - !ReduceAllDriver
         with:
           traversal_paths: ['m']
-      - !SortQL
```
JoanFM, line 11: Please add it back

JoanFM (changes requested):
> Since https://github.com/jina-ai/jina/pull/1719/files is merged, the tests of this pr can run through.
> 
> > If the logic is counting on the fact that the default request_size at index time is set to 100, I would set it explicitly in these tests.
> 
> I think if we would move this to crud testing, we would have to move other features which work with crud as well (e.g. docidcache)
> Delete and update is required in many features - like index and search.
> Therefore, whenever we develop a new feature, we need to test it not only with index and search but also with update and delete.
> Putting all these test cases into one folder would not be an option for us.

the point is these tests ensure that CRUD operations work well with sharding, so for me they should belong under a subfolder of crud.

Update and Delete are the operations that define CRUD . (docidcache is also needed for incremental indexing). Not a big deal though, but I think is how it makes it more clear, even the PR title seems to suggest this inmho

jina/resources/executors._merge_matches_topk.yml
```diff
@@ -8,10 +8,6 @@ requests:
...
-        with:
```
JoanFM, line 12: u may need to add rhe field: 'score__value`

JoanFM (changes requested):

hanxiao (approved):

PR #1718, opened 2021-01-17
feat: allow rank driver access info in tags

JoanFM:
**Changes introduced**
Right now, the Driver uses `required_keys` to extract the metainformation from Matches and Query to pass to the `Ranker` Executor.

However, there is no way the `Executor` can have access to the `metainfo` from `tags`. The user could choose to select `tags` but it would get all of them.

I propose to allow the user to access fields from tags using `required_keys` as `tags_*` (with a similar syntax as in QueryLangDriver)

**TODO**
- Allow the executors to have 2 sets of required_keys, 1 for queries and 1 for matches.
- Have these requred_keys None in the BaseClasses

@hanxiao @nan-wang @maximilianwerk does this feel like the way to go? I think this would be good for the Ranker Executor abstraction


jina/executors/rankers/__init__.py
```diff
@@ -39,6 +38,14 @@ class Chunk2DocRanker(BaseRanker):
     COL_DOC_CHUNK_ID = 'doc_chunk_hash'
     COL_SCORE = 'score'
 
+    required_keys = {'text'}
+    """
+    Set of required keys to be extracted from matches and query to fill the information of `query` and `chunk` meta information.
+    These are the set of keys to be extracted from `Document`.
+
+    Keys prefixed with `tags` will be directly extracted from `Document tags` as if they were direct arguments from documents.
```
nan-wang, line 46:
```suggestion
    Keys prefixed with `tags_` will be directly extracted from `Document.tags` as if they were direct arguments from documents.
```

jina/executors/rankers/__init__.py
```diff
@@ -39,6 +38,14 @@ class Chunk2DocRanker(BaseRanker):
     COL_DOC_CHUNK_ID = 'doc_chunk_hash'
     COL_SCORE = 'score'
 
+    required_keys = {'text'}
+    """
+    Set of required keys to be extracted from matches and query to fill the information of `query` and `chunk` meta information.
```
nan-wang, line 43: The inline doctring format: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html
```suggestion
    """set: Set of required keys to be extracted from matches and query to fill the information of `query` and `chunk` meta information.
```

nan-wang (changes requested):
I like this idea.

JoanFM:
> I like this idea.

My doubt now is if we should have the `executor` take all the tags and use them or we should do this trick so that the complexity is hidden?

nan-wang:
> > I like this idea.
> 
> My doubt now is if we should have the `executor` take all the tags and use them or we should do this trick so that the complexity is hidden?

`required_keys` is only used in the `Chunk2DocRanker`, right?

JoanFM:
> > > I like this idea.
> > 
> > 
> > My doubt now is if we should have the `executor` take all the tags and use them or we should do this trick so that the complexity is hidden?
> 
> `required_keys` is only used in the `Chunk2DocRanker`, right?

In any Ranker.

I also would like to have 2 set of keys, one for `matches` and one for `queries`

jina/drivers/rank/__init__.py
```diff
@@ -49,11 +48,10 @@ def _apply_all(self, docs: 'DocumentSet', context_doc: 'Document', *args,
             - Set the ``traversal_paths`` of this driver such that it traverses along the ``matches`` of the ``chunks`` at the level desired.
         """
 
-        # if at the top-level already, no need to aggregate further
-        query_meta = context_doc.get_attrs(*self.exec.required_keys)
-
+        query_meta = context_doc.get_attrs(*self.exec.required_keys) if hasattr(self.exec, 'required_keys') else None
         old_match_scores = {match.id: match.score.value for match in docs}
-        match_meta = {match.id: match.get_attrs(*self.exec.required_keys) for match in docs}
+        match_meta = {match.id: match.get_attrs(*self.exec.required_keys) for match in docs} if hasattr(self.exec, 'required_keys') else None
```
JoanFM, line 53: For https://github.com/jina-ai/jina-hub/pull/3670 it is important to make sure that the `required_keys` are obtained in the exact same order as in the `required_keys`

jina/types/document/__init__.py
```diff
@@ -450,8 +453,40 @@ def get_attrs(self, *args) -> Dict[str, Any]:
...
+
+            .. highlight:: python
+            .. code-block:: python
+
+                d = Document({'id': '123', 'hello': 'world', 'tags': {'good': 'bye'}})
+                res = d.get_attrs(*['id', 'hello', 'tags__good'])
+
+                assert res['id'] == '123'
+                assert res['hello'] == 'world'
+                assert res['good'] == 'bye'
```
maximilianwerk, line 469: I would prefer `assert res['tags_good'] == 'bye'`, but I guess having it as `good` is exactly the point, right? But what should happen if I do
```
 res = d.get_attrs(*['id', 'hello', 'tags__id'])
```
?

jina/drivers/rank/__init__.py
```diff
@@ -49,11 +48,10 @@ def _apply_all(self, docs: 'DocumentSet', context_doc: 'Document', *args,
             - Set the ``traversal_paths`` of this driver such that it traverses along the ``matches`` of the ``chunks`` at the level desired.
         """
 
-        # if at the top-level already, no need to aggregate further
-        query_meta = context_doc.get_attrs(*self.exec.required_keys)
-
+        query_meta = context_doc.get_attrs(*self.exec.required_keys) if hasattr(self.exec, 'required_keys') else None
```
maximilianwerk, line 51: Why this `if/else`? Why is it needed out of a sudden?

maximilianwerk (commented):
I don't see the benefit of choosing single fields in `tags` over getting the whole object back. Could you elaborate, why this is needed?

jina/types/document/__init__.py
JoanFM: yes, `tags__id` is a big problem. I think the nice way of assuming things from `tags` has the good thing that Executor developer does not need to know about `tags` and can keep a more "natural" language.

For instance, https://github.com/jina-ai/jina-hub/pull/3670 in this hub executor I propose, one needs to extract `features`. Should I know that they are stored in `tags`? Or is it better to allow the `Ranker` to talk its own language?

I see benefits and problems in every approach.

In the current approach there is a problem with `id` and `tags__id` because one would overwrite the other. It could be solved in a little hacky way fhough

jina/drivers/rank/__init__.py
JoanFM: My bad, this is an advancement of an idea I had, but I will wait for another PR.

I was thinking about allowing to have different set of `required_keys` for `query` and for `matches`, and therefore some of the rankers may not need one of the sets. 

I will remove for now

JoanFM: I also think is not very good to have `text` as default. I'd rather have the class as abstract than having such a strong default

JoanFM: But good point, removed now

jina/types/document/__init__.py
JoanFM: Handled with new logic

JoanFM:
> I don't see the benefit of choosing single fields in `tags` over getting the whole object back. Could you elaborate, why this is needed?

If u see the LightGBMRanker in the hub (a new PR has been opened). The most natural way to use a LambdaMart model would be to extract some features from Documents (which would live inside the tags object). if we allow this access, the user does not need have to manually extract from tags the fields, which is a detail of our Document type that we intend to hide for the executors developers.

JoanFM:
What do you think @nan-wang about this? is this better than just extracting the tags from document and have the executor developer access it from tags?

JoanFM:
See PR on LightGBMRanker here https://github.com/jina-ai/jina-hub/pull/3670

tests/unit/types/document/test_document.py
```diff
@@ -495,3 +492,27 @@ def test_update_exclude_field():
...
+    d = Document({'id': '123', 'text': 'document', 'feature1': 121, 'name': 'name', 'tags': {'id': 'identity', 'a': 'b', 'c': 'd'}})
+    d.score = NamedScore(value=42)
+
+    res = d.get_attrs(*['id', 'text', 'name', 'feature1', 'score__value', 'tags__c', 'tags__id', 'inexistant', 'tags__none', 'tags__inexistant'])
+
+    assert res['id'] == '123'
+    assert res['feature1'] == 121
+    assert res['name'] == 'name'
+    assert res['text'] == 'document'
+    assert res['c'] == 'd'
```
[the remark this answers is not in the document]
JoanFM: which feature you mean?

nan-wang (dismissed):
LGTM👍

jina/types/document/__init__.py
```diff
@@ -457,8 +458,40 @@ def get_attrs(self, *args) -> Dict[str, Any]:
...
+                assert res['id'] == '123' # true
+                assert res['tags__hello'] == 'world' # true
+                assert res['tags__good'] == 'bye' # true
+                assert res['tags__id'] == 'external_id' # true
         """
-        return {k: getattr(self, k) for k in args if hasattr(self, k)}
+
+        ret = {}
+        for k in args:
+            try:
```
florian-hoenicke, line 483: `hasattr()` can be used instead of try except

jina/types/document/__init__.py
```diff
@@ -457,8 +458,40 @@ def get_attrs(self, *args) -> Dict[str, Any]:
...
+                value = getattr(self, k)
+                ret[k] = value
+                continue
+            except AttributeError:
+                pass
+            try:
```
florian-hoenicke, line 489: should we not just let it fail if the `k` is neither an attribute nor a dunder key?

florian-hoenicke (changes requested):

nan-wang (dismissed):
LGTM👍

jina/types/document/__init__.py
nan-wang: Agree. `hasattr` is a better option here.

jina/types/document/__init__.py
JoanFM: no because some Rankers may rely in the order of the keys in the list. I would keep it here for the moment

jina/types/document/__init__.py
JoanFM: I did it for potential performance when I was checking access in tags, but is true for just Document is better to do so

jina/types/document/__init__.py
```diff
@@ -457,8 +458,27 @@ def get_attrs(self, *args) -> Dict[str, Any]:
...
+                res = d.get_attrs(*['id', 'tags__hello', 'tags__good', 'tags__id'])
+
+                assert res['id'] == '123' # true
+                assert res['tags__hello'] == 'world' # true
+                assert res['tags__good'] == 'bye' # true
+                assert res['tags__id'] == 'external_id' # true
         """
-        return {k: getattr(self, k) for k in args if hasattr(self, k)}
+
+        return {k: getattr(self, k) if hasattr(self, k) else dunder_get(self._pb_body, k) for k in args}
```
maximilianwerk, line 481: We should catch & log, when `dunder_get` fails here. Otherwise, one malformed document in the index can corrupt a lot of rankings/searches.

maximilianwerk (changes requested):
Please catch, if `dunder_get` raises an Exception.

jina/types/document/__init__.py
JoanFM: the semantics are directly that it returns a None, I can make it fail if it makes us more secure

nan-wang: The errors caused by a malformed document should not be caught by the `Document` type. Logging in a primitive type is also strange.

JoanFM: `DocumentSet` does warning in `_extract_docs()`.

I am not sure what is the best option, so whatever we agree works for me.

What agreement we take @nan-wang @maximilianwerk ?

nan-wang: I've no strong opinion of this. I'm ok with the following codes. @maximilianwerk 
```
            try:
                value = dunder_get(self._pb_body, k)
                ret[k] = value
                continue
            except (AttributeError, ValueError):
                defaultlogger.warning('some warning message')
                ret[k] = None
```

maximilianwerk: Ah I am sorry, I should have read the tests in the first place. I believe it can only be an `AttributeError`, non? See
```
    with pytest.raises(AttributeError):
        d.get_attrs(*['inexistant'])
```

I'd rather catch it, to be honest.

JoanFM: see the updated version

nan-wang (approved):
LGTM👍

PR #1719, opened 2021-01-17
test: sharding non existent

florian-hoenicke:
Needs to be fixed. We insert 201 documents but only get 101 back.


JoanFM:
> Needs to be fixed. We insert 201 documents but only get 101 back.

Do u use merge with traversal root?

JoanFM:
This is expected as the KVSearchDriver removes the non-found documents.

florian-hoenicke:
> This is expected as the KVSearchDriver removes the non-found documents.

In this case we index `random_docs(0, 201)` and query `random_docs(0, 220)` but just get `101` documents back. Expected would be to get `201` documents.

JoanFM:
> > This is expected as the KVSearchDriver removes the non-found documents.
> 
> In this case we index `random_docs(0, 201)` and query `random_docs(0, 220)` but just get `101` documents back. Expected would be to get `201` documents.

The expected result is that u will get in output as many documents as they are found in the `KVIndex`.  Do the 220 documents guarantee that 201 of them have `id` that had been previously indexed?

JoanFM:
> > This is expected as the KVSearchDriver removes the non-found documents.
> 
> In this case we index `random_docs(0, 201)` and query `random_docs(0, 220)` but just get `101` documents back. Expected would be to get `201` documents.

And if it is the case, u need `uses_after` to merge all the documents at root level, so that the Request is rebuilt.

What is happening in this test is that, the first shard finishing returns its results. And that is why u just get a portion, what u would maybe see is 2 respones

florian-hoenicke:
>The expected result is that u will get in output as many documents as they are found in the KVIndex. Do the 220 documents guarantee that 201 of them have id that had been previously indexed?

Yes

> And if it is the case, u need `uses_after` to merge all the documents at root level, so that the Request is rebuilt.

Good point. However, I tried it out with `uses_after='_merge',` as well and it did not work.

JoanFM:
> > The expected result is that u will get in output as many documents as they are found in the KVIndex. Do the 220 documents guarantee that 201 of them have id that had been previously indexed?
> 
> Yes
> 
> > And if it is the case, u need `uses_after` to merge all the documents at root level, so that the Request is rebuilt.
> 
> Good point. However, I tried it out with `uses_after='_merge',` as well and it did not work.

what traversal paths are used there?

florian-hoenicke:
As discussed with @JoanFM supporting sharding for plain kv-indexers would be usefull but still requires some conceptual thinking and might take some time to implement.
In this example we use `polling all` to retrieve results from two shards. 
The problem is that the ReduceDriver seems not to work on document level but only on matches.

JoanFM:
> As discussed with @JoanFM supporting sharding for plain kv-indexers would be usefull but still requires some conceptual thinking and might take some time to implement.
> In this example we use `polling all` to retrieve results from two shards.
> The problem is that the ReduceDriver seems not to work on document level but only on matches.

ReduceAllDriver ...

jina/drivers/__init__.py
```diff
@@ -238,6 +245,15 @@ def __init__(self, traversal_paths: Tuple[str] = ('c', 'r'), *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._traversal_paths = [path.lower() for path in traversal_paths]
 
+    def _apply_root(
+        self,
+        docs: 'DocumentSet',
+        field: str,
```
JoanFM, line 251: if we say is `None`, should we also hide `field`?

jina/drivers/reduce.py
```diff
@@ -29,13 +29,22 @@ def __call__(self, *args, **kwargs):
         self._traverse_apply(self.docs, *args, **kwargs)
         self.doc_pointers.clear()
 
+    def _apply_root(self, docs: 'DocumentSet', field: str, *args, **kwargs):
+        docs = []
+        for doc in self.docs:
+            docs.append(doc)
+        request = self.msg.request
+        request.body.ClearField(field)
+        request.docs.extend(docs)
```
JoanFM, line 38: so I guess you saw the problem I had?

JoanFM (dismissed):

tests/integration/sharding/test_search_non_existent.py
```diff
@@ -0,0 +1,61 @@
...
+    for j in range(start, end):
+        d = Document()
+        d.id = f'{j:0>16}'
+        d.tags['id'] = j
+        d.text = ''.join(random.choice(string.ascii_lowercase) for _ in range(10)).encode('utf8')
+        d.embedding = np.random.random([embed_dim])
+        yield d
+
+
+# fails #TODO fix related issue
```
cristianmtr, line 33: remove this? :wink:

tests/integration/sharding/test_search_non_existent.py
```diff
@@ -0,0 +1,61 @@
...
+    ) as index_flow:
+        index_flow.index(input_fn=random_docs(0, 3), request_size=1)
+
+    mock = mocker.Mock()
+    with Flow(read_only=True).add(
+            show_exc_info=True,
+            uses=os.path.join(cur_dir, 'yaml', yaml_file),
+            shards=2,
+            separated_workspace=True,
+            uses_after=os.path.join(cur_dir, 'yaml', 'merge_root.yml'),
```
cristianmtr, line 54: Shouldn't this `merge_root` be available in resources to all?

cristianmtr (changes requested):
some quick clarifications

tests/integration/sharding/test_search_non_existent.py
JoanFM: could make sense

florian-hoenicke: Nice idea. I moved it to the resources.

tests/integration/sharding/test_search_non_existent.py
florian-hoenicke: done

jina/drivers/__init__.py
```diff
@@ -200,6 +200,13 @@ def queryset(self) -> 'QueryLangSet':
         else:
             return []
 
+    @property
+    def docs(self):
+        if self.expect_parts > 1:
+            return (d for r in reversed(self.partial_reqs) for d in r.docs)
```
florian-hoenicke, line 206: @JoanFM do we reverse the requests in order to wait for the last one to arrive?

jina/drivers/reduce.py
florian-hoenicke: @JoanFM The problem was that the docs were stored in the `body.docs` as well as in `docs`?

jina/drivers/__init__.py
florian-hoenicke: `field` contains `"docs"`.
The `context_doc` is set to `None`

florian-hoenicke:
@hanxiao could you review this?
On this pr @JoanFM @maximilianwerk and me worked collaboratively together.
Means no one of us can approve it.

jina/drivers/__init__.py
JoanFM: it was like this, so to be honest not sure why it was there in the first place

hanxiao (approved):

PR #1720, opened 2021-01-17
test: sharding workspace

florian-hoenicke:
Needs to be fixed. If we don't use a separated workspace and insert 3 documents on 2 shards using 3 requests (1 doc each), only 1 document ends up being stored on one shard


JoanFM:
> Needs to be fixed. If we don't use a separated workspace and insert 3 documents on 2 shards using 3 requests (1 doc each), only 1 document ends up being stored on one shard

I think the problem is that separated workspace should be enforced, because otherwise they may collide. I would even vote to remove the separated workspace option and have this behavior as default

florian-hoenicke:
> I think the problem is that separated workspace should be enforced, because otherwise they may collide. I would even vote to remove the separated workspace option and have this behavior as default

I agree. This would make our lives  much easier.
@hanxiao Would you have any concern with removing the separated workspace attribute and setting it to `True` internally?

florian-hoenicke:
Actually, what about the gateway pod? Should it also be in a separated workspace?

JoanFM:
> Actually, what about the gateway pod? Should it also be in a separated workspace?

why does it affect? it is not parallelized? this parameter does not affect if 2 pods share workspace or not

florian-hoenicke:
You are right, it has no affect if we have `parallel =1`. So it is fine.

florian-hoenicke:
As discussed with @han we remove the `separated_workspace` parameter and replace it with `True` where ever it is used.

florian-hoenicke:
related pr:
https://github.com/jina-ai/jina/pull/1722

JoanFM:
> As discussed with @han we remove the `separated_workspace` parameter and replace it with `True` where ever it is used.

u mean substitute by a conditional on parallel?

florian-hoenicke:
> > As discussed with @han we remove the `separated_workspace` parameter and replace it with `True` where ever it is used.
> 
> u mean substitute by a conditional on parallel?

yes just in case `parallel` > 1

Issue #1606, opened 2021-01-06
🚨 BREAKING CHANGES 0.9.0 -> 1.0.0


nan-wang:
This issue is created to keep track of breaking changes between `0.9.0` and `1.0.0` (scheduled release on Feb. 15, 2021)

## Breaking Changes & Migration Guide

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1722 | yes | #1739 |
| #1739 | yes | |

- Remove `separate_workspace`. Instead of using `pea_workspace`, `shard_workspace` is defined based on the `workspace,  the `name` and the `pea_id`. Set `workspace` metas default to `None`. Introduce `root_workspace` and `root_name` to resolve the workspace issue in `CompoundExecutor` and `ref_indexer`. Use `complete_path` to find `uses_internal` resource to `bind` into `docker` when spinning a `pea` in a `container`

<table stype="margin-left:auto;margin-right:auto;">
<tr>
<td> </td>
<td> v0.9.0 </td>
<td> v1.0.0 </td>
</tr>
<tr>
<td> </td>
<td> <sub> 

```python
f = Flow().add(uses='NumpyIndexer', shards=2, separated_workspace=True)
```

</sub> </td>
<td> <sub> 

```python
f = Flow().add(uses='NumpyIndexer', shards=2)

# saved files
# ├── jina.executors.indexers.vector.NumpyIndexer-86342edc-1
# │   ├── jina.executors.indexers.vector.NumpyIndexer-86342edc
# │   └── jina.executors.indexers.vector.NumpyIndexer-86342edc.bin
# └── jina.executors.indexers.vector.NumpyIndexer-c4f2ea6c-2
#     └── jina.executors.indexers.vector.NumpyIndexer-c4f2ea6c
```

</sub> </td>
</tr>
</table>

---

| PR |  Rolled Out? | Override |
| --- | --- | --- |
| #1632 | yes | |

- Separate `Segmenter` from `Crafter`. `Segmenter` is used to segment each Document into several Documents. In comparison, `Crafter` is used to process each Document and the output is still only one Document. 
 
<table stype="margin-left:auto;margin-right:auto;">
<tr>
<td> </td>
<td> v0.9.0 </td>
<td> v1.0.0 </td>
</tr>
<tr>
<td> </td>
<td> <sub> 

```python
from jina.executors.crafters import BaseSegmenter
class DummySentencizer(BaseSegmenter):
    def craft(self, text, *args, **kwargs):
[...]

PR #1722, opened 2021-01-18
feat: remove separated workspace param

florian-hoenicke:
## Breaking Change #1606 
related discussion:
https://github.com/jina-ai/jina/pull/1720

**Changes introduced**

- [x]  Remove `separate_workspace`
- [x]  Remote `pea_workspace` meta
- [x] Set `workspace` metas default to `None` to allow understand the logic where the user does not explicitly provides a value. Useful to clarify workspace structure in the case of `CompoundExecutors`. (In CompoundExecutor, components will compute its workspace considering some logic on its `CompoundExecutor` continent.
- [x]  Have `shard_workspace` property to be defined functionally `depending` on the `workspace`, the `name` and the `pea_id`.
- [x] Adapt logic to set component workspace based on `CompoundExecutor`  
- [x] Set `root_workspace` and `root_name` to set to `root` executor needed for `CompoundExecutor` and `ref_indexer`.
- [x] Have strong set of tests for different configurations of `workspaces` (Single, with Compound, with Compound but with specific workspace for its `components`, with `ref_indexer`, with `ref_indexer` inside a `Compound`
- [x] Use `complete_path` to find `uses_internal` resource to `bind` into `docker` when spinning a `pea` in a `container`


JoanFM (commented):
maybe we need to remove `pea_workspace` as well

jina/executors/metas.py
```diff
@@ -144,7 +136,7 @@
 
     .. note::
 
-        ``separated_workspace``, ``pea_workspace`` and ``pea_id`` is set in a way that when the executor ``A`` is used as
+        ``pea_workspace`` and ``pea_id`` is set in a way that when the executor ``A`` is used as
```
JoanFM, line 139: pea_workspace should as well be removed I think as now the workspace is uniquely identified

jina/jaml/parsers/executor/legacy.py
```diff
@@ -12,14 +12,14 @@ class LegacyParser(VersionedYAMLParser):
...
-            if meta_config.get('separated_workspace', False) is True and meta_config['pea_id'] != -1:
+            if meta_config['pea_id'] != -1:
                 if 'pea_id' in meta_config and isinstance(meta_config['pea_id'], int):
                     work_dir = meta_config['pea_workspace']
                     dump_path = os.path.join(work_dir, f'{meta_config["name"]}.{"bin"}')
                     if os.path.exists(dump_path):
                         return dump_path
                 else:
-                    raise BadWorkspace('separated_workspace=True but pea_id is unset or set to a bad value')
+                    raise BadWorkspace('pea_id is unset or set to a bad value')
```
JoanFM, line 22: Bad error message, may u say that the path does not exist?

jina/parsers/peapods/pea.py
```diff
@@ -38,7 +38,7 @@ def mixin_pea_parser(parser):
     # hidden CLI used for internal only
 
     gp.add_argument('--pea-id', type=int, default=-1,
-                    help='the id of the storage of this pea, only effective when `separated_workspace=True`'
+                    help='the id of the storage of this pea`'
```
JoanFM, line 41: u say, needed when multiple peas run in parallel inside a same Pod

JoanFM (changes requested):

jina/jaml/parsers/executor/legacy.py
```diff
@@ -12,14 +12,14 @@ class LegacyParser(VersionedYAMLParser):
     @staticmethod
     def _get_dump_path_from_config(meta_config: Dict):
         if 'name' in meta_config:
-            if meta_config.get('separated_workspace', False) is True and meta_config['pea_id'] != -1:
+            if meta_config['pea_id'] != -1:
                 if 'pea_id' in meta_config and isinstance(meta_config['pea_id'], int):
```
JoanFM, line 16: remove this if, the if inside should be enough I think

jina/executors/metas.py
florian-hoenicke: good idea. Let's do this in a dedicated pr.

jina/jaml/parsers/executor/legacy.py
florian-hoenicke: I think the message wants to say that the pea_id is not set to a value the path creation can work with.
However, I think for this pr we should just remove the separated_workspace attribute.

jina/parsers/peapods/pea.py
florian-hoenicke: lets do the documentation in a separated ticket

jina/jaml/parsers/executor/legacy.py
JoanFM: pea_id is unset or is not an integer is the condition, but I feel this condition was there to support the case where `separated_workspace` was True but `pea_id` was not coherent.

JoanFM (changes requested):
Not ready to be merged, we need to add some set of integration tests

jina/parsers/peapods/pea.py
JoanFM: let's add it here one second

tests/unit/drivers/test_cache_driver.py
```diff
@@ -70,8 +70,11 @@ def test_cache_driver_tmpfile(tmpdir, test_metas):
 
 
 def test_cache_driver_from_file(tmpdir, test_metas):
```
JoanFM, line 72: tmpdir should not be passed then?

JoanFM (changes requested):

tests/integration/ref_indexer/compound-index.yml
```diff
@@ -0,0 +1,7 @@
+!Flow
+with:
+  show_exc_info: true
```
theUnkownName, line 3: I thought the `show_exc_info` goes in the pods only? or is the other way around?

JoanFM: I think from here it may be passrd to pods but not sure

jina/executors/__init__.py
```diff
@@ -242,30 +240,52 @@ def config_abspath(self) -> str:
...
+    @property
+    def workspace_name(self):
+        return self.name
+
+    @property
+    def _workspace(self):
+        """ Property to access `workspace` if existing or default to `./`. Useful to provide good interface when
+        using executors directly in python.
+
+        .. highlight:: python
```
cristianmtr, line 257: This snippet doesn't seem relevant

florian-hoenicke: What would you change?

jina/executors/__init__.py
```diff
@@ -241,30 +240,52 @@ def config_abspath(self) -> str:
...
+        .. code-block:: python
+
+            with NumpyIndexer() as indexer:
+                indexer.touch()
+
+        :return: returns the workspace property of the executor or default to './'
+        """
+        return self.workspace or './'
```
maximilianwerk, line 265: I am totally surprised, that this works.
After digging around a little: Seems like I never thought about how boolean expressions work in python. I am totally surprised, that
```
a = 'a' or 'b'
=> a = 'a' (First `True` evaluated element)
a = 'a' and 'b'
=> a = 'b' (Last `True` evaluated element)
a = '' or 'b'
=> a = 'b' (First `True` evaluated element, which is the second one)
a = '' and 'b'
=> a = '' (First `False` evaluated element)
```
When you think how `and` and `or` are implemented (via `__and__` and `__or__`) this makes total sense. I am just amazed... A consequence is:
```
def my_func():
    return 'haha'
if ['a'] and my_func():
    ....
```
breaks down to
```
if 'haha':
    ...
```
which is obviously `True`.

jina/executors/__init__.py
```diff
@@ -241,30 +240,52 @@ def config_abspath(self) -> str:
         """
         return self.get_file_from_workspace(f'{self.name}.yml')
 
+    @staticmethod
+    def get_shard_workspace(workspace_folder: str, workspace_name: str, pea_id: int) -> str:
+        # TODO (Joan, Florian). We would prefer not to keep `pea_id` condition, but afraid many tests rely on this
+        return os.path.join(workspace_folder, f'{workspace_name}-{pea_id}') if pea_id > 0 else os.path.join(workspace_folder)
```
maximilianwerk, line 246: couldn't it be just `... else workspace_folder`?

maximilianwerk: And furthermore, if the whole line would just be
```
if pea_id < 0:
    pea_id = 0
return os.path.join(workspace_folder, f'{workspace_name}-{pea_id}')
```
we might be able to heavily simplify the parsers, non? If we have shards, it makes total sense, that we have the `pea_id` in the workspace name. But if we don't have shards but only one `Pea`, it still makes total sense to add the `pea_id` to the workspace name (in my opinion).

maximilianwerk (commented):
Good job, great refactoring. Just one point. I'd be happy to be convinced, that your implementation is better than my suggestion. :)

jina/executors/__init__.py
JoanFM: > couldn't it be just `... else workspace_folder`?

oh sure!!

JoanFM: About the second point, I agree, but I was afraid that it would break current logic and a huge amount of tests, and thus we decided to keep it like this as in the `TODO` is expressd

JoanFM:
> Good job, great refactoring. Just one point. I'd be happy to be convinced, that your implementation is better than my suggestion. :)

Your suggestion is definitely better, we went conservative on the amount of things we would break

JoanFM:
> > Good job, great refactoring. Just one point. I'd be happy to be convinced, that your implementation is better than my suggestion. :)
> 
> Your suggestion is definitely better, we went conservative on the amount of things we would break

I would suggest to give it a try in a different PR

jina/executors/__init__.py
maximilianwerk: Ah I should read the comments as well... That's why I almost never write comments: I also never read them :D

Fine I'll approve

JoanFM, line 246:
```suggestion
        return os.path.join(workspace_folder, f'{workspace_name}-{pea_id}') if pea_id > 0 else workspace_folder
```

JoanFM (changes requested):

maximilianwerk (approved):
LGTM

jina/executors/__init__.py
cristianmtr: I'd remove it, right?

JoanFM (approved):

PR #1723, opened 2021-01-18
chore: update add() use case in readme

Yongxuanzhang:
The old example of `.add(uses='docker://jinahub/pretrained-cnn:latest') ` is outdated since pretrained-cnn:latest doesn't exist in jinahub and will cause error. 
Change it with the most downloaded image `.add(uses='docker://jinahub/pod.encoder.dummy_mwu_encoder:0.0.6-0.9.3')`.


nan-wang (approved):
LGTM👍

PR #1726, opened 2021-01-19
feat: optimizer yaml interface

maximilianwerk:
This will enable using a .yml file for running the optimizer.

- Added jsonlines parsing to the `index_lines` and `search_lines`
  - Please have a look if this is fine. Especially, the change to `data_type=DataInputType.AUTO`
- Added simple JAML parsing for any objects which don't need special handling.

Regarding the JAML parsing: I am well aware that the current solution is not in line with how Flows, Executors and Drivers are handled. In case the current solution is not wanted due to broken consistency (or consequences which I have not understood yet), we could also write a simple parser, which is used, whenever no other "special" parser is needed. I would be super happy about better informed input here @hanxiao @JoanFM


jina/optimizers/flow_runner.py
```diff
@@ -70,6 +96,8 @@ def run(
         """
 
         self._setup_workspace(workspace)
+        self._reset_callback()
```
JoanFM, line 99: we discouraged using this construction right?

jina/flow/__init__.py
```diff
@@ -104,7 +104,7 @@ def index_lines(self, lines: Iterator[str] = None, filepath: str = None, size: i
         """
         from ..clients.sugary_io import _input_lines
         return self._get_client(**kwargs).index(_input_lines(lines, filepath, size, sampling_rate, read_mode),
-                                         on_done, on_error, on_always, data_type=DataInputType.CONTENT, **kwargs)
+                                         on_done, on_error, on_always, data_type=DataInputType.AUTO, **kwargs)
```
JoanFM, line 107: what does AUTO do?

jina/optimizers/flow_runner.py
JoanFM: it seems a smell that maybe evaluation callback should not be owned by this?

jina/optimizers/flow_runner.py
```diff
@@ -78,15 +106,25 @@ def run(
...
 
+    def _reset_callback(self):
+        self.callback = self.callback.get_fresh_callback()
+
+    def get_evaluations(self):
+        return self.callback.get_mean_evaluation()
+
 
-class MultiFlowRunner:
+class MultiFlowRunner(FlowRunner):
```
JoanFM, line 116: Like a lot this pattern

JoanFM (commented):
I do not understand why current JAML parser does not work?

jina/flow/__init__.py
maximilianwerk: It first tries `DOCUMENT` and in case it fails falls back to `CONTENT`.

jina/optimizers/flow_runner.py
maximilianwerk: Totally, I'll fix it. Thanks for pointing out.

maximilianwerk:
> I do not understand why current JAML parser does not work?

Because the current JAML parser works ONLY for classes, which inherit from `BaseDriver`, `BaseFlow` or `BaseExecutor`. See [this function](https://github.com/jina-ai/jina/blob/d7fdc477db9fae6ebefb21e1c8fbd6a8f35ade21/jina/jaml/parsers/__init__.py#L9). We could have another `SimpleParseableBaseClass` which has a simple parser and works if inherited from. 

But on the other hand, that is exactly, what `JAMLCompatible` is for and thus instead of having a `raise NotImplementedError(f'{cls!r} does not implement YAML parser')` in the above mentioned function, we could implement a default simple parser. That would be my preferred solution, but before implementing this, I wanted to double check if there are some obvious no-gos.

JoanFM:
> > I do not understand why current JAML parser does not work?
> 
> Because the current JAML parser works ONLY for classes, which inherit from `BaseDriver`, `BaseFlow` or `BaseExecutor`. See [this function](https://github.com/jina-ai/jina/blob/d7fdc477db9fae6ebefb21e1c8fbd6a8f35ade21/jina/jaml/parsers/__init__.py#L9). We could have another `SimpleParseableBaseClass` which has a simple parser and works if inherited from.
> 
> But on the other hand, that is exactly, what `JAMLCompatible` is for and thus instead of having a `raise NotImplementedError(f'{cls!r} does not implement YAML parser')` in the above mentioned function, we could implement a default simple parser. That would be my preferred solution, but before implementing this, I wanted to double check if there are some obvious no-gos.



> > I do not understand why current JAML parser does not work?
> 
> Because the current JAML parser works ONLY for classes, which inherit from `BaseDriver`, `BaseFlow` or `BaseExecutor`. See [this function](https://github.com/jina-ai/jina/blob/d7fdc477db9fae6ebefb21e1c8fbd6a8f35ade21/jina/jaml/parsers/__init__.py#L9). We could have another `SimpleParseableBaseClass` which has a simple parser and works if inherited from.
> 
> But on the other hand, that is exactly, what `JAMLCompatible` is for and thus instead of having a `raise NotImplementedError(f'{cls!r} does not implement YAML parser')` in the above mentioned function, we could implement a default simple parser. That would be my preferred solution, but before implementing this, I wanted to double check if there are some obvious no-gos.

What I understand now, is that you do not want a `Parser`?

It seems that this may be a thoughtful design to guarantee that every `JAML` interface is backed by a `CLI` one, so maybe we should go for implementing our `parser`?

maximilianwerk:
Since we just need a trivial parser for `OptunaOptimizer`, `FlowRunner` and `OptimizationParsers` and the parser might look the same for all three (for now) I wondered, whether I should just implement a simple `DefaultParser`. And I think, that is what I will do straight away and then we see if it is a good design or not.

jina/optimizers/flow_runner.py
maximilianwerk: Moved the callback refreshing to optimizer. Now it makes much more sense.

JoanFM:
LGTM!

JoanFM (approved):

PR #1732, opened 2021-01-20
fix(hub): fix label insertion into dockerfile

### LINKED ISSUE #1731: Improper modification of Dockerfile when building for the hub
This is related to issue that appeared [here](https://github.com/jina-ai/jina-hub/pull/2480). From the decoded URL we can see the modifications that jina makes to the Dockerfile:

```
docker://localhost/v1.40/build?t=jinahub/pod.crafter.albumentationscrafter:0.0.1-0.9.18&q=False&nocache=False&rm=True&forcerm=False&pull=True&dockerfile=.dockerfile.85ba69d05f9914e2aef3438da3fd8b99512ce32e&dockerfile=FROM jinaai/jina AS base
LABEL ai.jina.hub.name="AlbumentationsCrafter" \      
ai.jina.hub.description="Apply transformations from the Albumentations package to the image." \      
ai.jina.hub.type="pod" \      
ai.jina.hub.author="Tadej Svetina (@tadejsv)" \      
ai.jina.hub.url="https://jina.ai" \      
ai.jina.hub.documentation="https://github.com/jina-ai/jina-hub" \      
ai.jina.hub.version="0.0.1" \      
ai.jina.hub.vendor="Jina AI Limited" \      
ai.jina.hub.license="apache-2.0" \      
ai.jina.hub.avatar="None" \      
ai.jina.hub.platform="['linux/amd64']" \      
ai.jina.hub.keywords="['image', 'transform', 'augmentation', 'albumentations']" \      
ai.jina.hub.manifest_version="1" \      
ai.jina.hub.kind="crafter" \      
ai.jina.hub.jina_version="0.9.18"
# setup the workspace
COPY . /workspace
WORKDIR /workspace

# install the third-party requirements
RUN pip install -r requirements.txt --no-cache-dir

# for testing the image
FROM base
LABEL ai.jina.hub.name="AlbumentationsCrafter" \      
ai.jina.hub.description="Apply transformations from the Albumentations package to the image." \      
ai.jina.hub.type="pod" \      
ai.jina.hub.author="Tadej Svetina (@tadejsv)" \      
ai.jina.hub.url="https://jina.ai" \      
ai.jina.hub.documentation="https://github.com/jina-ai/jina-hub" \      
ai.jina.hub.version="0.0.1" \      
ai.jina.hub.vendor="Jina AI Limited" \      
ai.jina.hub.license="apache-2.0" \      
ai.jina.hub.avatar="None" \      
ai.jina.hub.platform="['linux/amd64']" \      
ai.jina.hub.keywords="['image', 'transform', 'augmentation', 'albumentations']" \      
ai.jina.hub.manifest_version="1" \      
ai.jina.hub.kind="crafter" \      
ai.jina.hub.jina_version="0.9.18"RUN pip install pytest && pytest

FROM base
LABEL ai.jina.hub.name="AlbumentationsCrafter" \      
ai.jina.hub.description="Apply transformations from the Albumentations package to the image." \      
ai.jina.hub.type="pod" \      
ai.jina.hub.author="Tadej Svetina (@tadejsv)" \      
ai.jina.hub.url="https://jina.ai" \      
ai.jina.hub.documentation="https://github.com/jina-ai/jina-hub" \      
ai.jina.hub.version="0.0.1" \      
ai.jina.hub.vendor="Jina AI Limited" \      
ai.jina.hub.license="apache-2.0" \      
ai.jina.hub.avatar="None" \      
ai.jina.hub.platform="['linux/amd64']" \      
ai.jina.hub.keywords="['image', 'transform', 'augmentation', 'albumentations']" \      
ai.jina.hub.manifest_version="1" \      
ai.jina.hub.kind="crafter" \      
ai.jina.hub.jina_version="0.9.18"ENTRYPOINT ["jina", "pod", "--uses", "config.yml"]
```

Two issues here:
1. **serious**: Jina does not add/perserve line break after adding labels - see this line (`RUN` should start on a new line)
```
ai.jina.hub.jina_version="0.9.18"RUN pip install pytest && pytest
```
2. Labels should only be inserted after the last `FROM` statement, they are not needed elsewhere.

tadejsv:
Fixes #1731 .

@JoanFM , can you come up with a test for this?

### DIFF
--- a/jina/docker/hubio.py
+++ b/jina/docker/hubio.py
@@ -307,13 +307,14 @@ class HubIO:
                     dockerfile_path = self._check_completeness()['Dockerfile']
                     self._freeze_jina_version()
 
+                    labels = {_label_prefix + k: str(v) for k, v in self.manifest.items()}
                     streamer = self._raw_client.build(
                         decode=True,
                         path=self.args.path,
                         tag=self.tag,
                         pull=self.args.pull,
                         dockerfile=dockerfile_path,
-                        labels={k: str(v) for k, v in self.manifest.items()},
+                        labels=labels,
                         rm=True
                     )
 
@@ -341,7 +342,7 @@ class HubIO:
                                                        tag=self.tag,
                                                        pull=self.args.pull,
                                                        dockerfile=dockerfile_path,
-                                                       labels={k: str(v) for k, v in self.manifest.items()},
+                                                       labels=labels,
                                                        rm=True)
 
                 # success
@@ -534,7 +535,7 @@ class HubIO:
                     fp.write('\n'.join(new_requirements))
 
     def _check_completeness(self) -> Dict:
-        self.dockerfile_path = get_exist_path(self.args.path, 'Dockerfile')
+        dockerfile_path = get_exist_path(self.args.path, 'Dockerfile')
         manifest_path = get_exist_path(self.args.path, 'manifest.yml')
         self.config_yaml_path = get_exist_path(self.args.path, 'config.yml')
         self.readme_path = get_exist_path(self.args.path, 'README.md')
@@ -551,7 +552,7 @@ class HubIO:
         test_glob = glob.glob(os.path.join(self.args.path, 'tests/test_*.py'))
 
         completeness = {
-            'Dockerfile': self.dockerfile_path,
+            'Dockerfile': dockerfile_path,
             'manifest.yml': manifest_path,
             'config.yml': self.config_yaml_path,
             'README.md': self.readme_path,
--- a/jina/docker/hubio.py
+++ b/jina/docker/hubio.py
@@ -534,7 +534,7 @@ class HubIO:
                     fp.write('\n'.join(new_requirements))
 
     def _check_completeness(self) -> Dict:
-        dockerfile_path = get_exist_path(self.args.path, 'Dockerfile')
+        self.dockerfile_path = get_exist_path(self.args.path, 'Dockerfile')
         manifest_path = get_exist_path(self.args.path, 'manifest.yml')
         self.config_yaml_path = get_exist_path(self.args.path, 'config.yml')
         self.readme_path = get_exist_path(self.args.path, 'README.md')
@@ -551,7 +551,7 @@ class HubIO:
         test_glob = glob.glob(os.path.join(self.args.path, 'tests/test_*.py'))
 
         completeness = {
-            'Dockerfile': dockerfile_path,
+            'Dockerfile': self.dockerfile_path,
             'manifest.yml': manifest_path,
             'config.yml': self.config_yaml_path,
             'README.md': self.readme_path,
--- a/jina/docker/hubio.py
+++ b/jina/docker/hubio.py
@@ -304,7 +304,7 @@ class HubIO:
 
             with TimeContext(f'building {colored(self.args.path, "green")}', self.logger) as tc:
                 try:
-                    self._check_completeness()
+                    dockerfile_path = self._check_completeness()['Dockerfile']
                     self._freeze_jina_version()
 
                     streamer = self._raw_client.build(
@@ -312,7 +312,8 @@ class HubIO:
                         path=self.args.path,
                         tag=self.tag,
                         pull=self.args.pull,
-                        dockerfile=self.dockerfile_path_revised,
+                        dockerfile=dockerfile_path,
+                        labels={k: str(v) for k, v in self.manifest.items()},
                         rm=True
                     )
 
@@ -339,7 +340,8 @@ class HubIO:
                 image, log = self._client.images.build(path=self.args.path,
                                                        tag=self.tag,
                                                        pull=self.args.pull,
-                                                       dockerfile=self.dockerfile_path_revised,
+                                                       dockerfile=dockerfile_path,
+                                                       labels={k: str(v) for k, v in self.manifest.items()},
                                                        rm=True)
 
                 # success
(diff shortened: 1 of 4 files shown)


JoanFM:
> Fixes #1731 .
> 
> @JoanFM , can you come up with a test for this?

You can add a unittest only for this small function

tests/unit/docker/test_hub_build.py
```diff
@@ -86,3 +86,36 @@ def test_jina_version_freeze_no_jina_dependency(requirements, tmpdir):
...
+revised_dockerfile = """FROM test AS base
+RUN test
+
+FROM base
+LABEL ai.jina.hub.label1="l"
+LABEL ai.jina.hub.label2="l m n"
+RUN test"""
+
+
+def test_get_revised_dockerfile(tmpdir):
```
JoanFM, line 107: can you please add a test with the old standard to prove output is expected?

JoanFM (changes requested):

tests/unit/docker/test_hub_build.py
tadejsv: Old standard?

JoanFM: I mean the one currently on cookiecutter

JoanFM:
Hey @rutujasurve94 , @deepankarm, may you take a look at this PR?

tadejsv:
Change of strategy. It actually makes much more sense to feed labels directly to `label` arg in Docker SDK `build()`, rather than hacking the dockerfile.

This is the way it should have been done from the beginning.

jina/docker/hubio.py
```diff
@@ -532,7 +534,7 @@ def _freeze_jina_version(self) -> None:
                     fp.write('\n'.join(new_requirements))
 
     def _check_completeness(self) -> Dict:
-        dockerfile_path = get_exist_path(self.args.path, 'Dockerfile')
+        self.dockerfile_path = get_exist_path(self.args.path, 'Dockerfile')
```
JoanFM, line 537: please, do not set an attribute at this level.

JoanFM (changes requested):
I agree totally with the approach to pass `labels` to the `build` method of `sdk`. This should be however tested using a `mock`. I think similar strategy to testing `docker_kwargs` could be used

jina/docker/hubio.py
tadejsv: Why not? A few lines bellow a couple of other attributes are set

tadejsv:
@JoanFM what exactly should be tested? That passing a dict with labels does indeed result in an image with labels does not need to be tested, that is docker SDK functionality. The only thing remaining is to test that manifest is correct - but that's out of scope for this PR (if it's not already tested).

jina/docker/hubio.py
JoanFM: Yes but something to be avoided

JoanFM: With best effort attributes should be acquired in init time. Otherwise thr patterns of if hasattr() appear.

JoanFM:
> @JoanFM what exactly should be tested? That passing a dict with labels does indeed result in an image with labels does not need to be tested, that is docker SDK functionality. The only thing remaining is to test that manifest is correct - but that's out of scope for this PR (if it's not already tested).

Good point, maybe test that labels are indeed properly passed with a mock that assrtts how the external service is called

jina/docker/hubio.py
tadejsv: Ok, I agree with you here. 

However, moving all attribute setting to init would require a more thorough redesign of this module (no fewer than 5 other attributes are set in `_check_completeness` function), which I think is out of scope of this PR - I am just addressing one issue within the current framework.

tadejsv:
@JoanFM I am not sure how to do suck a mock test

jina/docker/hubio.py
JoanFM: Okey lets keep it here, I would just avoid not to add yet another one

tadejsv: Understood. Technically I am just replacing setting `revised_dockerfile_path` attr (which appeared originally later in the same function) with `dockerfile_path`, so net 0 of attributes are set :)

jina/docker/hubio.py
```diff
@@ -339,7 +340,8 @@ def build(self) -> Dict:
                 image, log = self._client.images.build(path=self.args.path,
                                                        tag=self.tag,
                                                        pull=self.args.pull,
-                                                       dockerfile=self.dockerfile_path_revised,
+                                                       dockerfile=self.dockerfile_path,
+                                                       labels={k: str(v) for k, v in self.manifest.items()},
```
JoanFM, line 344: is self.manifest not enough?

jina/docker/hubio.py
```diff
@@ -312,7 +312,8 @@ def build(self) -> Dict:
                         path=self.args.path,
                         tag=self.tag,
                         pull=self.args.pull,
-                        dockerfile=self.dockerfile_path_revised,
+                        dockerfile=self.dockerfile_path,
```
JoanFM, line 315: create a local labels if needed before and use the same in both `build` calls

JoanFM (changes requested):

JoanFM:
Hey @tadejsv , inspired by `test_pass_arbitrary_kwargs` inside `tests/unit/peapods/runtimes/container/test_container_runtime.py` you can see how you can try to mock the `client` returned from `docker.from_env` and assert the call parameters

jina/docker/hubio.py
tadejsv: No, everything needs to be serialised into a string

tadejsv:
Ok test added, ready to merge. PS: I think the test was not actually necessary, because the `_check_docker_image` already checks for labels - so as long as there is a test of `push()`, labels are checked as well.

JoanFM (approved):
Thank you very much! Great contribution

PR #1745, opened 2021-01-21
docs: some cleanup

cristianmtr:
Some minor fixes I did while reading a bit through the docs

### DIFF
--- a/docs/chapters/flow/index.md
+++ b/docs/chapters/flow/index.md
@@ -210,7 +210,7 @@ with f:
 ```
 
 - `input_fn` is an `Iterator[bytes]`, each of which corresponds to the representation of a Document with bytes.
-- `output_fn` is the callback function after each request, and takes a `Request` protobuf as its only input.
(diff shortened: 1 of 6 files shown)


JoanFM (approved):

docs/chapters/flow/index.md
```diff
@@ -224,7 +224,7 @@ def input_fn():
 input_fn = (b's' for _ in range(10))
 ```
 
-> Please note that the current Flow API does not support using `index()` and `search()` together in the same `with` scope. This is because the workflow of `index()` and `search()` are usually different and you cannot use one workflow for both tasks.
+> Please note that the current Flow API does not support using `index()` and `search()` together in the same `with` scope. This is because the workflow of `index()` and `search()` are usually different, and you cannot use one workflow for both tasks.
```
deepankarm, line 227: If itis about proper closure of Peas/Pods, it makes sense. If it is about multiple clients under the same scope, that is allowed.

image

cristianmtr: Interesting. However, the only thing i changed was a comma `,` lol

PR #1753, opened 2021-01-21
fix(driver): fix stateless driver dump

hanxiao:


tests/unit/executors/test_dump_executor_with_drivers.py
```diff
@@ -0,0 +1,39 @@
...
+
+def test_dump_excutor_with_drivers(tmpdir):
+    a = BaseExecutor.load_config(f'{cur_dir}/yaml/route.yml')
+    a.touch()
+    a._drivers['ControlRequest'][0].idle_dealer_ids = ('hello', 'there')
+    a.save(str(tmpdir / 'a.bin'))
+
+    print(a._drivers)
+
+    b = BaseExecutor.load(str(tmpdir / 'a.bin'))
```
JoanFM, line 35: but the point is that is should never have `drivers` right?

PR #1759, opened 2021-01-22
refactor: avoid import drivers from types

JoanFM:


JoanFM:
I think we can invert the logic and have a QueryLang created from a Driver instead of passing a Driver to QueryLang.

jina/types/querylang/__init__.py
```diff
@@ -32,31 +31,29 @@ class QueryLang:
             from jina.drivers.querylang.slice import SliceQL
 
             s = SliceQL(start=3, end=4)
```
nan-wang, line 33:
```suggestion
```

jina/types/querylang/__init__.py
```diff
@@ -32,31 +31,29 @@ class QueryLang:
...
 
     A :class:`QueryLang` object (no matter how it is constructed) can be converted to
-    protobuf object or back to driver object by using:
+    protobuf object by using:
 
         .. highlight:: python
         .. code-block:: python
 
             # to protobuf object
             s.as_pb_object
```
nan-wang, line 48:
```suggestion
            ql.as_pb_object
```

nan-wang (changes requested):
I agree upon that the types should not import drivers because conceptually drivers are one level higher than the types. However, I'm not sure using tuples for initializing `QueryLang` is a good solution. For me, it doesn't feel pythonic.

JoanFM:
> I agree upon that the types should not import drivers because conceptually drivers are one level higher than the types. However, I'm not sure using tuples for initializing `QueryLang` is a good solution. For me, it doesn't feel pythonic.

I agree it does not feel the best way, I think what we could do is to have a `namedtuple` and have this creation come from the `Driver`?

Maybe the best solution would be to put `QueryLang` in a different module (in between `types` and `drivers` so that both `Drivers` and `Types` can depend on it. (I think there we could also fit all the `dunderkeys` syntax and so?

JoanFM:
> I agree upon that the types should not import drivers because conceptually drivers are one level higher than the types. However, I'm not sure using tuples for initializing `QueryLang` is a good solution. For me, it doesn't feel pythonic.

Hey @nan-wang can you see if you like the new changes?

hanxiao (approved):

PR #1762, opened 2021-01-22
fix(importer): fix jina check

### LINKED ISSUE #1526: `jina check` is broken
**Describe the bug**

jina check can not detect any broken hub executors even though they are warned during *first* imports.

**Describe how you solve it**

Maybe something related to the import orders in `jina/__init__.py`

---

**Environment**

**Screenshots**

nan-wang:
resolved #1526 
- [x] Add unit tests for mocking the loading errors in `import_module`
- [x] Refactor the `import_classes` to improve the readability
- [x] Remove the unused `targets` argument from `import_classes`.  The only usage of `targets`argument is removed in #1759


jina/importer.py
```diff
@@ -26,62 +92,32 @@ def import_classes(namespace: str, targets=None,
 
     import os, re
```
maximilianwerk, line 93: Unrelated to this PR, but why is this done here? If there is a magical reason, please add a comment about it in the code.

jina/importer.py
```diff
@@ -26,62 +92,32 @@ def import_classes(namespace: str, targets=None,
...
+    _import_type = _namespace2type.get(namespace)
+    if _import_type is None:
         raise TypeError(f'namespace: {namespace} is unrecognized')
 
-    from setuptools import find_packages
-    import pkgutil
-    from pkgutil import iter_modules
+    _targets = _get_targets(targets)
+    if _targets is None:
+        raise TypeError(f'target must be a set, but received {targets!r}')
```
maximilianwerk, line 106: This can never happen, since `_get_targets` will never return `None`. Regardless, I'd rename `_get_targets` into `_parse_targets` and do any error handling there.

jina/importer.py
```diff
@@ -26,62 +92,32 @@ def import_classes(namespace: str, targets=None,
...
-            if (namespace != 'jina.hub' and not info.ispkg) or (namespace == 'jina.hub' and info.ispkg):
-                modules.add('.'.join([namespace, pkg, info.name]))
-
-    # filter
-    ignored_module_pattern = r'\.tests|\.api|\.bump_version'
-    modules = {m for m in modules if not re.findall(ignored_module_pattern, m)}
-
+    modules = _get_modules(namespace)
+    if not modules:
+        return {}
```
maximilianwerk, line 115: Shouldn't that deserve a proper logging/exception throwing? Further down the function, we would throw an Exception, if not all `_targets` are handled. But now we don't. Feels strange, even thought I must admit, I did not totally went into the details of this function.

maximilianwerk (changes requested):
I did not understood the changes in details. Anyhow, I like the refactoring into smaller functions. The `import_classes` functions is still to complex for my taste, but that might be OK. I added some comments on code fragments, that can be removed/should be changed.

jina/importer.py
nan-wang: A warning message is added.

jina/importer.py
nan-wang: As `targets` is no longer used after #1759, I've removed this part.

jina/importer.py
nan-wang: removed.

jina/importer.py
nan-wang: After removing `targets`, the return becomes clearer and there is no longer exception thrown out by the function.

jina/importer.py
```diff
@@ -11,141 +12,58 @@
...
+    if _import_type is None:
         raise TypeError(f'namespace: {namespace} is unrecognized')
 
-    from setuptools import find_packages
-    import pkgutil
-    from pkgutil import iter_modules
+    _imported_property = namespace.split('.')[-1]
+    _is_imported = getattr(IMPORTED, _imported_property)
+    if import_once and _is_imported:
+        warnings.warn(f'{namespace} has already imported. If you want to re-imported, please set `import_once=False`',
```
JoanFM, line 41: `If you want to reimport it` And where should I set `import_once=False`?

nan-wang: a copy-paste error...

hanxiao (approved):

PR #1764, opened 2021-01-23
feat: optimizer cli

florian-hoenicke:


jina/optimizers/__init__.py
```diff
@@ -155,6 +155,13 @@ def optimize_flow(
         return result_processor(study)
 
 
+def run_optimizer(args):
+    from .flow_runner import SingleFlowRunner
+    with open(args.uses) as f:
+        optimizer = JAML.load(f)
+    result = optimizer.optimize_flow()
```
JoanFM, line 162: is the result a OptimizerResultProcessor???

Then in the cli we should call the print method?

tests/integration/optimizers/optimizer_conf.yml
```diff
@@ -0,0 +1,14 @@
+!OptunaOptimizer
+version: 1
+with:
+  flow_runner: !SingleFlowRunner
+    with:
+      flow_yaml: 'tests/integration/optimizers/flow.yml'
```
JoanFM, line 6: doesn't it capture relative paths? we should maybe use `complete_path` maybe we can achieve similar experience as with other `uses`

tests/integration/optimizers/optimizer_conf.yml
```diff
@@ -0,0 +1,14 @@
...
+      overwrite_workspace: True
+      documents: 'tests/integration/optimizers/data.jsonlines'
+      request_size: 1
+      task: 'search_lines'
+  evaluation_callback: !MeanEvaluationCallback {}
+  parameter_yaml: 'tests/integration/optimizers/parameter.yml'
+  workspace_base_dir: './optimizer_workspace'
```
JoanFM, line 13: this should be parametrized to a tmpdir right?

jina/optimizers/__init__.py
```diff
@@ -155,6 +155,13 @@ def optimize_flow(
         return result_processor(study)
 
 
+def run_optimizer(args):
```
JoanFM, line 158: Just a reminder to add better type hinting in this module

jina/optimizers/__init__.py
```diff
@@ -105,22 +106,24 @@ class FlowOptimizer(JAMLCompatible):
...
-        sampler: str = 'TPESampler',
-        direction: str = 'maximize',
-        seed: int = 42,
+            self,
+            flow_runner: 'FlowRunner',
+            parameter_yaml: str,
+            evaluation_callback: 'OptimizerCallback',
+            n_trials: int,
+            workspace_base_dir: str = '',
+            output_file=None,
```
JoanFM, line 115: missing type hint

tests/integration/optimizers/test_optimizer.py
```diff
@@ -16,56 +19,80 @@
...
+def test_yaml(tmpdir, use_output_file):
     jsonlines_file = os.path.join(tmpdir, 'docs.jsonlines')
+    output_file = os.path.join(tmpdir, 'results', 'best_parameters.yml')
     optimizer_yaml = f'''!FlowOptimizer
 version: 1
 with:
   flow_runner: !SingleFlowRunner
     with:
-      flow_yaml: 'tests/integration/optimizers/flow.yml'
+      flow_yaml: {os.path.join('tests', 'integration', 'optimizers', 'flow.yml')}
```
JoanFM, line 85: does it need to be absolute? this will not work from other places than `jina` root right?

florian-hoenicke: yes the test needs to be run from the root

jina/optimizers/__init__.py
florian-hoenicke: thanks, done

jina/optimizers/__init__.py
florian-hoenicke: thanks, done

tests/integration/optimizers/optimizer_conf.yml
florian-hoenicke: was just a draft

tests/integration/optimizers/optimizer_conf.yml
florian-hoenicke: done

jina/optimizers/__init__.py
florian-hoenicke: The results are getting logged already.
Therefore, I think it would be redundant to print them again.
```
[I 2021-01-26 08:59:38,188] A new study created in memory with name: no-name-19a1acf1-c8d6-4493-95b2-1cce2c3961ea
[I 2021-01-26 08:59:39,942] Trial 0 finished with value: 0.0 and parameters: {'JINA_DUMMYCRAFTER_PARAM1': 0, 'JINA_DUMMYCRAFTER_PARAM2': 1, 'JINA_DUMMYCRAFTER_PARAM3': 2}. Best is trial 0 with value: 0.0.
[I 2021-01-26 08:59:41,678] Trial 1 finished with value: 1.0 and parameters: {'JINA_DUMMYCRAFTER_PARAM1': 0, 'JINA_DUMMYCRAFTER_PARAM2': 1, 'JINA_DUMMYCRAFTER_PARAM3': 1}. Best is trial 1 with value: 1.0.
[I 2021-01-26 08:59:43,481] Trial 2 finished with value: 0.0 and parameters: {'JINA_DUMMYCRAFTER_PARAM1': 0, 'JINA_DUMMYCRAFTER_PARAM2': 1, 'JINA_DUMMYCRAFTER_PARAM3': 2}. Best is trial 1 with value: 1.0.
[I 2021-01-26 08:59:45,229] Trial 3 finished with value: 1.0 and parameters: {'JINA_DUMMYCRAFTER_PARAM1': 0, 'JINA_DUMMYCRAFTER_PARAM2': 1, 'JINA_DUMMYCRAFTER_PARAM3': 1}. Best is trial 1 with value: 1.0.
[I 2021-01-26 08:59:46,984] Trial 4 finished with value: 0.0 and parameters: {'JINA_DUMMYCRAFTER_PARAM1': 0, 'JINA_DUMMYCRAFTER_PARAM2': 1, 'JINA_DUMMYCRAFTER_PARAM3': 2}. Best is trial 1 with value: 1.0.
```
Also you can specify the `output_file` in the yaml config to specify the path where the results should be stored.

jina/optimizers/__init__.py
```diff
@@ -112,7 +113,7 @@ def __init__(
             evaluation_callback: 'OptimizerCallback',
             n_trials: int,
             workspace_base_dir: str = '',
-            output_file=None,
+            output_file: str = None,
```
JoanFM, line 116: Optional since it is None

JoanFM (changes requested):

jina/optimizers/__init__.py
florian-hoenicke: thanks, done

tests/integration/optimizers/test_optimizer.py
JoanFM: is there an option to have it `relative` to the `current` file? the same way other `tests` do?

jina/optimizers/__init__.py
```diff
@@ -105,22 +107,24 @@ class FlowOptimizer(JAMLCompatible):
...
+            direction: str = 'maximize',
+            seed: int = 42,
     ):
         """
         :param flow_runner: `FlowRunner` object which contains the flows to be run.
         :param parameter_yaml: yaml container the parameters to be optimized
         :param evaluation_callback: The callback object, which stores the evaluation results
         :param n_trials: evaluation trials to be run
         :param workspace_base_dir: directory in which all temporary created data should be stored
+        :param output_file: directory in which the results are stored
```
cristianmtr, line 127:
```suggestion
        :param output_dir: directory in which the results are stored
```

jina/optimizers/__init__.py
```diff
@@ -105,22 +107,24 @@ class FlowOptimizer(JAMLCompatible):
...
-        sampler: str = 'TPESampler',
-        direction: str = 'maximize',
-        seed: int = 42,
+            self,
+            flow_runner: 'FlowRunner',
+            parameter_yaml: str,
+            evaluation_callback: 'OptimizerCallback',
+            n_trials: int,
+            workspace_base_dir: str = '',
+            output_file: Optional[str] = None,
```
cristianmtr, line 116:
```suggestion
            output_dir: Optional[str] = None,
```

jina/optimizers/__init__.py
florian-hoenicke: nice catch

jina/optimizers/__init__.py
```diff
@@ -105,22 +107,24 @@ class FlowOptimizer(JAMLCompatible):
...
-        workspace_base_dir: str = '',
-        sampler: str = 'TPESampler',
-        direction: str = 'maximize',
-        seed: int = 42,
+            self,
+            flow_runner: 'FlowRunner',
+            parameter_yaml: str,
+            evaluation_callback: 'OptimizerCallback',
+            n_trials: int,
+            workspace_base_dir: str = '',
```
maximilianwerk, line 115: Please don't do edit wars in indentation. I guess that happened due to auto code formatting, but we want to go towards black in the future anyhow, which will do the indentation that was here before.

jina/optimizers/__init__.py
florian-hoenicke: good finding. It has to be `output_file`. The docstring was wrong.

jina/optimizers/__init__.py
```diff
@@ -169,4 +175,19 @@ def optimize_flow(self, **kwargs) -> 'ResultProcessor':
             sampler = getattr(optuna.samplers, self._sampler)(seed=self._seed, **kwargs)
         study = optuna.create_study(direction=self._direction, sampler=sampler)
         study.optimize(self._objective, n_trials=self._n_trials)
-        return ResultProcessor(study)
+
+        result_processor = ResultProcessor(study)
+        if self._output_file:
+            result_processor.save_parameters(self._output_file)
+        return result_processor
```
maximilianwerk, line 182: I would prefer doing this outside as it was before. The reason is separation of concerns. The optimizer should optimize and the outside world should decide what it wants to do with the results. Putting both in here make code harder to maintain in the long run, since we make the optimizer overly complex.

jina/optimizers/__init__.py
```diff
@@ -169,4 +175,19 @@ def optimize_flow(self, **kwargs) -> 'ResultProcessor':
...
+
+def run_optimizer_cli(args: 'Namespace'):
+    """Used to run the optimizer from command line interface.
+
+    :param args: arguments passed via cli
+    """
+    from .flow_runner import SingleFlowRunner
+    with open(args.uses) as f:
+        optimizer = JAML.load(f)
+    optimizer.optimize_flow()
```
maximilianwerk, line 193: I'd rather have the output path in the `args` instead of the .yml and make the result saving here explicit, if the output path is given.

tests/integration/optimizers/test_optimizer.py
```diff
@@ -16,56 +19,80 @@
...
         groundtruth_doc = Document(content='hello')
         yield doc, groundtruth_doc
 
 
-def test_optimizer(tmpdir):
+@pytest.mark.parametrize('use_output_file', (True, False))
+def test_optimizer(tmpdir, use_output_file):
     eval_flow_runner = SingleFlowRunner(
-        flow_yaml='tests/integration/optimizers/flow.yml',
+        flow_yaml=os.path.join('tests', 'integration', 'optimizers', 'flow.yml'),
```
maximilianwerk, line 53: This is done for our Windows users, right? I find it seriously ugly and waaaay less readable. Do we care in the rest of the code base about it? I thought no.

maximilianwerk (changes requested):
I would definitely separate optimization and output handling. Otherwise, I like it a lot.

tests/integration/optimizers/flow.yml
```diff
@@ -5,5 +5,5 @@ env:
   JINA_DUMMYCRAFTER_PARAM2_VAR: ${{JINA_DUMMYCRAFTER_PARAM2}}
   JINA_DUMMYCRAFTER_PARAM3_VAR: ${{JINA_DUMMYCRAFTER_PARAM3}}
 pods:
-  - uses: pods/craft.yml
-  - uses: pods/evaluate.yml
+  - uses: $JINA_OPTIMIZER_CRAFTER_FILE
+  - uses: $JINA_OPTIMIZER_EVALUATOR_FILE
```
maximilianwerk, line 9: We don't care about this anywhere else in the codebase, do we? That makes our whole tests sooo much more ugly. If this is the sacrifice for supporting windows, I'd rather not support it @JoanFM

florian-hoenicke: yes this is not so frequently used at other places. I can change it back to have it consistent with the rest

jina/optimizers/__init__.py
JoanFM: I agree

tests/integration/optimizers/test_optimizer.py
florian-hoenicke: I know that it is in general good practice to use os.path.join to be agnostic to the platform. I started a discussion on slack.

tests/integration/optimizers/flow.yml
maximilianwerk: And in all our examples we never used environment variables for defining pods. I see your problem with the backslash, but the solution shouldn't be environment variables, but rather a during parsing, that the path resolution in ignorant to the operating system.

jina/optimizers/__init__.py
maximilianwerk: I saw, that quite some people started editing indentation at hanging parameters in function definitions to go 8 spaces deep instead of 4. So you can keep it, even thought I find it ugly.

tests/integration/optimizers/test_optimizer.py
JoanFM: we care almost everywhere I think, at least we used to use `os.path.join` everywhere

jina/jaml/parsers/default/v1.py
```diff
@@ -13,6 +13,7 @@ def parse(self, cls: Type['JAMLCompatible'], data: Dict) -> 'JAMLCompatible':
         :param cls: target class type to parse into, must be a :class:`JAMLCompatible` type
         :param data: flow yaml file loaded as python dict
         """
+        data = JAML.expand_dict(data, None)
```
JoanFM, line 16: can we use a different variable to avoid updating variables in place?

jina/parsers/optimizer.py
```diff
@@ -0,0 +1,11 @@
+from .base import set_base_parser
+from .peapods.base import mixin_base_ppr_parser
+
+
+def set_optimizer_parser(parser=None):
```
JoanFM, line 5: type hinting

jina/parsers/optimizer.py
```diff
@@ -0,0 +1,11 @@
...
+    if not parser:
+        parser = set_base_parser()
+    mixin_base_ppr_parser(parser)
+    parser.add_argument('--uses', type=str, help='path to a yaml file which defines an optimizer')
+    parser.add_argument('--output_file', type=str, help='path to a yaml file to store the output')
```
JoanFM, line 10: `--output-file`, then it transforms it to `args.output_file`. But this way is aligned with style of other arguments

JoanFM (dismissed):

jina/parsers/optimizer.py
florian-hoenicke: thanks!

jina/optimizers/__init__.py
florian-hoenicke: You mean defining the directory instead of the file? This would mean the user can not define the file name right?

JoanFM: I think is about separating clearly optimization and output results processing.

tests/integration/optimizers/flow.yml
florian-hoenicke: True. If we solve this in a centralised way, we don't have to care about this anymore.

jina/parsers/optimizer.py
florian-hoenicke: thanks, done

jina/jaml/parsers/default/v1.py
florian-hoenicke: done

tests/integration/optimizers/test_optimizer.py
florian-hoenicke: I changed it back for now

jina/optimizers/__init__.py
florian-hoenicke: done

jina/optimizers/__init__.py
florian-hoenicke: Yes! We have to do something against edit wars! I use the pep8 formatter. We could also all agree on using pep8. So we won't have different formatting.

jina/parsers/optimizer.py
```diff
@@ -1,12 +1,14 @@
...
+
+def set_optimizer_parser(parser: 'ArgumentParser' = None):
     if not parser:
         parser = set_base_parser()
-
     mixin_base_ppr_parser(parser)
-    parser.add_argument('--uses', type=str, help='a yaml file represents a optimizer')
-
+    parser.add_argument('--uses', type=str, help='path to a yaml file which defines an optimizer')
+    parser.add_argument('--output-dir', type=str, help='path to a yaml file to store the output')
```
maximilianwerk, line 13: I believe Han just reformatted all argument parsing help string towards upper-casing at the beginning of the help, sentencing it and starting the sentence with a "The"

maximilianwerk (approved):

PR #1774, opened 2021-01-25
test: logging level

Yongxuanzhang:
Test the logging level with os variable 'JINA_LOG_LEVEL' setting.
Related to https://github.com/jina-ai/jina/pull/1744.

### DIFF
--- a/tests/unit/logging/test_logging.py
+++ b/tests/unit/logging/test_logging.py
@@ -38,6 +38,23 @@ def test_logging_default():
             assert len(logger.handlers) == 2
 
 
+def test_logging_level():
+    with JinaLogger('test_logger', log_config=os.path.join(cur_dir, 'yaml/file.yml')) as logger:
+        log(logger)
+        assert logger.logger.level == LogVerbosity.from_string('INFO')
+
+    os.environ['JINA_LOG_LEVEL'] = 'SUCCESS'
+    with JinaLogger('test_logger', log_config=os.path.join(cur_dir, 'yaml/file.yml')) as logger:
+        log(logger)
+        assert logger.logger.level == LogVerbosity.from_string(os.environ['JINA_LOG_LEVEL'])
+
+    del os.environ['JINA_LOG_LEVEL']
+
+    with JinaLogger('test_logger', log_config=os.path.join(cur_dir, 'yaml/file.yml')) as logger:
+        log(logger)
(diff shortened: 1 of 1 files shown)


tests/unit/logging/test_logging.py
```diff
@@ -37,6 +38,23 @@ def test_logging_default():
             assert len(logger.handlers) == 2
 
 
+def test_logging_level():
+    with JinaLogger('test_logger', log_config=os.path.join(cur_dir, 'yaml/file.yml')) as logger:
+        log(logger)
+        assert logger.logger.level == LogVerbosity.from_string('INFO')
+
+    os.environ['JINA_LOG_LEVEL'] = 'SUCCESS'
```
JoanFM, line 46: add a try catch mechanism that ensures thr os env is properly deleted but at same time captures the assertion error

JoanFM (changes requested):

tests/unit/logging/test_logging.py
JoanFM: or split the test so that u can use fixtures to guarantee this

hanxiao (changes requested):
this is not a valid test, as in your test assertion you write the same code as in the `logger`

```python
        if 'JINA_LOG_LEVEL' in os.environ:
            verbose_level = LogVerbosity.from_string(os.environ['JINA_LOG_LEVEL'])
        self.logger.setLevel(verbose_level.value)
```

This is the same as having the code `a=1` and then test it via `assert a==1`, it is almost meaningless in most cases.

The correct way is to test the emission of the logger, and see if it is suppressed given the current loglevel.

tests/unit/logging/test_logging.py
Yongxuanzhang: Split the test and use fixtures in the latest commit. Thanks!

Yongxuanzhang:
> this is not a valid test, as in your test assertion you write the same code as in the `logger`
> 
> ```python
>         if 'JINA_LOG_LEVEL' in os.environ:
>             verbose_level = LogVerbosity.from_string(os.environ['JINA_LOG_LEVEL'])
>         self.logger.setLevel(verbose_level.value)
> ```
> 
> This is the same as having the code `a=1` and then test it via `assert a==1`, it is almost meaningless in most cases.
> 
> The correct way is to test the emission of the logger, and see if it is suppressed given the current loglevel.

Revised in the latest commit.

PR #1778, opened 2021-01-25
feat: synced arguments of index-lines and search-lines

maximilianwerk:
The functions `index_lines` and `search_lines` have a different order of the same parameters. This is pretty confusing. Furthermore, the `FlowRunner` supports now all `Flow` entry functions.

In case unit tests fail, I'll fix them. Beware @JoanFM this might also impact the examples.


jina/optimizers/flow_runner.py
```diff
@@ -22,48 +22,66 @@ def run(
...
 
 class SingleFlowRunner(FlowRunner):
-    """Module to define and run a flow."""
+    """Module to define and run a flow.
+       `documents` maps to a parameter of the `execution_method`, depending on the method.
+       To understand to which parameter `documents` is mapped see `EXECUTION_METHOD_PARAMETER_MAPPINGS`.
+       Anyhow, the default values can be overriden by the `**kwargs` argument of the `run` method.
+    """
+
+    EXECUTION_METHOD_PARAMETER_MAPPINGS = {
```
JoanFM, line 37: I do not like this much? can't we rely on simply having `input_generators`? I think is not a good thing to `depend` on sugary IO

maximilianwerk: If we want to support the `.jsonlines` construct we would need to write a jsonline reader outside once again, even thought `Flow` already support it. I would rather leverage all the available methods on `Flow` level instead of mirroring them here once again.

maximilianwerk: Anyhow, I agree, that this is super ugly.

jina/optimizers/__init__.py
```diff
@@ -140,14 +141,57 @@ def _trial_parameter_sampler(self, trial):
...
-            )
+            trial_parameters[param.jaml_variable] = FlowOptimizer._suggest(param, trial)
 
         trial.workspace = self._workspace_base_dir + '/JINA_WORKSPACE_' + '_'.join([str(v) for v in trial_parameters.values()])
 
         return trial_parameters
 
+    @staticmethod
+    def _suggest(param, trial):
+
```
florian-hoenicke, line 152: what do you think about having something linke this in order to get rid of the redundant code?:
```
class_to_method = {
    IntegerParameter: trial.suggest_int,
    FloatParameter: trial.suggest_float,
    ...
}
return class_to_method[param.class](**param)
```

jina/optimizers/flow_runner.py
maximilianwerk: The solution now puts the user in charge. I agree with you, that the whole data parsing needs a refactoring at some point in time to put it in a dedicated place.

jina/optimizers/__init__.py
maximilianwerk: This will be a bit more subtle. We have to also do the parameter mapping. and `**param` does not work, as long as the parameter is now dict. That would result in the same code as before with the `to_optuna_args` function, non?

florian-hoenicke: We could also use the `vars` build in like: `**vars(param)` 
As a down site other unwanted parameters are transferred as well and `name` is not defined. However, with some small refactoring on the parameters it might be possible.
```
vars(IntegerParameter(1,2,3,True)) == {
    'parameter_name': '', 
    'jaml_variable': 'JINA_NONE_', 
    'low': 1, 'high': 2, 'step_size': 3, 
    'log': True, 
    'optuna_method': 'suggest_int'
}
```

I just wanted to put this solution for discussion. There is no request for change from my site.

florian-hoenicke (approved):

jina/optimizers/parameters.py
```diff
@@ -35,16 +35,6 @@ def __init__(
...
-        return {
-            'name': self.jaml_variable,
-            'low': self.low,
-            'high': self.high,
-            'step': self.step_size,
-            'log': self.log,
-        }
 
 
 class FloatParameter(OptimizationParameter):
```
JoanFM, line 40: just to know, we are not going to do the separation of "parameters" and "sampling" before 1.0 right?

jina/optimizers/__init__.py
maximilianwerk: Hmm I prefer stupid easily readable code above elegant non-readable code for such kind of parsing. There is no complexity in the `ifelse`, which would be circumvented when doing anything more elaborated.

jina/optimizers/parameters.py
maximilianwerk: Was not planned. There is a reason, optuna keeps them together: Even if we separate them, they will still be glued together, since there is no point having a `Uniform` sampler for categorical parameters. There it would be a `DiscreteUniform` one, which optuna exactly does.

PR #1789, opened 2021-01-26
docs: added docstring for paramters

maximilianwerk:
Removed `FloatParameter` since `suggest_float` in `optuna` is anyhow just a wrapper around the three float sampling algorithms and I'd rather not forward this "hack".

Furthermore, added docstrings.

### DIFF
--- a/jina/optimizers/parameters.py
+++ b/jina/optimizers/parameters.py
@@ -4,6 +4,8 @@ from ..jaml import JAML, JAMLCompatible
 
 
 class OptimizationParameter(JAMLCompatible):
+    """Base class for all optimization parameters."""
+
     def __init__(
         self,
         parameter_name: str = "",
@@ -19,6 +21,11 @@ class OptimizationParameter(JAMLCompatible):
 
 
 class IntegerParameter(OptimizationParameter):
+    """Used for optimizing integer parameters with the Jina Optimizer.
+       For a detailed information about sampling and usage see
+       https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial.suggest_int
+    """
+
     def __init__(
         self,
         low: int,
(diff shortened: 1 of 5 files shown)


jina/optimizers/parameters.py
```diff
@@ -37,38 +44,36 @@ def __init__(
         self.log = log
```
JoanFM, line 44: Why do I need to use `log` when optimizing `IntegerParameters`?

jina/optimizers/parameters.py
```diff
@@ -37,38 +44,36 @@ def __init__(
...
-        **kwargs,
-    ):
-        super().__init__(*args, **kwargs)
-        self.low = low
-        self.high = high
-        self.step_size = step_size
-        self.log = log
-
-
 class UniformParameter(OptimizationParameter):
```
JoanFM, line 47: why can't we use `uniform` sampling also for `Integers`?

maximilianwerk: In some sense, it is what the normal integer sampling does. But you don't call it uniform sampling, but rather choosing a value with a certain step size. I would not try to invent the wheel a second time here, but go with what the people that seriously thought about hyper parameter optimization decided. We can change, if we see it does not fit well for us and we see user requests to change it. But for now, I wouldn't deem us smarter then the optuna folks.

jina/optimizers/parameters.py
maximilianwerk: If you want a logarithmic scale instead of a uniform sampling via step-size. And here you got me: We are not consistent about how we create parameters and sampling.

jina/optimizers/parameters.py
```diff
@@ -19,6 +21,11 @@ def __init__(
 
 
 class IntegerParameter(OptimizationParameter):
+    """Used for optimizing integer parameters with the Jina Optimizer.
+       For a detailed information about sampling and usage see
```
florian-hoenicke, line 25: information is always plural.
```suggestion
       For detailed information about sampling and usage see
```

jina/optimizers/parameters.py
```diff
@@ -37,38 +44,36 @@ def __init__(
...
-        super().__init__(*args, **kwargs)
-        self.low = low
-        self.high = high
-        self.step_size = step_size
-        self.log = log
-
-
 class UniformParameter(OptimizationParameter):
+    """Used for optimizing float parameters with the Jina Optimizer with uniform sampling.
+       For a detailed information about sampling and usage see
```
florian-hoenicke, line 49: same for the other strings

nan-wang (approved):
LGTM👍

PR #1793, opened 2021-01-26
fix(daemon): fix pydantic model on List[str]

hanxiao:


daemon/models/custom.py
```diff
@@ -59,6 +59,8 @@ class _PydanticConfig(BaseConfig):
 
 
 def build_pydantic_model(model_name: str, module: str):
+    from jina.parsers import helper
+    helper._SHOW_ALL_ARGS = True
```
JoanFM, line 63: isn't this change too global? I think it should be handled via an `argument` in the parsers or removed.

JoanFM (changes requested):

PR #1800, opened 2021-01-27
test: optimizer multiflow

florian-hoenicke:


JoanFM (commented):
Could it also be added with `docker` Pods?

tests/integration/optimizers/test_optimizer.py
```diff
@@ -58,14 +58,86 @@ def test_optimizer(tmpdir, config):
...
+            flow_yaml='tests/integration/optimizers/flow1.yml',
+            documents=document_generator(10),
+            request_size=1,
+            execution_method='search',
+        ),
+        SingleFlowRunner(
+            flow_yaml='tests/integration/optimizers/flow2.yml',
+            documents=document_generator(10),
+            request_size=1,
+            execution_method='search',
```
maximilianwerk, line 73: I would put the second flow first and remove the evaluator from the second one. This demonstrates, that we only can leverage one evaluator. 

And we could also achieve this by just setting `execution_method='index'` for the first flow. Then we don't even need two flow files.

tests/integration/optimizers/test_optimizer.py
```diff
@@ -58,14 +58,86 @@ def test_optimizer(tmpdir, config):
...
+with:
+  flow_runner: !MultiFlowRunner
+    with:
+      flows: 
+        - !SingleFlowRunner
+          with:
+            flow_yaml: 'tests/integration/optimizers/flow1.yml'
+            documents: {jsonlines_file}
+            request_size: 1
+            execution_method: 'search_lines'
```
maximilianwerk, line 100: if this would be `index_lines`, it would already do what I suggest above.

maximilianwerk (changes requested):
I believe we can do the test with one flow file and once call index and once search. This also leverages our index/search setup.

tests/integration/optimizers/test_optimizer.py
florian-hoenicke: nice idea! I changed it accordingly

tests/integration/optimizers/test_optimizer.py
florian-hoenicke: changed, thanks

tests/integration/optimizers/test_optimizer.py
```diff
@@ -60,18 +60,20 @@ def test_optimizer_single_flow(tmpdir, config):
...
-        ),
-        SingleFlowRunner(
-            flow_yaml='tests/integration/optimizers/flow2.yml',
-            documents=document_generator(10),
-            request_size=1,
-            execution_method='search',
-        )
+        [
+            SingleFlowRunner(
+                flow_yaml='tests/integration/optimizers/flow.yml',
```
JoanFM, line 65: why not making it independent from where the tests is run. this way is easier to move tests around eventually

maximilianwerk: That is a legacy from me in the optimizer integration tests. Shame on me. I'll fix it.

maximilianwerk (approved):
LGTM

PR #1804, opened 2021-01-27
fix: add warning

JoanFM:


JoanFM:
I will change a test to make it more robust to sharding in remote

jina/parsers/helper.py
```diff
@@ -1,8 +1,11 @@
 import argparse
 import os
 import uuid
+from jina.logging import default_logger
```
hanxiao, line 4: move this inside `if`

JoanFM: Done!

theUnkownName (approved):

Issue #1543, opened 2020-12-26
async input_fn


ace-kay-law-neo:
It would be nice if the input_fn attribute could be an async method (or generator). This is especially useful if the data is retrieved via an async library (like motor for mongodb).
```
async def input_fn():
    cursor = db.documents.find({ })
    async for entry in cursor:
        with Document() as d:
            d.text = entry["text"]
        yield d

flow.index(input_fn=input_fn)
```

janandreschweiger:
I agree with that @ace-kay-law-neo. This feature would be helpful.

Nevertheless, I think you could iterate over the cursor syncronously with cursor.delegate.

ace-kay-law-neo:
Thank you very much @janandreschweiger!

ace-kay-law-neo:
I reopened this issue, because we now use purely asynchronous libraries within the input_fn.

hanxiao:
make sense.

hanxiao:
Hi all, I'm working on this feature, just to clarify some thoughts here.

The high-level interface is fine, but on the lower-level, we use `batch_iterator` for request generation, which fills a number of documents to each request. Hence,
- from this batch to the next batch, they are "sequential & sync".
- inside each batch, the document filing is "async"
Is that what you pictured?

hanxiao:
@ace-kay-law-neo @janandreschweiger fyi this feature is added and rolled out in `v0.9.23`. Example: https://github.com/jina-ai/jina#asynchronous-input

Issue #1540, opened 2020-12-25
Return results for asyncFlow.search


janandreschweiger:
Hey Jina team,

we've added you framework to our fastapi server. Unfortunately, one can only access the search-results via the on_done callback.
Therefore, it would be great if the results of the asyncFlow.search method would be returned directly.

This example illustrates our dilemma:
```
@router.post('/search')
def search(query: str):
    res = await asyncFlow.search_lines([query]) # should return search-result
    return res
```

hanxiao:
i see, should be easy to fix.

janandreschweiger:
Thank you @hanxiao!

hanxiao:
@janandreschweiger please note that #1816 reverts #1541

PR #1816, opened 2021-01-29
feat(async): AsyncClient & AsyncFlow's return is async generator

hanxiao:
This is a breaking change to AsyncFlow API #1606 

- as a continuation of #1543 , this PR introduces return of an AsyncFlow as async generator (PEP525);
- now AsyncFlow are "fully async" in the sense that both `input_fn` and return are now async, (the flow itself is intrinsically async as well)
- note, this reverts the #1540, as `return` can not be used inside an `async generator`. Also keeping all results in a list is pretty trivial and users can do it themselves by looping over the async generator.


JoanFM:
This should be called `on_done`?

PR #1817, opened 2021-01-29
feat: check yaml definition matches instanciated class in LegacyParser

davidbp:
Currently when a jina app is executed it might read from a YAML file that specifies an encoder. If an argument from the encoder is misspelled (or does not exist) the code crashes but no error or Warning is presented to the user.

### DIFF
--- a/jina/jaml/parsers/executor/legacy.py
+++ b/jina/jaml/parsers/executor/legacy.py
@@ -161,5 +161,3 @@ class LegacyParser(VersionedYAMLParser):
         if hasattr(data, 'components'):
             r['components'] = data.components
         return r
-
-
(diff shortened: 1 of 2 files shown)


jina/jaml/parsers/executor/legacy.py
```diff
@@ -115,3 +161,5 @@ def dump(self, data: 'BaseExecutor') -> Dict:
         if hasattr(data, 'components'):
             r['components'] = data.components
         return r
+
```
JoanFM, line 164: can u remove these 2 spaces?

JoanFM (changes requested):

JoanFM (approved):

PR #1821, opened 2021-01-30
feat: flow delete via id

florian-hoenicke:


jina/clients/request/helper.py
```diff
@@ -15,6 +15,8 @@ def _build_doc_from_content():
         if isinstance(data, Document):
             # if incoming is already primitive type Document, then all good, best practice!
             return data, DataInputType.DOCUMENT
+        elif request_type == RequestType.DELETE:
```
JoanFM, line 18: I think having data directly passed to Document and have the client of this function care about the format of it is a cleaner solution?

JoanFM: also the signature of this method claims to build a document from data irrespective of the request type, so maybe should be someone else's concern to properly format it?

florian-hoenicke:
I close this pr because there is a new one:
https://github.com/jina-ai/jina/pull/1823

Issue #1777, opened 2021-01-25
Hub builds broken


tadej-redstone:
It seems that hub builds are broken (see [here](https://github.com/jina-ai/jina-hub/pull/2773) for example), and the error seems to be related to #1732. I suspect that there's some problem with label encoding to strings (some special characters such as `). But this needs to be checked - unfortunately with the current error reporting the whole stack is not printed.

Broadly, this shows two more issues:
- testing needs to be updated, because neither unit nor integration PR caught this
- error reporting needs to be improved, because without seeing the whole stack it is not possible to pinpoint the issue in some cases.

JoanFM:
hey @tadej-redstone , let's first revert #1732 and then try to see how the problem can be better solved?

tadej-redstone:
I think first it should be checked what the problem is. You can go ahead and revert, but the final solution will still probably involve the same design (passing labels to docker sdk)

JoanFM:
> I think first it should be checked what the problem is. You can go ahead and revert, but the final solution will still probably involve the same design (passing labels to docker sdk)

totally agree

tadej-redstone:
Experimenting locally I think I found the problem and solution: `dockerfile_path` should simply equal `Dockerfile`. But it would still be great to have tests that catch this.

PR #1826, opened 2021-02-01
feat: add support to multistage dockerfile 1825

### LINKED ISSUE #1825: Enable the `jina hub build` to use Dockerfile with multi-stage building
**Describe the feature**
Enable the `jina hub build` to use Dockerfile with multi-stage building

**Your proposal**

---

**Environment**

**Screenshots**

nan-wang:
resolved #1825

- Raise `DockerLoginFailed` error instead of swallowing exceptions when login failed
- Fix a bug in `_image_version_exists`
- Unify the field name to `jina-version` in both `manifest.yml` and [https://hubapi.jina.ai/images](https://hubapi.jina.ai/images). 
- Update the sample manifest in `tests/unit/docker/test_hubapi.py`


tadej-redstone:
Hi @nan-wang , I believe it would be better to pass labels using the `label` arg to docker SDK, as was attempted in #1732 , and to avoid the modification of the dockerfile.

The only issue was #1777. I believe I discovered the problem there (see my last comment), however the deeper problem is that the test didn't catch anything.

jina/docker/hubio.py
```diff
@@ -627,12 +627,14 @@ def _get_revised_dockerfile(self, dockerfile_path: str, manifest: Dict) -> str:
         # modify dockerfile
         revised_dockerfile = []
         with open(dockerfile_path) as fp:
-            for l in fp:
+            _last_from_ln = None
+            for _ln, l in enumerate(fp):
```
Yongxuanzhang, line 631: Just curious about what does _ln and l stand for(like lines, labels?) here?

nan-wang: i used it for `line_numbers` :)

nan-wang:
The `build()` function in `HubIO` class is quite complicated. I'll try to do refactoring on it and reduce the complexity a bit.

nan-wang:
Due to the conflicts with the hub bookkeeping, I will withdraw this PR for the time being.

PR #1851, opened 2021-02-03
fix: default value of workspace dir

maximilianwerk:
Better to take current local as default instead of `/` implicitly


Yongxuanzhang (approved):
LGTM 👍

PR #1853, opened 2021-02-03
refactor: cache runtime optimization

florian-hoenicke:


jina/executors/indexers/cache.py
```diff
@@ -67,19 +66,18 @@ def __init__(self, index_filename: Optional[str] = None, field: Optional[str] =
...
-
         # optimization. don't duplicate ids
         if self.field != ID_KEY:
             data = kwargs.get(DATA_FIELD, None)
-            if data is None:
-                raise ValueError(f'Got None from CacheDriver')
-            self.query_handler.content_hash.append(data)
+        else:
+            data = doc_id
+        self.query_handler.id_to_cache_val[doc_id] = data
```
cristianmtr, line 79: So with this change we always store both the `id`s and the `content_hash`s, right? We are no longer optimizing the case of caching by `id`, so that we only store the `id`s and not the latter?

cristianmtr: Right, so now they're being duplicated. 

Can we make it so that we don't?

tests/unit/drivers/test_cache_driver.py
```diff
@@ -76,8 +76,8 @@ def test_cache_driver_from_file(tmpdir, test_metas):
     folder = os.path.join(test_metas["workspace"])
     bin_full_path = os.path.join(folder, filename)
     docs = list(random_docs(10, embedding=False))
-    pickle.dump([doc.id for doc in docs], open(f'{bin_full_path}.ids', 'wb'))
-    pickle.dump([doc.content_hash for doc in docs], open(f'{bin_full_path}.cache', 'wb'))
+    pickle.dump({doc.id: i for i, doc in enumerate(docs)}, open(f'{bin_full_path}.ids', 'wb'))
+    pickle.dump({doc.content_hash: i for i, doc in enumerate(docs)}, open(f'{bin_full_path}.cache', 'wb'))
```
cristianmtr, line 80: Should this be `i`? I think it should be the field by which we cache, right?

cristianmtr: Cause here you have

```
self.query_handler.id_to_cache_val[doc_id] = data
```

cristianmtr (changes requested):
some questions

jina/executors/indexers/cache.py
florian-hoenicke: Nice thoughts about memory efficiency.
In our case, we don't duplicate the `doc_id` just the reference (`id(doc_id) == id(data)`).
`sys.getsizeof(doc_id) == 65` and the reference is just 8 bytes on 64bit systems.
We wouldn't save much memory and have to think if we would like to trade 8 bytes per document against more complexity. I think it would not be worth it for the cache.

tests/unit/drivers/test_cache_driver.py
florian-hoenicke: Nice catch! 
That was confusing. I changed it.
The test was passing because it just checks if it can load the files and not if the values make sense.

jina/executors/indexers/cache.py
cristianmtr: Ok, fair enough. But then the code and comments should be cleaned / removed / updated

jina/executors/indexers/cache.py
```diff
@@ -92,39 +90,34 @@ def query(self, data, *args, **kwargs) -> Optional[bool]:
...
+        return data in self.query_handler.cache_val_to_id
 
-        return status
 
-    def update(self, keys: Iterable['UniqueId'], values: Iterable[any], *args, **kwargs):
+    def update(self, keys: Iterable[str], values: Iterable[any], *args, **kwargs):
         """Update cached documents.
         :param keys: list of Document.id
         :param values: list of either `id` or `content_hash` of :class:`Document`"""
         # if we don't cache anything else, no need
```
cristianmtr, line 92: here

jina/executors/indexers/cache.py
```diff
@@ -67,19 +66,18 @@ def __init__(self, index_filename: Optional[str] = None, field: Optional[str] =
...
 
-    def add(self, doc_id: 'UniqueId', *args, **kwargs):
+    def add(self, doc_id: str, *args, **kwargs):
         """Add a document to the cache depending on `self.field`.
 
         :param doc_id: document id to be added
         """
-        self.query_handler.ids.append(doc_id)
-
         # optimization. don't duplicate ids
```
cristianmtr: here

cristianmtr (changes requested):

jina/executors/indexers/cache.py
florian-hoenicke: done

jina/executors/indexers/cache.py
florian-hoenicke: That one is still valid.
If we only cache the id, the update should not do anything.
The content is getting updated but the id will stay the same.

cristianmtr (approved):

Issue #1705, opened 2021-01-15
hello-world result is empty 


hanxiao:
**Describe the bug**

1. no image is showed @cristianmtr 
2. Prec/Recall are all zero @JoanFM (probably due to refactoring on the evaluator)

![image](https://user-images.githubusercontent.com/2041322/104789658-fb223a00-5795-11eb-8d93-eace2bf446ab.png)

**Describe how you solve it**

Recent PR on KVIndexer may be the reason

i've tested the following two commits on master and i can confirm the no image is due to `merge=False`.

![image](https://user-images.githubusercontent.com/2041322/104790130-f90cab00-5796-11eb-967a-63992f79a286.png)

When fixining it, please add a test assert to cover (1) (2) so that such implict break wont happen again.

---

**Environment**

**Screenshots**

hanxiao:
im reverting #1701 for now. Note this is a tri-partite problem. 
- `is_merge` makes the image lost; 
- refactoring on evaluator makes PR/Recall all-zero; 
- no test/assert can detect those two.

cristianmtr:
Why revert my fix? Just make `is_merge=true` for this example. By default, `is_merge` should be `false`

JoanFM:
> Why revert my fix? Just make `is_merge=true` for this example. By default, `is_merge` should be `false`

Agree, lets get it back in

PR #1855, opened 2021-02-04
fix: is_merge should be False by default

cristianmtr:
As agreed, `is_merge` should be `False` by default. See https://github.com/jina-ai/jina/issues/1705#issuecomment-762079812

### DIFF
--- a/jina/drivers/search.py
+++ b/jina/drivers/search.py
@@ -49,7 +49,7 @@ class KVSearchDriver(BaseSearchDriver):
             - K is the top-k
     """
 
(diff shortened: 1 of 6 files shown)


JoanFM (changes requested):
make sure u change the parameter for the `helloworld` example, since this broke it

cristianmtr:
> make sure u change the parameter for the `helloworld` example, since this broke it

Ran it locally with the newest changes, and it works

JoanFM (approved):

PR #1865, opened 2021-02-04
fix(grpc): fix zmqlet lifetime in grpcRuntime

hanxiao:


jina/peapods/runtimes/asyncio/grpc/__init__.py
```diff
@@ -28,3 +30,4 @@ async def async_cancel(self):
 
     async def async_run_forever(self):
         await self.server.wait_for_termination()
```
JoanFM, line 32: should there be a `try catch`?

hanxiao: no 
image
https://docs.jina.ai/api/jina.peapods.runtimes.base.html?highlight=baseruntime#jina.peapods.runtimes.base.BaseRuntime

PR #1872, opened 2021-02-04
refactor: remove unique id

florian-hoenicke:


jina/clients/request/helper.py
```diff
@@ -67,4 +67,5 @@ def _add_docs_groundtruths(req, batch, data_type, _kwargs):
 
 
 def _add_ids(req, batch):
-    req.ids.extend(batch)
+    string_ids = (str(doc_id) for doc_id in batch)
```
JoanFM, line 70: why is this needed? id is supposed to be already a str?

florian-hoenicke: I think it can be an integer. Or where is it converted before?
Let me check if this is needed.

florian-hoenicke: Okay, I checked it.
If I run
```index_flow.delete(input_fn=range(0,3))```
then I get the ids as integers. So they are not converted.
image

Comment (JoanFM) [reply jina/clients/request/helper.py:70]: ok so are we allowing that right now?

File: jina/drivers/rank/__init__.py [hunk 2]
```diff
@@ -66,7 +65,7 @@ def _sort_matches_in_place(self, context_doc: 'Document', match_scores: 'np.ndar
         cm = context_doc.matches
         cm.build()
         for str_match_id, score in match_scores:
-            match_id = UniqueId(str_match_id)
+            match_id = str_match_id
```
JoanFM, line 68: iterate directly on match id

JoanFM (changes requested):

jina/clients/request/helper.py
florian-hoenicke: Yes, we allow to delete by id.
In addition, since we allow integer ids for the constructor when creating a document, we allow integers for deleting documents as well. Otherwise this might become a source of client site errors.
The int to str conversion is now very simplistic it ist just `str(num)`

jina/drivers/rank/__init__.py
florian-hoenicke: you are so right! Thanks. This statement is not needed anymore

tests/unit/drivers/test_cache_driver.py
```diff
@@ -146,7 +146,7 @@ def test_cache_content_driver_same_content(tmpdir, test_metas):
     doc1.text = new_string
     doc1.update_content_hash()
     with BaseExecutor.load(filename) as executor:
-        executor.update([UniqueId(1)], [doc1.content_hash])
+        executor.update(['1'], [doc1.content_hash])
```
maximilianwerk, line 149: To be consistent with the id setting in generation, this should be `...[1]...`. Or change the indexed Documents to `Document(id='1')`

florian-hoenicke: changed

jina/drivers/rank/aggregate/__init__.py
```diff
@@ -105,9 +104,9 @@ def _apply_all(self, docs: 'DocumentSet',
         :return:
         """
 
-        match_idx = []  # type: List[Tuple[UniqueId, UniqueId, UniqueId, float]]
-        query_meta = {}  # type: Dict[UniqueId, Dict]
-        match_meta = {}  # type: Dict[UniqueId, Dict]
+        match_idx = []  # type: List[Tuple[str, str, str, float]]
```
cristianmtr, line 107: Why are these types assigned behind a comment?

florian-hoenicke: https://realpython.com/lessons/type-comments/

cristianmtr: Yes, but why are they comments in this case? They only reason to do that is if we want to support Python 2. Anyway, not a big deal

PR #1873, opened 2021-02-05
refactor: index performance

florian-hoenicke:


jina/executors/indexers/__init__.py
```diff
@@ -196,32 +196,13 @@ def flush(self):
...
-        for key_index, key in enumerate(keys):
-            if key not in existent_keys:
-                indices_to_drop.append(key_index)
-        if indices_to_drop:
-            self.logger.warning(
-                f'Key(s) {[keys[i] for i in indices_to_drop]} were not found in {check_path}. Continuing anyway...')
-        return indices_to_drop
+    def _filter_nonexistent_keys_values(self, keys: Iterable, values: Iterable, existent_keys: Iterable) -> Tuple[
+        Iterable, Iterable]:
+        filtered_list = [[key, value] for key, value in zip(keys, values) if key in existent_keys]
```
JoanFM, line 201: Can it be better to return a Tuple? for inmutability?

cristianmtr, line 201: Should we not still check the length of the two match?

cristianmtr: `zip()` swallows that problem

florian-hoenicke: In what scenario would we have different lengths?

cristianmtr: Don't remember the exact reason it came up. But we had that check before to fix some issue that I can't recall

florian-hoenicke: do you mean doing:
```
tuple([key_value[0] for key_value in filtered_list])
```
I can do it. An argument against this would be that we use lists almost everywhere and no tuples.
What do you think?

florian-hoenicke: Okay, let me check if I find out if the reason still exists.

JoanFM: no tuple, just using ( ) would make it inmutable right? not important so forget about it

hanxiao: note, all sanity check should be done at the driver level

florian-hoenicke: Before my refactoring of the cache, the cache was using `_filter_nonexistent_keys_values` as well. 
Could it be related to the previous data structure in the cache?

florian-hoenicke: using () we would create a generator object. Actually, I started with just using generators here to make it super performant. But then I realized that the vector indexer converts everything into numpy array anyways.
So we have to call `list` on the key generator which is not so nice: `np.array(list(keys), (np.str_, self.key_length))`

florian-hoenicke:
All tests passed. 
Can it be merged?

jina/executors/indexers/__init__.py
maximilianwerk: I believe he meant `[(key, value)...]`, not?

maximilianwerk, line 201: :heart:

cristianmtr: We can't do the check if keys exist in the driver because we don't have access to the existing keys, do we?

maximilianwerk (approved):
LGTM

PR #1875, opened 2021-02-05
feat: added base embedding evaluator

maximilianwerk:
Enables using `BaseEmbeddingEvaluator`s out of the box. Furthermore, updated the other driver resources: deleted non existing ones and added the `BaseTextEvaluator`


nan-wang (approved):
LGTM👍

PR #1883, opened 2021-02-05
fix(types): fix extend in docset

hanxiao:


jina/types/document/multimodal.py
```diff
@@ -44,10 +44,17 @@ def __init__(self, document: Optional[DocumentSourceType] = None,
...
         if chunks or modality_content_map:
             if chunks:
-                granularities = [chunk.granularity for chunk in chunks]
-                if len(set(granularities)) != 1:
+                g = {c.granularity for c in chunks}
+                if len(g) != 1:
                     raise BadDocType('Each chunk should have the same granularity.')
                 self.chunks.extend(chunks)
+
+                # in case chunks have granularity defined, override
```
JoanFM, line 52: a warning should be raised if something is overriden for the user?

hanxiao: this is your original logic in the code, which I found hard to understand.

hanxiao: i keep it as it is as your original does not raise warning

hanxiao: >this is your original logic in the code, which I found hard to understand.

to be clear, I'm against the logic that a child granularity can decide parent's granularity

PR #1888, opened 2021-02-06
fix(driver): compoundindexer should turn on is_merge

hanxiao:
without `is_merge=True`, all scored info is lost due to copy override


JoanFM:
@hanxiao @cristianmtr now I see why it was True by default, I think then we should set it back to True by default and find a fix for the problem of duplicated shapes.

PR #1897, opened 2021-02-08
test: fix test in flow, add back hw result check

hanxiao:


tests/integration/test_helloworld.py
```diff
@@ -32,7 +32,7 @@ def check_hello_world_results(html_path: str):
     evaluation = soup.find_all('h3')[0].text
     assert 'Precision@50' in evaluation
     assert 'Recall@50' in evaluation
-    evaluation_results = re.findall("\d+\.\d+", evaluation)
+    evaluation_results = re.findall(r'\d+\.\d+', evaluation)
     assert len(evaluation_results) == 2
     # not exact to avoid instability, but enough accurate to current results to raise some alarms
     assert float(evaluation_results[0]) > 65.0
```
JoanFM, line 38: can u set the 65.0 to 50.0? it is quite random the set of documents taken for the test and we were seeing times with lower value

JoanFM (changes requested):

PR #1900, opened 2021-02-08
fix: refactor doccache filename usage

cristianmtr:


jina/executors/indexers/cache.py
```diff
@@ -10,28 +12,38 @@
...
     The difference between a cache and a :class:`BaseKVIndexer` is the ``handler_mutex`` is released in cache,
     this allows one to query-while-indexing.
     """
 
     def __init__(self, *args, **kwargs):
+        """Create a new BaseCache."""
         super().__init__(*args, **kwargs)
 
     def post_init(self):
+        """Call the post-init."""
```
JoanFM, line 26: I think this does not make much sense right? I think may not be needed

jina/executors/indexers/cache.py
```diff
@@ -109,14 +125,15 @@ def delete(self, keys: Iterable[str], *args, **kwargs) -> None:
...
     def get_query_handler(self) -> CacheHandler:
-        return self.CacheHandler(self.index_abspath, self.logger)
+        """Get the CacheHandler."""
+        return self.CacheHandler(self.save_abspath, self.logger)
 
     def get_create_handler(self):
-        # not needed, as we use the queryhandler
+        """Not needed, as we use the queryhandler."""
         # FIXME better way to silence warnings
         return 1
```
JoanFM, line 139: why do not we just call `return self.get_query_handler()`?

JoanFM (changes requested):

jina/executors/indexers/cache.py
cristianmtr: Good point. Done now

jina/executors/indexers/cache.py
cristianmtr: Yes, will move the comment here

jina/executors/indexers/cache.py
```diff
@@ -43,14 +55,15 @@ def __init__(self, path, logger):
                 self.cache_val_to_id = dict()
 
         def close(self):
+            """Flushes the in-memory cache to pickle files."""
```
JoanFM, line 58: should super.close be called as well here?

jina/executors/indexers/cache.py
```diff
@@ -109,14 +125,13 @@ def delete(self, keys: Iterable[str], *args, **kwargs) -> None:
...
     def get_query_handler(self) -> CacheHandler:
-        return self.CacheHandler(self.index_abspath, self.logger)
+        """Get the CacheHandler."""
+        return self.CacheHandler(self.save_abspath, self.logger)
 
     def get_create_handler(self):
-        # not needed, as we use the queryhandler
-        # FIXME better way to silence warnings
-        return 1
+        """Get the CacheHandler."""
```
JoanFM, line 136: can we then use the proper access handler depending on the operation?

cristianmtr: No point. It would just make it more confusing, I think. We only need one.

jina/executors/indexers/cache.py
cristianmtr: It's called by the `DocCache` closing lifecycle

JoanFM (approved):

PR #1901, opened 2021-02-08
docs: docstrings for cache driver

cristianmtr:


jina/drivers/cache.py
```diff
@@ -9,11 +10,11 @@
...
 class BaseCacheDriver(BaseIndexDriver):
-    """A driver related to :class:`BaseCache`
-    """
+    """A driver related to :class:`BaseCache`."""
 
     def __init__(self, with_serialization: bool = False, *args, **kwargs):
-        """
+        """Create a new BaseCacheDriver.
+
         :param with_serialization: feed serialized Document to the CacheIndexer
```
theUnkownName, line 18: All params need :type: too

theUnkownName (changes requested):

davidbp (approved):

jina/drivers/cache.py
cristianmtr: They're in the method signature, right? That's just duplication of information

cristianmtr: And the linter doesn't complain about it

PR #1902, opened 2021-02-08
fix(flow): pea/pod start on the Flow level should not block

hanxiao:


JoanFM (commented):
What is the motivation behind this feature?

hanxiao:
> What is the motivation behind this feature?

Run this on master, u will see
https://github.com/jina-ai/jina/pull/1902/files#diff-cd5dc30e812afe32d2c9ebde15530d33b4a6140085b28f58759aa7aa1a8e9102

PR #1905, opened 2021-02-08
fix: size assertion for first add call

maximilianwerk:
Before this change, a mismatch between vector and key shapes during the very first `add` call would not have raised the `ValueError`.


nan-wang (approved):
LGTM👍

PR #1906, opened 2021-02-08
docs(checker): update docstring for checker

Yongxuanzhang:


jina/checker.py
```diff
@@ -14,9 +14,14 @@
...
 
 class ImportChecker:
-    """Check all executors, drivers and handler functions in the package. """
+    """Check all executors, drivers and handler functions in the package."""
 
     def __init__(self, args: 'argparse.Namespace'):
+        """
+        Create a new :class:`ImportChecker`.
+
+        :param args: args provided by the CLI.
```
theUnkownName, line 23: Could you maybe add the `type` for the `params`?

theUnkownName (changes requested):

jina/checker.py
hanxiao: @theUnkownName as long as the param get type hint in the function signature, there is no need to write `:param type` manually in docstring

theUnkownName: We were just having that discussion on the Slack channel @hanxiao

PR #1913, opened 2021-02-09
fix(resources): fix yaml config in resources

hanxiao:


Yongxuanzhang (commented):
tests/unit/test_driver_yaml.py::test_resource_executor is failing since executors._route.yml, executors._unique.yml, executors._unique_content.yml are removed. Shall we also remove the corresponding tests?

PR #1914, opened 2021-02-09
docs: improve docstrings on drivers

cristianmtr:


jina/drivers/multimodal.py
```diff
@@ -47,23 +52,25 @@ def positional_modality(self) -> List[str]:
...
+        """From a dictionary ``content_by_modality`` it returns the arguments in the proper order so that they can be
         passed to the executor.
+
          :param content_by_modality: a dictionary of `Document content` by modality name
+         :return: list of input arguments as np arrays
         """
         return [content_by_modality[modality] for modality in self.positional_modality]
 
-    def _apply_all(
+    def apply_all(
```
JoanFM, line 64: do not change the `name of the function

JoanFM (changes requested):

jina/drivers/multimodal.py
cristianmtr: Mistake. Wanted to use PyCharm to create the template, but it only does it for public methods. Reverted

cristianmtr: Thanks for catching that

PR #1916, opened 2021-02-09
docs(proto): update proto docstring

Yongxuanzhang:
- [ ] Cannot find usage of JinaRPC.call. Not sure how to write docstrings for this method.

### DIFF
--- a/jina/proto/jina_pb2_grpc.py
+++ b/jina/proto/jina_pb2_grpc.py
@@ -6,16 +6,14 @@ from . import serializer as jina__pb2
 
 
 class JinaRPCStub(object):
-    """*
-    jina gRPC service.
+    """
+    Jina gRPC service.
+
+    :param channel: A grpc.Channel.
     """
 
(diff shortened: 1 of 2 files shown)


JoanFM (commented):
I believe these files are autogenerated by protobuf when compiling the .proto file, so I think we should not add docstrings here

Yongxuanzhang:
Oh ok, if so I will close this PR. How about jina_pb2_grpc.py? @JoanFM  I thought we created those classes.

JoanFM:
> Oh ok, if so I will close this PR. How about jina_pb2_grpc.py? @JoanFM I thought we created those classes.

The same, jina_pb2 are autogenerated

Yongxuanzhang:
> > Oh ok, if so I will close this PR. How about jina_pb2_grpc.py? @JoanFM I thought we created those classes.
> 
> The same, jina_pb2 are autogenerated

Got it. Thanks!

PR #1925, opened 2021-02-10
feat: precommit hooks for docstrings

cristianmtr:
- [x] Add pre-commit hook for linting docstrings
- [x] Use agreed-upon configuration
- [x] Update CONTRIBUTING.md
- [x] Updated docs on docstrings https://github.com/jina-ai/docs/pull/83

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+repos:
+-   repo: https://github.com/terrencepreilly/darglint
+    rev: v1.5.8
+    hooks:
+    - id: darglint
+      args:
+      - --message-template={path}:{line} {msg_id} {msg}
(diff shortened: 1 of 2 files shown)


hanxiao:
is this mandatory, i.e. if not using `pre-commit` will it fail/block on the PR?

maximilianwerk:
> is this mandatory, i.e. if not using `pre-commit` will it fail/block on the PR?

No, it is not. `pre-commit` will only be executed, when you have installed it consciously. I would still recommend all core developers to use it, to not get diverging quality in the checked parts.

`pre-commit` is mostly for way faster feedback to the developer, when you don't want to push and wait for CI to tell you, what you did wrong.

maximilianwerk (approved):
LGTM

Issue #1891, opened 2021-02-07
Problems with shape and repeated fields under `MergeFrom` (is_merge option in KVSearchDriver)


JoanFM:
**Describe the bug**
We have been dealing with problems at `KVSearchDriver` with the `is_merge` option. After PR #1888  I understood the neeed for `Merge` in KVSearchDriver.

Reading the documentation from protobuf one can see:

```
virtual void Message::MergeFrom(        const Message & from)
Merge the fields from the given message into this message.

Singular fields will be overwritten, if specified in from, except for embedded messages which will be merged. Repeated fields will be concatenated. The given message must be of the same type as this message (i.e. the exact same class).
```

**Point 1**
Maybe the problem is that `shape` should not be a repeated field but an `NdArray`

**Point 2**
This merge will be dangerous when `chunks` and `matches` already exist (as any other repeated field). The avoidance to duplicate the `matches` or `chunks` whould be handled inside `Document` type?

JoanFM:
@cristianmtr

cristianmtr:
Any thoughts on why it would be `repeated`?

hanxiao:
> Any thoughts on why it would be repeated?

it is default Protobuf merge behavior. see proto docs

hanxiao:
there is no such Proto def as `NdArray`, I actually defined `NdArray`, if you change `shape` to `NdArray`, you just nest it, not solving the problem

PR #1926, opened 2021-02-10
feat(types): add update method to document types

hanxiao:
this solves #1891


nan-wang:
This is a potential Breaking change because the API of `KVSearchDriver` is changed.

PR #1927, opened 2021-02-11
docs(helper): add missing helper docstrings

Yongxuanzhang:


jina/helper.py
```diff
@@ -904,7 +948,10 @@ def run_async(func, *args, **kwargs):
     """
 
     class RunThread(threading.Thread):
+        """Create a running thread when in Jupyter notebook."""
```
maximilianwerk, line 951: This class is private to the function. Having docstrings is OK, but I'd still rename it to `_RunThread` in order to highlight, that it is private.

jina/helper.py
```diff
@@ -904,7 +948,10 @@ def run_async(func, *args, **kwargs):
...
+
         def run(self):
+            """Run function for thread."""
```
maximilianwerk, line 954:
```suggestion
            """Runs given `func` asynchronously."""
```

jina/helper.py
```diff
@@ -743,11 +757,41 @@ def typename(obj):
 
 
 def rsetattr(obj, attr: str, val):
```
maximilianwerk, line 759: `rsetattr` and `rgetattr` are not used anywhere in the codebase anymore. Let's remove them.

maximilianwerk (changes requested):
I'd rather remove not used functions.

jina/helper.py
Yongxuanzhang: Done

jina/helper.py
Yongxuanzhang: Done

jina/helper.py
Yongxuanzhang: Done

nan-wang (approved):
LGTM👍

PR #1933, opened 2021-02-12
refactor: move queryset from drivers to types

JoanFM:
**Changes introduced**
Move QuerySet logic to `types` so that it can be made a regular way of accessing Documents and other primitive types `attributes`.

Plan to use this change in #1718.


jina/drivers/control.py
```diff
@@ -6,7 +6,7 @@
 from google.protobuf.json_format import MessageToJson
 
 from . import BaseDriver
-from .querylang.queryset.dunderkey import dunder_get
+from jina.types.querylang.queryset.dunderkey import dunder_get
```
maximilianwerk, line 9: it should still be a relative import, e.g. `from ..types...`

maximilianwerk (changes requested):
While in principle, I would love to have absolute imports everywhere, our current convention is:

- relative imports in `jina/`
- absolute imports in `tests/`

Apart from that, I like the PR.

jina/drivers/control.py
JoanFM: oh yes, pycharm auto refactor

maximilianwerk (approved):

Issue #1932, opened 2021-02-11
the design of `_traverse_apply` makes all BaseRecursiveDriver low-efficient


hanxiao:
**Describe the bug**

*Highest priority*, must solve now.

The current design combination of `_apply_all` and `_traverse_apply` in `BaseRecursiveDriver` results in a low efficient recursion, in particular when each document has only a small number of chunks. The `_apply_all` will only apply on that small documents, making the CPU/GPU extremely data-hungry. 

This is inefficient. The only reason that this problem was not paid enough attention is that the problem only becomes obvious when you have multiple granularities. Single granularity, i.e. directly working on root-level won't reveal it.

This inefficient design also introduces further hacky implementations of the driver. One example is the `Cache` in `EncodeDriver` which is unnecessary if `apply_all` can work on all defined granularity in one-shot. Note that this temporally solves the data-hungry problem for `EncodeDriver` but all other drivers facing the same problem.

**Describe how you solve it**

- redesign `_traverse_apply`, move it to `Document` 
- `_apply_all` iterates over the **concatenation of iterators** in one shot
- remove hacky implementation in `EncodeDriver`

---

**Environment**

**Screenshots**

JoanFM:
Would you mean, creating a DocumentSet with all the Documents where Driver should apply in advance? 

Or would promote the CacheDocumentSet of Encoder solve also the problem?

hanxiao:
> Or would promote the CacheDocumentSet of Encoder solve also the problem?

no, as I said this cacheencodedriver is a result of the bad design of `traverse_apply`. It does not solve the problem that other driver is facing. 

 After all done, i will delete this logic from the driver. We will rely on `@batching` decorator on the executor level as it should be.

> Would you mean, creating a DocumentSet with all the Documents where Driver should apply in advance?

The following modification is expected:

`traverse_apply` -> move to DocSet, rename to `traverse`, return *one iterator* of `Document`
`apply_all`-> decouple from `traverse`, loop over the iterator
`apply_root`-> delete

FYI, this is strongly related to the previous idea I mentioned here: https://github.com/jina-ai/jina/issues/1326 unfortunately this was never implemented

hanxiao:
algorithm logic-wise & design-wise i do not worry at all. 

I worry about the huge number of unit tests that deeply tangled with the `traverse_apply` function.

JoanFM:
> > Or would promote the CacheDocumentSet of Encoder solve also the problem?
> 
[...]

PR #1938, opened 2021-02-12
refactor(driver): use traverse from document type

hanxiao:
this is the continuation of #1937 to solve #1932 

I've already layout the new implementation of `RecursiveDriver`, please continue on

- [ ] refactoring all drivers using this new logic
- [ ] retiring & refactoring test


nan-wang:
The returning type of `traverse()` of `DocumentSet` is changed in this PR. This makes this PR a potential breaking change.

PR #1946, opened 2021-02-15
docs(readme): rm duplicate wikipedia example

alexcg1:
Max and I both added the wiki example so it got added twice! This PR:
- removes Max's entry
- expands title of my own to include "NLP" for better explanation


nan-wang (approved):
LGTM👍

PR #1947, opened 2021-02-15
feat: have required keys for queries and matches and as instance attr

JoanFM:
**Changes introduced**
Allow the Ranker to have different set of keys extracted from Documents for Query and Matches


tests/unit/drivers/rank/aggregate/test_aggregate_matches_rank_driver.py
```diff
@@ -27,8 +31,7 @@ def exec_fn(self):
 
 class MockLengthRanker(Chunk2DocRanker):
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.required_keys = {'length'}
+        super().__init__(query_required_keys=('length', ), match_required_keys=('length', ), *args, **kwargs)
```
cristianmtr, line 34: Maybe add a test with more than one required key?

JoanFM: any idea? the interpretability of this test would be lost

nan-wang: What about adding `discount` and `price`? We can set ranking score by `discount * price`.

```
```

nan-wang (changes requested):
Can we make this backward compatible? There are pod images in the hub using `required_keys`. https://github.com/search?p=1&q=org%3Ajina-ai+required_keys&type=Code

tests/unit/drivers/rank/aggregate/test_aggregate_matches_rank_driver.py
JoanFM: Great idea @nan-wang

JoanFM:
> Can we make this backward compatible? There are pod images in the hub using `required_keys`. https://github.com/search?p=1&q=org%3Ajina-ai+required_keys&type=Code

Let me try to have it backward compatible

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
```diff
@@ -48,21 +58,23 @@ def create_document_to_score():
...
+            match.id = str(match_id)
             parent_id = 10 * int(match_id)
-            match.parent_id = str(parent_id) * 8
+            match.parent_id = str(parent_id)
             match.length = int(match_id)
             # to be used by MaxRanker and MinRanker
             match.score.ref_id = chunk.id
             match.score.value = int(match_id)
+            match.tags['price'] = match.score.value
+            match.tags['discount'] = 0.5
```
nan-wang, line 77:
```suggestion
            match.tags['discount'] = self.discount
```

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
```diff
@@ -21,13 +21,23 @@ def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwa
...
+        super().__init__(query_required_keys=('tags__price', 'tags__discount'), match_required_keys=('tags__price', 'tags__discount'), *args, **kwargs)
+
+    def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwargs):
+        price = match_chunk_meta[match_idx[0][self.COL_DOC_CHUNK_ID]]['tags__price']
+        discount = match_chunk_meta[match_idx[0][self.COL_DOC_CHUNK_ID]]['tags__discount']
+        return match_idx[0][self.COL_PARENT_ID], \
+               price - price * discount
+
+
 class SimpleChunk2DocRankDriver(Chunk2DocRankDriver):
```
nan-wang, line 42: discount = 0.5

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
```diff
@@ -127,45 +139,25 @@ def create_chunk_chunk_matches_to_score():
...
-
 
-@pytest.mark.parametrize('keep_source_matches_as_chunks', [False, True])
-def test_chunk2doc_ranker_driver_max_ranker(keep_source_matches_as_chunks):
-    doc = create_document_to_score()
-    driver = SimpleChunk2DocRankDriver(keep_source_matches_as_chunks=keep_source_matches_as_chunks)
-    executor = MockMaxRanker()
-    driver.attach(executor=executor, runtime=None)
-    driver._traverse_apply(DocumentSet([doc, ]))
+    scale = 1 if not isinstance(executor, MockPriceDiscountRanker) else 0.5
```
nan-wang, line 150:
```suggestion
    scale = 1 if not isinstance(executor, MockPriceDiscountRanker) else MockPriceDiscountRanker.discount
```

nan-wang (changes requested):
LGTM👍

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
JoanFM: there is no self here right?

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
JoanFM: But here I want to shpw that is passed to the Ranker

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
JoanFM: but then the point of the test is lost I want to show how discount info arrives to the Ranker score function

tests/unit/drivers/rank/aggregate/test_chunk2doc_rank_drivers.py
JoanFM: here the same

nan-wang (approved):
LGTM👍

PR #1950, opened 2021-02-16
refactor: added fast traversal with structure

maximilianwerk:
This PR enables changes on chunks and matches level even with the `FastRecursiveMixin`.
Discussion point (opinions welcomed!):

- renaming of `docs` to `leaves` in `_apply_all`. Reasoning: it fits to leaves in the tree traversal and thus it is a better semantic naming, then the general `docs`. Anyhow, I am not totally convinced about the renaming.

some tests still needs fixing. will happen throughout the day.


tests/unit/drivers/test_encoder_driver.py
```diff
@@ -22,35 +23,23 @@ def docs_to_encode(num_docs):
...
+            super().__init__(*args, **kwargs)
 
-    @property
-    def is_last_batch(self):
-        return self.id_pass == self.total_passes
+        @batching(batch_size=batch_size)
+        def encode(self, data: Any, *args, **kwargs) -> Any:
+            if batch_size is not None and batch_size > 0:
+                assert len(data) <= batch_size
+            if batch_size == 5:
```
JoanFM, line 35: this test is very weird, very dependant in the exact value of batch_size?

JoanFM (changes requested):

tests/unit/drivers/test_encoder_driver.py
maximilianwerk: The reason is that only for 1,2,5,10 you can guarantee for 10 documents, that each batch has a certain size. That's why I put this test only for the case of `batch_size` 5.

jina/drivers/__init__.py
```diff
@@ -326,18 +327,33 @@ class FastRecursiveMixin:
...
 
     @property
     def docs(self) -> 'DocumentSet':
         from ..types.sets import DocumentSet
 
         if self.expect_parts > 1:
-            return DocumentSet(
-                (d for r in reversed(self.partial_reqs) for d in r.docs)
-            ).traverse(self._traversal_paths)
+            return DocumentSet((d for r in reversed(self.partial_reqs) for d in r.docs))
```
florian-hoenicke, line 354: We can change the constructor of DocumentSet to accept `Iterable` instead of `Sequence`. 
Or we handover a list instead of the generator.

florian-hoenicke (changes requested):

jina/drivers/__init__.py
maximilianwerk: It needs to be a `Sequence` in order to be able to delete and index elements.

hanxiao (changes requested):
I don't understand the change on FastRecursiveMixin, there is nothing wrong with it.

- what's the reason behind the change from DocumentSet to Iterable of DocumentSet? This PR basically reverts the effort of #1938 and entangled the traversal & applying again.
- The real problem is not about `FastRecursiveMixin`, those Drivers that can fit with `FastRecursiveMixin` are good already. The problem is about the remaining drivers. However, I don't see other significant changes on the legacy `RecursiveMixin`.
- I'm against calling it `leaves`, recursion and applying are separated in the `FastRecursiveMixin` that's the idea in #1932.
- In general, `FastRecursiveMixin` in #1938 is good enough already.

**If there are some new traversal, create a new `Mixin` may be a better way to go than changing the already working `FastRecursiveMixin`**, maybe create a new mixin called `MutableRecursiveMixin`

JoanFM:
The point is to have only one source of traversal logic, and the more generic one that can be used for every driver is the one returning Iterable[DocumentSet]. Flattening before or afterwards have very little implications in my opinion. The cost of having two traversal logics can make even more painful further optimizations or mantainance in my opinion.

jina/drivers/__init__.py
```diff
@@ -362,6 +364,49 @@ def _traverse_rec(self, docs, parent_doc, parent_edge_type, path, *args, **kwarg
             self._apply_all(docs, parent_doc, parent_edge_type, *args, **kwargs)
 
 
+class SetRecursiveMixin:
+    """
+     The full datastructure version of :class:`RecursiveMixin`, to be mixed in with :class:`BaseRecursiveDriver`
```
JoanFM, line 369: correct name in docstring

JoanFM, line 369: this one is not to be used in Index and EncodeDrivrr right? Why is this change applied directly on RecursiveMixin (and maybe renaming it?)

JoanFM (dismissed):

jina/drivers/__init__.py
maximilianwerk: Because I want to refactor it piece by piece. Finally, the `RecursiveMixin` should be deleted.

jina/drivers/evaluate.py
```diff
@@ -5,14 +5,15 @@
...
+from . import BaseExecutableDriver
 from ..types.querylang.queryset.dunderkey import dunder_get
 from .search import KVSearchDriver
 from ..types.document import Document
 from ..types.document.helper import DocGroundtruthPair
+from ..types.sets.doc_groundtruth import DocumentGroundtruthSequence
 
 
-class BaseEvaluateDriver(RecursiveMixin, BaseExecutableDriver):
+class BaseEvaluateDriver(BaseExecutableDriver):
```
JoanFM, line 16: I think this should also inherit `ContextAwareRecursiveDriver` by having `DocGroundTruthPair` implement ` traversal`

jina/types/sets/chunk.py
```diff
@@ -48,7 +48,7 @@ def append(self, document: 'Document', **kwargs) -> 'Document':
         return chunk
 
     @property
-    def parent_doc(self) -> 'Document':
+    def reference_doc(self) -> 'Document':
```
JoanFM, line 51: I think the access through `parent_doc` still makes sense in `ChunkSet`

JoanFM: we can have `reference_doc` as common accessor for `MatchSet` and `ChunkSet`

jina/types/sets/traversable.py
```diff
@@ -0,0 +1,65 @@
+from typing import Iterable
+import itertools
+
+if False:
+    from ..document.traversable import Traversable
+
+
+class TraversableSequence:
```
JoanFM, line 8: Like a lot this idea!

tests/unit/drivers/test_reduce_all_driver.py
```diff
@@ -67,3 +68,27 @@ def validate(req):
         flow.index(input_fn=input_fn, on_done=response_mock)
 
     response_mock.assert_called()
+
+
+def test_reduce_all_root_chunks(mocker, docs):
```
JoanFM, line 73: This should go with best possible effort into an `integration` test. Try not to add tests with `Flow` in unittests

JoanFM: It can be done in another PR though

JoanFM (commented):
It is looking really good!

jina/types/sets/chunk.py
maximilianwerk: That's the point. Unify the doc in the level above.

jina/drivers/evaluate.py
maximilianwerk: The point is, that the `__call__` function itself is overwriten. So nothing from the `ContextAwareRecursiveMixin` remains. For the sake of "that's what I want to do here as a higher concept" we could keep the mixin, but it would not change anything.

jina/types/sets/chunk.py
JoanFM: But we can keep the sugar for parent, Think this ChunkSet could be used by extrrnal users.

jina/drivers/__init__.py
```diff
@@ -469,6 +365,17 @@ def __init__(self,
         super().__init__(*args, **kwargs)
         self._traversal_paths = [path.lower() for path in traversal_paths]
 
+    @property
+    def docs(self) -> 'DocumentSet':
```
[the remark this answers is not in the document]
maximilianwerk: :+1:

tests/unit/drivers/test_reduce_all_driver.py
maximilianwerk: While I see your point, this was here before. Not part of this PR for now, to not obfuscate more, what actually happens in this PR.

jina/drivers/evaluate.py
JoanFM: then remove okey!

jina/drivers/__init__.py
```diff
@@ -305,93 +320,41 @@ def __getstate__(self) -> Dict[str, Any]:
...
-        :param docs: a list of :class:`jina.Document` objects to work on; they could come from ``matches``/``chunks``.
-        :param context_doc: the owner of ``docs``
-        :param field: where ``docs`` comes from, either ``matches`` or ``chunks``
-        :param *args: *args
-        :param **kwargs: **kwargs
-        """
-
-    def __call__(self, *args, **kwargs):
-        """Call the Driver.
+        """Apply function works on a list of list of docs, modify the docs in-place.
```
florian-hoenicke, line 348: Is it a list of docs instead of a list of list of docs?

jina/types/sets/traversable.py
```diff
@@ -0,0 +1,65 @@
...
+        return _traverse_all()
+
+    def traverse_flatten(self, traversal_paths: Iterable[str]) -> 'TraversableSequence':
+        """
+        Returns a single flattened :class:``DocumentSet`` with all Documents, that are reached
+        via the :param:``traversal_paths``.
+
+        :param traversal_paths: a list of string that represents the traversal path
+        """
+        leaves = self.traverse(traversal_paths)
```
theUnkownName, line 60: Aren't the leaves already flatten in this step?

jina/drivers/reduce.py
```diff
@@ -23,73 +22,65 @@ class ReduceAllDriver(RecursiveMixin, BaseRecursiveDriver):
...
-            doc.embedding = np.concatenate(self.doc_pointers[doc.id], axis=0)
-        else:
-            if doc.id not in self.doc_pointers:
-                self.doc_pointers[doc.id] = [doc.embedding]
-            else:
-                self.doc_pointers[doc.id].append(doc.embedding)
+        last_request_documents = self.req.docs.traverse_flatten(self._traversal_paths)
+        self._concat_apply(last_request_documents, doc_pointers)
+
+    def _collect_embeddings(self, docs: 'DocumentSet'):
```
florian-hoenicke, line 78: Quick question about this function:
Do we have duplicates in the document Set?

JoanFM: we do not guarantee Documents in DocumentSet are unique, so is not a real set actually

jina/drivers/search.py
```diff
@@ -61,24 +61,25 @@ def __init__(self, is_update: bool = True, traversal_paths: Tuple[str] = ('m'),
...
 
-    def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
-        miss_idx = []  #: missed hit results, some search may not end with results. especially in shards
-        for idx, retrieved_doc in enumerate(docs):
-            serialized_doc = self.exec_fn(retrieved_doc.id)
-            if serialized_doc:
-                r = Document(serialized_doc)
-                if self._is_update:
-                    retrieved_doc.update(r)
+    def _apply_all(self, doc_sequences: Iterable['DocumentSet'], *args, **kwargs) -> None:
```
florian-hoenicke, line 64: It would be nice to have less nested loops and if-statements.

florian-hoenicke (changes requested):

jina/drivers/reduce.py
florian-hoenicke: thanks

jina/drivers/reduce.py
```diff
@@ -23,73 +22,65 @@ class ReduceAllDriver(RecursiveMixin, BaseRecursiveDriver):
...
+        doc_pointers = defaultdict(list)
+        for doc in docs:
+            doc_pointers[doc.id].append(doc.embedding)
```
florian-hoenicke, line 81: Where do we know the embeddings are appended in the right order?

maximilianwerk: Ouch, fair point. But this problem is not introduced with the changes of this PR. I create an issue.

maximilianwerk: Ah no, we rely, that the chunks are in the right order for all documents.

jina/types/sets/traversable.py
maximilianwerk: No they are not. They are `Iterable[TraversableSequence]`, where each `TraversableSequence` is a set of matches or chunks.

jina/drivers/__init__.py
maximilianwerk: It is an `Iterable[DocumentSet]` which is an `Iterable` over Chunks or Matches. Witch are a list of Documents.

jina/drivers/search.py
maximilianwerk: True. This function will be refactored in a next step anyhow, where we enable batching on the Executor level.

jina/drivers/__init__.py
florian-hoenicke: then, we should change the docstring I guess:
`iterable over docsets`

jina/helper.py
```diff
@@ -282,12 +282,25 @@ def random_name() -> str:
     return '_'.join(random.choice(_random_names[j]) for j in range(2))
 
 
+_used_ports = set()
```
JoanFM, line 285: this should not remain here, when is this closed?

maximilianwerk:
In order to get this PR running, I changed the `random_port` generation. I made sure, that the already sampled ports are saved and are not sampled once again.

JoanFM:
> In order to get this PR running, I changed the `random_port` generation. I made sure, that the already sampled ports are saved and are not sampled once again

I think this is very dangerous

jina/helper.py
maximilianwerk: Hmm why not? if we used up all 10000 possible ports, that's bad, but until then it should be fine, no?

maximilianwerk:
> > In order to get this PR running, I changed the `random_port` generation. I made sure, that the already sampled ports are saved and are not sampled once again
> 
> I think this is very dangerous

Why? Can you elaboratE?

jina/helper.py
JoanFM: because this is global. If a daemon is running forever, it will eventually consume all of these?

JoanFM:
> > > In order to get this PR running, I changed the `random_port` generation. I made sure, that the already sampled ports are saved and are not sampled once again
> > 
> > 
> > I think this is very dangerous
> 
> Why? Can you elaboratE?

It is a global variable, when does it refresh? a daemon or someone running loading jina and running jina infintely will run out of ports.

JoanFM:
> > > > In order to get this PR running, I changed the `random_port` generation. I made sure, that the already sampled ports are saved and are not sampled once again
> > > 
> > > 
> > > I think this is very dangerous
> > 
> > 
> > Why? Can you elaboratE?
> 
> It is a global variable, when does it refresh? a daemon or someone running loading jina and running jina infintely will run out of ports.

U can do this by mocking this function in test if u wish, but I do not think it should be in the framework

maximilianwerk:
> 
> U can do this by mocking this function in test if u wish, but I do not think it should be in the framework

Done

JoanFM (dismissed):
@nan-wang breaking changes here, the traversal_paths for `RankDrivers` will need to change

maximilianwerk:
> @nan-wang breaking changes here, the traversal_paths for `RankDrivers` will need to change

Absolutely right!

maximilianwerk:
@JoanFM I needed to change some imports in order to make the `mock` of `random_port` work effectively everywhere. That is a subtle change and I will dig up more reading material for the team abort import structuring in the future.

JoanFM (approved):

nan-wang (approved):
LGTM👍

daemon/models/custom.py
```diff
@@ -38,9 +39,9 @@ def _get_pydantic_fields(parser: Callable[..., 'argparse.ArgumentParser']):
                     default_factory=random_identity,
                     example=a['default'],
                     description=a['help'])
-            elif a['default_factory'] == random_port.__name__:
+            elif a['default_factory'] == helper.random_port.__name__:
```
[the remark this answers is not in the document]
maximilianwerk: But it is not mockable, if the import happens before the mock was defined.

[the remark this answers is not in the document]
maximilianwerk: And efficiency in these imports is not important from real importance. I first want to see in numbers, how much efficiency increases, before we keep these dirty imports.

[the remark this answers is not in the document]
JoanFM: agree

PR #1953, opened 2021-02-16
feat: ranking evaluate driver may need more than one field

JoanFM:
**Changes introduced**
Motivation of this PR comes from the realization that `NDCGEvaluator` is not applicable, since more than one field may be needed for it.

`RankingEvaluateDriver` now can access multiple fields of the `matches` of both `Document` and `GroundTruth`. 

For instance, for NDCGEvaluator, we may need:
    fields = ('tags__id', 'score__value') 

So that the NDCGEvaluator would get
desired = (id, relevance), (id, relevance)
actual = (id, score), (id, score) so that it can be properly computed


jina/drivers/evaluate.py
```diff
@@ -90,28 +91,35 @@ def extract(self, doc: 'Document') -> Any:
...
+            The differences with `:class:FieldEvaluateDriver` are:
+                - More than one field is allowed. For instance, for NDCGComputation you may need to have both `ID` and `Relevance` information.
+                - The fields are extracted from the `matches` of the `Documents` and the `Groundtruth` so it returns a sequence of values.
+    :param *args:
+    :param **kwargs:
     """
 
     def __init__(self,
-                 field: str = 'tags__id',
+                 fields: Tuple[str] = ('tags__id',),
```
nan-wang, line 122: Add a `deprecate` warning to make it backward compatible.

```
from ..helper import deprecated_alias
    @deprecated_alias(filed=('fields', 0)
```

jina/drivers/evaluate.py
```diff
@@ -90,28 +91,35 @@ def extract(self, doc: 'Document') -> Any:
...
+        self.fields = fields
 
     def extract(self, doc: 'Document'):
-        r = [dunder_get(x, self.field) for x in doc.matches]
-        # flatten nested list but useless depth, e.g. [[1,2,3,4]]
-        return list(numpy.array(r).flat)
+        if len(self.fields) > 1:
+            ret = [tuple(dunder_get(x, field) for field in self.fields) for x in doc.matches]
+        else:
+            ret = [dunder_get(x, self.fields[0]) for x in doc.matches]
```
nan-wang, line 132: `self.fields=[]` will lead to an exception

nan-wang: Having the returning type depending on a class member variable doesn't look good. Can we always return a list of tuples?

nan-wang: Remember to keep this part backward compatible.

nan-wang (changes requested):

jina/drivers/evaluate.py
JoanFM: It is an option, but it will make it quite dirty for the executors? I also do not like it so much

JoanFM: Can u check the changes? I did it backward compatible, but I do not check that there is not None in the Tuple (since it is not an Optional as per signature I am not sure it is so useful to check.

The same way we do not check if `field` is not None

jina/drivers/evaluate.py
```diff
@@ -105,33 +108,46 @@ def extract(self, doc: 'Document') -> Any:
...
+        super().__init__(*args, **kwargs)
+        self.fields = fields
 
-    def extract(self, doc: 'Document'):
-        """Extract the field from the Document's matches.
+    @property
+    def single_field(self):
+        if isinstance(self.fields, str):
+            return self.fields
+        elif len(self.fields) == 1:
```
bwanglzu, line 137: how do you think
```python
elif isinstance(self.fields, tuple):
    # do something
```
I don't see there is a logic if else relationship between `if isinstance and elif len()`

JoanFM: The problem is that some of the Executors' interfaces for backward compatibility expect a List[Union[str, float, int]], and it would not work if a Tupe of single element is an input

bwanglzu: okay got it!

bwanglzu (approved):
LGTM!

nan-wang (approved):
LGTM👍

deepankarm (approved):

PR #1966, opened 2021-02-17
tests: fix wrongly used mock

maximilianwerk:
This is the right usage of `mocker.Mock` with a validation function. It generates way better tracebacks, when the tests fail.

Furthermore, the correct naming would have been `wraps` instead of `wrap`.

There are still tests, which do it via calling the mock inside the validation function like `tests/distributed/test_join_local_from_remote/test_integration.py`. Anyhow, they work correctly, but don't give a bad traceback. Can be refactored how it is done in this PR.


tests/__init__.py
```diff
@@ -47,3 +47,9 @@ def rm_files(file_paths):
             elif os.path.isdir(file_path):
                 shutil.rmtree(file_path, ignore_errors=False, onerror=None)
 
+
+def validate_callback(mock, validate_func):
```
cristianmtr, line 51: Could we perhaps create a new instance of `Mocker` here and avoid spamming it in the code of the tests?

maximilianwerk: I don't get the suggestion :thinking:

maximilianwerk:
I have not a lot of experience around these tests, but could someone check, if my fix for `tests/unit/flow/test_flow_skip.py::test_bad_flow_skip_handle_join` is correct or if the test itself is failing. Thanks! @JoanFM

tests/__init__.py
rutujasurve94: We could pass `mocker` instead of `mock` as an argument and use `mocker.patch()` or `mocker.patch.object`.

cristianmtr: Do we need to create `m = mocker.Mock()` in the body of the test? Can't we create it inside this `validate_cb` instead, for each situation?

maximilianwerk: Yes, because it is send to the Flow as the `on_done` callback and collects the callback data there.

tests/unit/flow/test_flow_skip.py
```diff
@@ -103,18 +104,18 @@ def validate(req):
...
 
-    on_error_mock = mocker.Mock(wrap=validate)
+    on_error_mock = mocker.Mock()
 
     # always test two times, make sure the flow still works after it fails on the first
     with f:
         f.index(['abbcs', 'efgh'], on_error=on_error_mock)
 
-    on_error_mock.assert_called()
+    validate_callback(on_error_mock, validate)
```
nan-wang, line 121: Do we need to test twice here?

nan-wang (approved):
LGTM👍

tests/__init__.py
JoanFM: we could maybe offer this as a fixture that yields the mock and when finished passes the callback through this validation?

JoanFM: This would need to be very well documented however

PR #1968, opened 2021-02-18
docs(jaml): add missing docstrings

Yongxuanzhang:


theUnkownName (dismissed):
LGTM!

jina/jaml/parsers/default/v1.py
```diff
@@ -5,6 +5,7 @@
 
 
 class V1Parser(VersionedYAMLParser):
+    """V1 default parser, used as the default parser for objects except driver, executor and flow."""
```
JoanFM, line 8: Are we sure is not used for Flow if version is set to 1?

JoanFM (changes requested):
Could you check the questions I asked? You can add prints and load yamls of different objects to see if they are called for different objects

jina/jaml/__init__.py
```diff
@@ -105,6 +105,15 @@ def load_no_tags(stream, **kwargs):
     def expand_dict(d: Dict, context: Union[Dict, SimpleNamespace, None] = None,
                     resolve_cycle_ref=True,
                     resolve_passes: int = 3) -> Dict[str, Any]:
+        """
+        Expand variables from YAML file.
+
+        :param d: flow yaml file loaded as python dict
```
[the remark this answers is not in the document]
Yongxuanzhang: I think you're correct, 'flow' is not correct. I cannot remember why I add 'flow' here.

jina/jaml/parsers/default/v1.py
Yongxuanzhang: To my understanding, this is the V1 default parser, and we have another v1 parser for flow. 
image

Yongxuanzhang: I updated the expression in the new commit.

jina/jaml/helper.py
```diff
@@ -33,6 +39,13 @@ def get_hashable_key(self, key):
         return key
 
     def construct_mapping(self, node, deep=True):
+        """
+        Build the mapping between
+
+        :param node: the node to traverse
+        :param deep:
```
theUnkownName, line 46: Can you add more info to this parameters?

jina/jaml/helper.py
```diff
@@ -33,6 +39,13 @@ def get_hashable_key(self, key):
...
+        :return:
```
theUnkownName, line 47: Also maybe add info for the return type

theUnkownName (changes requested):
Add more info to parameteres

jina/jaml/helper.py
Yongxuanzhang: 👍

jina/jaml/helper.py
Yongxuanzhang: 👍

theUnkownName (approved):
Great job!

PR #1970, opened 2021-02-18
feat: doccache can handle multiple fields

cristianmtr:


jina/executors/indexers/cache.py
```diff
@@ -59,28 +65,29 @@ def close(self):
             pickle.dump(self.id_to_cache_val, open(self.path + '.ids', 'wb'))
             pickle.dump(self.cache_val_to_id, open(self.path + '.cache', 'wb'))
 
-    supported_fields = [ID_KEY, CONTENT_HASH_KEY]
-    default_field = ID_KEY
+    default_fields = [ID_KEY]
 
-    def __init__(self, index_filename: Optional[str] = None, field: Optional[str] = None, *args, **kwargs):
+    def __init__(self, index_filename: Optional[List[str]] = None, fields: Optional[List[str]] = None, *args, **kwargs):
```
JoanFM, line 70: `index_filename` is a `List[str]`?

JoanFM: I would like better for `fields` to be tuples?

jina/drivers/cache.py
```diff
@@ -21,23 +21,20 @@ class BaseCacheDriver(BaseIndexDriver):
...
         self.with_serialization = with_serialization
         super().__init__(*args, **kwargs)
-        self.field = None
 
     def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
-        self.field = self.exec.field
-
         if self._method_name == 'update':
-            self.exec_fn([d.id for d in docs], [d.id if self.field == ID_KEY else d.content_hash for d in docs])
+            values = [''.join(d.get_attrs(*self.exec.fields).values()) for d in docs]
```
JoanFM, line 27: If from here u can get a List of Tuples, I think the Tuple can be hashed? Could this may be a better solution?

jina/executors/indexers/cache.py
cristianmtr: Yes, fixed

jina/drivers/cache.py
cristianmtr: What do you mean?

jina/drivers/cache.py
```diff
@@ -21,23 +21,20 @@ class BaseCacheDriver(BaseIndexDriver):
...
+            self.exec_fn([d.id for d in docs],
+                         values)
```
florian-hoenicke, line 29: Interesting code cov says, the `self._method_name == 'update'`- case is not tested.
Is it right?

cristianmtr: It's tested here (mocked) : https://github.com/jina-ai/jina/blob/e6adf2b62e62255ab30f2c5b3e769bd79b7c985a/tests/unit/drivers/test_cache_driver.py#L206

and here: https://github.com/jina-ai/jina/blob/e6adf2b62e62255ab30f2c5b3e769bd79b7c985a/tests/integration/doccache/test_crud_cache.py#L203

jina/drivers/cache.py
```diff
@@ -21,23 +21,20 @@ class BaseCacheDriver(BaseIndexDriver):
...
         else:
             for d in docs:
-                data = d.id
-                if self.field == CONTENT_HASH_KEY:
-                    data = d.content_hash
-                result = self.exec[data]
+                value = ''.join(d.get_attrs(*self.exec.fields).values())
```
florian-hoenicke, line 32: having a cache value like this can cause unwanted collisions.
Example:
```
d1.a = '123'
d1.b = '456'
d2.a = '1'
d2.b = '23456'
```

both documents would have the same cache value

florian-hoenicke (changes requested):

jina/drivers/cache.py
florian-hoenicke: I would suggest creating a hash value from the list

florian-hoenicke: You can create the hash using:
```
hash(('123', '456'))
-3037693913774805154
hash(('1', '23456'))
-8391814287327595382
```

florian-hoenicke: Also the values should not be converted to strings since the conversion changes the hash value
```
hash(('1', 23456))
8044249239569849110
hash(('1', '23456'))
-8391814287327595382
```

jina/executors/indexers/cache.py
```diff
@@ -59,28 +69,24 @@ def close(self):
...
-        """
+    def __init__(self, index_filename: Optional[str] = None, fields: Optional[Tuple[str]] = None, *args, **kwargs):
         if not index_filename:
             # create a new temp file if not exist
             index_filename = tempfile.NamedTemporaryFile(delete=False).name
         super().__init__(index_filename, *args, **kwargs)
-        self.field = field or self.default_field
-        if self.field not in self.supported_fields:
-            raise ValueError(f"Field '{self.field}' not in supported list of {self.supported_fields}")
+        fields = fields or self.default_fields
```
florian-hoenicke, line 79: we try to avoid assignments to the same variable.

jina/executors/indexers/cache.py
```diff
@@ -59,28 +69,24 @@ def close(self):
...
+        # order shouldn't matter
+        self.fields = sorted(fields)
```
florian-hoenicke, line 81: Why do we have the `self.fields` attribute? Is it really necessary?

jina/drivers/cache.py
cristianmtr: Good idea. I adapted the logic to use `hash`. Theoretically this can still fail if 

```
d1.a = '123'
d1.c = '456'
d2.a = '123'
d2.b = '456'
```

jina/executors/indexers/cache.py
cristianmtr: We need it to know which fields we cache by, no?

jina/executors/indexers/cache.py
cristianmtr: Done, thanks

jina/executors/indexers/cache.py
florian-hoenicke: I think the driver should provide the hash of all fields and the executor does not even have to know about fields and just takes keys and values.

cristianmtr: We've discussed this (@maximilianwerk  @JoanFM ). For now fields are passed by the Executor in the YAML. Having them in the driver would mean rewriting the entire `on: requests`.

jina/drivers/cache.py
```diff
@@ -55,11 +50,21 @@ def on_miss(self, req_doc: 'Document', value: str) -> None:
...
         """Call when cache is hit for a document.
 
-        :param req_doc: the document in the request and hitted in the cache
+        :param req_doc: the document in the request and hit in the cache
         :param hit_result: the hit result returned by the cache
         """
         pass
 
+    @staticmethod
+    def hash(doc: 'Document', fields: List[str]) -> str:
```
florian-hoenicke, line 59: buildin method `hash` should not be shadowed.

florian-hoenicke (changes requested):

jina/drivers/cache.py
cristianmtr: Done

jina/drivers/cache.py
```diff
@@ -55,11 +50,21 @@ def on_miss(self, req_doc: 'Document', value: str) -> None:
...
 
+    @staticmethod
+    def hash_doc(doc: 'Document', fields: List[str]) -> str:
+        """Calculate hash by which we cache.
+
+        :param doc: the Document
+        :param fields: the list of fields
+        :return: the hash value of the fields
+        """
+        return str(hash(tuple((doc.get_attrs(*fields).values()))))
```
hanxiao, line 66: depending on the context, `hash` here may not be the good function, it may give different values on different runs.  as @JoanFM pointed out before

cristianmtr: Pointed out where? His comment above is just about hashing a tuple. Should we use one of these https://docs.python.org/3.7/library/hashlib.html ? Maybe sha256?

cristianmtr: Adapted to use that. Recheck

hanxiao (approved):

nan-wang:
I'd prefer to add backward compatible support to `field`, or else it is a breaking change.

PR #1972, opened 2021-02-18
tests: improve testing in hubio

JoanFM:


extra-requirements.txt
```diff
@@ -71,4 +71,6 @@ python-multipart:           http, devel, test, daemon
 aiofiles:                   devel, cicd, http, test, daemon
 pytest-custom_exit_code:    cicd, test
 bs4:                        test
-aiostream:                  devel, cicd
\ No newline at end of file
+aiostream:                  devel, cicd
+click:                      test
+cookiecutter:               test
```
deepankarm, line 76: `cookiecutter` is already listed above

JoanFM: can u suggest removing?

jina/docker/hubapi/remote.py
```diff
@@ -76,24 +76,24 @@ def _fetch_docker_auth(logger) -> Tuple[str, str]:
...
+        }
+        response = requests.get(url=f'{hubapi_url}', headers=headers)
+        if response.status_code == requests.codes.ok:
+            json_response = json.loads(response.text)
+            username = base64.b64decode(json_response['docker_username']).decode('ascii')
+            password = base64.b64decode(json_response['docker_password']).decode('ascii')
+            logger.debug(f'Successfully fetched docker creds for user')
+            return username, password
+        else:
+            logger.error(f'failed to fetch docker credentials. status code {response.status_code}')
```
nan-wang, line 91:
```suggestion
        if response.status_code != requests.codes.ok:
            logger.error(f'failed to fetch docker credentials. status code {response.status_code}')
        json_response = json.loads(response.text)
        username = base64.b64decode(json_response['docker_username']).decode('ascii')
        password = base64.b64decode(json_response['docker_password']).decode('ascii')
        logger.debug(f'Successfully fetched docker creds for user')
        return username, password
```

jina/docker/hubio.py
```diff
@@ -67,8 +67,10 @@ def _load_docker_client(self):
             # low-level client
             self._raw_client = APIClient(base_url='unix://var/run/docker.sock')
 
-    def new(self) -> None:
-        """Create a new executor using cookiecutter template."""
+    def new(self, no_input: bool = False) -> None:
+        """Create a new executor using cookiecutter template.
+            :param no_input: Argument to avoid prompting dialogue (just to be used for testing)
+        """
```
nan-wang, line 73: reference: https://docs.jina.ai/chapters/docstring/docstring.html#multi-line-docstrings-example-of-a-function

```suggestion
        """
        Create a new executor using cookiecutter template.
        
        :param no_input: Argument to avoid prompting dialogue (just to be used for testing)
        """
```

nan-wang (changes requested):

tests/unit/docker/hubapi/test_remote.py
```diff
@@ -0,0 +1,55 @@
...
+        return Path(str(tmpdir))
+
+    monkeypatch.setattr(requests, 'post', _mock_post)
+    monkeypatch.setattr(Path, 'home', _mock_home)
+    _register_to_mongodb(JinaLogger('test_mongodb_register'), summary)
+
+    request_args = m.call_args_list[0][1]
+    assert request_args['url'] == 'https://hubapi.jina.ai/push'
+    assert request_args['headers'] == {'Accept': 'application/json', 'authorizationToken': f'{DUMMY_ACCESS_TOKEN}'}
+    assert request_args['data'] == json.dumps(summary)
```
nan-wang, line 55: TBH, I didn't see the point of this test. I think a more useful test is to mock different responses and check all the `if-else` works as expected.

JoanFM: well is to ensure that the data sent in the request is the expected one. Since I am not familiar with the functionality I added these tests more as a non regression guarantee than anything else after seeing that a lot of them were not covered.

JoanFM: the if - else conditions only affect the loggers, so I am trying to force them to happen, but very little I can check

nan-wang: fair point.

nan-wang (approved):

extra-requirements.txt
```diff
@@ -51,7 +51,7 @@ protobuf>=3.13.0:           core
 grpcio>=1.33.1:             core
 pyyaml>=5.3.1:              core
 tornado>=5.1.0:             core
-cookiecutter:               hub, devel, cicd
+cookiecutter:               hub, devel, cicd, test
```
hanxiao, line 54:
```suggestion
cookiecutter:               hub, devel, cicd
```

JoanFM: I am adding tests that require these dependencies.

JoanFM: If I remove this, it won't pass

nan-wang (dismissed):
LGTM👍

extra-requirements.txt
hanxiao: no, `cicd` is the correct tag.
`test` is all "testsuit"-dependent packages

hanxiao (changes requested):
remove `test` from extra-requirement, see my comment above

extra-requirements.txt
JoanFM: ah ok

JoanFM:
> remove `test` from extra-requirement, see my comment above

Done!

PR #1976, opened 2021-02-18
docs(drivers): add missing docstrings

Yongxuanzhang:


jina/drivers/evaluate.py
```diff
@@ -132,12 +128,19 @@ def __init__(self,
 
     @property
     def single_field(self):
+        """Get single field."""
```
JoanFM, line 131: Can u add here the comment. Property to guarantee compatibility when only one field is provided either as a string or as a unit length tuple

JoanFM (changes requested):

jina/drivers/evaluate.py
Yongxuanzhang: Sure! Thanks!

jina/drivers/evaluate.py
```diff
@@ -68,13 +68,11 @@ def _apply_all(
             evaluation.ref_id = groundtruth.id
 
     def extract(self, doc: 'Document') -> Any:
-        """Extracting the to-be-evaluated field from the document.
-        Drivers inherit from :class:`BaseEvaluateDriver` must implement this method.
+        """Extract the to-be-evaluated field from the document. Drivers inherit from :class:`BaseEvaluateDriver` must implement this method.
```
theUnkownName, line 71: I think the first line of the docstring must be one short sentence.

PR #1979, opened 2021-02-19
fix: update the resource file for BaseRankingEvaluator

nan-wang:
fix: update the resource file for BaseRankingEvaluator


florian-hoenicke:
Can we have a small test or adjust an existing test?

hanxiao (changes requested):
im not sure this is a valid change, `tags.id` is not always there unless you fill it. `BaseRankingEvaluator` should not depend on a field that not exist.

nan-wang:
@hanxiao @JoanFM @florian-hoenicke Instead of changing the resource file, I've just added a unit test to show the issue. 

In https://github.com/jina-ai/jina/blob/master/jina/drivers/evaluate.py#L124, we provide the default value for fields as `('tags__id',)`.

However, in the resource file, https://github.com/jina-ai/jina/blob/master/jina/resources/executors.requests.BaseRankingEvaluator.yml#L17, we provide the default value for fields as `['id']`. 

This inconsistency might introduce confusion for the users. I think Han's argument is valid, and therefore suggest to remove the default value in the resource file and change the default value in `RankEvaluateDriver` to `('id', )`

jina/drivers/evaluate.py
```diff
@@ -144,7 +146,7 @@ def extract(self, doc: 'Document'):
         Extract values of the matches from documents with fields as keys.
 
         :param doc: Documents to be extracted.
-        :return: Extracted data.
+        :return: a list of tuples consisted of the values from the fields.
```
cristianmtr, line 149:
```suggestion
        :return: a list of tuples consisting of the values from the fields.
```

nan-wang: good catch!

JoanFM (approved):

PR #1982, opened 2021-02-19
feat(document): add data_uri2blob converter

hanxiao:


JoanFM (changes requested):
This needs to be tested

PR #1983, opened 2021-02-19
test: add test for log formatter

JoanFM:


cristianmtr (approved):

jina/logging/formatter.py
```diff
@@ -6,6 +6,9 @@
 from .profile import used_memory
 from ..helper import colored
 
+if False:
```
Yongxuanzhang, line 9: Just curious how is 'if False' used here

Yongxuanzhang (approved):
LGTM!

jina/logging/formatter.py
JoanFM: It is used to have `flake` not complaint when the class of the `type hint` is not imported while avoid the extra import. 

You can see is a `trick` used all over the place

Yongxuanzhang: Thanks!

PR #1990, opened 2021-02-19
fix: CI should never skip last step

JoanFM:
The CI was successfull and I was allowed to merge even with failing tests.

According to this issue in github runner https://github.com/actions/runner/issues/777 and documentation on github runners I believe this is the way to go


JoanFM:
image (15)

JoanFM:
Propose to add this action https://github.com/marketplace/actions/workflow-conclusion-action

PR #1993, opened 2021-02-20
ci: try to add codecov for daemon

JoanFM:


nan-wang:
It seems not working as expected. https://codecov.io/gh/jina-ai/jina/tree/master/jina

PR #1997, opened 2021-02-20
fix: test a fix, fails locally

JoanFM:


hanxiao (changes requested):
this line is required as Document is recursive, the pydantic model has postpone the building process for it. https://pydantic-docs.helpmanual.io/usage/postponed_annotations/

PR #1999, opened 2021-02-20
fix(document): fix wav type sniff

Yongxuanzhang:
This kind of hardcode check is not the best solution.
Wonder if we can change the audio/wav -> audio/x-wav when we construct the `Documents`?

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -573,6 +573,9 @@ class Document(ProtoTypeMixin):
         """
         if value in mimetypes.types_map.values():
             self._pb_body.mime_type = value
+        if value == 'audio/wav':
+            self._pb_body.mime_type = 'audio/x-wav'
+            return
         elif value:
             # given but not recognizable, do best guess
             r = mimetypes.guess_type(f'*.{value}')[0]
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -571,8 +571,7 @@ class Document(ProtoTypeMixin):
         :param value: the acceptable MIME type, raise ``ValueError`` when MIME type is not
                 recognizable.
         """
-        if value == 'audio/wav':
-            value = 'audio/x-wav'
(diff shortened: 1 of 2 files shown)


jina/types/document/__init__.py
```diff
@@ -571,6 +571,8 @@ def mime_type(self, value: str):
         :param value: the acceptable MIME type, raise ``ValueError`` when MIME type is not
                 recognizable.
         """
+        if value == 'audio/wav':
```
JoanFM, line 574: I agree with ur comment, this is a big code smell.

Who is assigning these mime type in the example?

JoanFM (changes requested):

jina/types/document/__init__.py
Yongxuanzhang: I guess it's different browsers. Check this one. I also tried on firefox and there are no issues.
https://stackoverflow.com/questions/44891157/why-wav-format-doesnt-have-same-mimetype-in-different-browsers

Yongxuanzhang: But the time lenght is not showing properly. lol
image

JoanFM: I have seen that mime type accept an add_mapping or something like this, maybe is a better way to handle this.

Yongxuanzhang: I changed the code to `mimetypes.add_type('audio/wav', '.wav')` in `@mime_type.setter`.Or I create a new function like miietype_loader() to add missing types. But for now I only see 'audio/wav' is missing.

hanxiao (changes requested):
this line should not be in the document

Yongxuanzhang:
> this line should not be in the document

I can think of several solutions atm, like change 'audio/wav' to 'audio/x-wav' in `converter` `guessmimes` function.  Or use mimetypes.add_type('audio/wav', '.wav') like the latest commit. There should be other decent ways...

PR #2000, opened 2021-02-20
test(logging): add profiling test

Yongxuanzhang:


tests/unit/logging/test_profile.py
```diff
@@ -0,0 +1,37 @@
...
+    yield
+    default_logger.logger.propagate = False
+
+
+def test_logging_profile_profiling(caplog, default_logger_propagate):
+    @profiling
+    def foo():
+        print(1)
+    foo()
+    assert "memory" in caplog.text
```
JoanFM, line 21: please use single quote

tests/unit/logging/test_profile.py
```diff
@@ -0,0 +1,37 @@
...
+def default_logger_propagate():
+    default_logger.logger.propagate = True
+    yield
+    default_logger.logger.propagate = False
+
+
+def test_logging_profile_profiling(caplog, default_logger_propagate):
+    @profiling
+    def foo():
+        print(1)
```
JoanFM, line 19: we could allocate some memory here or sleep for some time to showcase how it is used and how the memory and some time is returned.

Tests tend to be good way to clarify behavior

JoanFM (changes requested):

tests/unit/logging/test_profile.py
Yongxuanzhang: Sure! I add some changes. Not sure if this is ok

hanxiao:
for some reason, merging this PR breaks the `test_logging` ci, reverting now

PR #2001, opened 2021-02-21
test(logging): test pipelogger

Yongxuanzhang:
PipeLogger is used in cli -> jina log. I wanted to run 'jina log' to cover the test of PipeLogger but seems doesn't work.


JoanFM (changes requested):
I think the PipeLogger and this option of cli should disappear

hanxiao:
PipeLogger is deprecated and should be removed

PR #2007, opened 2021-02-22
ci: add darglint to ci

bwanglzu:
Add darglint to perform incremental docstring check.


.github/workflows/ci.yml
```diff
@@ -69,6 +69,24 @@ jobs:
...
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - id: files
+        uses: jitterbit/get-changed-files@v1
+      - name: docstring check with darlint
+        run: |
+          pip install darglint
+          for changed_file in ${{ steps.files.outputs.all }}; do
```
cristianmtr, line 85: Where is this coming from?

cristianmtr (changes requested):
Is there a way to test it in this PR with a dummy file that we can then remove?

.github/workflows/ci.yml
JoanFM: from the previous step of the github action, see the `github action workflow` documentation

cristianmtr:
To test

[] add malformed python file
[] add random text file to check that it ignores it

bwanglzu:
added a `test.py`, 
image

cristianmtr:
Issue: it seems like it only checks changed files on the last **sha**, not on entire PR level.

bwanglzu:
@cristianmtr now it should works, can you take a look? I'll remove the unused files.

.github/workflows/ci.yml
```diff
@@ -69,6 +69,27 @@ jobs:
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - id: file_changes
+        uses: jitterbit/get-changed-files@v1
+      - name: docstring check with darlint
+        run: |
+          pip install darglint
+          pip install pydocstyle
+          for changed_file in ${{ steps.file_changes.outputs.all }}; do
```
cristianmtr, line 88: Why did you remove the echo? Now in the check you don't see which file it's referring to

```
Successfully installed pydocstyle-5.1.1 snowballstemmer-2.1.0
removeme.py:1 in public class `X`:
        D101: Missing docstring in public class
Error: Process completed with exit code 1.
```

This can be difficult since we have the same filenames across folders

cristianmtr (changes requested):
Also, can you try one other thing:

run the commands on all the files at the same time so we get the status on all of them? As it is, it stops at the first error.

.github/workflows/ci.yml
```diff
@@ -69,6 +69,27 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - id: file_changes
+        uses: jitterbit/get-changed-files@v1
+      - name: docstring check with darlint
```
cristianmtr, line 84:
```suggestion
      - name: docstring check with darglint and pydocstyle
```

cristianmtr (approved):

Yongxuanzhang (approved):

PR #2011, opened 2021-02-22
test: added more test logging

maximilianwerk:
So many things.
1. python3.7.10 and ubuntu18.04 do not work well together for us. We get errors just from using this combination.
2. refactored all mocks in the way, that the validate provides way better readability in the logs, if a test fails.
3. removed `-n 1` parameter from `pytest` calls. Reasoning: if `-n` is set, `-s -v` will be ignored and we get less verbose logs.
4. fixed some further tests here and there
5. timeout and try again does not work well together in the current setup (and even at all). We should wait for [this](https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/117) feature and move the retry to the github-action itself. I don't see pytest supporting this properly ever. The best workaround I have seen so far is (https://github.community/t/re-run-jobs/16145/11) which is still bad.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -17,7 +17,7 @@ jobs:
       !startsWith(github.event.head_commit.message, 'chore') &&
       !startsWith(github.event.head_commit.message, 'build: hotfix') &&
       !endsWith(github.event.head_commit.message, 'reformatted by jina-dev-bot')
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
@@ -93,7 +93,7 @@ jobs:
 
   update-docker:
     needs: update-doc
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
@@ -150,7 +150,7 @@ jobs:
 
   prep-testbed:
     needs: update-doc
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
@@ -177,7 +177,7 @@ jobs:
 
   core-test:
     needs: prep-testbed
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
@@ -197,6 +197,7 @@ jobs:
           docker pull docker.pkg.github.com/jina-ai/jina/jina:test-pip
           docker tag docker.pkg.github.com/jina-ai/jina/jina:test-pip jinaai/jina:test-pip
           python -m pip install --upgrade pip
+          python -m pip install wheel
           pip install ".[cicd,test,daemon]" --no-cache-dir
           jina check
           export JINA_LOG_LEVEL="ERROR"
@@ -208,14 +209,14 @@ jobs:
           if [[ "${{ matrix.test-path }}" == "tests/distributed/test_against_external_daemon/" ]]; then
             docker build -f Dockerfiles/debianx.Dockerfile -t jinaai/jina:test-daemon .
             docker run --name jinad -v /var/run/docker.sock:/var/run/docker.sock --network=host -d jinaai/jina:test-daemon
-            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml -n 1 --timeout=120 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=180 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
             docker rm -f jinad
           else
             SUB='daemon'
             if [[ "${{ matrix.test-path }}" == *"$SUB"* ]]; then
-              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=daemon --cov-report=xml -n 1 --timeout=120 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=daemon --cov-report=xml --timeout=180 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
             else
-              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml -n 1 --timeout=120 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=180 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
             fi
           fi
           SUB='daemon'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,17 +139,18 @@ jobs:
       - name: Test
         id: test
         run: |
+          pip freeze
           if [[ "${{ matrix.test-path }}" == "tests/distributed/test_against_external_daemon/" ]]; then
             docker build -f Dockerfiles/debianx.Dockerfile -t jinaai/jina:test-daemon .
             docker run --name jinad -v /var/run/docker.sock:/var/run/docker.sock --network=host -d jinaai/jina:test-daemon
-            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml -n 1 --timeout=300 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=300 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
             docker rm -f jinad
           else
             SUB='daemon'
             if [[ "${{ matrix.test-path }}" == *"$SUB"* ]]; then
-              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=daemon --cov-report=xml -n 1 --timeout=300 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=daemon --cov-report=xml --timeout=300 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
             else
-              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml -n 1 --timeout=300 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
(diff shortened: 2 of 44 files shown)


jina/resources/logging.daemon.yml
```diff
@@ -17,14 +16,3 @@ configs:
     host: # when not given then record it locally, /dev/log on linux /var/run/syslog on mac
     port: # when not given then record it locally,  /dev/log on linux /var/run/syslog on mac
     formatter: PlainFormatter
-  FluentHandler:
```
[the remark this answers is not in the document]
maximilianwerk: Added it already back locally again. With the next push it will be back.

scripts/get-all-test-paths.sh
```diff
@@ -6,10 +6,10 @@ declare -a array1=( "tests/unit/*.py" "tests/integration/*.py" "tests/distribute
 declare -a array2=( $(ls -d tests/{unit,integration,distributed}/*/ | grep -v '__pycache__' ))
 dest1=( "${array1[@]}" "${array2[@]}" )
 
-declare -a array1=( "tests/daemon/unit/*.py" "tests/daemon/integration/*.py" )
+declare -a array1=( "tests/daemon/unit/*.py" )
```
JoanFM, line 9: I would leave it to make sure if we add them that they run

JoanFM (commented):
If it works, make sure to apply it to cd. 

Then it would be good to modularize better our workflows

scripts/get-all-test-paths.sh
maximilianwerk: Hmm I am not convinced. This creates a github action for each time we run the tests which uses quite some computation power. Even thought we can use as much github-actions as we like, since we are open source, I'd at least not run meaningless commands the whole time. But that is just my opinion and I can add it back.

JoanFM: are u sure it creates a gthub action it not found?

JoanFM: if it does create the action yes I would remove

maximilianwerk: Yes it creates the action.

JoanFM: then kill it heavily

jina/peapods/zmq/__init__.py
```diff
@@ -276,10 +276,13 @@ async def send_message(self, msg: 'Message', sleep: float = 0, **kwargs):
...
-            self.bytes_recv += msg.size
             self.msg_recv += 1
-            if callback:
-                return callback(msg)
+            if msg is not None:
+                self.bytes_recv += msg.size
+                if callback:
+                    return callback(msg)
+            else:
+                self.logger.error('received and empty message')
```
JoanFM, line 285:
```suggestion
                self.logger.error('received an empty message')
```

JoanFM (approved):

tests/unit/docker/test_hub_build_level.py
```diff
@@ -40,18 +40,18 @@ def test_hub_build_level_pass(monkeypatch, test_workspace, docker_image):
...
+                                               os.path.join(cur_dir, 'yaml/test-joint.yml'), 60000, True,
                                                JinaLogger('unittest'))
 
     assert expected_failed_levels == failed_levels
 
 
 def test_hub_build_level_fail(monkeypatch, test_workspace, docker_image):
     args = set_hub_build_parser().parse_args(['path/hub-mwu', '--push', '--host-info', '--test-level', 'FLOW'])
-    expected_failed_levels = [BuildTestLevel.POD_NONDOCKER, BuildTestLevel.POD_DOCKER, BuildTestLevel.FLOW]
+    expected_failed_levels = [BuildTestLevel.POD_DOCKER, BuildTestLevel.FLOW]
```
maximilianwerk, line 51: Can someone verify, that this change makes sense? @rutujasurve94 @JoanFM
I am not sure about the consequences of this change and only did it, since the test expected it.

JoanFM: Yes I do not see why it should Fail in the other one

.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       - name: Set up Python 3.7
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.7.9
```
JoanFM, line 63: we should test with the minimal supported version I think

jina/peapods/zmq/__init__.py
```diff
@@ -276,10 +276,13 @@ async def send_message(self, msg: 'Message', sleep: float = 0, **kwargs):
...
-            self.bytes_recv += msg.size
             self.msg_recv += 1
-            if callback:
-                return callback(msg)
+            if msg is not None:
+                self.bytes_recv += msg.size
+                if callback:
+                    return callback(msg)
+            else:
+                self.logger.error('Received message is empty.')
```
maximilianwerk, line 285: I am not sure about this one. In the end it is "just" better logging information in CI and will most likely not bring any benefit for real flows. Should I leave it in? Or rather catch an `AttributeError`?

jina/clients/base.py
```diff
@@ -148,7 +149,7 @@ async def _get_results(self,
                                       logger=self.logger)
                         p_bar.update(self.args.request_size)
                         yield resp
-        except KeyboardInterrupt:
+        except (KeyboardInterrupt, CancelledError):
```
maximilianwerk, line 152: This one brings better environment cleanup in CI, when tests are failing.

.github/workflows/ci.yml
```diff
@@ -138,17 +139,18 @@ jobs:
       - name: Test
         id: test
         run: |
+          pip freeze
           if [[ "${{ matrix.test-path }}" == "tests/distributed/test_against_external_daemon/" ]]; then
             docker build -f Dockerfiles/debianx.Dockerfile -t jinaai/jina:test-daemon .
             docker run --name jinad -v /var/run/docker.sock:/var/run/docker.sock --network=host -d jinaai/jina:test-daemon
-            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml -n 1 --timeout=600 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=300 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
```
maximilianwerk, line 146: I will remove the `-n 1`, since CI logs way more verbose when we remove it.

maximilianwerk (commented):
Some questions and clarifications

jina/peapods/zmq/__init__.py
nan-wang: I think it is fine, however, we need to change the function return to `Optional['Message]`

nan-wang (approved):
LGTM👍

tests/distributed/test_against_external_daemon/test_single_instance.py
```diff
@@ -131,7 +131,7 @@ def docker_image():
 
 @pytest.mark.parametrize('silent_log', [True, False])
 @pytest.mark.parametrize('parallels', [1, 2, 3])
-def test_l_r_l_with_upload(silent_log, parallels, docker_image, mocker):
+def test_l_r_l_with_upload_docker(silent_log, parallels, docker_image, mocker):
```
JoanFM, line 134: the docker `suffix` I would remove. We use docker to "simulate" remote instances but is not  a "docker" or "docker pod" feature we are testing

tests/distributed/test_join_local_from_remote/test_integration.py
```diff
@@ -30,21 +31,22 @@ def client():
...
     assert_request(method='get',
                    url=f'http://localhost:8000/flows/{flow_id}')
 
     assert_request(method='delete',
                    url=f'http://localhost:8000/flows/{flow_id}',
                    payload={'workspace': False})
-    m.assert_called_once()
+
+    mock.assert_called_once()
+    validate_callback(mock, validate_resp)
```
florian-hoenicke, line 52: what about providing the validate_callback function an additional parameter to assert how often it is called? `validate_callback(mock, validate_resp, calls)`

tests/integration/encode_driver_batching/test_encode_driver_batching.py
```diff
@@ -159,8 +161,14 @@ def fail_if_error(resp):
...
     with Flow().add(uses=executor_yml_file) as f:
         f.search(input_fn=document_generator(num_docs, num_chunks, num_chunks_chunks),
                  request_size=request_size,
-                 on_done=validate_response,
-                 on_error=fail_if_error)
+                 on_done=on_done_mock,
+                 on_error=on_error_mock)
+
+    validate_callback(on_done_mock, validate_response)
+    on_error_mock.assert_not_called()
```
florian-hoenicke, line 174: validate_callback(on_done_mock, validate_response, calls=0)

florian-hoenicke (dismissed):

.github/workflows/cd.yml
```diff
@@ -208,14 +209,14 @@ jobs:
...
             docker run --name jinad -v /var/run/docker.sock:/var/run/docker.sock --network=host -d jinaai/jina:test-daemon
-            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml -n 1 --timeout=120 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=180 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
             docker rm -f jinad
           else
             SUB='daemon'
             if [[ "${{ matrix.test-path }}" == *"$SUB"* ]]; then
-              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=daemon --cov-report=xml -n 1 --timeout=120 -v --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
+              pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=daemon --cov-report=xml --timeout=180 -v -s --ignore-glob='tests/integration/hub_usage/dummyhub*' ${{ matrix.test-path }}
```
bwanglzu, line 217: when the subfolder is `daemon` we don't need  `--ignore-glob='tests/integration/hub_usage/dummyhub*'`

bwanglzu (changes requested):
minor things, great work!

bwanglzu (approved):

bwanglzu:
not blocking , but  when the subfolder is daemon we don't need --ignore-glob='tests/integration/hub_usage/dummyhub*'

tests/distributed/test_against_external_daemon/test_single_instance.py
maximilianwerk: Reason is, that there is another test with the exact same name, but not using docker. I wanted to distinguish them. So rather `_remote`?

JoanFM: where is that other test?

JoanFM: yes I think rempte shows better the spirit

tests/distributed/test_join_local_from_remote/test_integration.py
maximilianwerk: I thought about this back and forth. Had even implemented it three times and removed it again. Reasoning: Checking the amount of calls is a thing on mock via `mock.assert_called_once()` or `assert mock.call_count == `. I did not wanted to put tooo much magic in this function, but only, that it guarantees, the validate will be called with at least one set of parameters. Anyhow, I can see writing our own standardized mock usability function would create benefit in similar usage across the code base. Would adding it as a follow-up PR make sense?

maximilianwerk:
> not blocking , but when the subfolder is daemon we don't need --ignore-glob='tests/integration/hub_usage/dummyhub*'

I see your point. Let's put this into another ticket to get this merged ASAP.

maximilianwerk:
> I think it is fine, however, we need to change the function return to Optional['Message]

Totally agree. The function signature should have been like this already beforehand... :D

hanxiao (dismissed):

nan-wang (approved):

JoanFM (approved):

hanxiao (approved):

PR #2016, opened 2021-02-23
test(logging): add profiling test

Yongxuanzhang:
The previous test may fail, so change it to make sure it could pass and meanwhile illustrate the functionality.


nan-wang (approved):
LGTM👍

PR #2025, opened 2021-02-23
fix: removing problematic mime type assignment

davidbp:


JoanFM (dismissed):
Let's make sure tests pass, but good to go if they do

JoanFM (approved):

PR #2037, opened 2021-02-24
feat: add support for multistage dockerfile

nan-wang:
- [x] add support for multistag dockerfile
- [x] refactoring the way of adding labels to the docker images as suggested in #1732


tadejsv:
Any chance modifying the Dockerfile could be replaced by passing `labels` to docker SDK call :) ?

JoanFM:
> Any chance modifying the Dockerfile could be replaced by passing `labels` to docker SDK call :) ?

I hope we will find the time to do this

tests/unit/docker/test_hub_build.py
```diff
@@ -86,3 +86,32 @@ def test_jina_version_freeze_no_jina_dependency(requirements, tmpdir):
...
+
+    with pytest.raises(BaseException, match='labels all good'):
+        hubio.build()
+
+
+@pytest.mark.timeout(360)
+def test_hub_build_multistage():
+    args = set_hub_build_parser().parse_args(
+        [os.path.join(cur_dir, 'hub-mwu-multistage'), '--test-uses', '--raise-error'])
+    HubIO(args).build()
```
JoanFM, line 117: this should go in integration test

nan-wang: fair point

nan-wang:
> Any chance modifying the Dockerfile could be replaced by passing `labels` to docker SDK call :) ?

@tadejsv  tYes, I've adopted your PR #1732 into this one. ❤️

jina/docker/hubio.py
```diff
@@ -309,15 +309,18 @@ def build(self) -> Dict:
 
             with TimeContext(f'building {colored(self.args.path, "green")}', self.logger) as tc:
                 try:
-                    self._check_completeness()
+                    _check_result = self._check_completeness()
                     self._freeze_jina_version()
 
+                    _dockerfile_path = _check_result['Dockerfile']
```
tadejsv, line 315: So I think here it should be `_dockerfile_path = 'Dockerfile'`, or alternatively something like `_check_result['Dockerfile'].split('/')[-1]` (to be better prepared for different dockerfile names, such as the one for GPU).

This is because in the call to the client below, `path=self.args.path` is passed, which means that the `dockerfile` location will be relative to that `path`. However in `_check_completness` the dockerfile path is obtained by joining `self.args.path` and `Dockerfile`. I think this was the reason that #1732 broke the hub builds.

tadejsv: Also, back then the issue was that CI didn't catch this error for some reason. I hope now it will

nan-wang: hmm, the tests actually work locally. I've put them into the integration tests and let's see how ci says.

tadejsv: Aaaah I figured it out! 

The reason test worked and hub build broke was because in tests `args.path` is specified with an absolute path - for example `hub_build_test` uses `os.path.join(cur_dir, 'hub-mwu')`, where `cur_dir` is an absolute path, while building in Jina Hub is done using relative paths, e.g. `encoders/audio/MFCCTimbreEncoder/`. 

Tests worked absolutely fine, but they were only done for the absolute path - where both `path` and `dockerfile` are absolute paths. If both are relative, then docker SDK tries to join them together, which doesn't work unless `dockerfile` is really just the name of the file inside `args.path`.

So yes, extracting the path using `os.path.basename` as you are doing now is the right way (btw you have to do it in another location in the file, that's why the CI is failing). All that should be added is a test where `args.path` is passed as a relative path, and we are good to go :D

nan-wang: Great catch! Added another test on the relative pathes.

tests/integration/hub_usage/test_hub_usage.py
```diff
@@ -148,3 +148,11 @@ def test_hub_build_push_push_again(monkeypatch, mocker):
         # try and push same version again should fail with `--no-overwrite`
         args = set_hub_build_parser().parse_args([str(cur_dir) + '/hub-mwu', '--push', '--host-info', '--no-overwrite'])
         HubIO(args).build()
+
+
+@pytest.mark.timeout(360)
+@pytest.mark.parametrize('dockerfile_path', [os.path.join(cur_dir, 'hub-mwu-multistage'), 'hub-mwu-multistage'])
```
tadejsv, line 154: Try replacing this with
```python
hub_mwu_ms_abspath = os.path.join(cur_dir, 'hub-mwu-multistage')

@pytest.mark.timeout(360)
@pytest.mark.parametrize('dockerfile_path', [hub_mwu_ms_abspath, os.path.relpath(hub_mwu_ms_abspath, os.getcwd())])
```

I think now the CI for relative path is failing due to difference between `cur_dir` and `cwd`, this should fix it.

tests/integration/hub_usage/test_hub_usage.py
```diff
@@ -148,3 +151,13 @@ def test_hub_build_push_push_again(monkeypatch, mocker):
         # try and push same version again should fail with `--no-overwrite`
         args = set_hub_build_parser().parse_args([str(cur_dir) + '/hub-mwu', '--push', '--host-info', '--no-overwrite'])
         HubIO(args).build()
+
+
+@pytest.mark.timeout(360)
+@pytest.mark.parametrize('dockerfile_path',
+                         [os.path.join(cur_dir, 'hub-mwu-multistage'),
+                          os.path.relpath('hub-mwu-multistage', os.getcwd())])
```
tadejsv, line 159: I think this line should be `os.path.relpath(os.path.join(cur_dir, 'hub-mwu-multistage'), os.getcwd())`, then it should work

tests/integration/hub_usage/test_hub_usage.py
nan-wang: fixed

tests/integration/hub_usage/test_hub_usage.py
nan-wang: fixed

PR #2038, opened 2021-02-24
fix: delete invalid entries before dumping indexer

JoanFM:


jina/executors/indexers/vector.py
```diff
@@ -59,6 +60,7 @@ def __init__(self,
             self._size = ref_indexer._size
             # point to the ref_indexer.index_filename
             # so that later in `post_init()` it will load from the referred index_filename
+            self.valid_indices = ref_indexer.valid_indices
```
JoanFM: if this does not have to be serialized, it should go in post init

jina/executors/indexers/vector.py
```diff
@@ -70,15 +72,36 @@ def __init__(self,
...
+            os.remove(self.index_abspath)
+            os.rename(self.index_abspath+'tmp', self.index_abspath)
+
+            valid_key_bytes = np.frombuffer(self.key_bytes, dtype=(np.str_, self.key_length))[self.valid_indices].tobytes()
+            self.key_bytes = valid_key_bytes
+            self._size = len(valid)
+            self.valid_indices = valid
+            # self._int2ext_id = self._int2ext_id[valid]
+
+    def __setstate__(self, d):
```
JoanFM, line 101: why do we need to override?

JoanFM: I see why, but isnit it better to delete on store? why would we store data we actually do not want?

cristianmtr:
Will open separate PR

PR #2042, opened 2021-02-25
test(checker): add importchecker test

Yongxuanzhang:


tests/unit/test_checker.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def test_importchecker(tmpdir):
+    args = argparse.Namespace()
+    args.cli = 'check'
+    tmp_exec_file = os.path.join(str(tmpdir), 'tmp_exec.yml')
+    tmp_exec_driver = os.path.join(str(tmpdir), 'tmp_driver.yml')
+    args.summary_exec = tmp_exec_file
+    args.summary_driver = tmp_exec_driver
+    ImportChecker(args)
```
JoanFM, line 14: what is being tested here?

Yongxuanzhang: I want to test ImportChecker from checker.py and _print_dep_tree_rst from importer.py. ImportChecker will call _print_dep_tree_rst to write executors and drivers to tmp files. 
What could be the best practice do you think to test it?
Like, assert the content in text files? Or is it possible to use mock?

JoanFM: did not u say it is not used anywhere?

U could do check a little the text file or use a mock and see the arguments passed.

Yongxuanzhang: I thought it's a private function, but actually, it's used outside of that file. And usually, we will do `Jina check` without these `summary_exec`, `summary_driver`. I think that's the reason they are not tested before. But they offer the ability to output executors, drivers to files.

JoanFM: so lets test the output of the functionality

Yongxuanzhang: Sure. Will do!

PR #2056, opened 2021-02-26
feat: enable to select dockerfile

### LINKED ISSUE #1822: Enable the user to choose which Dockerfile to use
**Describe the feature**
`jina hub build` does **NOT** offer the option for choosing Dockfile 

**Your proposal**
`jina hub build --file gpu.Dockfile`
`jina hub build -f gpu.Dockfile`

---

**Environment**

**Screenshots**

nan-wang:
Add support to `jina hub build -f gpu.Dockerfile my/awesome/encodeer`
resolved #1822 
- [x] Add feature and unit tests
- [x] Add integration tests
- [x] Refactoring the integration tests to avoid having multiple `hub-mwu`


jina/parsers/hub/build.py
```diff
@@ -14,6 +14,8 @@ def mixin_hub_build_parser(parser):
                                            'zero or more yaml config, '
                                            'zero or more Python file. '
                                            'All files in this directory will be shipped into a Docker image')
+    gp.add_argument('-f', '--file', type=str, default='Dockerfile',
```
cristianmtr, line 17: I think you can make the argument of type file. Can we try that?

cristianmtr (approved):

jina/parsers/hub/build.py
JoanFM: the argument must be a file right? since it may come from CLI u would only be able to pass str but no file object?

nan-wang: good point. I'll fix this.

PR #2064, opened 2021-03-01
test(indexer): add exception test

Yongxuanzhang:


bwanglzu:
@Yongxuanzhang dont copy-past, use pytest [parameterized test](https://docs.pytest.org/en/stable/parametrize.html)

Yongxuanzhang:
> @Yongxuanzhang dont copy-past, use pytest [parameterized test](https://docs.pytest.org/en/stable/parametrize.html)

Which ones should we use  pytest parameterized test?

Yongxuanzhang:
Will close this one if we want to exclude exceptions

PR #2098, opened 2021-03-01
feat: add single decorator to guarantee no batching

JoanFM:
**Changes introduced**
This adds a decorator to ensure that input to a function is given one instance at a time.

This can help having the same Driver for executors with and without batching.

Idea being: 
- Driver enables the usage of batching.
- Is up to the executor to decorate themselves with @batching or @single to decide if they work on batches or one instance at a time.

Even if the feature may not last, it can be useful to enable the feature in the driver and keep migrating executors slowly into new batching mode without blocking the changes in the core


hanxiao:
isn't single just a shorthand for `@batching(batch_size=1)`?

jina/executors/decorators.py
```diff
@@ -381,3 +396,54 @@ def arg_wrapper(*args, **kwargs):
...
+                    final_result.append(r)
+
+            return _merge_results_after_batching(final_result, merge_over_axis)
+
+        return arg_wrapper
+
+    if func:
+        return _single(func)
+    else:
+        return _single
```
Yongxuanzhang, line 449: Just curious is it possible/or under which situation to reach this line? For example, not decorating a function?

Yongxuanzhang:
> isn't single just a shorthand for `@batching(batch_size=1)`?

I guess it's different? For example, input is [1,1,1,1]. 
`@batching(batch_size=1)` -> [[1], [1], [1], [1]]
`@single` -> [1, 1, 1 ,1]

Am I correct? @JoanFM . And is this what we want for single?

JoanFM:
> > isn't single just a shorthand for `@batching(batch_size=1)`?
> 
> I guess it's different? For example, input is [1,1,1,1].
> `@batching(batch_size=1)` -> [[1], [1], [1], [1]]
> `@single` -> [1, 1, 1 ,1]
> 
> Am I correct? @JoanFM . And is this what we want for single?
> If this is the case, are there any benefits of this decorating? I feel like the input and output is the same as batching without batch_size. And if we use @single we will iterate the data one by one and still return the data as one single instance.

exactly

JoanFM:
> isn't single just a shorthand for `@batching(batch_size=1)`?

no, it guarantees that it arrives as a single instance and not as an array of size 1

jina/executors/decorators.py
```diff
@@ -381,3 +396,51 @@ def arg_wrapper(*args, **kwargs):
         return _batching(func)
     else:
         return _batching
+
+
+def single(func: Callable[[Any], np.ndarray] = None,
+           split_over_axis: int = 0,
```
nan-wang, line 402: `split_over_axis` is not used.

JoanFM: You are right, removed

nan-wang (approved):
LGTM👍

PR #2100, opened 2021-03-01
fix: hanging ci tests and one retry works.

maximilianwerk:
Still not perfect, but I want to test, if it does the job in CI. The problem is:

The outer most shell of Flow as a ContextManager was not properly set up. When there was an Exception during `__enter__` the state was not cleaned up and hanging processes in some abandoned Pods were the consequence.


jina/flow/base.py
```diff
@@ -424,7 +424,21 @@ def __call__(self, *args, **kwargs):
         return self.build(*args, **kwargs)
 
     def __enter__(self):
-        return self.start()
+        class CatchAllCleanupContextManager():
```
JoanFM, line 427: a try catch exit should be enough right?

maximilianwerk: No it is not. The point of the context manager is, that it will execute the cleanup code, even thought the main code is interrupted

.github/workflows/ci.yml
```diff
@@ -142,14 +142,14 @@ jobs:
           if [[ "${{ matrix.test-path }}" == "tests/distributed/test_against_external_daemon/" ]]; then
             docker build -f Dockerfiles/debianx.Dockerfile -t jinaai/jina:test-daemon .
             docker run --name jinad -v /var/run/docker.sock:/var/run/docker.sock --network=host -d jinaai/jina:test-daemon
-            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=300 -v -s ${{ matrix.test-path }}
+            pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --cov=jina --cov-report=xml --timeout=180 -v -s ${{ matrix.test-path }}
```
JoanFM, line 145: do not set to such a low value, I have seen locally tests taking more than 250 seconds

JoanFM (changes requested):

.github/workflows/ci.yml
maximilianwerk: In core? or in the hub? It was set to 120 until two weeks ago. I prefer setting it lower again and increase if we see things failing again.

JoanFM: in core, see the PR I opened where this changed, I saw locally tests running in 250 seconds. 180 is not realistic

maximilianwerk: But this timeout is per test and not for all 25 tests as in your PR.

JoanFM: no I meant for each test

maximilianwerk: The timeout you are referring to is `timeout-minutes: 15`

JoanFM: See #2057

maximilianwerk: Yes. There 25 tests pass in 251 seconds. And this is the total time of all tests. But the 180s is the time per test.

JoanFM: ah okey

hanxiao (approved):

PR #2101, opened 2021-03-01
ci(codecov): exclude lines

Yongxuanzhang:
This ticket aims at removing the `codecov` detection on some specific lines, (e.g. NotImplementedError). For this error in the base class, we don't have to write tests to cover it. Exclude them will also increase the test coverage.

File introduced: [.coveragerc](https://coverage.readthedocs.io/en/latest/config.html)

1. Can we exclude these exceptions?
2. Can we put this file in the root folder?
3. If so, any other we codes want to exclude?


.coveragerc
```diff
@@ -0,0 +1,5 @@
+# .coveragerc to control coverage.py
```
JoanFM, line 1: is this supposed to affect the coverage locally or the coverage in codecov. If it is locally, u should try to create 2 reports with and without this file and see if there is difference

Yongxuanzhang: Should be " coverage in codecov".

Yongxuanzhang: But this may not be the correct way to configure

Yongxuanzhang: related issue: https://github.com/codecov/codecov-python/issues/107

Yongxuanzhang:
Seems like it could work, not sure if this file can work in other folders. I think it is better if I put it under .github?
[Report](https://codecov.io/gh/jina-ai/jina/src/1eda9df0c1b0da54dd8405b0270847f0359edd6c/jina/executors/indexers/__init__.py)
image

bwanglzu:
@Yongxuanzhang i think it's better to be placed in `.github`, btw can you add a short description to the ticket?

Yongxuanzhang:
> @Yongxuanzhang i think it's better to be placed in `.github`, btw can you add a short description to the ticket?

1. I tried to put in `.github` but didn't work. 
2. Sure

JoanFM:
> Seems like it could work, not sure if this file can work in other folders. I think it is better if I put it under .github?
> @bwanglzu @JoanFM
> [Report](https://codecov.io/gh/jina-ai/jina/src/1eda9df0c1b0da54dd8405b0270847f0359edd6c/jina/executors/indexers/__init__.py)
> image

I think is okey wherever u want

florian-hoenicke:
I think in general, it is needed to exclude lines.
For this specific case, I think it can be tested.
```
with pytest.raises(NotImplementedError):
    obj.my_method()
```

bwanglzu:
@florian-hoenicke I personally think it makes no sense to unit test abstract methods, since the "real logic" will be implemented in subclasses.

bwanglzu (approved):
LGTM, let's see how other colleagues think about it.

JoanFM:
> > LGTM, let's see how other colleagues think about it.
> 
> Sure! I will leave it to receive more comments.

To me is good. If u find is better to keep codecov.yml in the root of the project feel free to open a PR

PR #2104, opened 2021-03-02
test: crud api

florian-hoenicke:
Besides the tests, this pr contains fixes for the REST API as well.
  - [x] changed `from starlette.responses import StreamingResponse` to `from fastapi.responses import StreamingResponse`
 - [x] added 'text', 'embedding' and 'buffer' to the pydantic model.
 - [x] remove none values from the requests


tests/integration/crud/rest/test_rest.py
```diff
@@ -0,0 +1,205 @@
...
+    else:
+        path = Path(os.environ['JINA_REST_DIR'])
+    bin_files = list(path.glob('*.bin'))
+    assert len(bin_files) > 0
+    for index_file in bin_files:
+        index = BaseIndexer.load(str(index_file))
+        assert index.size == num_indexed_docs
+
+
+@pytest.mark.parametrize('flow_file, compound', [
```
nan-wang, line 93: `compound` variable is not necessary. It is coupled with the choice of flow configuration file.
```suggestion
@pytest.mark.parametrize('flow_file, compound', [
```

nan-wang (changes requested):

nan-wang:
BTW, I just found `executors.requests.CompoundIndexer.yml` is out-dated. The `UpdateRequest` part is wrong.

jina/peapods/runtimes/asyncio/rest/app.py
```diff
@@ -199,13 +200,13 @@ async def delete_api(body: JinaDeleteRequestModel):
...
         """
-        Collect the protobuf message converts protobuf message to a dictionary.
+        Collect the protobuf message converts protobuf messages to strings.
 
         :param req_iter: request iterator
         :yield: result
         """
         async for k in servicer.Call(request_iterator=req_iter, context=None):
-            yield MessageToDict(k)
+            yield json.dumps(MessageToDict(k))
```
JoanFM, line 209: why do we want a string and not a dict?

JoanFM: I think the proper way is to return a `dictionary` and let the processing into documents to the client.

JoanFM:
Hey @florian-hoenicke,

would you provide some information and context related to the changes, since according to the title it seems to be just a test, but some changes are seen

jina/peapods/runtimes/asyncio/rest/app.py
JoanFM: This is also a potential breaking change if someone was using its own client?

florian-hoenicke: Actually the api endpoint is broken on master. This one of the fix I did
I could not get it to run returning dictionaries in the StreamingResponse.
Reproducting steps:
```
from jina import Document
with Flow(restful=True, port_expose=8089).add() as f:
    f.block()
```
Go to `http://localhost:8089/docs#/CRUD/index_api_index_post` -> try out -> execute

florian-hoenicke:
> Hey @florian-hoenicke,
> 
> would you provide some information and context related to the changes, since according to the title it seems to be just a test, but some changes are seen

Yes, you are right. I added the required information to the description

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -51,6 +55,7 @@ class JinaRequestModel(BaseModel):
         Field(..., example=[Document().dict()])
     request_size: Optional[int] = default_request_size
     mime_type: Optional[str] = ''
+    # TODO mime type is contained in documents already
```
cristianmtr, line 58: So is this `mime_type` used? I can't see how

florian-hoenicke: I also don't see a reason to keep it there and I will create more tickets for further issues I discovered.

tests/integration/crud/rest/test_rest.py
florian-hoenicke: Thanks, good idea! I removed it and always look in both directories when doing the validation.

florian-hoenicke:
> BTW, I just found `executors.requests.CompoundIndexer.yml` is out-dated. The `UpdateRequest` part is wrong.

I had a look and could not find what would be the issue with the `UpdateRequest`.
We use the VectorIndexDriver for updates when setting `mothod: update`. Is that what you mean? In that case it is as expected.

  on:
    UpdateRequest:
      - !VectorIndexDriver
        with:
          method: update
          executor: vecidx
          traversal_paths: [ 'r' ]
      - !KVIndexDriver
        with:
          method: update
          executor: docidx
          traversal_paths: [ 'r' ]
```

File: jina/peapods/runtimes/asyncio/rest/app.py (hunk 2 above)
Comment (deepankarm) [reply jina/peapods/runtimes/asyncio/rest/app.py:209]:
```suggestion
            yield MessageToJson(k)
```

deepankarm: Also, please add `media_type` to all `StreamingResponse`
```python
return StreamingResponse(result_in_stream(request_generator(**bd)),
                                 media_type="application/json")
```

jina/peapods/runtimes/asyncio/rest/models.py
deepankarm: Please note that, this follows the caller arguments from the [request_generator](https://github.com/jina-ai/jina/blob/master/jina/clients/request/__init__.py#L24)

deepankarm (changes requested):

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -24,12 +24,16 @@ def build_model_from_pb(name: str, pb_model: Callable):
...
-
     dp = MessageToDict(pb_model(), including_default_value_fields=True)
-
     all_fields = {k: (name if k in ('chunks', 'matches') else type(v), Field(default=v)) for k, v in dp.items()}
     if pb_model == QueryLangProto:
         all_fields['parameters'] = (Dict, Field(default={}))
+    if pb_model == DocumentProto:
+        # these fields are defined as oneof in the jina.proto
+        # therefore, they are not instantiated when using pb_model()
+        all_fields['buffer'] = (str, Field()) #base64 encoded
```
deepankarm, line 34: Any changes in proto would need a change here. This method should be generic & independent on proto level fields. I would fix this in a different PR.

nan-wang:
> > BTW, I just found `executors.requests.CompoundIndexer.yml` is out-dated. The `UpdateRequest` part is wrong.
> 
> I had a look and could not find what would be the issue with the `UpdateRequest`.
> We use the VectorIndexDriver for updates when setting `mothod: update`. Is that what you mean? In that case it is as expected.
> 
> on:
> UpdateRequest:
> - !VectorIndexDriver
> with:
> method: update
> executor: vecidx
> traversal_paths: [ 'r' ]
> - !KVIndexDriver
> with:
> method: update
> executor: docidx
> traversal_paths: [ 'r' ]

Yes, this is what I mean

deepankarm:
@florian-hoenicke I have fixed the issues with the response and the pydantic models in https://github.com/jina-ai/jina/pull/2121. We can focus only on tests in this PR.

florian-hoenicke:
@deepankarm really cool. Let's merge your pr first.

nan-wang (dismissed):
LGTM👍

jina/peapods/runtimes/asyncio/rest/app.py
```diff
@@ -18,6 +18,21 @@
 from .....types.request import Request
 
 
+# TODO instead of this function make sure that only one 'oneof field' is used.
```
deepankarm, line 21: Pydantic's `oneof` implementation is based on python `Union` and represents something different than the protobuf implementation. We need to have custom logic in `protobuf_to_pydantic_model` to make `oneof` pick only one field.

https://github.com/samuelcolvin/pydantic/issues/656
https://github.com/samuelcolvin/pydantic/issues/619

florian-hoenicke: Good to see that others are interested in this issue as well. 
Until they don't support this feature, we have to go with this approach I guess.

florian-hoenicke:
I fixed an issue with the `oneof` type in the protobuf definition.
The problem was that always all optional fields were initialized in the pydantic model.
Providing more than one of the optional fields in the document creation makes it fail.
As discussed with @deepankarm, there are feature requests to add the support for a more elegant solution.
https://github.com/samuelcolvin/pydantic/issues/656
https://github.com/samuelcolvin/pydantic/issues/619

Validating the request to check that the client obeys the oneof rules should be done in a separate ticket.
https://github.com/jina-ai/jina/issues/2132

deepankarm (approved):

PR #2102, opened 2021-03-02
feat(binarypb): delete on dump

cristianmtr:
- [x] basic tests pass
- [x] add benchmark

---

Benchmark

delete_on_dump=True, entries=100000. took 1.7688062191009521 seconds
delete_on_dump=False, entries=100000. took 0.001649618148803711 seconds

Code: 

```python
@pytest.mark.skipif('GITHUB_WORKFLOW' in os.environ, reason='skip the network test on github workflow')
@pytest.mark.parametrize('delete_on_dump', [True, False])
def test_binarypb_benchmark(test_metas, delete_on_dump):
    entries = 100000
    nr_to_update = 10000
    keys = np.arange(entries)
    values = np.random.randint(0, 10, size=entries).astype(bytes)

    with BinaryPbIndexer(metas=test_metas, delete_on_dump=delete_on_dump) as idxer:
        idxer.add(keys, values)
        idxer.save()
        assert idxer.size == entries
        save_abspath = idxer.save_abspath

    new_values = np.random.randint(0, 10, size=nr_to_update).astype(bytes)

    with BaseIndexer.load(save_abspath) as idxer:
        idxer.update(keys[:nr_to_update], new_values)
        time_now = time.time()
        idxer.save()

    time_end = time.time()
    print(f'{delete_on_dump=}, entries={entries}. took {time_end - time_now} seconds')

```

PR #2106, opened 2021-03-02
refactor(binarypb): handlers are context managers

cristianmtr:
To be merged only **after** https://github.com/jina-ai/jina/pull/2102

### DIFF
--- a/jina/executors/indexers/keyvalue.py
+++ b/jina/executors/indexers/keyvalue.py
@@ -14,7 +14,10 @@ HEADER_NONE_ENTRY = (-1, -1, -1)
 
 
 class BinaryPbIndexer(BaseKVIndexer):
-    """Simple Key-value indexer."""
+    """Simple Key-value indexer that writes to disk
+
+    :param delete_on_dump: whether to delete the entries that were marked as 'deleted'
+    """
 
     class WriteHandler:
         """
@@ -28,12 +31,36 @@ class BinaryPbIndexer(BaseKVIndexer):
             self.body = open(path, mode)
             self.header = open(path + '.head', mode)
 
-        def close(self):
-            """Close the file."""
+        def __init__(self, path, mode):
+            self.path = path
+            self.mode = mode
+            print(f'### writehandler mode = {self.mode}')
+
+        def __enter__(self):
+            print(f'### WriteHandler enter. mode = {self.mode}, path = {self.path}')
+            self.body = open(self.path, self.mode)
+            self.header = open(self.path + '.head', self.mode)
+            return self
+
+        def __exit__(self, exc_type, exc_val, exc_tb):
+            self._flush()
+            self._close()
+
+        def _close(self):
+            # TODO do we need to make sure
             self.body.close()
             self.header.close()
 
-        def flush(self):
+        def close(self):
+            """Close the file."""
+            if getattr(self, 'body', None) and not self.body.closed:
+                self.body.flush()
+                self.body.close()
+            if getattr(self, 'header', None) and not self.header.closed:
+                self.header.flush()
+                self.header.close()
+
+        def _flush(self):
             """Clear the body and header."""
             self.body.flush()
             self.header.flush()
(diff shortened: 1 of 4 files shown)


jina/executors/indexers/keyvalue.py
```diff
@@ -2,9 +2,11 @@
 __license__ = "Apache-2.0"
 
 import mmap
+import os
 from typing import Iterable, Optional
 
 import numpy as np
+from tensorflow.python.autograph.pyct.anno import keys
```
JoanFM, line 9: tensorflow is not a basic dependency of us

cristianmtr: Don't know how that got in there. Some PyCharm auto-import maybe

jina/executors/indexers/keyvalue.py
```diff
@@ -13,24 +14,49 @@
...
+
+        def _close(self):
+            # TODO do we need to make sure
             self.body.close()
             self.header.close()
 
-        def flush(self):
+        def close(self):
+            """Close the file."""
+            if self.__dict__.get('body') and not self.body.closed:
```
JoanFM, line 52: more pythonic way is to do `getattr(self, body, None)`

jina/executors/indexers/keyvalue.py
```diff
@@ -13,24 +14,49 @@
...
+                self.body.flush()
+                self.body.close()
+            if self.__dict__.get('header') and not self.header.closed:
```
JoanFM, line 55: same here

cristianmtr:
Closing this, as it is not possible to have context handlers Read/Write Handlers in BinaryPb.

Why?

We are relying on `write_handler`, `query_handler` fields to be callable in the `close` of the `BaseIndexer`, with a `.close()` call of their own. This would not be possible in the ctx mger approach.

Also, using a ctx mger is less efficient for `query`, as we would need to always rebuild the `query_handler` for every `query` call.

JoanFM:
> Closing this, as it is not possible to have context handlers Read/Write Handlers in BinaryPb.
> 
> Why?
> 
> We are relying on `write_handler`, `query_handler` fields to be callable in the `close` of the `BaseIndexer`, with a `.close()` call of their own. This would not be possible in the ctx mger approach.
> 
> Also, using a ctx mger is less efficient for `query`, as we would need to always rebuild the `query_handler` for every `query` call.

Are we sure this is impossible? Isn't maybe the problem that BaseIndexer is assuming something that is not true?

cristianmtr:
> > Closing this, as it is not possible to have context handlers Read/Write Handlers in BinaryPb.
> > Why?
> > We are relying on `write_handler`, `query_handler` fields to be callable in the `close` of the `BaseIndexer`, with a `.close()` call of their own. This would not be possible in the ctx mger approach.
> > Also, using a ctx mger is less efficient for `query`, as we would need to always rebuild the `query_handler` for every `query` call.
> 
> Are we sure this is impossible? Isn't maybe the problem that BaseIndexer is assuming something that is not true?

Could be. If we do go down that path, it would require a rewrite of `BaseIdxer` though. Maybe that's a bit outside our current scope

Also, the statement about speed/performance would still hold. We would be opening the ReadHandler on every update, for each document

JoanFM:
> > > Closing this, as it is not possible to have context handlers Read/Write Handlers in BinaryPb.
> > > Why?
> > > We are relying on `write_handler`, `query_handler` fields to be callable in the `close` of the `BaseIndexer`, with a `.close()` call of their own. This would not be possible in the ctx mger approach.
> > > Also, using a ctx mger is less efficient for `query`, as we would need to always rebuild the `query_handler` for every `query` call.
> > 
> > 
> > Are we sure this is impossible? Isn't maybe the problem that BaseIndexer is assuming something that is not true?
> 
> Could be. If we do go down that path, it would require a rewrite of `BaseIdxer` though. Maybe that's a bit outside our current scope
> 
> Also, the statement about speed/performance would still hold. We would be opening the ReadHandler on every update, for each document

there may be a way to keep it as context manager and load it only when needed, but maybe it needs further refactoring

JoanFM:
At least we need to surround the usage with try catches

PR #2107, opened 2021-03-02
refactor: try to remove reduce driver

JoanFM:


JoanFM:
Ok to remove ReduceDriver and keep the usage of _merge backward compatible we should make it work as _pass.

However, I think is betrer to have _merge work as _merge_root and have explicitly _pass when wanted.

What do u think @nan-wang ?

tests/distributed/test_join_local_from_remote/flow.yml
```diff
@@ -20,7 +20,7 @@ pods:
     port_expose: 8000
     needs: [gateway]
   - name: join_all
-    uses: _merge
+    uses: _pass
```
nan-wang, line 23: What's the motivation for changing this? I think `_merge` also works.

nan-wang (changes requested):

tests/distributed/test_join_local_from_remote/flow.yml
JoanFM: To keep the usage as before. The difference being.

If _pass is used, the request of a branch is passed to the gateway as they come, if use _merge, it waits for the other part

bwanglzu:
what is the reason for removing `ReduceDriver`, can you.give a bit of context?

JoanFM:
> what is the reason for removing `ReduceDriver`, can you.give a bit of context?

It is very opinionated, but it seems to me missleading to see `_merge` (alias of Reduce) do the same as `Route`? I expect _merge to behave as `_merge_root`. I added an integration test that I think can make more clear what is the different behaviors between these drivers

jina/resources/executors._merge.yml
```diff
@@ -4,7 +4,11 @@ metas:
   name: merge
 requests:
   on:
-    [SearchRequest, TrainRequest, IndexRequest, DeleteRequest, UpdateRequest]:
-      - !ReduceDriver {}
+    [SearchRequest, TrainRequest, IndexRequest]:
+      - !ReduceAllDriver
+        with:
+          traversal_paths: ['r']
```
nan-wang, line 10: I think this should be 'c' by default
```suggestion
          traversal_paths: ['r']
```

tests/distributed/test_join_local_from_remote/test_integration.py
```diff
@@ -33,8 +33,9 @@ def client():
 def test_flow(docker_compose, doc_to_index, client, mocker):
 
     def validate_resp(resp):
-        assert len(resp.search.docs) == 1
+        assert len(resp.search.docs) == 2
```
nan-wang, line 36: This is wrong because the reducing by default takes place on `r`

tests/integration/merge_pass/test_merge_pass_messages.py
```diff
@@ -0,0 +1,72 @@
...
+    def validate(resp):
+        assert len(resp.index.docs) == 1
+        # only the second part of the message is passed by _pass
+        assert resp.index.docs[0].tags['id'] == 'slow'
+
+    f = Flow().add(name='pod0', uses='_pass').add(name='pod1', uses='!SlowCrafter', needs=['gateway']).add(name='pod2',
+                                                                                                           uses='_pass',
+                                                                                                           needs=[
+                                                                                                               'pod0',
+                                                                                                               'pod1'])
```
nan-wang, line 26:
```suggestion
    f = (Flow()
         .add(name='pod0', uses='_pass')
         .add(name='pod1', uses='!SlowCrafter', needs=['gateway'])
         .add(name='pod2', uses='_pass', needs=['pod0', 'pod1']))
```

jina/resources/executors._merge.yml
nan-wang: using `r` makes it identical as `_merge_root`

nan-wang (changes requested):

nan-wang:
> > what is the reason for removing `ReduceDriver`, can you.give a bit of context?
> 
> It is very opinionated, but it seems to me missleading to see `_merge` (alias of Reduce) do the same as `Route`? I expect _merge to behave as `_merge_root`. I added an integration test that I think can make more clear what is the different behaviors between these drivers

I see your points. In this case, I don't see the value of keeping `_merge` as we already have all sets of `_merge_chunks`, `_merge_root` and etc. What about removing `_merge`?

JoanFM:
> > > what is the reason for removing `ReduceDriver`, can you.give a bit of context?
> > 
> > 
> > It is very opinionated, but it seems to me missleading to see `_merge` (alias of Reduce) do the same as `Route`? I expect _merge to behave as `_merge_root`. I added an integration test that I think can make more clear what is the different behaviors between these drivers
> 
> I see your points. In this case, I don't see the value of keeping `_merge` as we already have all sets of `_merge_chunks`, `_merge_root` and etc. What about removing `_merge`?

Yes, I would remove it, just thought about keeping for backwards compatibility

jina/resources/executors._merge.yml
JoanFM: That is the point

tests/integration/merge_pass/test_merge_pass_messages.py
```diff
@@ -0,0 +1,89 @@
...
+
+    mock.assert_called_once()
+    validate_callback(mock, validate)
+
+
+def test_flow_merge_chunks(mocker):
+    def validate(resp):
+        assert len(resp.index.docs) == 1
+
+    f = Flow().add(name='pod0', uses='_pass').add(name='pod1', uses='_pass', needs=['gateway']).add(name='pod2',
```
bwanglzu, line 78: shall we factor out all the flows in tests and turn it into a fixture with parameters?

JoanFM: can u do a PR on top of this branch with this?

bwanglzu: ignore it.. didn't notice it's an integration test

JoanFM:
@nan-wang can this be merged?

nan-wang (approved):
LGTM👍

PR #2110, opened 2021-03-03
refactor: prepare changes to have batching for every executor

JoanFM:
**Changes introduced**
The changes introduced include:

- [x] Changes in the way attributes are accessed in `Document`
- [x] Availability to extract full set of attributes from `DocumentSet` (beyond only embeddings so that it can be used for batching everywhere)
- [x] Extract multiple content from document in different batches (if I extract text and id from DocumentSet I want to extract a batch of text in one side and a batch of id in another)
- [x] Better testing and flat results for batching
- [x] Start integration test showing how different executors should implement batching


jina/types/document/__init__.py
```diff
@@ -506,18 +538,61 @@ def get_attrs(self, *args) -> Dict[str, Any]:
                 ret[k] = None
         return ret
 
+    def get_attrs_values(self, *args) -> List[Any]:
```
maximilianwerk, line 541: Personally, I would prefer `def get_attrs_values(self, attributes: List[str]) -> Dict[Any]:` instead of using the magic `*args`.

And as a return value rather a Dict, to not rely on the order. But that might also be stupid, depending, where this function is usually called.

jina/types/document/__init__.py
```diff
@@ -315,7 +321,7 @@ def update(self, source: 'Document',
 
         if exclude_fields is None:
             if include_fields:
-                exclude_fields = tuple(f for f in self.non_empty_fields if f not in include_fields)
+                exclude_fields = set(f for f in self.non_empty_fields if f not in include_fields)
             else:
                 exclude_fields = self.non_empty_fields
```
maximilianwerk, line 326: Not sure, if that is a problem, but in the if, the type of `excluded_fields` is `Set[..]` and in the else it is `Tuple[..]`.

jina/types/document/__init__.py
```diff
@@ -884,5 +1007,8 @@ def plot(self, output: str = None,
 
     @property
     def non_empty_fields(self) -> Tuple[str]:
-        """Return the set fields of the curren"""
+        """Return the set fields of the current document that are not empty
+
+        :return: the set of non-empty fields
```
maximilianwerk, line 1012: It's a tuple, not a set.

jina/types/sets/document.py
```diff
@@ -155,21 +165,34 @@ def all_contents(self) -> Tuple['np.ndarray', 'DocumentSet']:
         """
         return self._extract_docs('content')
 
-    def _extract_docs(self, attr: str) -> Tuple['np.ndarray', 'DocumentSet']:
-        contents = []
+    def _extract_docs(self, *attr: str) -> Tuple['np.ndarray', 'DocumentSet']:
```
maximilianwerk, line 168: Again, I prefer having it more explicit above using `*attr`. I guess, you want to keep it for backward compatibility?

maximilianwerk (commented):
Some comments. I haven't checked tests yet.

jina/types/document/__init__.py
JoanFM: oh yes, I forgot to remove this, I tried to change to set but some tests failed. Will revert

jina/types/document/__init__.py
JoanFM: yes forgot to revert

jina/types/sets/document.py
JoanFM: backward compatibility and potential use, u would prefer having a List directly?

jina/types/document/__init__.py
JoanFM: there is a version that does this (returning a Dict), this one is intended to extract directly the values in one shot by the drivers to have it ready for batching. In the same spirit as `documentSet.all_embeddings`

jina/types/sets/document.py
maximilianwerk: Yes, I'd prefer a list.

JoanFM: In this case where passing a single element would be the most common pattern I kind of like this variable length args though, any feelings about this @jina-ai/engineering ?

JoanFM: This will become public when we start using batching extensively btw

jina/types/sets/document.py
```diff
@@ -155,21 +165,34 @@ def all_contents(self) -> Tuple['np.ndarray', 'DocumentSet']:
...
+        list_of_contents_output = len(attr) > 1
+        if list_of_contents_output:
```
davidbp, line 170: Do you need this `if` `else` ?

Can't you simply do...
```
     contents = [[] for at in attr]
```

if `attr` is not  present then `contents = []`
If attr has 3 elements then `contents = [[],[],[]]`

davidbp (commented):
Probably there is no need to do the if/else first block

jina/types/sets/document.py
JoanFM: Good!

JoanFM: If there is one element, it does `[[]]` which is what I want to avoid

davidbp: You can do that filtering 
```
    contents = [[] for at in attr if len(attr)>1]
```

JoanFM: Done!

tests/integration/executors_batching/test_encoder_executors_batching.py
```diff
@@ -0,0 +1,82 @@
...
+    @as_ndarray
+    @single
+    def encode(self, data, *args, **kwargs):
+        assert isinstance(data, str)
+        return np.random.random(EMBED_SIZE)
+
+
+@pytest.mark.parametrize('encoder', [DummyEncoderTextSingle(), DummyEncoderTextBatching()])
+def test_batching_encode_text(encoder):
+    docs = DocumentSet([Document(text=f'text-{i}') for i in range(15)])
```
cristianmtr, line 41: I just ran this with empty content/text and the test still passes. Is that a corner case we want?

cristianmtr: Seems that `.text` is empty string by default when you don't assign anything to the `Document`. I think it should be `None`, right?

tests/unit/types/document/test_document.py
```diff
@@ -499,7 +499,8 @@ def test_get_attr():
                   'tags': {'id': 'identity', 'a': 'b', 'c': 'd'}})
     d.score = NamedScore(value=42)
 
-    required_keys = ['id', 'text', 'tags__name', 'tags__feature1', 'score__value', 'tags__c', 'tags__id', 'tags__inexistant', 'inexistant']
+    required_keys = ['id', 'text', 'tags__name', 'tags__feature1', 'score__value', 'tags__c', 'tags__id',
```
cristianmtr, line 502: This is some formatting issue

tests/integration/executors_batching/test_encoder_executors_batching.py
JoanFM: could be, but that is another issue in my opinion?

jina/executors/decorators.py
```diff
@@ -199,14 +199,17 @@ def _get_total_size(full_data_size, batch_size, num_batch):
...
-def _merge_results_after_batching(final_result, merge_over_axis: int = 0):
+def _merge_results_after_batching(
+    final_result, merge_over_axis: int = 0, flatten: bool = True
+):
     if len(final_result) == 1:
         # the only result of one batch
         return final_result[0]
 
-    if len(final_result) and merge_over_axis is not None:
+    if len(final_result):
```
nan-wang, line 209:
```suggestion
    if final_result:
```

nan-wang: check out [PEP-8](https://www.python.org/dev/peps/pep-0008/)

jina/types/sets/document.py
```diff
@@ -107,11 +106,18 @@ def append(self, doc: 'Document') -> 'Document':
...
+        """Shortcut to :meth:`append`, do not override this method.
+
+        :param doc: the document to add to the set
+        :return: Appended list.
+        """
         return self.append(doc)
 
     def extend(self, iterable: Iterable['Document']) -> None:
-        """Extend an iterable to :class:`DocumentSet`."""
+        """Extend an iterable to :class:`DocumentSet`.
```
nan-wang, line 117:
```suggestion
        """
        Extend the :class:`DocumentSet` by appending all the items from the iterable.
```

jina/types/sets/document.py
```diff
@@ -138,7 +144,11 @@ def build(self):
         self._docs_map = {d.id: d for d in self._docs_proto}
 
     def sort(self, *args, **kwargs):
-        """Sort the list of :class:`DocumentSet`."""
+        """Sort the list of :class:`DocumentSet`.
```
nan-wang, line 147:
```suggestion
        """
        Sort the items of the :class:`DocumentSet` in place.
```

jina/types/sets/document.py
```diff
@@ -161,35 +171,54 @@ def all_contents(self) -> Tuple['np.ndarray', 'DocumentSet']:
...
                 docs_pts.append(doc)
             else:
                 bad_docs.append(doc)
 
-        contents = np.stack(contents) if contents else None
+        if list_of_contents_output:
+            for i in range(len(contents)):
+                contents[i] = np.stack(contents[i])
+        else:
+            contents = np.stack(contents) if contents else None
```
nan-wang, line 200: I would like to refactoring it a bit to make it easier to read. Plus, when `list_of_contents_output==False`, we need to take care the case `doc.get_attrs_values(*fields)` returns an empty list.
```
        if list_of_contents_output:
            for doc in self:
                content = doc.get_attrs_values(*fields)
                if content is None:
                    bad_docs.append(doc)
                    continue
                for i, c in enumerate(content):
                    contents[i].append(c)
                docs_pts.append(doc)
            for i, attr_list in enumerate(contents):
                contents[i] = np.stack(attr_list)
        else:
            for doc in self:
                content = doc.get_attrs_values(*fields)
                if not content:
                    bad_docs.append(doc)
                    continue
                contents.append(content)
                docs_pts.append(doc)
            contents = np.stack(contents) if contents else None
```

jina/types/sets/document.py
```diff
@@ -161,35 +171,54 @@ def all_contents(self) -> Tuple['np.ndarray', 'DocumentSet']:
...
 
         if bad_docs and docs_pts:
```
nan-wang, line 202: If all the docs are in `bad_docs`, we won't raise any warning?
```suggestion
        if bad_docs and docs_pts:
```

nan-wang (changes requested):

jina/executors/decorators.py
```diff
@@ -244,6 +250,7 @@ def batching(
     :param ordinal_idx_arg: the location of the ordinal indexes argument. Needed for classes
             where function decorated needs to know the ordinal indexes of the data in the batch
             (Not used when label_on is used)
+    :param flatten_output: Flag to determine if a result of list of lists needs to be flattened in output
```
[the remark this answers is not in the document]
JoanFM: I did not find the usage, I think as we keep applying these batches to different Driver we will see the needs I think

jina/types/sets/document.py
JoanFM: Done, u are right!

jina/types/sets/document.py
JoanFM: Done!

tests/integration/executors_batching/test_craft_executors_batching.py
```diff
@@ -0,0 +1,215 @@
...
+        assert isinstance(blob, np.ndarray)
+        return {'blob': blob}
+
+
+@pytest.mark.parametrize(
+    'crafter', [DummyCrafterBlobSingle(), DummyCrafterBlobBatching()]
+)
+def test_batching_blob_one_argument(crafter):
+    docs = DocumentSet(
+        [Document(blob=np.array([[i, i, i, i, i], [i, i, i, i, i]])) for i in range(15)]
```
nan-wang, line 114:
```suggestion
        [Document(blob=np.array([[i] * 5, [i] * 5])) for i in range(15)]
```

nan-wang (commented):
minor comments

nan-wang (approved):
LGTM

PR #2113, opened 2021-03-04
refactor: refactor rankers, move logic to driver

JoanFM:
**Changes introduced**
First steps to refactor Ranker interfaces.

Goals:
- [x] Have a single interface for Chunk2DocRankers (Just need to inherit one function). Move extra grouping and sorting logic into the Driver

Steps:
- [ ] Pass this PR
- [ ] Remove TFIDF and BM25Rankers from the jina-hub
- [ ] Refactor BiMatchRanker and SimpleAggregateRanker to not depend on functions that have disappeared.
- [ ] Enable batching for Rankers

### DIFF
--- a/jina/drivers/rank/aggregate/__init__.py
+++ b/jina/drivers/rank/aggregate/__init__.py
@@ -68,14 +68,6 @@ class BaseAggregateMatchesRankerDriver(BaseRankDriver):
                     m.chunks.append(chunk_matches_by_id[match_chunk_id])
             query.matches.append(m)
 
-    def group_by_doc_id(self, match_idx):
-        """
-        Group the ``match_idx`` by ``doc_id``.
-        :return: an iterator over the groups.
-        :rtype: :class:`Chunk2DocRanker`.
-        """
-        return self._group_by(match_idx, Chunk2DocRanker.COL_PARENT_ID)
-
     @staticmethod
     def _group_by(match_idx, col_name):
         # sort by ``col
@@ -113,15 +105,13 @@ class BaseAggregateMatchesRankerDriver(BaseRankDriver):
         :type match_idx: np.ndarray.
         :param query_chunk_meta: The meta information of the query chunks, where the key is query chunks' ``chunk_id``,
             the value is extracted by the ``query_required_keys``.
-        :type query_chunk_meta: Dict.
         :param match_chunk_meta: The meta information of the matched chunks, where the key is matched chunks'
             ``chunk_id``, the value is extracted by the ``match_required_keys``.
-        :type match_chunk_meta: Dict.
         :return: A [N x 2] numpy ``ndarray``, where the first column is the matched documents' ``doc_id`` (integer)
                 the second column is the score/distance/metric between the matched doc and the query doc (float).
         :rtype: np.ndarray.
         """
-        _groups = self.group_by_doc_id(match_idx)
+        _groups = self._group_by(match_idx, Chunk2DocRanker.COL_PARENT_ID)
         r = []
         for _g in _groups:
             score = self.exec_fn(_g, query_chunk_meta, match_chunk_meta)
--- a/jina/drivers/rank/aggregate/__init__.py
+++ b/jina/drivers/rank/aggregate/__init__.py
@@ -3,7 +3,7 @@ from collections import defaultdict, namedtuple
 
 import numpy as np
 
-from ....executors.rankers import Chunk2DocRanker, COL_STR_TYPE
+from ....executors.rankers import Chunk2DocRanker
 from ....types.document import Document
 from ....types.score import NamedScore
 
@@ -11,6 +11,8 @@ from .. import BaseRankDriver
 
 if False:
     from ....types.sets import DocumentSet
+
+COL_STR_TYPE = 'U64'  #: the ID column data type for score matrix
 
 
 class BaseAggregateMatchesRankerDriver(BaseRankDriver):
@@ -65,6 +67,66 @@ class BaseAggregateMatchesRankerDriver(BaseRankDriver):
                 for match_chunk_id in parent_id_chunk_id_map[doc_id]:
                     m.chunks.append(chunk_matches_by_id[match_chunk_id])
             query.matches.append(m)
+
+    def group_by_doc_id(self, match_idx):
+        """
+        Group the ``match_idx`` by ``doc_id``.
+        :return: an iterator over the groups.
+        :rtype: :class:`Chunk2DocRanker`.
+        """
+        return self._group_by(match_idx, Chunk2DocRanker.COL_PARENT_ID)
+
+    @staticmethod
+    def _group_by(match_idx, col_name):
+        # sort by ``col
+        _sorted_m = np.sort(match_idx, order=col_name)
+        _, _doc_counts = np.unique(_sorted_m[col_name], return_counts=True)
+        # group by ``col``
+        return np.split(_sorted_m, np.cumsum(_doc_counts))[:-1]
+
+    @staticmethod
+    def _sort_doc_by_score(r):
+        """
+        Sort a list of (``doc_id``, ``score``) tuples by the ``score``.
+        :param r: List of Tuples with document id and score
+        :type r: List[Tuple[Any, Any]]
+        :return: A `np.ndarray` in the shape of [N x 2], where `N` in the length of the input list.
+        :rtype: np.ndarray
+        """
+        r = np.array(r, dtype=[
+            (Chunk2DocRanker.COL_PARENT_ID, COL_STR_TYPE),
+            (Chunk2DocRanker.COL_SCORE, np.float64)]
+                     )
+        return np.sort(r, order=Chunk2DocRanker.COL_SCORE)[::-1]
+
+    def _score(self, match_idx: 'np.ndarray', query_chunk_meta: Dict, match_chunk_meta: Dict) -> 'np.ndarray':
+        """
+        Translate the chunk-level top-k results into doc-level top-k results. Some score functions may leverage the
+        meta information of the query, hence the meta info of the query chunks and matched chunks are given
+        as arguments.
+
+        :param match_idx: A [N x 4] numpy ``ndarray``, column-wise:
+                - ``match_idx[:, 0]``: ``doc_id`` of the matched chunks, integer
+                - ``match_idx[:, 1]``: ``chunk_id`` of the matched chunks, integer
+                - ``match_idx[:, 2]``: ``chunk_id`` of the query chunks, integer
+                - ``match_idx[:, 3]``: distance/metric/score between the query and matched chunks, float
+        :type match_idx: np.ndarray.
+        :param query_chunk_meta: The meta information of the query chunks, where the key is query chunks' ``chunk_id``,
+            the value is extracted by the ``query_required_keys``.
+        :type query_chunk_meta: Dict.
+        :param match_chunk_meta: The meta information of the matched chunks, where the key is matched chunks'
+            ``chunk_id``, the value is extracted by the ``match_required_keys``.
+        :type match_chunk_meta: Dict.
+        :return: A [N x 2] numpy ``ndarray``, where the first column is the matched documents' ``doc_id`` (integer)
+                the second column is the score/distance/metric between the matched doc and the query doc (float).
+        :rtype: np.ndarray.
+        """
+        _groups = self.group_by_doc_id(match_idx)
+        r = []
+        for _g in _groups:
+            score = self.exec_fn(_g, query_chunk_meta, match_chunk_meta)
+            r.append((_g[0][Chunk2DocRanker.COL_PARENT_ID], score))
+        return self._sort_doc_by_score(r)
 
 
 class Chunk2DocRankDriver(BaseAggregateMatchesRankerDriver):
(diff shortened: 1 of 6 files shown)


jina/drivers/rank/aggregate/__init__.py
```diff
@@ -66,6 +68,56 @@ def _insert_query_matches(self,
...
+    @staticmethod
+    def _sort_doc_by_score(r):
+        """
+        Sort a list of (``doc_id``, ``score``) tuples by the ``score``.
+        :param r: List of Tuples with document id and score
+        :type r: List[Tuple[Any, Any]]
+        :return: A `np.ndarray` in the shape of [N x 2], where `N` in the length of the input list.
+        :rtype: np.ndarray
+        """
+        r = np.array(r, dtype=[
```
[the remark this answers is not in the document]
JoanFM: the tuple is to allow accessing column by name

florian-hoenicke (approved):

jina/executors/rankers/__init__.py
```diff
@@ -160,22 +126,17 @@ class Match2DocRanker(BaseRanker):
...
-        :type match_meta: Dict
-        :return: A `np.ndarray` in the shape of [N x 2] where `N` is the length of
-            the `old_match_scores`. Semantic: [[match_id, new_score]]
-        :rtype: np.ndarray
+        Calculates the new scores for matches and returns them. Returns an iterable of the scores to be assined to the matches.
+        The returned scores need to be returned in the same order as the input `:param old_match_scores`.
+
+        :param old_match_scores: Contains old scores in a list
+        :param query_meta: Dictionary containing all the query meta information requested by the `required_keys` class_variable.
+        :param match_meta: List containing all the matches meta information requested by the `required_keys` class_variable. Sorted in the same way as `old_match_scores`
```
nan-wang, line 140:
```suggestion
        :param match_meta: List containing all the matches meta information requested by the `match_required_keys` class_variable. Sorted in the same way as `old_match_scores`
```

jina/resources/multimodal/pods/weighted_ranker.py
```diff
@@ -10,14 +10,32 @@ class WeightedRanker(Chunk2DocRanker):
 
     required_keys = {'weight'}
```
nan-wang, line 11: update this line to `query_required_keys` and `match_required_keys`

jina/executors/rankers/__init__.py
```diff
@@ -70,80 +91,25 @@ class Chunk2DocRanker(BaseRanker):
 
     def score(
         self, match_idx: 'np.ndarray', query_chunk_meta: Dict, match_chunk_meta: Dict
-    ) -> 'np.ndarray':
+    ) -> float:
```
nan-wang, line 94: Instead of passing all the chunks with different parent ids, we assume all the input chunks have the same parent id. In this case, `match_idx[: 0]` is no longer necessary and can be removed.

nan-wang (changes requested):
Love this change!

jina/executors/rankers/__init__.py
JoanFM: Yes. I like the idea, but I am not sure if it is overkill to delete the first column, should I only select the other 3?

What is the best way to do this with zero runtime cost in numpy?

jina/drivers/rank/__init__.py
```diff
@@ -82,29 +76,36 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
...
             matches = doc.matches
-            old_match_scores = {match.id: match.score.value for match in matches}
-            match_meta = (
-                {match.id: match.get_attrs(*self._exec_match_keys) for match in matches}
-                if self._exec_match_keys
-                else None
-            )
+            num_matches = len(matches)
+            old_match_scores = []
+            extract_match_meta = self._exec_match_keys is not None
```
bwanglzu, line 81: can we rename this variable? `extract_match_meta` sounds like a function name, while it's actually a boolean variable.

jina/drivers/rank/aggregate/__init__.py
```diff
@@ -68,6 +70,62 @@ def _insert_query_matches(
...
+        # sort by ``col
+        _sorted_m = np.sort(match_idx, order=col_name)
+        _, _doc_counts = np.unique(_sorted_m[col_name], return_counts=True)
+        # group by ``col``
+        return np.split(_sorted_m, np.cumsum(_doc_counts))[:-1]
+
+    @staticmethod
+    def _sort_doc_by_score(r):
+        """
+        Sort a list of (``doc_id``, ``score``) tuples by the ``score``.
```
bwanglzu, line 84: can you add an empty line between method description and it's parameter section in docstring? It's not reported an error because it's an internal method.

bwanglzu (commented):
what's the motivation for moving methods into drivers?

jina/drivers/rank/__init__.py
JoanFM: can u suggest a name inline?

jina/drivers/rank/aggregate/__init__.py
JoanFM: can u suggrst the change inline?

jina/executors/rankers/__init__.py
JoanFM: @nan-wang , in this case it is not so easy, because we are creating a numpy array of a custom tuple, so is not easy to access by column. I would go with this solutioin and try to refactor this later, maybe there is some optimization to be done here

bwanglzu (approved):
LGTM!

jina/drivers/rank/__init__.py
```diff
@@ -60,7 +54,7 @@ def __init__(
         reverse: bool = False,
         traversal_paths: Tuple[str] = ('r',),
         *args,
-        **kwargs
+        **kwargs,
```
nan-wang, line 57: Any argument to do this?

jina/drivers/rank/__init__.py
```diff
@@ -82,29 +76,36 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
...
-            cm[match_id].score = NamedScore(
-                value=score, op_name=op_name, ref_id=context_doc.id
-            )
+        ref_doc_id = matches._ref_doc.id
+
+        for match, score in zip(matches, match_scores):
+            match.score = NamedScore(value=score, op_name=op_name, ref_id=ref_doc_id)
 
-        cm.sort(key=lambda x: x.score.value, reverse=True)
+        matches.sort(key=lambda x: x.score.value, reverse=True)
```
nan-wang, line 111:
```suggestion
        matches.sort(key=lambda x: x.score.value, reverse=self.reverse)
```

jina/drivers/rank/aggregate/__init__.py
```diff
@@ -68,6 +70,63 @@ def _insert_query_matches(
...
+
+        :param r: List of Tuples with document id and score
+        :type r: List[Tuple[Any, Any]]
```
nan-wang, line 87:
```suggestion
        :type r: List[Tuple[np.str_, np.float64]]
```

jina/executors/rankers/__init__.py
nan-wang: fair points. Maybe we can add a `TODO` here.

nan-wang (changes requested):

jina/drivers/rank/__init__.py
JoanFM: I guess is just black formatting?

jina/drivers/rank/__init__.py
JoanFM: We do not have a `reverse` parameter. I think  for now every `Ranker` is scoring per `relevance` so it is fair to set it to `True`

jina/drivers/rank/__init__.py
```diff
@@ -82,29 +76,36 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
...
+                if needs_match_meta:
+                    match_meta.append(match.get_attrs(*self._exec_match_keys))
 
             # if there are no matches, no need to sort them
             if not old_match_scores:
                 continue
 
-            new_match_scores = self.exec_fn(query_meta, old_match_scores, match_meta)
-            self._sort_matches_in_place(doc, new_match_scores)
+            new_scores = self.exec_fn(old_match_scores, query_meta, match_meta)
```
Yongxuanzhang, line 92: why change the name from `new_match_scores` to `new_scores` here?
We have `old_match_scores` would it be better to use `new_match_scores`?

Yongxuanzhang:
Shall we merge the master to pass ci?

tests/unit/drivers/rank/test_matches2doc_rank_drivers.py
```diff
@@ -29,15 +29,9 @@ def __init__(self, *args, **kwargs):
...
-        return np.array(
-            new_scores,
-            dtype=[(self.COL_MATCH_ID, np.object), (self.COL_SCORE, np.float64)],
-        )
+    def score(self, old_match_scores, query_meta, match_meta):
+        new_scores = [-abs(m['length'] - query_meta['length']) for m in match_meta]
+        return new_scores
 
 
 def create_document_to_score():
```
Yongxuanzhang, line 37: The way to generate data is a little bit confusing.

1. doc and match ID: Why doc.id = '1' * 20? In the comments the doc_id is '1' and the id for matches are 2,3,4,5. But in fact, we set them to match.id = str(match_id) * match_length; What's the reason?
2. Why `match.length = match_score`?

JoanFM: 1.the *20 responds to changes that have been applied to Document ID and continuous refactoring. 

2. I thinks is by test design, but this test is too complicated to be honest

jina/drivers/rank/__init__.py
JoanFM: yes, we could call it back to nee

JoanFM:
> what's the motivation for moving methods into drivers?

To have a much simpler interface. Every executor has a very neat and clear interface except Chunk2DocRanker.

Also like this we can hide some complexity about grouping from the executor developer

tests/unit/drivers/rank/test_matches2doc_rank_drivers.py
Yongxuanzhang: `match.score.value = match_score` this line seems doesn't work?

Yongxuanzhang: > 1.the *20 responds to changes that have been applied to Document ID and continuous refactoring.
> 
> 1. I thinks is by test design, but this test is too complicated to be honest

😂

JoanFM: where is this code?

Yongxuanzhang: In the test, https://github.com/jina-ai/jina/blob/f0b6a44045f7fccca05a34020fd42981ce34dc4e/tests/unit/drivers/rank/test_matches2doc_rank_drivers.py#L61

Yongxuanzhang: The value of the score cannot be set this way? In the test, this value is 0, thus old_matches_socres is [0.0,0.0,0.0,0.0]

JoanFM: there maybe an error in this test, you are welcome to refactor

Yongxuanzhang: > there maybe an error in this test, you are welcome to refactor

The refactor is in this PR, this makes more sense I think. We compare the abs diff between the length of matches with the query. Those with the same value will be sorted by id asending.
https://github.com/jina-ai/jina/pull/2127

jina/drivers/rank/__init__.py
nan-wang: We do have. 
https://github.com/jina-ai/jina/pull/2113/files/bade8af72d03b6690bdee454907f49c676283125#diff-5263f3d40f3d59bd6ea5d52f39b61f12a708f1e44ec18487347bfc57b803bea0R60

nan-wang:
```
ds = DocumentSet([Document(text=f'{np.random.randint(0, 10)}') for i in range(3)])
ds.sort(key=lambda x: dunder_get(x, 'text'), reverse=True)
# {'id': '1040e3b6-808e-11eb-a397-00e04c6b0562', 'mimeType': 'text/plain', 'text': '9'}
# {'id': '1040ec8a-808e-11eb-a397-00e04c6b0562', 'mimeType': 'text/plain', 'text': '7'}
# {'id': '1040e8d4-808e-11eb-a397-00e04c6b0562', 'mimeType': 'text/plain', 'text': '4'}

ds.sort(key=lambda x: x.text, reverse=False)
# {'id': '1040e8d4-808e-11eb-a397-00e04c6b0562', 'mimeType': 'text/plain', 'text': '4'}
# {'id': '1040ec8a-808e-11eb-a397-00e04c6b0562', 'mimeType': 'text/plain', 'text': '7'}
# {'id': '1040e3b6-808e-11eb-a397-00e04c6b0562', 'mimeType': 'text/plain', 'text': '9'}
```

nan-wang (dismissed):
LGTM👍

PR #2119, opened 2021-03-06
Improve Formatter

chunyuema:
1. Modified the formatter.py using deepcopy to replace copy
2.Aligned comments with code to ensure matching

### DIFF
--- a/jina/logging/formatter.py
+++ b/jina/logging/formatter.py
@@ -1,6 +1,6 @@
 import json
 import re
-from copy import copy
+from copy import deepcopy
 from logging import Formatter
 
 from .profile import used_memory
@@ -15,11 +15,11 @@ class ColorFormatter(Formatter):
 
     MAPPING = {
         'DEBUG': dict(color='white', on_color=None),  # white
-        'INFO': dict(color='white', on_color=None),  # cyan
+        'INFO': dict(color='cyan', on_color=None),  # cyan
         'WARNING': dict(color='yellow', on_color='on_grey'),  # yellow
         'ERROR': dict(color='red', on_color=None),  # 31 for red
         'CRITICAL': dict(color='white', on_color='on_red'),  # white on red bg
-        'SUCCESS': dict(color='green', on_color=None),  # white on red bg
+        'SUCCESS': dict(color='green', on_color=None),  # green
     }  #: log-level to color mapping
 
     def format(self, record):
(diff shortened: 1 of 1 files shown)


jina/logging/formatter.py
```diff
@@ -29,9 +29,10 @@ def format(self, record):
         :param record: A LogRecord object
         :returns: Formatted LogRecord with level-colour MAPPING to add corresponding colour.
         """
-        cr = copy(record)
+        cr = deepcopy(record)
```
JoanFM, line 32: hey @chunyuema,

Why do we need deepcopy in this case?

chunyuema: Hi, thank you for the question. I suggested using deepcopy mostly because the record is a LogRecord object. Making a copy of the object should avoid most of the accidental overwrite of the original copy. However, I was a little worried on effect of codes from line 88-89 on the original object. I think even though it seems to be sufficient to just use copy for now, it might still be a better practice to make a completely independent copy of the LogRecord object into cr so that the integrity of the original copy would be preserved as much as possible. This might become more relevant if more operations are added with the cr variable.

Yongxuanzhang: Hi @chunyuema, will deepcopy harm the performance?

chunyuema: Hi, this is a good question. I think deepcopy will definitely be slower than just copy. I need to do more study on the code base to see how much using deepcopy impacts the performance.

Yongxuanzhang: Sure. Since the `formatter` is used in `logger`, which is used extensively. I'm not sure if this will harm the performance, haven't got time to look into it. To my understanding, we should use `deepcopy` only when necessary.

And it's good that you can raise this suggestion.

jina/logging/formatter.py
```diff
@@ -15,11 +15,11 @@ class ColorFormatter(Formatter):
 
     MAPPING = {
         'DEBUG': dict(color='white', on_color=None),  # white
-        'INFO': dict(color='white', on_color=None),  # cyan
+        'INFO': dict(color='cyan', on_color=None),  # cyan
```
maximilianwerk, line 18: Please rather change the comment instead of the color. Thanks :)

Yongxuanzhang:
Hi @chunyuema, are you still working on this?

chunyuema:
> Hi @chunyuema, are you still working on this?

Hi, I am not actively working on this issue as for now :)

Yongxuanzhang:
Ok, you can reopen it when you're back on this. ;)

PR #2121, opened 2021-03-07
feat(schema): generate pydantic based jsonschema for any jina proto

deepankarm:


florian-hoenicke (approved):

florian-hoenicke:
I like the conversion from protobuf to pydantic. It is more flexible in contrast to just adding the oneof-fields manually. Good work! @deepankarm

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -1,11 +1,153 @@
...
+    oneof_fields = defaultdict(list)
+
+    if isinstance(protobuf_model, Descriptor):
+        model_name = protobuf_model.name
+        protobuf_fields = protobuf_model.fields
+    elif isinstance(protobuf_model, GeneratedProtocolMessageType):
+        model_name = protobuf_model.DESCRIPTOR.name
+        protobuf_fields = protobuf_model.DESCRIPTOR.fields
+
+    if model_name.endswith('Proto') and model_name in PROTO_TO_PYDANTIC_MODELS:
```
JoanFM, line 76: this assumption is needed? isnt enough to remove this endswith? I think it adds a big assumption for the future

JoanFM (changes requested):

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -1,11 +1,153 @@
...
+    FieldDescriptor.TYPE_SFIXED32: float,
+    FieldDescriptor.TYPE_SFIXED64: float,
+    FieldDescriptor.TYPE_BYTES: bytes,
+    FieldDescriptor.TYPE_STRING: str,
+    FieldDescriptor.TYPE_ENUM: Enum,
+    FieldDescriptor.TYPE_MESSAGE: None,
+}
+
+
+def protobuf_to_pydantic_model(
```
JoanFM, line 52: is it possible to break this down in smaller functions?

JoanFM (changes requested):

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -1,11 +1,153 @@
...
+
+            if f.message_type.name == 'Struct':
+                # Proto Field Type: google.protobuf.Struct
+                field_type = Dict
+                default_value = {}
+
+            elif f.message_type.name == 'Timestamp':
+                # Proto Field Type: google.protobuf.Timestamp
+                field_type = datetime
+                default_value = datetime.now()
```
JoanFM, line 105: Isn't it better to put an stupid default value? Like this we can see if the timestamp is real or not? Normally timestamp is needed to measure some stuff about the presence of an object in a Pod, like this we try the "false positives" of seeing a meaningful timestamp

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -1,11 +1,153 @@
...
+
+            elif f.message_type.name.endswith('Proto'):
```
JoanFM, line 107: I would just have an else here. I know it is true that now every Proto type ends with Proto, but is it really needed to introduce this assumption? At least we can add a static method in types describing if that is a Proto message, so that the logic is not spread in the codebase

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -1,11 +1,153 @@
...
+
+    if isinstance(protobuf_model, Descriptor):
+        model_name = protobuf_model.name
+        protobuf_fields = protobuf_model.fields
+    elif isinstance(protobuf_model, GeneratedProtocolMessageType):
+        model_name = protobuf_model.DESCRIPTOR.name
+        protobuf_fields = protobuf_model.DESCRIPTOR.fields
+
+    if model_name.endswith('Proto') and model_name in PROTO_TO_PYDANTIC_MODELS:
+        return PROTO_TO_PYDANTIC_MODELS[model_name]
```
JoanFM, line 77: Is this just an optimization?

Is it worth it the fact of introducing a global variable?

deepankarm: This helps us in 2 ways
1. Keeping this global variable adds all schemas to openapi docs. Hence, auto-rendering of all Jina protos in jsonschema for both swagger & redoc
2. Easier handling of recursive schemas.

jina/peapods/runtimes/asyncio/rest/models.py
deepankarm: `Timestamp` field needs further polishing. `google.protobuf.Timestamp` describes time in `seconds` & `nanos`, where `datetime.now()` is in a different format. Since this schema is only exposed to the users for viewing, for now, I would prefer to keep it as-is for now and upgrade as and when required.

jina/peapods/runtimes/asyncio/rest/models.py
deepankarm: I couldn't find a better way to validate `MessageDescriptor` objects in protobuf except for the name field. It is true that a simple `else` would also work, but that brings issues if we introduce more predefined modules in `jina.proto`.

JoanFM: can we at least centralize the logic somewhere in types, so that it is kept in a single place

deepankarm: I have removed the name check on `Proto` for now.

JoanFM (approved):

Issue #2092, opened 2021-03-01
Add code snippet with redoc api docs


lusloher:
Code snippet for API invocation needs to be added for `curl` & `python requests`.

<p></p>

yah. I had added this. The goal is to have automatic code-generation in our redoc based docs <https://api.jina.ai/daemon/#operation/_status_status_get>

<p></p>

<p></p>

Like - <http://apidocs.directfreight.com/#tag/boards>

<p></p>

<p></p>

I had spent an hour on this, but faced a few issues, so it still exists in the backlog

lusloher:
Link to feature: https://jinaai.productboard.com/feature-board/planning/features/6941839

PR #2130, opened 2021-03-09
docs(openapi): autogenerate curl commands with redoc

deepankarm:
Fixes https://github.com/jina-ai/jina/issues/2092
image


deepankarm:
We should check for black/docstring only for added/modified files -  https://github.com/jina-ai/jina/pull/2131

nan-wang (approved):
LGTM👍

PR #2136, opened 2021-03-09
refactor: missing optional types

florian-hoenicke:


jina/clients/request/__init__.py
```diff
@@ -29,8 +29,8 @@ def request_generator(
     data: GeneratorSourceType,
     request_size: int = 0,
     mode: RequestType = RequestType.INDEX,
-    mime_type: str = None,
-    queryset: Union[AcceptQueryLangType, Iterator[AcceptQueryLangType]] = None,
+    mime_type: Optional[str] = None,
+    queryset: Union[AcceptQueryLangType, Iterator[AcceptQueryLangType], None] = None,
```
JoanFM, line 33: Optional is the good way to see Union with None

JoanFM (commented):
is this ticket actually checking if None is properly handled and thus is a valid default?

florian-hoenicke:
> is this ticket actually checking if None is properly handled and thus is a valid default?

No, it is just on a syntactical level to make sure to add `Optional` wherever it is obviously needed.

jina/clients/request/__init__.py
florian-hoenicke: Would you prefer to either have only Union types or only Optional types in the code?

florian-hoenicke:
> > > is this ticket actually checking if None is properly handled and thus is a valid default?
> > 
> > 
> > No, it is just on a syntactical level to make sure to add `Optional` wherever it is obviously needed. But it is not checked if the default initialization with None can be removed.
> 
> It woule be good to eventually check if they are really Optional or is just a false default value, but that may be overkill

I checked for some cases but can not confidently say they are always needed.
I could create a follow-up ticket to go more into the semantics.

jina/clients/request/__init__.py
JoanFM: If there is Union including None, Optional is a better option.

JoanFM: It is more informative about wehter None is acceptable or not (it should be done properly but we have a lot of places where Optional is not really true).

Also Optional is a useful functional construct

JoanFM: So in this case is Optional[Union[..., ...]]

jina/executors/decorators.py
```diff
@@ -226,8 +226,8 @@ def _merge_results_after_batching(
 
 
 def batching(
-    func: Callable[[Any], np.ndarray] = None,
-    batch_size: Union[int, Callable] = None,
+    func: Optional[Callable[[Any], np.ndarray]] = None,
+    batch_size: Union[int, Callable, None] = None,
```
JoanFM, line 230: This needs to be Optional[Union[int, Callable]]

jina/executors/decorators.py
```diff
@@ -341,8 +341,8 @@ def arg_wrapper(*args, **kwargs):
 
 
 def batching_multi_input(
-    func: Callable[[Any], np.ndarray] = None,
-    batch_size: Union[int, Callable] = None,
+    func: Optional[Callable[[Any], np.ndarray]] = None,
+    batch_size: Union[int, Callable, None] = None,
```
JoanFM, line 345: same here

jina/executors/indexers/__init__.py
```diff
@@ -37,17 +37,19 @@ class BaseIndexer(BaseExecutor):
...
+
+    :param index_filename: the name of the file for storing the index, when not given metas.name is used.
+    :param args:  Additional positional arguments
+    :param kwargs: Additional keyword arguments
     """
 
     def __init__(
-        self, index_filename: str = None, key_length: int = 36, *args, **kwargs
+        self,
+        index_filename: Optional[str] = None,
```
JoanFM, line 48: just comment, but I think this one is not a real Optional

jina/flow/base.py
```diff
@@ -229,7 +229,7 @@ def needs_all(self, name: str = 'joiner', *args, **kwargs) -> 'BaseFlow':
 
     def add(
         self,
-        needs: Union[str, Tuple[str], List[str]] = None,
+        needs: Union[str, Tuple[str], List[str], None] = None,
```
JoanFM, line 232: Optional here

jina/flow/mixin/crud.py
```diff
@@ -121,13 +121,13 @@ def search_ndarray(
     )
     def index_lines(
         self,
-        lines: Union[Iterable[str], TextIO] = None,
-        filepath: str = None,
-        size: int = None,
-        sampling_rate: float = None,
+        lines: Union[Iterable[str], TextIO, None] = None,
```
JoanFM, line 124: and here

jina/flow/mixin/crud.py
```diff
@@ -383,11 +383,11 @@ def search_files(
     )
     def search_lines(
         self,
-        lines: Union[Iterable[str], TextIO] = None,
-        filepath: str = None,
-        field_resolver: Dict[str, str] = None,
-        size: int = None,
-        sampling_rate: float = None,
+        lines: Union[Iterable[str], TextIO, None] = None,
```
JoanFM, line 386: here

jina/peapods/zmq/__init__.py
```diff
@@ -39,7 +39,7 @@ def __init__(
         self,
         args: 'argparse.Namespace',
         logger: 'JinaLogger' = None,
```
JoanFM, line 41: here missing an Optional but I think these defaults are wrong

JoanFM (changes requested):

jina/clients/request/__init__.py
florian-hoenicke: I checked the documentation.
https://docs.python.org/3/library/typing.html
It says `You can use Optional[X] as a shorthand for Union[X, None].`
So, the python docs don't care which way is used.
I still think Optional would be more obvious when you look at it.

jina/drivers/delete.py
```diff
@@ -1,14 +1,24 @@
...
+
 from . import BaseExecutableDriver
 
 
 class DeleteDriver(BaseExecutableDriver):
     """Drivers inherited from this Driver will bind :meth:`delete` by default """
 
-    def __init__(self, executor: str = None, method: str = 'delete', *args, **kwargs):
+    def __init__(
+        self, executor: Optional[str] = None, method: str = 'delete', *args, **kwargs
```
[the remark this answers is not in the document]
florian-hoenicke: Yes, it is the name of the sub-executor.
The `BaseExecutableDriver` also requires a string.

jina/drivers/__init__.py
```diff
@@ -196,7 +196,7 @@ class BaseDriver(JAMLCompatible, metaclass=DriverType):
     def __init__(self, priority: int = 0, *args, **kwargs):
         self.attached = False  # : represent if this driver is attached to a
         # :class:`jina.peapods.runtimes.zmq.zed.ZEDRuntime` (& :class:`jina.executors.BaseExecutor`)
-        self.runtime = None  # type: Optional['ZEDRuntime']
+        self.runtime: Optional['ZEDRuntime'] = None
```
JoanFM, line 199: we avoid adding type hints outside arguments

JoanFM (changes requested):

jina/peapods/zmq/__init__.py
florian-hoenicke: I added the default annotation to have it at least correct on syntactical level

jina/flow/mixin/crud.py
florian-hoenicke: done

jina/flow/base.py
florian-hoenicke: done

jina/executors/decorators.py
florian-hoenicke: done

jina/executors/decorators.py
florian-hoenicke: done

JoanFM (approved):

PR #2143, opened 2021-03-10
feat: distributed(remote) pea in pod

bwanglzu:
- [x] research argument parsing at pod level when `parallel` is greater than 1.
- [x] add remote pea in pods.
- [x] add unit tests.
- [x] add integration tests.


jina/parsers/peapods/pod.py
```diff
@@ -63,12 +63,14 @@ def mixin_base_pod_parser(parser):
...
-    # hosts for peas
+    # hosts for peas, when parallel greater than 1 each pea should have it's own host
 
     gp.add_argument(
         '--peas-hosts',
         type=KVAppendAction,
         metavar='KEY=VALUE',
         nargs='*',
-        help='The hosts of the pea when parallel greater than 1'
+        help='''The hosts of the peas when parallel greater than 1,
```
JoanFM, line 73: Each Pea has a pod is not true, I think the Peas not present in this map will get the `host` from the `Pods` arguments, as well as Head and Tail Peas

JoanFM: Key is not the host id, key is the `pea_id`

JoanFM (changes requested):

JoanFM (approved):
LGTM! Great job!

PR #2144, opened 2021-03-10
fix: add delete() for cached property

cristianmtr:
- [x] provide overridden method for `del` for `@cached_property`
- [x] correct closing of query handler in BinaryPb

NOTE

@florian-hoenicke @JoanFM @maximilianwerk 

Closing the query_handler is now coherent, but makes the update and delete ops slower. This is a trade-off we need to discuss perhaps.

Before, we were only closing the qhandler's body but still relying on its header to get the keys. That was a bit of smell due to the design of the Write/ReadHandlers in BinaryPb. Once we call close we shouldn't be able to access it, but we are. 

Do we want to keep the old way (less readable, but more efficient), or this way (more readable, less efficient)?

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -972,6 +972,13 @@ class cached_property:
         value = obj.__dict__[f'CACHED_{self.func.__name__}'] = self.func(obj)
         return value
 
+    def __delete__(self, obj):
+        cached_value = obj.__dict__.get(f'CACHED_{self.func.__name__}', None)
+        if cached_value is not None:
+            if hasattr(cached_value, 'close'):
+                cached_value.close()
+            del obj.__dict__[f'CACHED_{self.func.__name__}']
+
 
 def get_now_timestamp():
     """
--- a/jina/executors/indexers/keyvalue.py
+++ b/jina/executors/indexers/keyvalue.py
@@ -210,13 +210,13 @@ class BinaryPbIndexer(BaseKVIndexer):
         keys, values = self._filter_nonexistent_keys_values(
             keys, values, self.query_handler.header.keys()
         )
+        del self.query_handler
+        self.handler_mutex = False
         if keys:
             self._delete(keys)
             self.add(keys, values)
 
     def _delete(self, keys: Iterable[str]) -> None:
-        self.query_handler.close()
-        self.handler_mutex = False
         for key in keys:
             self.write_handler.header.write(
                 np.array(
@@ -229,9 +229,6 @@ class BinaryPbIndexer(BaseKVIndexer):
                     ],
                 ).tobytes()
             )
-
-            if self.query_handler:
-                del self.query_handler.header[key]
             self._size -= 1
 
     def delete(self, keys: Iterable[str], *args, **kwargs) -> None:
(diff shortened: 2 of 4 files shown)


jina/helper.py
```diff
@@ -972,6 +972,13 @@ def __get__(self, obj, cls):
         value = obj.__dict__[f'CACHED_{self.func.__name__}'] = self.func(obj)
         return value
 
+    def __delete__(self, obj):
+        cached_value = obj.__dict__.get(f'CACHED_{self.func.__name__}', None)
+        if cached_value is not None:
+            if hasattr(cached_value, 'close'):
```
florian-hoenicke, line 978: is there another way to check if the cached_value has to be closed.
Something with `isinstance()`

cristianmtr: What would be the instance? We have different classes that could be handlers here

florian-hoenicke: just brainstorming. If we can not make any assumptions about the handler class, then we have to keep it like this.

JoanFM: not really a fan of this to be honest. We assume that close is an special method, which may not be true

JoanFM: I think this is a code smell that we are caching something we should not cache

JoanFM (changes requested):

cristianmtr:
@JoanFM (not sure why I can't quote your comments)

> I think this is a code smell that we are caching something we should not cache

Yes, that is a smell buried deep in Jina unfortunately. We are caching the query handler, for example in here https://github.com/jina-ai/jina/blob/640daf4d389768be216dac4125ef4e837ee65d23/jina/executors/indexers/__init__.py#L85

I see why this is necessary as we want to avoid the performance hit of having to build it up every time.

> not really a fan of this to be honest. We assume that close is an special method, which may not be true

Hm I see. Is there any alternative you can think of? We could check that the field name is something like 'query handler' etc., but that would be just as bad.

We are making a similar assumption here: https://github.com/jina-ai/jina/blob/640daf4d389768be216dac4125ef4e837ee65d23/jina/executors/indexers/__init__.py#L169

JoanFM:
> @JoanFM (not sure why I can't quote your comments)
> 
> > I think this is a code smell that we are caching something we should not cache
> 
> Yes, that is a smell buried deep in Jina unfortunately. We are caching the query handler, for example in here
> 
> https://github.com/jina-ai/jina/blob/640daf4d389768be216dac4125ef4e837ee65d23/jina/executors/indexers/__init__.py#L85
> 
> I see why this is necessary as we want to avoid the performance hit of having to build it up every time.
> 
> > not really a fan of this to be honest. We assume that close is an special method, which may not be true
> 
> Hm I see. Is there any alternative you can think of? We could check that the field name is something like 'query handler' etc., but that would be just as bad.
> 
> We are making a similar assumption here:
> 
> https://github.com/jina-ai/jina/blob/640daf4d389768be216dac4125ef4e837ee65d23/jina/executors/indexers/__init__.py#L169

Can't we maybe have a private class that can control if a handler is to be reread or recomputed or something?

cristianmtr:
> Can't we maybe have a private class that can control if a handler is to be reread or recomputed or something?

Not sure I understand what you mean. Another class decorator, like `cached_property`?

maximilianwerk:
> Can't we maybe have a private class that can control if a handler is to be reread or recomputed or something?

What we could do is to have an argument for the decorator, which says, if it is closable of not in order to make it explicit instead of implicit. Anyhow, the current solution is fine in my opinion.

JoanFM (approved):

maximilianwerk (approved):
LGTM

PR #2146, opened 2021-03-10
fix(pydantic): add oneof validations for correct document creation

### LINKED ISSUE #2132: REST API - "oneof" validation
The behavior of the `oneof` type in the protobuf definition can not be mapped to pydantic models.
There are request to address this issue:
samuelcolvin/pydantic#656
samuelcolvin/pydantic#619

As long as this feature does not exist, we should make our custom solution.

Please validate the request to check that the client obeys the `oneof` rules.

deepankarm:
Fixes #2132 
Note:  https://github.com/samuelcolvin/pydantic/discussions/2504


jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -54,15 +54,60 @@ class CamelCaseConfig(BaseConfig):
...
+def _get_oneof_validator(oneof_fields: List, oneof_key: str) -> Callable:
+    """
+    Pydantic root validator classmethod generator to confirm only one oneof field is passed
+
+    :param oneof_fields: list of field names for oneof
+    :type oneof_fields: List
+    :param oneof_key: oneof key
+    :type oneof_key: str
+    :return: classmethod for Pydantic model validation
+    :rtype: Callable
```
florian-hoenicke, line 66: just a small note, we don't define the types in the docstring again

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -54,15 +54,60 @@ class CamelCaseConfig(BaseConfig):
...
+    """
+
+    def oneof_validator(cls, values):
+        if len(set(oneof_fields).intersection(set(values))) > 1:
+            raise ValueError(
+                f'only one field among {oneof_fields} can be set for key {oneof_key}!'
```
florian-hoenicke, line 72: can we write a test here to check for value errors?

tests/unit/peapods/runtimes/asyncio/rest/test_models.py
```diff
@@ -73,45 +69,50 @@ def test_all_fields_in_document_proto():
...
-        'DocumentProto'
-    ]().schema()['definitions']['DocumentProto']['properties']
+    document_proto_properties_alias = PROTO_TO_PYDANTIC_MODELS.DocumentProto().schema()[
+        'definitions'
+    ]['DocumentProto']['properties']
     for i in ['contentHash', 'levelName', 'parentId', 'mimeType']:
         assert i in document_proto_properties_alias
 
 
 def test_oneof():
```
florian-hoenicke, line 79: I think each case should be a separated test

tests/unit/peapods/runtimes/asyncio/rest/test_models.py
```diff
@@ -73,45 +69,50 @@ def test_all_fields_in_document_proto():
...
+    assert doc.buffer == b'abc'
+    assert 'text' not in doc.dict()
+    assert 'blob' not in doc.dict()
+
+    doc = PROTO_TO_PYDANTIC_MODELS.DocumentProto(
+        blob=PROTO_TO_PYDANTIC_MODELS.NdArrayProto()
+    )
+    assert doc.blob == PROTO_TO_PYDANTIC_MODELS.NdArrayProto()
+    assert 'text' not in doc.dict()
+    assert 'buffer' not in doc.dict()
```
florian-hoenicke, line 97: Can we add a test for the content attribute as well?
```
    doc = PROTO_TO_PYDANTIC_MODELS.DocumentProto(content=b'abc')
    assert doc.buffer == b'abc'
    assert 'text' not in doc.dict()
    assert 'blob' not in doc.dict()
    ```
    
    The document creation works the same way:
    
Document(content='my text')
<jina.types.document.Document id=8a2788b2-823f-11eb-b8b8-faffc24bcbdb mimeType=text/plain text=my text at 140330282190992>

File: tests/unit/peapods/runtimes/asyncio/rest/test_models.py (hunk 3 above)
Comment (florian-hoenicke) [inline tests/unit/peapods/runtimes/asyncio/rest/test_models.py:79]: Testing a case with oneof types in different levels would be nice.
For instance testing a Document with some content and an embedding

File: jina/peapods/runtimes/asyncio/rest/app.py [hunk 5]
```diff
@@ -18,21 +18,6 @@
 from .....types.request import Request
 
 
-# TODO instead of this function make sure that only one 'oneof field' is used.
```
florian-hoenicke, line 21: Very cool that you found a way to get rid of this function call. It simplifies the implementation of the endpoint.

florian-hoenicke (changes requested):

jina/peapods/runtimes/asyncio/rest/models.py
```diff
@@ -54,15 +54,60 @@ class CamelCaseConfig(BaseConfig):
     allow_population_by_field_name = True
 
 
+def _get_oneof_validator(oneof_fields: List, oneof_key: str) -> Callable:
```
JoanFM, line 57: instead of using this, u can use `bind` to bind parameters to method.

`from functools import partial`

JoanFM (changes requested):

tests/unit/peapods/runtimes/asyncio/rest/test_models.py
deepankarm: - `PROTO_TO_PYDANTIC_MODELS` is trying to replicate every proto definition in `jina.proto`, but not things in `Document()`

```python
from jina.proto.jina_pb2 import DocumentProto
DocumentProto(content='abc')
# ValueError: Protocol message DocumentProto has no "content" field.
# `oneof` key 'content' cannot be set

DocumentProto(text='abc').content
# AttributeError: content
# `oneof` key 'content' doesn't get set if we set the oneof fields

from jina.peapods.runtimes.asyncio.rest.models import PROTO_TO_PYDANTIC_MODELS
PROTO_TO_PYDANTIC_MODELS.DocumentProto(text='abc').content
# AttributeError: 'DocumentProto' object has no attribute 'content'
# Like proto, we too don't set content 
```

- `jina.types.document.Document()` can read from the proto directly or `bytes` or `str` or `Dict`. Pydantic models help us send the doc as str or dict to `Document()` for REST based requests, I should add tests for the below.

```python
DocumentSourceType = TypeVar(
    'DocumentSourceType', jina_pb2.DocumentProto, bytes, str, Dict
)

str -> PROTO_TO_PYDANTIC_MODELS.DocumentProto().json()
Dict -> PROTO_TO_PYDANTIC_MODELS.DocumentProto().dict()
```
- The oneof key setting happens in `Document()` for both proto & pydantic model.
```python
proto_doc = DocumentProto(text='abc')
Document(proto_doc).content
# 'abc'

pydantic_doc_dict = PROTO_TO_PYDANTIC_MODELS.DocumentProto(text='abc').dict()
Document(pydantic_doc_dict).content
# 'abc'
```

deepankarm: @hanxiao please let me know if this assumption is not right.

jina/peapods/runtimes/asyncio/rest/models.py
deepankarm: We also need to change the qualname for each call. We can leave this for now, as it is easier to read

deepankarm:
**Still missing scenario:**
 
`Document` allows custom dict key-value pairs and updates them to `tags`

```python
d = Document({'id': '123', 'hello': 'world', 'tags': {'good': 'bye'}})

assert d.id == '123'  # true
assert d.tags['hello'] == 'world'  # true
assert d.tags['good'] == 'bye'  # true
```

Pydantic model skips these key-value pairs, which means if the client's (frontend) json payload has additional key-value pairs, they wouldn't get passed to `Document()`
```python
PROTO_TO_PYDANTIC_MODELS.DocumentProto(hello='world')
# DocumentProto(id='', content_hash='', granularity=0, adjacency=0, level_name='', parent_id='', chunks=[], weight=0.0, length=0, matches=[], mime_type='', uri='', tags={}, location=[], offset=0, embedding=None, score=None, modality='', evaluations=[])
```

tests/unit/peapods/runtimes/asyncio/rest/test_models.py
deepankarm: Good point. I'll add this in a later pr.

deepankarm:
> **Still missing scenario:**
> 
> `Document` allows custom dict key-value pairs and updates them to `tags`
> 
> ```python
> d = Document({'id': '123', 'hello': 'world', 'tags': {'good': 'bye'}})
> 
> assert d.id == '123'  # true
> assert d.tags['hello'] == 'world'  # true
> assert d.tags['good'] == 'bye'  # true
> ```
> 
> Pydantic model skips these key-value pairs, which means if the client's (frontend) json payload has additional key-value pairs, they wouldn't get passed to `Document()`
> 
> ```python
> PROTO_TO_PYDANTIC_MODELS.DocumentProto(hello='world')
> # DocumentProto(id='', content_hash='', granularity=0, adjacency=0, level_name='', parent_id='', chunks=[], weight=0.0, length=0, matches=[], mime_type='', uri='', tags={}, location=[], offset=0, embedding=None, score=None, modality='', evaluations=[])
> ```

This is done.

JoanFM (approved):

florian-hoenicke (approved):

PR #2148, opened 2021-03-10
refactor: returns to return

florian-hoenicke:
A small correction for the docstrings.
We had cases where we were using `:returns:` in the docstring, but it has to be `:return:`

Resolves https://github.com/jina-ai/jina/issues/2149


Yongxuanzhang (approved):
LGTM 👍

PR #2150, opened 2021-03-11
test: add delete on dump to more

cristianmtr:
- [x] add `delete_on_dump: true` to more tests
- [x] fix trying to close possibly closed file handler

### DIFF
--- a/jina/executors/indexers/keyvalue.py
+++ b/jina/executors/indexers/keyvalue.py
@@ -30,13 +30,17 @@ class BinaryPbIndexer(BaseKVIndexer):
 
         def close(self):
             """Close the file."""
-            self.body.close()
-            self.header.close()
+            if not self.body.closed:
+                self.body.close()
+            if not self.header.closed:
+                self.header.close()
 
         def flush(self):
             """Clear the body and header."""
-            self.body.flush()
-            self.header.flush()
+            if not self.body.closed:
+                self.body.flush()
+            if not self.header.closed:
+                self.header.flush()
 
     class ReadHandler:
         """
@@ -68,7 +72,8 @@ class BinaryPbIndexer(BaseKVIndexer):
 
         def close(self):
             """Close the file."""
-            self._body.close()
+            if not self._body.closed:
+                self._body.close()
 
     def __getstate__(self):
         # called on pickle save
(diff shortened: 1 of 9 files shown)


jina/executors/indexers/keyvalue.py
```diff
@@ -30,13 +30,17 @@ def __init__(self, path, mode):
 
         def close(self):
             """Close the file."""
-            self.body.close()
-            self.header.close()
+            if not self.body.closed:
+                self.body.close()
+            if not self.header.closed:
+                self.header.close()
```
maximilianwerk, line 36: Looks like a smell. Why is `close` called multiple times?

Yongxuanzhang: We have body and header I think? Both need to be closed. What would be the best way?

jina/executors/indexers/keyvalue.py
```diff
@@ -30,13 +30,17 @@ def __init__(self, path, mode):
 
         def close(self):
             """Close the file."""
-            self.body.close()
-            self.header.close()
+            if not self.body.closed:
```
Yongxuanzhang, line 33: Is it possible that self.body.closed is true? But actually, it's not closed yet?
Like something else set this to true. Or it's not setable.

jina/executors/indexers/keyvalue.py
cristianmtr: It might be called several times since it depends on what state the Indexer is in when it's closed and whether the `delete_on_dump` is set to `True`. One of the tests to which I added `del_on_dump: true` in this PR exhibited this error

jina/executors/indexers/keyvalue.py
cristianmtr: I don't think so

```
In [1]: f = open('a','w')

In [2]: f.closed
Out[2]: False

In [3]: f.closed = True
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-dcb2d4390ee6> in <module>
----> 1 f.closed = True

AttributeError: attribute 'closed' of '_io.TextIOWrapper' objects is not writable

```

maximilianwerk (approved):
LGTM

PR #2153, opened 2021-03-11
docs: fix docstring rankdriver

JoanFM:


jina/drivers/rank/aggregate/__init__.py
```diff
@@ -149,10 +149,6 @@ class Chunk2DocRankDriver(BaseAggregateMatchesRankerDriver):
                 |    |- matches: {granularity: k}
                 |
                 |-matches: {granularity: k-1} (Ranked according to Ranker Executor)
-
-        .. note::
```
Yongxuanzhang, line 153: Why remove this. It's not correct?

JoanFM: It changed to traverse on root.

Yongxuanzhang: To my understanding, for this driver, we pass in the document, the document has several chunks, and chunks have matches. Then we traverse all the chunks of the document to do some ranking(score). Is it correct? 
If so, is it a traversal on chunks or root?
If not, what would be the case?

JoanFM: This is correct, therefore the docs in input of the apply_all is the root

Yongxuanzhang: 😂  Ok. I think it's fine to remove it. The docstring above is clear and understandable to me.

Yongxuanzhang (approved):

PR #2154, opened 2021-03-11
feat: DumpRequest, splitting Indexers into DBMS and Query

cristianmtr:
Replaced by https://github.com/jina-ai/jina/pull/2260. 

do NOT close yet


jina/drivers/search.py
```diff
@@ -77,6 +77,7 @@ def _apply_all(
                 []
             )  #: missed hit results, some search may not end with results. especially in shards
             for idx, retrieved_doc in enumerate(docs):
+                # some requests would be with old, some with new (in the same batch)
```
JoanFM, line 80: what does this sentence mean?

maximilianwerk: It doesn't make any sense anymore. It was a thought we had in the middle of the designing.

jina/drivers/control.py
```diff
@@ -175,4 +178,6 @@ def __init__(
         super().__init__(executor, method, *args, **kwargs)
 
     def __call__(self, *args, **kwargs):
-        self.exec_fn(self.req.path)
\ No newline at end of file
+        fn = getattr(self.exec, 'reload_compound')
```
JoanFM, line 181: Adding a `thread` into a `Pea`, is this 100% needed?

JoanFM: this can harm also the process I think, also when is this joined?

JoanFM: Just sharing thoughts

cristianmtr: We're not sure this is 100% the way to go. But we do need an async way to start building the next version of the indexers (which means loading from the dump path, building the index etc.). This needs to be async so it doesn't block (so queries can still come through). Then when it's done, on the next `__call__`, it will just do the switch (memory reference passing, depending on each indexer implementation) and continue.

JoanFM: What about making it blocking and just have the traffic go to another Pea?

JoanFM: if when u reload u will have to lock the resource, it will be blocked also

cristianmtr: The switching will be very fast (as it happens after the pre-loading is done, which is the task of building the inner query handler in whatever format it needs to be for that specific Indexer))

cristianmtr: > What about making it blocking and just have the traffic go to another Pea?

Is this possible? Isn't this the other approach of having some MetaPod/Pea?

JoanFM: Well, it would involve a lot of work yes.

JoanFM: > The switching will be very fast (as it happens after the pre-loading is done, which is the task of building the inner query handler in whatever format it needs to be for that specific Indexer))

If it is very fast, why do we need to be async or in a different thread?

cristianmtr: That's reload.

JoanFM: will this reload block the index for keep being indexed and updated?

cristianmtr: THat's why we need to make the reload async/its own thread.

JoanFM: but I mean, they compete for the same resource? how does this Pea still serve queries or indexes or updates while is being reloaded?

cristianmtr:
Replaced by https://github.com/jina-ai/jina/pull/2260. 

do NOT close yet

PR #2159, opened 2021-03-11
refactor: match parent method

florian-hoenicke:
The parent class had a different function interface compared to the overridden method.
resolves https://github.com/jina-ai/jina/issues/2161


maximilianwerk (approved):
LGTM

PR #2166, opened 2021-03-14
test: ignore resources folder for pytest and codecov

bwanglzu:
resources folder has some python scripts, while most of them are `yml`, should not be included in pytest and codecov.


Yongxuanzhang (approved):
LGTM 👍

cristianmtr (approved):

PR #2169, opened 2021-03-15
fix: stacking bytes changes to np.bytes_

JoanFM:


jina/types/sets/document.py
```diff
@@ -196,7 +196,8 @@ def extract_docs(
                     contents[i].append(c)
                 docs_pts.append(doc)
             for i in range(len(contents)):
-                contents[i] = np.stack(contents[i])
+                if not isinstance(contents[i][0], bytes):
+                    contents[i] = np.stack(contents[i])
```
nan-wang, line 200:
```
for idx, c in enumerate(contents):
    if not c:
        continue
    if not isinstance(c[0], bytes):
        contents[idx] = np.stack(c)
```

jina/types/sets/document.py
```diff
@@ -205,7 +206,10 @@ def extract_docs(
                     continue
                 contents.append(content)
                 docs_pts.append(doc)
-            contents = np.stack(contents) if contents else None
+
+            is_bytes = len(contents) > 0 and isinstance(contents[0], bytes)
+            if not is_bytes:
+                contents = np.stack(contents) if contents else None
```
nan-wang, line 212:
```suggestion
            if not contents:
                contents = None
            elif not isinstance(contents[0], bytes):
                contents = np.stack(contents)
```

nan-wang: I think this is doing the same logic with less `if`s

nan-wang (changes requested):
minor changes

nan-wang (approved):
LGTM👍

PR #2141, opened 2021-03-10
feat: add batching capability for segmenters

JoanFM:

PR #2172, opened 2021-03-16
fix: segment driver

Yongxuanzhang:
Need to adapt all exec to add @single or @batching?
Related:
issue: https://github.com/jina-ai/examples/issues/429
reason: https://github.com/jina-ai/jina/pull/2141
solutions:
https://github.com/jina-ai/examples/pull/430
https://github.com/jina-ai/examples/pull/431


Yongxuanzhang:
Should be fixed in batching/single

PR #2177, opened 2021-03-16
refactor: remove useless if

davidbp:
Removing an if statement that is not needed


cristianmtr:
What's happening in that git history? All those commits for one line change?

cristianmtr (approved):

PR #2178, opened 2021-03-16
fix: fix multimodal example

JoanFM:


jina/resources/multimodal/pods/segmenter.py
```diff
@@ -22,6 +24,7 @@ def craft(self, tags):
 class BiSegmenter(Segmenter):
     """Segmenter for multimodal example."""
 
+    @single_multi_input(num_data=2, flatten_output=False)
```
Yongxuanzhang, line 27: This setting increases the learning time for developers. They not only need to learn which exec to choose, but also determine how to decorate the method.
Is it possible to maintain the previous code and behavior, and we only add these decorators when needed?

JoanFM: I would love to do so, but the cost for this would mean having an extra set of Drivers, (with an explosion of Drivers).

One thing I wanted to do was to see if we can add the decorators automatically.

But yes, it is a problem from this point of view. 

What can we do about this @nan-wang ?

Yongxuanzhang:
Shall we merge this?

Yongxuanzhang (approved):

hanxiao:
reject. as `stack_contents` is unnecessary

JoanFM:
> reject. as `stack_contents` is unnecessary

How will it be handled then?

PR #2189, opened 2021-03-16
fix: hub_build_level  test

Yongxuanzhang:
This is a test PR from the forked repo.
For PRs from fork repo, tests/unit/docker/test_hub_build_level.py will fail. The direct reason is that the list is empty. 
`image_name = list(filter(lambda image: _filter_repo_tag(image), images))[0]`

- [ ] One solution is that we could catch this exception and skip the two tests.
- [x] Another is to pull an image before using it.
- [ ] Refactor the hubIO tests?

This issue needs to dig deeper. This means that for PR from fork repo, before this test there are no images in the testing environment, but for our PRs, there are images in env. 
Then there must be some other tests not working as expected, take this for example: 
https://github.com/jina-ai/jina/blob/4578fd65c4b5477f1fd2ddf723aaa7a7e9cbc22d/tests/unit/docker/test_hub_build.py#L12

This test is not asserted so we won't know even it fails. If it **fails to pull the images**, the test of this PR will fail for sure.

### DIFF
--- a/tests/unit/docker/test_hub_build_level.py
+++ b/tests/unit/docker/test_hub_build_level.py
@@ -27,48 +27,45 @@ def docker_image():
         else:
             return False
 
+    img_name = 'jinahub/pod.dummy_mwu_encoder:0.0.6'
     client = docker.from_env()
+    client.images.pull(img_name)
     images = client.images.list()
-    try:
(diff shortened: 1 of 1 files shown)


nan-wang (approved):
LGTM👍

PR #2190, opened 2021-03-16
docs: spelling correction

florian-hoenicke:


jina/parsers/client.py
```diff
@@ -36,7 +36,7 @@ def mixin_client_cli_parser(parser):
     gp.add_argument(
         '--top-k',
         type=int,
-        help='The number of results will be returned. Sorted by their scores descendingly.',
+        help='The number of results will be returned. Sorted by their scores in descending order.',
```
JoanFM, line 39: We should remove this part, it is not guaranteed to be returned sorted in descending order

JoanFM (changes requested):

jina/parsers/client.py
florian-hoenicke: wow okay, I didn't know that. Interesting.
Is it because the rankers can change the sorting?

JoanFM: well, it is because it depends if u sort or not them, or if they come right after a Vector Search (where score is distance)

jina/executors/indexers/__init__.py
```diff
@@ -176,15 +176,15 @@ def write_handler(self):
             return r
 
     def get_query_handler(self):
-        """Get a *readable* index handler when the ``index_abspath`` already exist, need to be overrided"""
+        """Get a *readable* index handler when the ``index_abspath`` already exist, need to be overridden"""
```
bwanglzu, line 179: just curious, should it be `overwritten`? checkout [this link](https://english.stackexchange.com/questions/88400/when-to-use-override-and-overwrite)

JoanFM (approved):

jina/executors/indexers/__init__.py
cristianmtr: You override a method, not overwrite it

cristianmtr: https://stackoverflow.com/a/30330202  https://stackoverflow.com/a/8962940

jina/parsers/client.py
```diff
@@ -36,7 +36,7 @@ def mixin_client_cli_parser(parser):
     gp.add_argument(
         '--top-k',
         type=int,
-        help='The number of results will be returned. Sorted by their scores descendingly.',
+        help='At maximum k results are returned.',
```
cristianmtr, line 39:
```suggestion
        help='The maximum nr of results to return.',
```

PR #2197, opened 2021-03-17
feat: online update poc

JoanFM:


jina/peapods/runtimes/zmq/base.py
```diff
@@ -14,11 +14,27 @@ def __init__(self, args: 'argparse.Namespace'):
         self.ctrl_addr = Zmqlet.get_ctrl_address(
             self.args.host, self.args.port_ctrl, self.args.ctrl_with_ipc
         )[0]
+        self.other_pea_sync_addr = 'tcp://0.0.0.0:55333'
```
JoanFM, line 17: I think this pattern shows that maybe the best way to do so would be to use a broker

PR #2198, opened 2021-03-17
⚠️ Refactor rename length attribute to siblings - breaking change

florian-hoenicke:
As requested by @hanxiao, the length attribute needs to be changed to siblings.
This implies  breaking changes in the 
- examples
- hub
- documentation
resolves https://github.com/jina-ai/jina/issues/2201


jina/types/document/__init__.py
```diff
@@ -205,22 +205,23 @@ def __init__(
...
+        :return: number of chunks generated from the parent document
         """
-        # TODO(Han): rename this to siblings as this shadows the built-in `length`
 
         return self._pb_body.length
 
-    @length.setter
-    def length(self, value: int):
-        self._pb_body.length = value
+    @siblings.setter
```
JoanFM, line 217: If we aim for having this enforced to be the number of siblings. This may need to be handled at ChunkSet level, where everytime a Document is added,   we may need to make sure that the complete set updates their siblings value? I am not sure if that would come at too much cost? But if we aim for having such a behavior I would enforce it.

Otherwise I would even remove this field from our Proto

florian-hoenicke: You are right, it is never used.
I think if this attribute is meaningful to someone, they can have it within the tags of the document.

cristianmtr: Yeah it doesn't seem to be used anywhere. And just using tags would be sufficient, as Florian said

nan-wang (approved):
LGTM👍 This won't be considered as a Breaking change because deleting fields from protobuf won't hurt. If the document length is not used anywhere, we can delete it safely.

florian-hoenicke:
@nan-wang I wonder if some of our users might have used it in the past. So it might be good for them to know that this attribute does not exist anymore.
Also, some tests of the hub need to be fixed after the release. The actual hub components don't need this attribute, but some tests are setting it.

florian-hoenicke:
@Yongxuanzhang found out that the `BiMatchRanker` is using the `length` attribute via reflection.

PR #2204, opened 2021-03-18
fix: use import extensions

JoanFM:


bwanglzu (commented):
will this change raise ModuleNotFoundError?

nan-wang (approved):
LGTM👍

Issue #2193, opened 2021-03-17
Regression test for all three hello-world demos


hanxiao:
**Describe the feature**

hello-world examples are important, if it breaks for a first-time user, then we lose this user.

The current tests on hello-world is not enough, it does not check (1) if index/search is successful (2) the sanity of retrieved results. 

**Your proposal**

Please add regression test for three hello-world.

---

**Environment**

**Screenshots**

Tasks: 
- [x] helloworld 
- [x] multimodal
- [x] chatbot
- [x] ci/cd pipeline

JoanFM:
Example for multimodal can be taken from reverted change in #2178. It was reverted for other reasons, but the test would still be valid

bwanglzu:
glad to work on it with other engineers.

JoanFM:
Multimodal helloworld has already been added, please add the other integration tests in the same subfolders

JoanFM:
One thing to be aware of.

Adding these tests, make the tests dependant on the `hub` submodule, which will bring to a situation where the tests in core do not pass because the hub needs an update and the hub will not be updatable because the new core version is not released

JoanFM:
@bwanglzu can u make sure the helloworld examples tests are moved to their own folder independent of integration tests? Like this we can easily skip it when necessary much easier

bwanglzu:
@JoanFM I will

PR #2205, opened 2021-03-18
test: regression test for three examples

bwanglzu:
part of #2193 :

1. Move `multimodal` regression test into `system` folder.
2. Clean up `multimodal` regression test.
3. add `fashion` regression test similar as `multimodal`.

Reason for creating a new `system` folder:

1. regression test is another level of test above integration.
2. > can u make sure the helloworld examples tests are moved to their own folder independent of integration tests? Like this we can easily skip it when necessary much easier cc @JoanFM 

`chatbot` example is breaking, will try to fix it & add test in another PR.


tests/system/fashion/test_fashion.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture
+def query_document(test_query_image):
+    return Document(
+        {
+            'weight': 0.5,
+            'uri': test_query_image,
+            'granularity': 1,
```
JoanFM, line 31: granularity is 0 in this case

tests/system/chatbot/test_chatbot.py
```diff
@@ -0,0 +1,55 @@
...
+        ['--workdir', str(tmpdir), '--unblock-query-flow']
+    )
+
+
+@pytest.fixture
+def query_document():
+    return Document(
+        {
+            'text': 'Is my dog safe from virus',
+            'granularity': 1,
```
JoanFM, line 25: same here

JoanFM (changes requested):

tests/system/fashion/test_fashion.py
```diff
@@ -0,0 +1,54 @@
...
+from tests import validate_callback
+
+
+@pytest.fixture
+def helloworld_args(tmpdir):
+    return set_hw_parser().parse_args(['--workdir', str(tmpdir)])
+
+
+@pytest.fixture
+def test_query_image():
```
JoanFM, line 20: this is not used right?

tests/system/fashion/test_fashion.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.fixture
+def query_document(test_query_image):
+    return Document(content=np.random.rand(28, 28))
+
+
+def test_multimodal(helloworld_args, query_document, mocker):
```
JoanFM, line 29: the name of the test is not correct

tests/system/fashion/test_fashion.py
```diff
@@ -0,0 +1,54 @@
...
+    """Regression test for multimodal example."""
```
JoanFM, line 30: wrong annotation

JoanFM (changes requested):

Yongxuanzhang (commented):
Very good test! Learned a lot! 😄

bwanglzu:
@JoanFM fixed

tests/system/fashion/test_fashion.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.fixture
+def query_document(test_query_image):
```
rutujasurve94, line 25: `text_query_image` argument is not being used here

PR #2208, opened 2021-03-18
refactor: make some functions private

JoanFM:


jina/peapods/zmq/__init__.py
```diff
@@ -280,13 +280,6 @@ def recv_message(
             if callback:
                 return callback(msg)
 
-    def clear_stats(self):
```
deepankarm, line 283: This would get used once we start using `profile_logger` properly.

deepankarm (approved):

Issue #2112, opened 2021-03-03
Image Logger (Driver / Indexer)


lusloher:
# Feature improvement ticket

<p></p>

Background discussed can be found here: <https://jinaai.slack.com/archives/C0183GCN8AK/p1613652002117100>

This would be useful for debugging image processing (similar to `LogInfoDriver`)

Would require the implementation of a new Driver and a new Indexer

<p></p>

An initial version of the driver can be found here: <https://github.com/jina-ai/TUM-workshop/blob/fe58151c87e5a56568f8093755b5d96ed2b1d5ee/components.py#L12>

<p></p>

There will be a task to add this to the core repo, along with tests. 

<p></p>

There should be a discussion about if this is indexer or driver or both. Please sync with Max initially to answer this question. If there is a need for further discussion pull in Joan and Nan. 

<p></p>

<p></p>

lusloher:
Link to subfeature: https://jinaai.productboard.com/feature-board/planning/features/7057227

JoanFM:
I think, maybe instead of extending the drivers for this purpose, we should encourage and expand the usage of `hanging` pods to analyze traffic in the middle of the Flow.

This could also be seen as a good debugging tool

cristianmtr:
> I think, maybe instead of extending the drivers for this purpose, we should encourage and expand the usage of `hanging` pods to analyze traffic in the middle of the Flow.
> 
> This could also be seen as a good debugging tool.

Like debugging breakpoints, or is there something else you had in mind?

JoanFM:
The same way we can have hanging Pods to be used potentially for evaluation, we can use these Pods to extract the messages from the Flow, to be analyzed separately?

PR #2209, opened 2021-03-18
feat: add pngtodisk driver

cristianmtr:
Add a PngToDiskDriver, that stores intermediate representations of pngs to the workspace. 

Useful in debugging.

Think of it like `LogInfoDriver`, but for images.

- [ ] decide where helper methods should be stored
- [x] add unit test
- [x] add integration test
- [x] allow specifying workspace in a better way

Closes https://github.com/jina-ai/jina/issues/2112


jina/drivers/control.py
```diff
@@ -158,3 +160,46 @@ def __call__(self, *args, **kwargs):
...
+    ) -> None:
+        def _move_channel_axis(
+            img: 'np.ndarray', channel_axis_to_move: int, target_channel_axis: int = -1
+        ) -> 'np.ndarray':
+            if channel_axis_to_move == target_channel_axis:
+                return img
+            return np.moveaxis(img, channel_axis_to_move, target_channel_axis)
+
+        def _load_image(blob: 'np.ndarray', channel_axis: int):
+            from PIL import Image
```
JoanFM, line 195: as an external import, it should be imported with the help of `ImportChecker`

JoanFM (commented):
If we go for this, I would add these kind of drivers into a `debug.py` new file. Would not merge with actual functional ones.

jina/drivers/control.py
cristianmtr: I can't find any usages of this class

It's added here but this method is never called

def check(args: 'Namespace'):
    """Check jina config, settings, imports, network etc"""
    from jina.checker import ImportChecker

    ImportChecker(args)

tests/integration/debug/components.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, channel_axis: int = -1, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.channel_axis = channel_axis
+
+    @single(slice_nargs=2)
+    def craft(self, buffer: bytes, uri: str, *args, **kwargs) -> Dict:
+        """
+        Read image file and craft it into image matrix.
+
+        Read the image from the given file path that specified in `buffer` and save the `ndarray` of the image in
```
theUnkownName, line 31: Super small typo: "That **is** specified"

tests/integration/debug/components.py
```diff
@@ -0,0 +1,95 @@
...
+
+    The coordinate is the same coordinate-system in the
+        :py:mode:`PIL.Image`.
+
+    :param target_size: Desired output size. If size
+        is a sequence like (h, w), the output size will
+        be matched to this. If size is an int, the
+        output will have the same height and width as
+        the `target_size`.
+    :param channel_axis: Axis for channel
```
theUnkownName, line 65: Maybe it's better to specify it's the color channel?

cristianmtr: This is copy-pasted from the Hub and only used in the test. I don't think we should modify them here

tests/integration/debug/components.py
cristianmtr: Also copy-pasted from the hub just for this test

nan-wang (dismissed):
LGTM👍

cristianmtr:
> LGTM

Reverted some changes I made while debugging CI and some unnecessary changes in control drivers

hanxiao (approved):

PR #2211, opened 2021-03-18
fix: black exclude reformat

florian-hoenicke:
in https://github.com/jina-ai/jina/pull/2199 we ignored generated protobuf files for docstring matching.
In this pr, we need to ignore the reformatting as well.


Yongxuanzhang (approved):
LGTM 👍

PR #2213, opened 2021-03-19
feat: remove excludeQL from CompoundIndexer in SearchRequest

theUnkownName:
Could we remove the ExcludeQL directive? Maybe it's needed to return the embedding at some point

### DIFF
--- a/jina/resources/executors.requests.CompoundIndexer.yml
+++ b/jina/resources/executors.requests.CompoundIndexer.yml
@@ -3,10 +3,6 @@ on:
     - !VectorSearchDriver
       with:
         executor: BaseVectorIndexer
-    - !ExcludeQL
-      with:
-        fields:
-          - embedding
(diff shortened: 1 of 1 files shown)


florian-hoenicke (changes requested):
I think this is good to let the user decide if they want to keep the embedding.
Would be good to add a test for this.

theUnkownName:
If I remove the `- !ExcludeQL` from the `SearchRequest` it seems it's not enough to be able to access the embedding. Since they were excluded on the `IndexRequest` there was nothing to exclude anyway. We would need to remove it from the Index too, but that would mean saving them twice, for the `VectorIndexDriver` and the `KVIndexDriver`, does this makes sense? Am I missing something?

theUnkownName:
This will be solved in a different way with the new Indexer approach, but for the moment this not needed.

PR #2218, opened 2021-03-22
test: add regression for chatbot

bwanglzu:
Given a system under test, `test_chatbot` depends on the `index` flow to be fired. While `f.block` blocked the process. I did some search and find the official plugin: [pytest-xprocess](https://github.com/pytest-dev/pytest-xprocess)

> pytest plugin for managing processes across test runs.

This plugin:

1. Bootstrap `hello_world` example in another process.
2. Monitor system logs to check if it is ready to test using `pattern matching`.
3. Run the test.
4. Kill the process.

This PR completes regression tests for three examples.

### DIFF
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -59,7 +59,7 @@ pytest-mock:                test
 pytest-cov:                 test
 pytest-repeat:              test
 pytest-asyncio:             test
-pytest-xprocess             test
+pytest-xprocess:            test
 flaky:                      test
 mock:                       test
 requests:                   http, devel, cicd, daemon
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -59,6 +59,7 @@ pytest-mock:                test
 pytest-cov:                 test
 pytest-repeat:              test
 pytest-asyncio:             test
+pytest-xprocess             test
 flaky:                      test
 mock:                       test
 requests:                   http, devel, cicd, daemon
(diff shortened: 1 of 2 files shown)


bwanglzu:
please **DO NOT MERGE** before CI fixed.

extra-requirements.txt
```diff
@@ -59,6 +59,7 @@ pytest-mock:                test
 pytest-cov:                 test
 pytest-repeat:              test
 pytest-asyncio:             test
+pytest-xprocess             test
```
maximilianwerk, line 62: must be `pytest-xprocess:    test`. Colon is missing

tests/system/chatbot/test_chatbot.py
```diff
@@ -0,0 +1,47 @@
...
+@pytest.fixture(autouse=True)
+def start_server(xprocess, chatbot_args):
+    class Starter(ProcessStarter):
+        pattern = "You should see a demo page opened in your browser"
+        args = ["jina", "hello", "chatbot"]
+        max_read_lines = 10000
+
+    xprocess.ensure("server", Starter)
+    yield
+    xprocess.getinfo("server").terminate()
```
nan-wang, line 40: This is not used.

nan-wang (changes requested):

tests/system/chatbot/test_chatbot.py
bwanglzu: @nan-wang it was used, as you can see we specified `autouse=True`, this fixture will be automatically triggered before we run the test

nan-wang (approved):
LGTM👍

PR #2221, opened 2021-03-22
fix: test expects file with contents

cristianmtr:
Test expects all `.py` files in its dir to have contents. Ensure they all do, and ensure it always checks the same files, no matter where you're running it from

### DIFF
--- a/tests/unit/flow/test_flow.py
+++ b/tests/unit/flow/test_flow.py
@@ -409,9 +409,9 @@ def test_index_text_files(mocker, restful, datauri_workspace):
     f = Flow(restful=restful, read_only=True).add(
         uses=os.path.join(cur_dir, '../yaml/datauriindex.yml'), timeout_ready=-1
     )
-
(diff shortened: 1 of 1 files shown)


tests/unit/flow/test_flow.py
```diff
@@ -409,9 +409,9 @@ def validate(req):
     f = Flow(restful=restful, read_only=True).add(
         uses=os.path.join(cur_dir, '../yaml/datauriindex.yml'), timeout_ready=-1
     )
-
+    files = os.path.join(cur_dir, '*.py')
```
florian-hoenicke, line 412: interesting that we use python files as text input.
Wouldn't it be better to create files in the tmpdir instead of reading from the program code

cristianmtr: We can just use the yaml files. We know these will always have content

nan-wang (approved):
LGTM👍

nan-wang:
good catch!

PR #2222, opened 2021-03-22
test: remove untested

cristianmtr:
This test was never checked in CI.

It is always broken for me, locally.

The usage of shards is tested in other places


theUnkownName (approved):

tests/unit/flow/test_flow_index.py
```diff
@@ -1,76 +0,0 @@
...
-
-
-@pytest.mark.skipif(
-    'GITHUB_WORKFLOW' in os.environ, reason='skip the network test on github workflow'
-)
-@pytest.mark.parametrize('restful', [False, True])
-def test_shards_insufficient_data(mocker, restful, docpb_workspace):
-    """THIS IS SUPER IMPORTANT FOR TESTING SHARDS
-
-    IF THIS FAILED, DONT IGNORE IT, DEBUG IT
```
bwanglzu, line 38: But why this test was added in the first place?

nan-wang: These are legacy testing codes. If the similar tests have been done somewhere else, feel free to deleteit

nan-wang (approved):
LGTM👍

PR #2224, opened 2021-03-22
feat: introduce replicas

florian-hoenicke:
Reload pod during flow execution.

A variety of upcoming features need query while indexing or query while training...
Our approach is to have a separation between the data layer for the crud operations and the query instances.
Following our approach, we need to create two query instances as replicas - one for answering the incoming queries and one for building up the new index.

We discussed 4 possible implementations each on a different abstraction level.

1. a thread within an executor
2. a thread within ZedRuntime
3. recreating entire peas
4. recreating entire pods

# Update 24.03.2021

@JoanFM came up with the idea of a new Replica abstraction level between pods and peas. The implementation is changed accordingly as shown in the following graphic:

described here
https://docs.google.com/presentation/d/1SdEopa7vjiHKjBfsnzha4WmRBXqWpSGTR4Njx00-XIE/edit#slide=id.gcba09c5536_0_0

- [x] replica level is introduced
- [x] replica can be reloaded (shut down and start)
- [x] we have a fixed number of replicas
- [x] when not using replicas, nothing is changed to our current pea configuration
- [x] always make the most simple configuration (never configure peas which are not needed)
- [x] no major breaking changes. All yaml configurations are still valid and behave the same way.
- [x] naming convention and logging for replica and pea are discussed and adjusted: 
e.g. recv ControlRequest  from ctl▸pod0/replica1/pea1/ZEDRuntime▸⚐
- [x] no errors are shown in the output


maximilianwerk:
I don't believe, this is the way to go. We need to solve this on `Pea` level and do a rolling update there. Here is the reasoning:
Blocking ALL requests (even for no time, but just waiting until Pod1 has finished) at the same time is not acceptable. Imagine a system with 10k queries per second. Holding them for 500ms would be huge and create spike traffic in the system afterwards. Especially, if we envision to do this once per minute or even more often.

florian-hoenicke:
As discussed with @maximilianwerk, @nan-wang and @hanxiao, we introduce a new layer of abstraction called replica between pod and pea level.
Therefore, an update can be done without changing the ports of the previous pod. Also, shards can be updated at once which would not have worked on pea level.

tests/integration/rolling_update/test_rolling_update.py
```diff
@@ -0,0 +1,64 @@
...
+        # search requests on the old peas
+        for i in range(5):
+            flow.search('my_text', on_done=callback)
+            time.sleep(1)
+
+        # reconnect
+        print('### second pod connecting')
+        start_time = time.time()
+        # switch traffic to new pod. Blocking request, but it is super fast since the new pod is there already.
+        flow.reconnect_pod('pod1')
```
cristianmtr, line 57: Shouldn't this done as part of the next query request? How do we know when the `.update_pod` is one, as it's asynchronous?

florian-hoenicke: The request would just change the port of the previous pea.
But this is not needed anymore.
We can have a call.

jina/peapods/pods/helper.py
```diff
@@ -8,26 +8,26 @@
...
+        if args.replicas > 1:
+            _args.replica_id = idx + 1  #: if it is parallel, then replica_id is 1-indexed
+            _args.pea_role = PeaRoleType.REPLICA
             _args.identity = random_identity()
             if _args.peas_hosts:
-                _args.host = _args.peas_hosts.get(str(_args.pea_id), args.host)
+                _args.host = _args.peas_hosts.get(str(_args.replica_id), args.host)
             if _args.name:
-                _args.name += f'/{_args.pea_id}'
+                _args.name += f'/replica_{_args.replica_id}'
```
JoanFM, line 26: Inside a regular Pod u would not have replicas right?

jina/peapods/replicas/__init__.py
```diff
@@ -0,0 +1,345 @@
...
+    _set_after_to_pass,
+    _copy_to_head_args,
+    _copy_to_tail_args,
+    _fill_in_host,
+)
+from ..peas import BasePea
+from ...enums import *
+
+
+class BaseReplica(ExitStack):
```
JoanFM, line 20: In order to have Replication transparent to the Flow, a Replica should fullfill Pod interface, thus being a Pod itself I think

JoanFM:
why is parallel changed to replica directly?

jina/peapods/pods/__init__.py
```diff
@@ -250,14 +323,19 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
 
     def join(self):
-        """Wait until all peas exit"""
+        """
+        # noqa: DAR201
+        Wait until all peas exit"""
         try:
             for p in self.peas:
                 p.join()
+            for p in self.replica_list:
```
JoanFM, line 332: I do not think Replicas should be something to be added inside a Pod. A Pod should be considered a replica, and therefore add a construction "around" Pods to handle Pods which are replicas of the same concept.

If this new construction follows the same interface as Pods, the changes are almost transparent to the Flow

JoanFM: So inside a Pod, one should still have `peas_args` as `shards` and that should not change imho

JoanFM: Also, keeping this hidden, we can profit the constructs of how Flow connects different Pods and constructs the Flow pipeline

jina/flow/base.py
```diff
@@ -981,10 +981,8 @@ def rolling_update_thread(self, pod_name):
 
         :param pod_name: pod to update
         """
-        print('start thread')
         x = threading.Thread(target=self.rolling_update, args=(pod_name,))
```
JoanFM, line 984: I think it should not be in a Thread. It should be blocking. After all, the Runtime receiving queries is in another Process, so it makes no sense inmho to add a thread here

jina/peapods/compoundpod/__init__.py
```diff
@@ -308,15 +309,15 @@ def wait_start_success(self) -> None:
             self.close()
             raise
 
-    def _enter_replica(self, replica: 'BaseReplica') -> None:
+    def _enter_replica(self, replica: 'Pod') -> None:
         self.replica_list.append(replica)
         self.enter_context(replica)
 
-    def _enter_pea(self, pea: 'BasePea') -> None:
+    def _enter_pea(self, pea: 'Pea') -> None:
```
JoanFM, line 316: it should not have `_enter_pea` the `CompoundPod` right?

jina/peapods/compoundpod/__init__.py
```diff
@@ -308,15 +309,15 @@ def wait_start_success(self) -> None:
...
         self.peas.append(pea)
         self.enter_context(pea)
 
-    def __enter__(self) -> 'BasePod':
+    def __enter__(self) -> 'Pod':
```
JoanFM, line 320: it is a `CompoundPod` the return type right?

jina/peapods/zmq/__init__.py
```diff
@@ -59,12 +59,15 @@ def __init__(
...
         self.msg_sent = 0
         self.is_closed = False
+        self.is_closing = False
         self.opened_socks = []  # this must be here for `close()`
         self.ctx, self.in_sock, self.out_sock, self.ctrl_sock = self._init_sockets()
         self._register_pollin()
 
         self.opened_socks.extend([self.in_sock, self.out_sock, self.ctrl_sock])
         if self.in_sock_type == zmq.DEALER:
+            if self.args.dealer_startup_wait_time > 0:
```
JoanFM, line 69: what is the rationale behind this waiting time?

florian-hoenicke: Currently, when starting all peas, the flow context uses the `wait_start_success` method to make sure all peas are started before sending the first data request.
Here, we close and start pods while data requests arrive. This means we can not ensure that the peas are ready when receiving their first request.
Therefore I made the wait_time configurable depending on the expected executor load time.
I think a better solution in the future would be to let the pea send the idle request after it is fully initialized.
So this is a workaround.

jina/peapods/compoundpod/__init__.py
florian-hoenicke: yes thanks

jina/peapods/compoundpod/__init__.py
florian-hoenicke: yes, changed, thanks

jina/peapods/zmq/__init__.py
JoanFM: The Runtime sends the idle request in its initialization right?

florian-hoenicke: Yes, the zmqlet of the pea. Not the flow context.

jina/peapods/compoundpod/__init__.py
```diff
@@ -0,0 +1,361 @@
...
+    def num_peas(self) -> int:
+        """
+        # noqa: DAR201
+        Get the number of running :class:`BaseReplica`"""
+        return len(self.replica_list)
+
+    def __eq__(self, other: 'Pod'):
+        return self.num_peas == other.num_peas and self.name == other.name
+
+    def start(self) -> 'Pod':
```
JoanFM, line 259: `CompoundPod` as return type

jina/peapods/compoundpod/__init__.py
JoanFM: It is still there

florian-hoenicke: yes one moment, I add the docstrings in the same commit

jina/peapods/compoundpod/__init__.py
florian-hoenicke: check

jina/flow/base.py
florian-hoenicke: If we don't use a thread, we can not run queries while reloading the replicas.
Another way of user interaction would be to start a flow server:
```
with flow:
    flow.block()
```
And then send data requests to the flow as well as rolling update requests.
This can be implemented in the close future.

I could move the threading to the in the user code in the test for now.

JoanFM: the problem is the need of a Flow object to wrap a client. The proper way should be to instantiate a client outside, and yes moving the thread in the user code should be the way to go

jina/flow/base.py
```diff
@@ -992,6 +981,7 @@ def rolling_update(self, pod_name):
         """
         for i, replica in enumerate(self._pod_nodes[pod_name].replica_list):
             replica.close()
+            time.sleep(1)
```
JoanFM, line 984: why is the sleep needed?

jina/flow/base.py
florian-hoenicke: okay done

jina/flow/base.py
florian-hoenicke: From time to time the test hangs. I thought it might be because some waiting between close and start is required. But that is not the reason. I will remove it.

jina/peapods/compoundpod/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+    _set_pod_args,
+    _set_after_to_pass,
+    _copy_to_head_args,
+    _copy_to_tail_args,
+    _fill_in_host,
+)
+from .. import Pea
+from .. import Pod
+from .. import BasePod
+from ...enums import *
```
JoanFM, line 18: try to avoid importing everything

jina/peapods/compoundpod/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+    def role(self) -> 'PodRoleType':
+        """
+        Return the role of this :class:`CompoundPod`.
+
+        :return: role type
+        """
+        return self.args.pod_role
+
+    @property
+    def is_singleton(self) -> bool:
```
JoanFM, line 62: I think `is_singleton` may not make sense in `CompoundPod`?

jina/peapods/pods/__init__.py
```diff
@@ -18,7 +18,27 @@
 
 
 class BasePod(ExitStack):
```
JoanFM, line 20: I think the `BasePod` interface should be richer. At the end, it should be transparent to the `Flow` which concrete Pod is instantiating, therefore, every method needed by `Flow` to talk to `Pods` should somehow be thought as part as this interface.

jina/peapods/zmq/__init__.py
```diff
@@ -215,6 +218,7 @@ def close(self):
             This method is idempotent.
 
         """
+        self.is_closing = True
```
JoanFM, line 221: Are we sure this is necessary? When u close, there won't be any more data being received. Another thing is if u need to consume the latest messages in a buffer ...

jina/flow/base.py
```diff
@@ -967,3 +972,15 @@ def identity(self, value: str):
...
     join = needs
+
+    def rolling_update(self, pod_name):
+        """
+        Update pods one after another - only used for compound pods.
+
+        :param pod_name: pod to update
+        """
+        for i, replica in enumerate(self._pod_nodes[pod_name].replica_list):
+            replica.close()
```
JoanFM, line 983: what about removing a `Pod` and restarting a new one with the same parameters? This way we have a more `functional` style and cleaner implementations?

jina/peapods/compoundpod/__init__.py
```diff
@@ -382,3 +382,17 @@ def is_ready(self) -> bool:
...
         )
+
+    def rolling_update(self):
+        """
+        Update all replicas of this compound pod.
+        """
+        for i in range(len(self.replica_list)):
+            replica = self.replica_list[i]
+            replica.close()
+            del replica
```
JoanFM, line 393: deleting this replica has not real effect right.

You should make sure the context of this replica is exited.

At the end we went for the functional solution of replacing Pod? I like it, but wasnt it a little more complicated?

florian-hoenicke: > deleting this replica has not real effect right.
True, it gets garbage collected anyways.

> You should make sure the context of this replica is exited.
Yes, it is done by calling `replica.close()`

> At the end we went for the functional solution of replacing Pod? I like it, but wasn't it a little more complicated?
Yes, I inserted a timeout and stuff but it did not solve the issue that from time to time the test was hanging.
Therefore, I just block the flow until the rolling update is done.
I will create a follow-up ticket to have query while reloading.

jina/flow/base.py
florian-hoenicke: Sounds good - done

jina/peapods/compoundpod/__init__.py
JoanFM: I would remove the delete.

As per query while reloading, the problem is not about havint this blocking or anything. 

I think the design problem is the need for Flow instance to query wrapping a client.

This should not be needed. We should separatr the client and the Flow instance

jina/peapods/zmq/__init__.py
florian-hoenicke: If it is closing, it should not send an idle request anymore. That is the only thing I wanted to adress.

JoanFM: But since this is blocking, this could never happen right? I am not sure about it. Have u seen the problem?

jina/executors/__init__.py
```diff
@@ -351,18 +352,27 @@ def config_abspath(self) -> str:
...
-        :param pea_id: Id of the pea,
+        :param workspace_folder: folder of the workspace.
+        :param workspace_name: name of the workspace.
+        :param pea_id: id of the pea
+        :param replica_id: id of the replica
 
         :return: returns the workspace of the shard of this Executor.
         """
-        return os.path.join(workspace_folder, f'{workspace_name}-{pea_id}')
+        if replica_id is None or replica_id == 'None':
```
cristianmtr, line 370: This should be checked in a test, right?

jina/flow/base.py
```diff
@@ -7,6 +7,7 @@
 import os
 import re
 import threading
+import time
```
cristianmtr, line 10: unused import

jina/peapods/compoundpod/__init__.py
```diff
@@ -0,0 +1,398 @@
...
+            _set_after_to_pass(args)
+            self.is_head_router = True
+            self.is_tail_router = True
+            replicas_args['head'] = _copy_to_head_args(args, PollingType.ANY.is_push)
+            replicas_args['tail'] = _copy_to_tail_args(args)
+            replicas_args['replicas'] = _set_pod_args(
+                args, replicas_args['head'], replicas_args['tail']
+            )
+
+        # TODO this case does properly not exist anymore
```
cristianmtr, line 159: Should this line be removed? What's this about?

cristianmtr (changes requested):
Missing more tests:

- test that the shards are created correctly with the expected name and contents in the replicas
- another test: index `data1`, search and assert; then change the data in the dump path from which it's reading, rolling update, search and assert contents have changed;

jina/executors/__init__.py
maximilianwerk: Since `replica_id` is `Optional[int]`, how can it ever be a string?

jina/flow/base.py
```diff
@@ -31,7 +32,8 @@
 
 __all__ = ['BaseFlow']
 
-from ..peapods import BasePod
+from ..peapods import Pod, BasePod
```
maximilianwerk, line 35: `BasePod` should go to a `if False:` block, since it is only used for type checking.

jina/flow/base.py
```diff
@@ -977,3 +982,18 @@ def dump(self, pod_name, path: str, shards: int, timeout=-1):
...
+
+    def rolling_update(self, pod_name):
+        """
+        Update pods one after another - only used for compound pods.
+
+        :param pod_name: pod to update
+        """
+
+        compound_pod = self._pod_nodes[pod_name]
+        if type(compound_pod) == CompoundPod:
```
maximilianwerk, line 994: `isinstance(compound_pod, CompoundPod)`?

jina/peapods/__init__.py
```diff
@@ -1,8 +1,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-from .peas import BasePea
+from .peas import BasePea as Pea
```
maximilianwerk, line 4: This disables importing of `BasePea` from `jina.peapods`. Is this intentional? I think it makes sense, just wanted to double check.

jina/peapods/compoundpod/__init__.py
maximilianwerk: :+1: please no star imports.

jina/peapods/compoundpod/helper.py
```diff
@@ -0,0 +1,186 @@
+import copy
```
maximilianwerk, line 1: Is this just C&P from the regular pod? If so, please let's de-duplicate code.

jina/peapods/pods/__init__.py
```diff
@@ -21,7 +21,27 @@
...
```
maximilianwerk, line 23: This class can take way more common code from the CompoundPod and The BasePod. There is too much duplication.

maximilianwerk (dismissed):
- please remove star imports (`from bla import *`)
- I agree with @cristianmtr that some more tests would be great
- The `CompoundPod` duplicates a lot of code, which might go into `BasePod`. I'd like to see a cleanup before we merge. What is your opinion @JoanFM

jina/peapods/pods/__init__.py
JoanFM: yes, the BasePod should have all the interface that the `Flow` needs to know to connect different `BasePods`

Yongxuanzhang:
Just a reminder, remove the emojis in the title before merging

jina/peapods/zmq/__init__.py
florian-hoenicke: yes if it is blocking then we don't need this. I removed it

JoanFM: Well, I am not sure, is the zmq closed when the Runtime receives terminate signal or when the Pea closes the process?

jina/peapods/compoundpod/helper.py
florian-hoenicke: In the first iteration, I wanted to keep the existing pod logic as it is and refactor it in a follow-up ticket.
That would make the review process easier. If you want, I can also combine the code.

jina/peapods/__init__.py
florian-hoenicke: I have no strong opinion on this. If you want I can change it.

jina/flow/base.py
florian-hoenicke: thanks, done

jina/flow/base.py
florian-hoenicke: done

jina/peapods/compoundpod/__init__.py
florian-hoenicke: nice catch - removed

jina/flow/base.py
florian-hoenicke: thanks, done

jina/peapods/compoundpod/__init__.py
florian-hoenicke: It would make sense if we allow that the number of replicas is == 1.
But since we don't support this case, I will remove it, thanks.

jina/peapods/compoundpod/__init__.py
florian-hoenicke: thanks, done

jina/peapods/pods/__init__.py
florian-hoenicke: I thought it might be more readable to have this separated and optimize it in a follow-up pr.
Now, I combined the common logic.

jina/executors/__init__.py
```diff
@@ -356,7 +356,7 @@ def get_shard_workspace(
         workspace_folder: str,
         workspace_name: str,
         pea_id: int,
-        replica_id: Optional[int] = None,
+        replica_id: Optional[int] = -1,
```
JoanFM, line 359: then it is not Optional.

However,  I prefer to have Optional than have a default -1 as invalid, since it is more explicit

JoanFM (changes requested):

jina/executors/__init__.py
JoanFM: But I see why is needed. While I am not a fan lets keep it but then remove the Optional

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+class CompoundPod(BasePod):
+    """A CompoundPod is a immutable set of pods, which run in parallel. They share the same input and output socket.
+    Internally, the peas of the pods can run with the process/thread backend. They can be also run in their own containers.
+    :param args: arguments parsed from the CLI
+    :param needs: pod names of preceding pods, the output of these pods are going into the input of this pod
+    """
+
+    def __init__(self, args: Union['argparse.Namespace', Dict], needs: Set[str] = None):
```
JoanFM, line 21: what is the logic for these args? arr they the args that will be applied to every Pod in the Compound? Is it already a specialized argparse that takes into account?

This does not seem clear

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,253 @@
...
+        super().__init__(args, needs)
+        self.replica_list = []  # type: List['Pod']
+        if isinstance(args, Dict):
+            # This is used when a Pod is created in a remote context, where peas & their connections are already given.
+            self.replicas_args = args
+        else:
+            self.replicas_args = self._parse_args(args)
+
+    @property
+    def first_pea_args(self) -> Namespace:
```
JoanFM, line 31: This should not be the first pea of the first replica. As for the Flow is concerned, this should be the arguments of the HeadPea of the replicas. Maybe the naming of these properties need to change

JoanFM: Also here u are mixing pea naming with what is supposed to be Pod arguments

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,253 @@
...
+        :return: arguments for all Peas and pods
+        """
+        args = {
+            'peas': ([self.replicas_args['head']] if self.replicas_args['head'] else [])
+            + ([self.replicas_args['tail']] if self.replicas_args['tail'] else []),
+            'replicas': self.replicas_args['replicas'],
+        }
+        return args
+
+    @property
```
JoanFM, line 141: Is this property needed for this Pod class?

JoanFM (changes requested):

jina/peapods/pods/compoundpod.py
JoanFM: and if it is needed, shouldnt it be the sum of all the inner Pods num peas?

jina/executors/__init__.py
```diff
@@ -368,7 +368,7 @@ def get_shard_workspace(
 
         :return: returns the workspace of the shard of this Executor.
         """
-        if replica_id == -1 or replica_id is None:
+        if replica_id == -1 or replica_id is None or replica_id == 'None':
```
JoanFM, line 371: why is it even considered? it should never be a string right?

JoanFM (changes requested):

jina/executors/__init__.py
florian-hoenicke: yes, I'm still investigating why there is a 'None'- string in some cases. It should not be required.

florian-hoenicke: okay done

jina/peapods/pods/compoundpod.py
florian-hoenicke: The `num_peas` function of the `BaseFlow` needs this function.

JoanFM: and what is it used for?

jina/peapods/pods/compoundpod.py
florian-hoenicke: thanks, I fixed the naming

jina/peapods/pods/compoundpod.py
florian-hoenicke: similar to the pod passing the arguments to the peas, the Compound pod passes the arguments to the pods

florian-hoenicke: I can make the docstring more clear

jina/executors/__init__.py
florian-hoenicke: thanks, done

jina/peapods/pods/compoundpod.py
JoanFM: but it will be the same argument for a single Pod reused for new Pods? will another argument Namespace will be created?

JoanFM (commented):
`BasePod` is an abstract class, should not have properties implemented for which it misses the `attributes`

tests/integration/rolling_update/test_rolling_update.py
```diff
@@ -80,6 +81,32 @@ def test_vector_indexer_thread(config):
             flow.search(Document(text='documents after rolling update'))
 
 
+def test_workspace(config, tmpdir):
+    with Flow().add(
+        name='pod1',
+        uses='tests/integration/rolling_update/yaml/simple_index_vector.yml',
```
JoanFM, line 87: use cur_dir pattern to access these yaml files?

jina/executors/__init__.py
florian-hoenicke: @cristianmtr I created a test for the workspace
@maximilianwerk after setting it as None the type got lost and then in some scenarios, it was interpreted as String. Instead of None, I use now -1. We do the same for the pea_id. It feels not so clean but works.

jina/peapods/pods/compoundpod.py
florian-hoenicke: the Namespace is deep copied

tests/integration/rolling_update/test_rolling_update.py
florian-hoenicke: thanks, done

florian-hoenicke:
> * please remove star imports (`from bla import *`)
> * I agree with @cristianmtr that some more tests would be great
> * The `CompoundPod` duplicates a lot of code, which might go into `BasePod`. I'd like to see a cleanup before we merge. What is your opinion @JoanFM

@maximilianwerk 
1. the two star imports are removed
2. I added 
- a test for the workspaces to verify that peas from a different replica have a different workspace.
- a test checking that the reloading does not lead into deadlock
- a test making sure all replicas are used when running search requests. First request goes to replica 1 second to replica 2...
3. CompoundPod code is now combined with the pod code.

jina/peapods/pods/compoundpod.py
JoanFM: but then, doesnt a conflict of Ports happen between replicas?

What info is in this args?

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,271 @@
...
+        """
+        Return the first non-head/tail pod args
+
+        :return: arguments of the first pod
+        """
+        # note this will be never out of boundary
+        return self.replicas_args['replicas'][0]
+
+    @property
+    def port_expose(self) -> int:
```
JoanFM, line 41: This syntax is very confusing. Why is this needed?

It feels weird to ask a CompoundPod what is its port exposed

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,271 @@
...
+        """Get the grpc port number
+
+
+        .. # noqa: DAR201
+        """
+        return self.first_pod_args.port_expose
+
+    @property
+    def host(self) -> str:
```
JoanFM, line 50: same here

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,271 @@
...
+            args.uses_after = '_pass'
+
+    def rolling_update(self):
+        """
+        Update all pods of this compound pod.
+        """
+        for i in range(len(self.replica_list)):
+            replica = self.replica_list[i]
+            replica.close()
+            del replica
```
JoanFM, line 266: remove this delete, since it does nothing

JoanFM (changes requested):

jina/peapods/pods/compoundpod.py
florian-hoenicke: thanks, I thought I removed it already

jina/peapods/pods/compoundpod.py
JoanFM: I suggest thinking what properties need to be common and what not.

Needing this is a clear code smell.

JoanFM: Well, I see now. It needs the `port_expose` of the `gateway` to know where to send `grpc requests`.

JoanFM: But in this case, it is arbirary to take the `first replica`.

jina/peapods/pods/compoundpod.py
```diff
@@ -0,0 +1,270 @@
...
+        )
+
+    @property
+    def head_args(self):
+        """
+        Get the arguments for the `head` of this BasePod.
+
+        :return: arguments of the head pea
+        """
+        if self.is_head_router and self.replicas_args['head']:
```
JoanFM, line 77: I think that for `CompoundPod` we need to assume that THERE ABSOLUTELY is a `head Pea` in front of them, so for now not need to do all these IF ELSE...

JoanFM: Same applies to `tailPea`

JoanFM (commented):
We need a deep thought on what properties and functions are common to every Pod, and which are specific to each of them, and how they are used in every case by the `Flow`.

jina/peapods/pods/compoundpod.py
florian-hoenicke: I see your point. There will be a better solution for sure. 
Currently, the pod is using the same pattern. I fear that it is beyond the scope of this pr.
If you have an idea how to quickly change this I would go for this. If it requires too much time to refactor, let's do a followup pr.

JoanFM: okey, but do the refactor PR soon after

jina/peapods/pods/compoundpod.py
florian-hoenicke: nice catch! I changed it

florian-hoenicke:
> We need a deep thought on what properties and functions are common to every Pod, and which are specific to each of them, and how they are used in every case by the `Flow`.

@JoanFM I would suggest we merge this and be aware that it can be optimized. Once we combined this with the dump implementation from @cristianmtr, we can do another iteration to see if we can flatten the design of the pods and replica pods a bit more.

JoanFM (dismissed):
Let's merge and iterate on it

JoanFM (changes requested):
Wait, we need unit tests for the new class CompoundPod.

We cannot introduce such a big component without unit tests

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -71,7 +71,10 @@ def _load_executor(self):
         """Load the executor to this runtime, specified by ``uses`` CLI argument."""
         try:
             self._executor = BaseExecutor.load_config(
-                self.args.uses, pea_id=self.args.pea_id, read_only=self.args.read_only
+                self.args.uses,
+                pea_id=self.args.pea_id,
+                replica_id=getattr(self.args, 'replica_id', -1),
```
maximilianwerk, line 76: If I interpret this right, then being inside the runtime of a Pod with one replica is not the same as being inside the runtime of the first replica inside a Pod with two replicas. I am not against it, but: Why? If I remember correctly, we had the same pattern for shards and then switched to having one shard is handled in the same way as having multiple.

(We don't need to change this now, I just wanted to know, whether this is by accident or by consciously choosing it.)

maximilianwerk (commented):
Nothing, that needs change now. Just for clarification

jina/peapods/runtimes/zmq/zed.py
JoanFM: If u have one Replica u use Pod and not CompoundPod.

JoanFM: is that what u mean?

JoanFM (approved):

jina/peapods/runtimes/zmq/zed.py
florian-hoenicke: We had a short call. The question was if we should have the same workspace pattern even if replicas are 1.
This means storing the index files in:
```
'vecidx-0-0'
'vecidx-0-1'
'vecidx-0-2'
```
instead of 
```
'vecidx-0'
'vecidx-1'
'vecidx-2'
```
This is on purpose in to not break existing Jina implementations.

JoanFM: I would not break. And since this will suffer quite some changes, I would be conservative

JoanFM: at least until this is more mature

JoanFM (approved):

PR #2227, opened 2021-03-23
tests: fix broken tests

hanxiao:
- removing `SyslogHandler` from the default handler YAML config as it keeps complaining "message too long" on MacOS and pollutes the error trace


cristianmtr:
You need to rebase on `master` to pass the tests about docstrings and black

bwanglzu (dismissed):
LGTM!

PR #2207, opened 2021-03-18
feat: support for sparse arrays in documents

davidbp:
This PR wants to complete sparse support, as discussed in #2182 .

PR #2233, opened 2021-03-25
fix: pytorch sparse ndarray constructor

bwanglzu:
Was working together with @davidbp on sparse ndarray. We were never been able to set/get value to the `SparseNdarray` with Pytorch backend. Since in the child class we forget to add `proto` to the constructor.

This PR fix bug in the constructor for Pytorch sparse ndarray, without `proto` the getter and setter will always be initialised as `null_proto`.


nan-wang:
Maybe you can add a test to show the issue

bwanglzu:
@nan-wang no problem, will add now

hanxiao:
wait, if pytorch sparse does not work, then why this test passed all the time? tests/unit/types/ndarray/test_ndarray.py L100 onwards?

bwanglzu:
@hanxiao it's related to our class inherence..normally we get/set value in `NdArray` (`generic.py`), then you'll see the issue.

davidbp:
The issue was not visible from the `NDArray` object but when a `Document` had to interact with it. This solves the issue.

davidbp (commented):
Looks good to me

davidbp (approved):

bwanglzu:
merged this PR since it's blocking #2207

PR #2234, opened 2021-03-25
fix: put siblings back

florian-hoenicke:
Removing the length attribute broke the BiMatchRanker.
This pr restores the length attribute and renames it to siblings.


jina/__init__.py
```diff
@@ -60,7 +60,7 @@
 
 # do not change this line manually
 # this is managed by proto/build-proto.sh and updated on every execution
-__proto_version__ = '0.0.79'
+__proto_version__ = '0.0.80'
```
Yongxuanzhang, line 63: just curious when will we bump it to 0.1.0?

Yongxuanzhang (commented):
Maybe need to add tests?

jina/proto/jina.proto
```diff
@@ -106,6 +106,9 @@ message DocumentProto {
     // the weight of this document
     float weight = 5;
 
+    // total number of siblings of this document (docs that are in the same granularity and have the same parent_id)
+    uint32 siblings = 6;
+
```
nan-wang, line 111: Please use a new place instead of reusing the previous one. This will lead a real breaking change.
```suggestion
    // total number of siblings of this document (docs that are in the same granularity and have the same parent_id)
    uint32 siblings = 24;

```

nan-wang (changes requested):

jina/__init__.py
florian-hoenicke: you are right, probably we should have updated to 1.0.0 already. I don't know how the protobuf versions are intended to use.

florian-hoenicke:
> Maybe need to add tests?

thanks, I added the test

nan-wang (dismissed):
LGTM👍

jina/drivers/segment.py
```diff
@@ -54,8 +54,9 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs):
 
     @staticmethod
     def _add_chunks(doc, chunks):
+        num_siblings = len(chunks)
```
Yongxuanzhang, line 57: I wonder if this way will get the number of chunks for this level? Because num_siblings is called in a for loop, the `docs_chunks` is from self.exec_fn(contents) (the segment() from executor attached). so `docs_chunks` is a list of chunks or batches of chunks.

What if we put this `num_siblings = len(chunks)` before the for loop (line 52)? Which best lines with what we want?

cristianmtr (changes requested):
Can you also add an integration test? Or adapt an existing one with segmentation to assert nr of siblings?

jina/drivers/segment.py
nan-wang, line 57:
```suggestion
        num_siblings = len(chunks) + len(doc.chunks)
```

nan-wang: What is problem do you want to solve at this point? Getting access to `len(doc.chunks)`?

maximilianwerk:
Rebased to fix broken test

jina/drivers/segment.py
Yongxuanzhang: I think my understanding was wrong. It would be better to add an integration test to validate the value is what we expect.

JoanFM (commented):
Should we recompute the siblings in the `ChunkSet` and add it in every `add` operation?

jina/drivers/segment.py
```diff
@@ -54,8 +54,9 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs):
 
     @staticmethod
```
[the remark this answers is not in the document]
florian-hoenicke: Now I added a test.

jina/types/sets/chunk.py
```diff
@@ -48,17 +56,41 @@ def append(self, document: 'Document', **kwargs) -> 'Document':
         chunk.update_content_hash()
         return chunk
 
+    def extend(self, iterable: Iterable['Document']) -> None:
```
florian-hoenicke, line 59: I added the sibling update functionality to the extend method. It is a bit inconsistent because append does not update the siblings. But if we would update this attribute for append as well it could slow down the system when having many chunks. We could argue that the system is slowed down anyways if we have a lot of chunks and have the `doc.siblings = len(self)` also in the append. 
Anyways, I think there is the goal to remove siblings in the long term. So thinking about all the cases might not be necessary

nan-wang:
> Can you also add an integration test? Or adapt an existing one with segmentation to assert nr of siblings?

@cristianmtr I've added an integration test

nan-wang (approved):
LGTM👍

PR #2236, opened 2021-03-25
feat: async control request

florian-hoenicke:
Work in progress...
When making rolling updates on replicas, we can not wait until all the peas are closed. This would lead to waiting time on the search requests.
```
with flow:
    flow.index()
    flow.rolling_update() <-  needs to ran asynchronously but  needs to send the control requests synchronized (TERMINATE, IDLE)
    flow.index()
```


florian-hoenicke:
not needed at the moment

PR #2242, opened 2021-03-26
fix: use list instead of dictionary for peas-hosts

### LINKED ISSUE #2241: Use list instead of key-value dictionary when specifying hosts of peas
**Describe the feature**
In #2143, the `--peas-hosts` argument acepts a key-value dictionary. For example, the following codes will start a pod with 3 peas. The pea with id `1` and `2` will be running on the host with IP `0.0.0.1` and `0.0.0.2` respectively. The last pea will be running on the local host. 

```
jina pod --peas-hosts 1:0.0.0.1 2:0.0.0.2 --parallel 3
```

This is a bit confusing. **All the peas in the same pod have exactly the same function** and therefore there is no need to specify hosts for every single pea. 

**Your proposal**
Pass a list instead of a key-value dictionary to `--peas-hosts` as below. The peas will be evenly distributed across all the hosts.

```
jina pod --peas-hosts 0.0.0.1 0.0.0.2 --parallel 3
```

---

**Environment**

**Screenshots**

nan-wang:
close #2241

### DIFF
--- a/jina/parsers/peapods/pod.py
+++ b/jina/parsers/peapods/pod.py
@@ -65,6 +65,7 @@ The polling strategy of the Pod (when `parallel>1`)
     gp.add_argument(
         '--peas-hosts',
         nargs='+',
+        type=str,
         help='''The hosts of the peas when parallel greater than 1.
         Peas will be evenly distributed among the hosts. By default, 
         peas are running in the same host as the pod.''',
--- a/jina/parsers/peapods/pod.py
+++ b/jina/parsers/peapods/pod.py
@@ -64,12 +64,8 @@ The polling strategy of the Pod (when `parallel>1`)
     )
     gp.add_argument(
         '--peas-hosts',
-        action=KVAppendAction,
-        metavar='KEY: VALUE',
-        nargs='*',
-        help='''The hosts of the peas when parallel greater than 1,
-        pea have a new host address if the pea_id present in the map.
-        otherwise pea host will be identical to the host of pod.
-        Represented as a key value pair in argument.
-        key is the pea_id, and value is the host address.''',
+        nargs='+',
+        help='''The hosts of the peas when parallel greater than 1.
+        Peas will be evenly distributed among the hosts. By default, 
+        peas are running in the same host as the pod. ''',
     )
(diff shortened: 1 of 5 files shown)


jina/parsers/peapods/pod.py
```diff
@@ -64,12 +64,8 @@ def mixin_base_pod_parser(parser):
...
         '--peas-hosts',
-        action=KVAppendAction,
-        metavar='KEY: VALUE',
-        nargs='*',
-        help='''The hosts of the peas when parallel greater than 1,
-        pea have a new host address if the pea_id present in the map.
-        otherwise pea host will be identical to the host of pod.
-        Represented as a key value pair in argument.
-        key is the pea_id, and value is the host address.''',
+        nargs='+',
```
deepankarm, line 67: Please add a `type` to the arg to get rid of the below error. Pydantic models are based on these types.
```
<class 'pydantic.errors.ConfigError'>
unable to infer type for attribute "peas_hosts"
```

nan-wang: `type=list` is not what we want. This will generate a list of lists.

bwanglzu (approved):
LGTM!

nan-wang:
@cristianmtr I won't consider this as a breaking change because it replaces #2143

florian-hoenicke (approved):

JoanFM (commented):
Do we agree in this implementation that Head and TailPea remain in the host of the Pod arg?

jina/parsers/peapods/pod.py
```diff
@@ -64,12 +64,9 @@ def mixin_base_pod_parser(parser):
...
+        type=str,
+        help='''The hosts of the peas when parallel greater than 1.
```
JoanFM, line 69: I think we need to add in the docs here that order matters, specially to reload an index

JoanFM (changes requested):

jina/parsers/peapods/pod.py
nan-wang: Would you please elaborate more on why the order matters? Do you mean that the naming convention of the workspaces might cause trouble when reloading the index? Because there is Pea id information in the naming.

nan-wang:
> Do we agree in this implementation that Head and TailPea remain in the host of the Pod arg?

Yes, so far I didn't see the points for making head- and tail-peas distributed in other hosts.

jina/parsers/peapods/pod.py
JoanFM: Yes, that if u change the order, the pea ids will be missmatched and u will have problems when reloading the index or the executor.

nan-wang: I see. I will make proper documentation for this.

PR #2243, opened 2021-03-26
feat: peas start at index 0

cristianmtr:
Trying to make all pea ids start at 0. 

This is a breaking change!


nan-wang:
Let's hold this one for the `1.2` because this will introduce quite some changes for the dashboard.

cristianmtr:
> Let's hold this one for the `1.2` because this will introduce quite some changes for the dashboard.

So merge it after the `1.1` release? When will that be? If we keep adding tests in `master` it would mean more work to adapt them to this PR

hanxiao (approved):

Issue #2235, opened 2021-03-25
Introduce replicas


florian-hoenicke:
Reload pod during flow execution.

A variety of upcoming features need query while indexing or query while training...
Our approach is to have a separation between the data layer for the crud operations and the query instances.
Following our approach, we need to create two query instances as replicas - one for answering the incoming queries and one for building up the new index.

We discussed 4 possible implementations each on a different abstraction level.

~1. a thread within an executor
2. a thread within ZedRuntime
3. recreating entire peas
4. recreating entire pods~
5. creating a new abstraction inside a Pod, on top of Peas -- **Replica**

# Update 24.03.2021

@JoanFM came up with the idea of a new Replica abstraction level between pods and peas. The implementation is changed accordingly as shown in the following slides: 

https://docs.google.com/presentation/d/1SdEopa7vjiHKjBfsnzha4WmRBXqWpSGTR4Njx00-XIE/edit?usp=sharing

- [x] replica level is introduced
- [x] replica can be reloaded (shut down and start)
- [ ] we have a fixed number of replicas
- [ ] when not using replicas, nothing is changed to our current pea configuration
- [ ] always make the most simple configuration (never configure peas which are not needed)
- [ ] rename parallel to shards
- [ ] when `use before` or `use after` is configured head and tail do not merge into pea
- [ ] no major breaking changes. All yaml configurations are still valid and behave the same way.
- [ ] documentation about replicas
- [ ] naming convention and logging for replica and pea are discussed and adjusted: 
e.g. recv ControlRequest  from ctl▸pod0/replica1/pea1/ZEDRuntime▸⚐
- [ ] peas close asynchronously using send_message_async, ongoing requests are not failing
- [ ] new peas are created asynchronously 
- [ ] replica can be terminated at any point of time without causing deadlocks.  https://github.com/jina-ai/jina/issues/2255 (needs to be merged into `master`)
- [ ] separated workspaces for replicas (take worspace from configuration and append '_{replica_id}'
- [ ] no errors are shown in the output
- [ ] flow.plot() for visualization of parallel and replicas (wait until implementation is done)
- [ ] REST endpoint for the rolling update functionality is made
- [ ] add replicas to cli auto complete
- [ ] rename parallel to shards

FionnD:
Following a chat with team leads / CTO this morning. We will _not_ mention replicas in 1.2 release notes, but rather way until they are correctly documented which will likely be in 1.3.

JoanFM:
[...]

PR #2257, opened 2021-03-29
feat: router cancel request

florian-hoenicke:
As part of https://github.com/jina-ai/jina/issues/2235
The following slides explain why this is needed:
https://docs.google.com/presentation/d/1SdEopa7vjiHKjBfsnzha4WmRBXqWpSGTR4Njx00-XIE/edit#slide=id.gcba09c5536_0_44


jina/proto/jina.proto
```diff
@@ -317,6 +317,7 @@ message RequestProto {
             TERMINATE = 0; // shutdown the BasePod
             STATUS = 1; // check the status of the BasePod
             IDLE = 3; // used in ROUTER-DEALER pattern, tells the router that the dealer is idle
+            BUSY = 4; // used in ROUTER-DEALER pattern, tells the router that the dealer is busy (or closed)
```
JoanFM, line 320: Is it needed to have a Predefined set of commands in our Proto? Can't just the command be a String or an Integer that is interpreted by some known code in some Jina enumerator? Otherwise, we may end up with very large set of Commands, why seems not needed and overkill

JoanFM, line 320: For what is described in the slides, I would call CLOSED, or CANCEL

JoanFM (approved):

JoanFM (changes requested):
Comments above

jina/proto/jina_pb2.py
```diff
@@ -6,6 +6,7 @@
 from google.protobuf import message as _message
```
florian-hoenicke, line 6: Black reformats the whole file. It should be excluded. Seems like the exclusion is not working.

jina/proto/jina.proto
JoanFM: now we have Request primitive data type that could handle this logic, mapping from string to enumerator to guarantee consistent access to Request type. Like this we can keep a simpler proto definition

florian-hoenicke: I thought since the request is doing the opposite of IDLE it should be called BUSY and can be used in other cases as well in the future when the dealer does not want to receive data requests for whatever reason.
But since I don't have a concrete example, I would be also fine with naming it differently.
I can rename it to CANCEL.

florian-hoenicke: What I like about enums is that it is more efficient in terms of request size and that it is not allowed to insert any values. Less chance of spelling mistakes.
Once we face the issue of having too many commands, we could rethink this.

florian-hoenicke: What do you think?

florian-hoenicke: Okay, I changed it to CANCEL

JoanFM: I think is the good point to rethink this, and embrace Primitive data type.

JoanFM: ok, let's do in the future

jina/peapods/zmq/__init__.py
```diff
@@ -255,14 +255,23 @@ def send_message(self, msg: 'Message'):
...
         self.bytes_sent += send_message(self.in_sock, msg, **self.send_recv_kwargs)
         self.msg_sent += 1
-        self.logger.debug(f'idle and i {self.identity} told the router')
+        self.logger.debug(
+            f'control message {command} with id {self.identity} is sent to the router'
+        )
+
+    def _send_idle_to_router(self):
+        """Tell the upstream router this dealer is idle """
+        self._send_control_to_router('IDLE')
```
JoanFM, line 270: Watch out, there is a risk of 'recursion'. SENT_IDLE_ROUTER calls SEND_MESSAGE which calls SEND_IDLE_ROUTER.

I think this should be refactored

JoanFM (changes requested):

jina/peapods/zmq/__init__.py
```diff
@@ -65,7 +65,7 @@ def __init__(
 
         self.opened_socks.extend([self.in_sock, self.out_sock, self.ctrl_sock])
         if self.in_sock_type == zmq.DEALER:
-            self._send_idle()
+            self._send_idle_to_router()
```
JoanFM, line 68: I think this send and the option, should be moved outside `send_message`.

JoanFM (changes requested):

jina/peapods/zmq/__init__.py
JoanFM: ah no, my bad

JoanFM (approved):
LGTM

jina/peapods/zmq/__init__.py
```diff
@@ -360,6 +369,9 @@ def close(self):
         .. note::
             This method is idempotent.
         """
+        if self.in_sock_type == zmq.DEALER:
+            print('### return cancel request', self.name)
+            self._send_cancel_to_router()
```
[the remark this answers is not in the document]
JoanFM: need to do `try` `catch` here to ignore potential Exceptions, (specially is quite possible router is down)

JoanFM (changes requested):

jina/peapods/zmq/__init__.py
florian-hoenicke: Ideally yes because if it is a send method it should be static and not bound to an instance.
In this case, it is a bit different since the method has the side effect of counting the requests and the bytes sent. This means, it has access and modifies the members.

jina/peapods/zmq/__init__.py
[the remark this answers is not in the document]
florian-hoenicke: Very good point. I think you found the reason why I saw some exceptions in the logs. Yes, let's ignore the errors silently.

jina/peapods/zmq/__init__.py
```diff
@@ -370,8 +370,10 @@ def close(self):
             This method is idempotent.
         """
         if self.in_sock_type == zmq.DEALER:
-            print('### return cancel request', self.name)
-            self._send_cancel_to_router()
+            try:
+                self._send_cancel_to_router()
+            except:
+                pass  # in case the router is down already it throws an exception which can be silently ignored
```
JoanFM, line 376: I would only silence TimeOutError which may be a normal way to end, but the other exceptions I would still log them. What have u been seing?

cristianmtr (dismissed):

cristianmtr:
Need to rebase and address some broken tests

cristianmtr:
One test still failing. Seems to be timing out again

cristianmtr (approved):

PR #2260, opened 2021-03-30
feat: dump and split indexers

cristianmtr:
Splits the Indexers into 

DBMS Indexers
- stores ids -> vectors, metadata
- dumps its data using the `dump` method, via the `DumpDriver`
- this will be the future interface for indexing/update/delete
- right now we only provide a DBMSIndexer based on BinaryPbIndexer (plan is to add Postgres https://github.com/jina-ai/jina-hub/pull/6897/files )

and

Query Indexers
- only meant to write-once, upon initializing
- will provide the query side of the operations

**NOTE**

- atm `Dump` is done via a Request, similar to `Index`/ `Update` etc. Plan is to replace this via the new send control request method, once #2257 is merged. This is NOT done in this PR since it would require to refactor the `send_control_msg` methods as that PR is already doing.

---

Replaces https://github.com/jina-ai/jina/pull/2154

Represents work for https://github.com/jina-ai/internal-tasks/issues/40


jina/drivers/__init__.py
```diff
@@ -477,6 +477,8 @@ def attach(self, executor: 'AnyExecutor', *args, **kwargs) -> None:
         if self._executor_name and isinstance(executor, CompoundExecutor):
             if self._executor_name in executor:
                 self._exec = executor[self._executor_name]
+            elif self._executor_name == executor.name:
```
JoanFM, line 480: Does a `CompoundExecutor` have a `name` property? Why is this intended? Is it intended for a special subclass of `CompoundExecutor`?

jina/drivers/dbms.py
```diff
@@ -0,0 +1,27 @@
+from jina.drivers.index import BaseIndexDriver
+from .. import Document
+
+if False:
+    from ..types.sets import DocumentSet
+
+
+# TODO move?
+def doc_without_embedding(d):
+    new_doc = Document()
```
JoanFM, line 10: Document offers the option to pass a `document` and do `copy` in the constructor

jina/executors/__init__.py
```diff
@@ -527,7 +529,10 @@ def inject_config(
         tmp = fill_metas_with_defaults(raw_config)
         tmp['metas']['pea_id'] = pea_id
         tmp['metas']['read_only'] = read_only
-
+        if kwargs.get('metas'):
```
JoanFM, line 532: I think then we need to add `metas` as Optional argument

jina/executors/dump.py
```diff
@@ -0,0 +1,132 @@
...
+    """The enum of dump formats"""
+
+    DEFAULT = 0
+    NUMPY = 1
+
+
+BYTE_PADDING = 4
+
+
+class DumpPersistor:
```
JoanFM, line 20: If it does not inherit from an Executor, it may need to be moved to some special folder. But I think these functions may belong in the Driver?

jina/executors/dump.py
```diff
@@ -0,0 +1,132 @@
...
+        shard_range = list(range(shards))
+        for shard_id in shard_range:
+            if shard_id == shard_range[-1]:
+                size_this_shard = size_per_shard + extra
+            else:
+                size_this_shard = size_per_shard
+            shard_path = os.path.join(path, str(shard_id))
+            shard_written = 0
+            os.makedirs(shard_path)
+            vectors_fh, metas_fh, ids_fh = DumpPersistor._get_file_handlers(shard_path)
```
JoanFM, line 54: these file handlers should be context managed

jina/drivers/dbms.py
cristianmtr: Yes, good point. I refactored this method use the constructor with copy=True directly.

jina/drivers/__init__.py
cristianmtr: Removed

jina/clients/request/helper.py
```diff
@@ -46,6 +47,14 @@ def _new_request_from_batch(_kwargs, batch, data_type, mode, queryset):
             _add_docs_groundtruths(req, batch, data_type, _kwargs)
         elif mode == RequestType.DELETE:
             _add_ids(req, batch)
+        elif mode == RequestType.RELOAD:
+            req.path = batch
+            req.workspace = kwargs.get('workspace')
```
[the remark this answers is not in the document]
cristianmtr: This will be changed once we have Florian's changes in #2257 . Then the DUMP request will be passed as a control request directly to the Pod. This is kept here to show that the Driver/Indexer combination works, as this was the way we did Requests up until now.

jina/executors/dump.py
cristianmtr: Moved to the `drivers/`

jina/executors/dump.py
cristianmtr: That would make this method even messier. Not sure it's worth the mess

jina/clients/request/helper.py
```diff
@@ -46,6 +47,11 @@ def _new_request_from_batch(_kwargs, batch, data_type, mode, queryset):
             _add_docs_groundtruths(req, batch, data_type, _kwargs)
         elif mode == RequestType.DELETE:
             _add_ids(req, batch)
+        # TODO make Dump a control request to be passed to the Pod directly
+        elif mode == RequestType.DUMP:
+            req.path = batch
+            req.shards = kwargs.get('shards')
```
JoanFM, line 53: I would prefer to pass these kwargs specifically, do not like to extract things from `kwargs`. Otherwise maybe we may need to change the Request Constructor?

jina/executors/indexers/dbms/keyvalue.py
```diff
@@ -0,0 +1,94 @@
+import pickle
+from typing import List, Optional
+
+from jina.drivers.dump import DumpPersistor, DumpTypes
```
JoanFM, line 4: Executors should not import from `Drivers` I think.

Is better to have a special module in executors for `dump`.

jina/drivers/dump.py
```diff
@@ -0,0 +1,158 @@
...
+        :param pea_id: the id of the pea (as part of the shards)
+        :return: the generators for the ids and for the vectors
+        """
+        path = os.path.join(path, pea_id)
+        ids_gen = DumpPersistor._ids_gen(path)
+        vecs_gen = DumpPersistor._vecs_gen(path)
+        return ids_gen, vecs_gen
+
+    @staticmethod
+    def import_metas(path: str, pea_id: str):
```
JoanFM, line 98: Are all these methods so generic for every potential DBMBSIndexer?

jina/clients/__init__.py
```diff
@@ -167,6 +170,39 @@ def delete(
...
+        :param path: path to the dump
+        :param formats: the formats into which to dump
+        :param shards: nr of shards in the dump
+        :param on_done: the function to be called when the :class:`Request` object is resolved.
+        :param on_error: the function to be called when the :class:`Request` object is rejected.
+        :param on_always: the function to be called when the :class:`Request` object is  is either resolved or rejected.
+        :param kwargs: additional parameters
+        :return: None
+        """
+        # TODO make Dump a control request to be passed to the Pod directly
```
maximilianwerk, line 193: That's exactly the point. I think this function shouldn't be here anymore, since the control request is send via the `Flow`, non? Or are these control requests still send via the client? @florian-hoenicke

jina/executors/indexers/dbms/keyvalue.py
```diff
@@ -0,0 +1,94 @@
...
+from jina.executors.indexers.dbms import BaseDBMSIndexer
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
+
+
+class DBMSBinaryPbIndexer(BinaryPbIndexer, BaseDBMSIndexer):
```
JoanFM, line 9: Will it be in our vision @maximilianwerk @cristianmtr @florian-hoenicke to split BinaryPbIndexer?

Right now with this construction, the separation is not fully there. As it is now, this class still has a `query` method. I would expect in the future this separation can be full

JoanFM: This class also suffers from the Diamond Problem.

JoanFM (commented):
Some comments from my side

jina/clients/request/helper.py
cristianmtr: I addressed this in the description and [here](https://github.com/jina-ai/jina/pull/2260#discussion_r604900705)

It's done like this until we have an established refactored way of sending control requests.

jina/drivers/dump.py
cristianmtr: We'll adapt as we add new ones, if required. But I think so, for now.

jina/executors/indexers/dbms/keyvalue.py
cristianmtr: No separation yet. Especially since this is not fully ready for production. But yes, it should be done at some point.

Yes, I realize this class is not fully "done" yet. Was going to ask if we have a pattern for adapting another Indexer's storage methods (BinaryPb's) into another one

jina/proto/jina.proto
```diff
@@ -266,6 +266,7 @@ message RequestProto {
         ControlRequestProto control = 5; // a control request
         DeleteRequestProto delete = 9; // a delete request
         UpdateRequestProto update = 10; // an update request
+        DumpRequestProto dump = 11; // an dump request
```
JoanFM, line 269: What is the view on how this may evolve? Any new Indexer that is added to the hub will need a new Dump Format?

cristianmtr: No, the dump format is universal right now. Check the enum types, `DumpFormats`. So right now you can dump and import into any. 

Yes, we can add specific formats where you can optimize the format for importing etc.

jina/drivers/dump.py
```diff
@@ -0,0 +1,158 @@
...
+        shard_range = list(range(shards))
+        for shard_id in shard_range:
+            if shard_id == shard_range[-1]:
+                size_this_shard = size_per_shard + extra
+            else:
+                size_this_shard = size_per_shard
+            shard_path = os.path.join(path, str(shard_id))
+            shard_written = 0
+            os.makedirs(shard_path)
+            vectors_fh, metas_fh, ids_fh = DumpPersistor._get_file_handlers(shard_path)
```
JoanFM, line 70: use these handlers as context managers

JoanFM (changes requested):

jina/proto/jina.proto
```diff
@@ -309,6 +310,19 @@ message RequestProto {
...
+     * Represents a dump request
+     */
+    message DumpRequestProto {
+        enum DumpTypes {
+            DEFAULT = 0;
+            NUMPY = 1;
+        }
+        string path = 1; // path for the index
+        int32 shards = 2; // nr of shards in the dump
+        repeated DumpTypes formats = 3; // the formats for which to optimize
```
JoanFM, line 323: right now, where is this actually used?

cristianmtr: It should be in the Request, because the user knows what formats the importing Executors supports. The format is about them, not the exporter. The exporting Indexer just passes its data to the DumpPersistor, which will translate it.

cristianmtr: Refactored it, as it was indeed not yet used.

JoanFM: but then better remove it.

JoanFM: But what is this `DumpPersistor`? Is it something attached to an `Executor`? Is it an `Executor` itself?

cristianmtr: > but then better remove it.

Why? It will be used in the future.

cristianmtr: > But what is this `DumpPersistor`? Is it something attached to an `Executor`? Is it an `Executor` itself?

It is a helper class called by an Indexer to dump its data

JoanFM: We will see if it is used in the future, I am not so sure it should be sent with the data.

Right now there is no such a pattern.

I have the feeling that DumpPersistor should evolve into a class to be attached to an Executor (both for the Indexer and the Query One).

Also, every person that develops something new in the hub, will require to do a change in the core? and in our proto definition?

cristianmtr: > Also, every person that develops something new in the hub, will require to do a change in the core? and in our proto definition?

This is only for some formats, like NumpyIndexer. 

We can delete it, if that's what we want. But pretty sure we agreed we'd want to add this sort of dump format optimization.

jina/executors/dump.py
```diff
@@ -0,0 +1,147 @@
...
+                size_this_shard = size_per_shard
+            DumpPersistor._write_shard_data(data, path, shard_id, size_this_shard)
+
+    @staticmethod
+    def _write_shard_data(data, path, shard_id, size_this_shard):
+        shard_path = os.path.join(path, str(shard_id))
+        shard_docs_written = 0
+        os.makedirs(shard_path)
+        vectors_fp, metas_fp, ids_fp = DumpPersistor._get_file_paths(shard_path)
+        with open(vectors_fp, 'wb') as vectors_fh:
```
JoanFM, line 67: u can use the 3 context managers in one line

JoanFM (changes requested):

jina/executors/indexers/dbms/keyvalue.py
maximilianwerk: Refactor common code to a Mixin which is used by both classes might be a proper approach.

JoanFM: yes, we can see it when this is in a more mature state , but we will need to remember to do so.

jina/executors/__init__.py
maximilianwerk: Shouldn't that be `if 'metas' in kwargs`?

jina/clients/__init__.py
```diff
@@ -169,13 +172,46 @@ def delete(
...
         )
 
+    def dump(
+        self,
+        path: str,
+        on_done: CallbackFnType = None,
+        on_error: CallbackFnType = None,
+        on_always: CallbackFnType = None,
+        shards: int = None,
+        formats: List[DumpTypes] = None,
```
maximilianwerk: I would find the interface cleaner, if that would be just `format` and if you want to export to different formats, you need to call multiple times. I see the point, that if you export to multiple formats in one go, it may be easier to optimize. But I don't see that use-case as predominant and thus, would prefer a cleaner interface.

jina/executors/indexers/query/compound.py
```diff
@@ -0,0 +1,28 @@
...
+    #     """Get a *readable* index handler when the ``index_abspath`` already exist, need to be overridden"""
+    #     return 0
+    #
+    # def get_add_handler(self):
+    #     """Get a *writable* index handler when the ``index_abspath`` already exist, need to be overridden"""
+    #     return 0
+    #
+    # def get_create_handler(self):
+    #     """Get a *writable* index handler when the ``index_abspath`` does not exist, need to be overridden"""
+    #     return 0
```
maximilianwerk, line 18: What should go in here? should this code be removed/uncommented or should some comment be added, that explains it?

jina/executors/indexers/query/keyvalue.py
```diff
@@ -0,0 +1,37 @@
+from jina.executors.dump import DumpPersistor
+from jina.executors.indexers.keyvalue import BinaryPbIndexer
+from jina.executors.indexers.query import BaseQueryIndexer
+
+
+class QueryBinaryPbIndexer(BinaryPbIndexer, BaseQueryIndexer):
+    """A write-once Key-value indexer."""
```
maximilianwerk, line 7: should we overwrite the `add`/`update`/`delete` method from the `BinaryPBIndexer` and disable them?

maximilianwerk (changes requested):
Please see the questions as real and not rhetoric questions. Not sure myself.

jina/clients/__init__.py
maximilianwerk: The format code now went elsewhere. Question remains the same.

cristianmtr: I thought we agreed to remove formats from the MVP, and just dump to default. We would then add this at another point, if we go that way.

jina/executors/__init__.py
```diff
@@ -569,8 +574,8 @@ def attach(self, runtime: 'ZEDRuntime', *args, **kwargs):
         :param args: Additional arguments.
         :param kwargs: Additional key word arguments.
         """
-        for v in self._drivers.values():
-            for d in v:
+        for list_of_drivers in self._drivers.values():
+            for d in list_of_drivers:
```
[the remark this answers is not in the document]
cristianmtr: Yep, good point

jina/executors/dump.py
```diff
@@ -0,0 +1,147 @@
...
+    """The enum of dump formats"""
+
+    DEFAULT = 0
+
+
+class DumpPersistor:
+    """
+    Class for creating and importing from dumps
+
+    Do NOT instantiate. Only provides static methods
```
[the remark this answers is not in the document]
cristianmtr: None. I initially envisioned things differently, but now there's no need to have a class for it. Will change

jina/executors/indexers/query/keyvalue.py
cristianmtr: Yes, that's the plan

jina/clients/__init__.py
maximilianwerk: :+1:

jina/executors/indexers/query/compound.py
cristianmtr: Removed.

jina/executors/indexers/query/keyvalue.py
cristianmtr: Done, check again if you can

jina/executors/dump.py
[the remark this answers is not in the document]
cristianmtr: Done, check again if you can

jina/executors/__init__.py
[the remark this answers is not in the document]
cristianmtr: Done, check again if you can

jina/executors/dump.py
cristianmtr: Done, check again if you can

jina/executors/indexers/dbms/keyvalue.py
cristianmtr: Done, check again if you can

jina/executors/indexers/query/__init__.py
```diff
@@ -0,0 +1,11 @@
+from jina.executors.indexers import BaseIndexer
+
+
+class BaseQueryIndexer(BaseIndexer):
+    """An indexer only for querying. It only reads once (at creation time)"""
```
JoanFM, line 5: If this is only loaded once at creation time, shouldn't `load_dump` be a `private` function, and be the `constructor` detecting if it should be loaded from `dump`?

JoanFM: I see this being the pattern, so I think there is no need to add this extra class just to add a `private method`.

Another thing is if there is a pattern on how the existence of `dump` is checked and initialized from it

jina/types/request/dump.py
```diff
@@ -0,0 +1,55 @@
...
+        .. # noqa: DAR201
+        """
+        return self.body.shards
+
+    @shards.setter
+    def shards(self, value):
+        self.body.shards = value
+
+    @property
+    def formats(self):
```
JoanFM, line 41: if formats are not used, let's remove it from here

JoanFM (changes requested):

jina/executors/indexers/query/keyvalue.py
```diff
@@ -0,0 +1,58 @@
...
+        self.write_handler.close()
+        self.handler_mutex = False
+        self.is_handler_loaded = False
+        del self.write_handler
+        # warming up
+        self.query('someid')
+        # the indexer is write-once
+        self.add = lambda *args, **kwargs: self.logger.warning(
+            f'Index {self.index_abspath} is write-once'
+        )
```
maximilianwerk, line 36: I think the code is cleaner, if you call `super.add(...)` instead of `self.add` in line 25. Then you can overwrite the `add` method in the same way as `update` and `delete`. Some sample code showcasing the behavior:
```
class A:
    def add(self):
        print("a")
        self.add()


class B(A):
    def add(self):
        print("b")

    def call_super(self):
        super().add()


B().call_super()
> a
> b
```

JoanFM: This is dangerous, what if u have an instance of `A`?

cristianmtr: Just did a lot of refactoring there. Extracted the writing into a BinaryPbMixin and then subclassed it for both the DBMS and the original Indexers. Can you check again?

jina/executors/indexers/query/keyvalue.py
```diff
@@ -0,0 +1,49 @@
...
+        self.add(list(ids), list(metas))
+        self.write_handler.flush()
+        self.write_handler.close()
+        self.handler_mutex = False
+        self.is_handler_loaded = False
+        del self.write_handler
+        # warming up
+        self.query('someid')
+        # the indexer is write-once
+        self.add = lambda *args, **kwargs: self.logger.warning(
```
JoanFM, line 25: add a TODO to remember to refactor these classes to avoid having `add` in this instance.

jina/executors/indexers/query/keyvalue.py
```diff
@@ -0,0 +1,49 @@
...
+        :param dump_path: the path of the dump"""
+        ids, metas = import_metas(dump_path, str(self.pea_id))
+        self.add(list(ids), list(metas))
+        self.write_handler.flush()
+        self.write_handler.close()
+        self.handler_mutex = False
+        self.is_handler_loaded = False
+        del self.write_handler
+        # warming up
+        self.query('someid')
```
JoanFM, line 23: why is there a need for `query`?

cristianmtr: This is the queryindexer. Or do you mean the warming up? It's to prepare the `read_handler` to be in-memory (as much as it can in the case of the BinaryPb -- only the header is actually fully loaded). It's basically front-loading any processing needed further on for querying

JoanFM: ah okey

jina/executors/indexers/query/keyvalue.py
cristianmtr: yep, done

jina/executors/indexers/query/__init__.py
cristianmtr: Hm yes we should add a mechanism to ensure it's not called twice

Right now: it would fail since it would call `add` and that's been disabled after the first call

Alternatives:

- private method only called internally
- delete it after it's used once (what we do with `add` now)

cristianmtr: I would still keep a Base class to have a hierarchy

jina/executors/compound.py
```diff
@@ -274,6 +274,7 @@ def components(self, comps: Callable[[], List]) -> None:
 
         :param comps: a function returns a list of executors
         """
+        print(f'### called set components')
```
JoanFM, line 277: remove print

JoanFM (changes requested):

jina/executors/indexers/query/keyvalue.py
maximilianwerk: > This is dangerous, what if u have an instance of `A`?

The recursive call of `self.add` was just for demonstration purpose. For sure, you shouldn't do that.

jina/executors/indexers/dbms/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+        raise NotImplementedError
+
+    def query(self, key: str, *args, **kwargs) -> Optional[bytes]:
+        """DBMSIndexers do NOT support querying
+
+        :param key: the key by which to query
+        :param args: not used
+        :param kwargs: not used
+        """
+        raise NotImplementedError('DBMSIndexers do not support querying')
```
maximilianwerk, line 60: I think in the moment, we remove the old indexers, this should go away also from the `BaseIndexer`. For now it is OK to keep it as it is.

jina/executors/indexers/dbms/keyvalue.py
```diff
@@ -0,0 +1,100 @@
...
+    def add(self, ids, vecs, metas, *args, **kwargs):
+        """Add to the DBMS Indexer, both vectors and metadata
+
+        :param ids: the ids of the documents
+        :param vecs: the vectors
+        :param metas: the metadata, in binary format
+        :param args: not used
+        :param kwargs: not used
+        """
+        vecs_metas = [pickle.dumps((vec, meta)) for vec, meta in zip(vecs, metas)]
```
maximilianwerk, line 52: Why do we not need the 
```
        if not any(keys):
            return
```
check here?

jina/executors/indexers/query/keyvalue.py
maximilianwerk: Now you definitely can handle the `add` the same way as `update` and `delete, since we only call `_add` above.

jina/executors/indexers/query/vector.py
```diff
@@ -0,0 +1,96 @@
...
+        :param dump_path: the path of the dump"""
+        ids, vecs = import_vectors(dump_path, str(self.pea_id))
+        self.add(ids, vecs)
+        self.write_handler.flush()
+        self.write_handler.close()
+        self.handler_mutex = False
+        self.is_handler_loaded = False
+        test_vecs = np.array([np.random.random(self.num_dim)], dtype=self.dtype)
+        assert self.query(test_vecs, 1) is not None
+        self.add = lambda *args, **kwargs: self.logger.warning(
```
maximilianwerk, line 48: I would prefer the same behavior as in the key value equivalent: rename current `add` to `_add` and handle the `add` as you handle `update` and `delete`.

jina/executors/indexers/query/vector.py
```diff
@@ -0,0 +1,96 @@
...
+            f'Index {self.index_abspath} is write-once'
+        )
+
+    def add(self, keys: Generator, vectors: Generator, *args, **kwargs) -> None:
```
maximilianwerk, line 52: rename to `_add` or even use a slightly different naming, if you have a better name at hand.

maximilianwerk (commented):
Well done. Not sure if the Mixin is overkill, but now it is nicely separated :)

I haven't looked yet at the dump message passing.

JoanFM:
Lets add a ticket to make sure to remove old indexers, or to complete properly the split between Index and QueryIndexers, this will change the way executors are serialized also ...

jina/proto/jina_pb2.py
```diff
@@ -16,1426 +17,2763 @@
...
-  ,
-  dependencies=[google_dot_protobuf_dot_timestamp__pb2.DESCRIPTOR,google_dot_protobuf_dot_struct__pb2.DESCRIPTOR,])
-
+    name='jina.proto',
```
florian-hoenicke, line 20: Don't reformat the file

florian-hoenicke (changes requested):

jina/proto/jina_pb2.py
cristianmtr: It's automatic, from precommit config. Does it matter?

jina/executors/indexers/query/vector.py
cristianmtr: Yes, I removed the `add` from `QueryIndexers` and am only using the private `_add`

jina/executors/indexers/query/vector.py
cristianmtr: Yep

see comment below

> Yes, I removed the add from QueryIndexers and am only using the private _add

jina/executors/indexers/dbms/keyvalue.py
cristianmtr: Thanks, missed it

cristianmtr:
> Lets add a ticket to make sure to remove old indexers, or to complete properly the split between Index and QueryIndexers, this will change the way executors are serialized also ...

Done, here https://github.com/jina-ai/jina/issues/2289

jina/proto/jina_pb2.py
florian-hoenicke: Yes, it means we would always have 4000 line commits when changing this file. Then it will always get the XL label.
I excluded the file from black.

cristianmtr: It's not excluded in the precommit, if that's what you meant. I added it now

florian-hoenicke: okay nice!

jina/executors/indexers/query/vector.py
```diff
@@ -0,0 +1,80 @@
...
+        ids, vecs = import_vectors(dump_path, str(self.pea_id))
+        self._add(ids, vecs)
+        self.write_handler.flush()
+        self.write_handler.close()
+        self.handler_mutex = False
+        self.is_handler_loaded = False
+        test_vecs = np.array([np.random.random(self.num_dim)], dtype=self.dtype)
+        assert self.query(test_vecs, 1) is not None
+
+    def _add(self, keys: Generator, vectors: Generator, *args, **kwargs) -> None:
```
nan-wang, line 49: To my understanding, we only write once for the query indexers. Why do we still need to implement `_add` here?

jina/flow/base.py
```diff
@@ -12,6 +12,9 @@
 from contextlib import ExitStack
 from typing import Optional, Union, Tuple, List, Set, Dict, TextIO
 
+from jina.peapods.zmq import send_ctrl_message
+from jina.types.message.dump import DumpMessage
+
```
nan-wang, line 17: unused imports

jina/types/request/dump.py
```diff
@@ -0,0 +1,37 @@
...
+        .. # noqa: DAR102
+
+
+        .. # noqa: DAR201
+        """
+        return self.body.shards
+
+    @shards.setter
+    def shards(self, value):
+        self.body.shards = value
```
nan-wang, line 37: The number of shards in dumping should be consistent with the shards in querying time, right?

tests/integration/dump/test_dump_dbms.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def error_callback(resp):
+        raise Exception('error callback called')
+
+    if benchmark:
+        basic_benchmark(
+            tmpdir, docs, _validate_results_nonempty, error_callback, nr_search
+        )
+
+    DUMP_PATH = os.path.join(str(tmpdir), 'dump_dir')
```
nan-wang, line 140:
```suggestion
    dump_path = os.path.join(str(tmpdir), 'dump_dir')
```

nan-wang (changes requested):
good job!

jina/executors/dump.py
```diff
@@ -0,0 +1,122 @@
...
+        raise Exception(f'path for dump {path} already exists. Not dumping...')
+    size_per_shard = size // shards
+    extra = size % shards
+    shard_range = list(range(shards))
+    for shard_id in shard_range:
+        if shard_id == shard_range[-1]:
+            size_this_shard = size_per_shard + extra
+        else:
+            size_this_shard = size_per_shard
+        _write_shard_data(data, path, shard_id, size_this_shard)
```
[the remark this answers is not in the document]
JoanFM: we do not want to mix parallelization at `executor/driver` level.

jina/executors/indexers/dbms/keyvalue.py
```diff
@@ -0,0 +1,88 @@
...
+            self._add(keys, vecs_metas)
+
+    def delete(self, ids, *args, **kwargs):
+        """Delete from the indexer by ids
+
+        :param ids: the ids of the Documents to delete
+        :param args: not used
+        :param kwargs: not used
+        """
+        ids = self._filter_nonexistent_keys(ids, self.query_handler.header.keys())
```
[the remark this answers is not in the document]
cristianmtr: You mean `_filter_`? It's used there too.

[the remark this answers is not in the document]
cristianmtr: I think it makes more sense to have the naming as such

+ `'Base' + Type` --> `BaseQueryIndexer`, `BaseDBMSIndexer`. Here it's more important that it's a `'Base'` indexer
+ `Type + Storage` --> `QueryNumpy`, `QueryBinaryPb`, `DBMSBinaryPb`. In this case a user is more likely to think of the type first, and then of the storage type.

Any preferences? @maximilianwerk @JoanFM

jina/executors/indexers/query/vector.py
```diff
@@ -0,0 +1,80 @@
+from typing import Generator, Iterable, Optional, Dict
+
+import numpy as np
+
+from jina.executors.dump import import_vectors
+from jina.executors.indexers.query import BaseQueryIndexer
+from jina.executors.indexers.vector import NumpyIndexer
+
+
+class QueryNumpyIndexer(NumpyIndexer, BaseQueryIndexer):
```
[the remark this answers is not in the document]
cristianmtr: I think it makes more sense to have the naming as such

+ `'Base' + Type` --> `BaseQueryIndexer`, `BaseDBMSIndexer`. Here it's more important that it's a `'Base'` indexer
+ `Type + Storage` --> `QueryNumpy`, `QueryBinaryPb`, `DBMSBinaryPb`. In this case a user is more likely to think of the type first, and then of the storage type.

Any preferences? @maximilianwerk @JoanFM

jina/executors/indexers/query/vector.py
```diff
@@ -0,0 +1,80 @@
...
+
+        :param dump_path: the path of the dump"""
+        ids, vecs = import_vectors(dump_path, str(self.pea_id))
+        self._add(ids, vecs)
+        self.write_handler.flush()
+        self.write_handler.close()
+        self.handler_mutex = False
+        self.is_handler_loaded = False
+        test_vecs = np.array([np.random.random(self.num_dim)], dtype=self.dtype)
+        assert self.query(test_vecs, 1) is not None
```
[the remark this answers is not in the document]
cristianmtr: Yes, otherwise the query handler is not loaded

jina/executors/indexers/query/vector.py
cristianmtr: Yes, because it uses a generator instead of a list (as in vanilla). I made a comment in the docstring of it

jina/types/request/dump.py
cristianmtr: Yes

jina/flow/base.py
cristianmtr: Cleand. Thanks!

tests/integration/dump/test_dump_dbms.py
cristianmtr: done

jina/executors/indexers/query/vector.py
```diff
@@ -0,0 +1,80 @@
...
+    def _post_init_wrapper(
+        self,
+        _metas: Optional[Dict] = None,
+        _requests: Optional[Dict] = None,
+        fill_in_metas: bool = True,
+    ) -> None:
+        super()._post_init_wrapper(_metas, _requests, fill_in_metas)
+        dump_path = _metas.get('dump_path')
+        if dump_path:
+            self.load_dump(dump_path)
```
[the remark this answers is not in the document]
cristianmtr: Did so. There is a problem with the Compound that it's not passing the metas of the `dump_path` down into the components. Added a TODO there

jina/executors/indexers/query/vector.py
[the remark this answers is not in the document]
maximilianwerk: While having no strong opinion, I prefer @nan-wang argument, to keep things together consistently. It helps the eye to get names faster, if the string `QueryIndexer` stays together.

tests/unit/dbms/test_dbms.py
```diff
@@ -0,0 +1,37 @@
+from jina.drivers.dbms import _doc_without_embedding
```
JoanFM, line 1: this unit test should be under `executors` not another level

JoanFM (changes requested):

jina/executors/indexers/query/vector.py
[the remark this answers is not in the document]
cristianmtr: Ok, done

tests/unit/dbms/test_dbms.py
cristianmtr: Ah yes, thanks. Done

JoanFM (approved):
LGTM as first step! Good work!

jina/executors/indexers/dbms/keyvalue.py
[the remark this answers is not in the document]
nan-wang: The implementation of `update` and `delete` functions are the same in `BinaryPbIndexer` and `BinaryPbDBMSIndexer`

jina/executors/indexers/query/vector.py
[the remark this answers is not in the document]
nan-wang:
```suggestion
       del self.write_handler
        assert self.query(test_vecs, 1) is not None
```

nan-wang (changes requested):

cristianmtr:
@nan-wang (can't quote your comments directly for some reason)

> The implementation of update and delete functions are the same in BinaryPbIndexer and BinaryPbDBMSIndexer

Update on DBMS is `def update(self, ids, vecs, metas, *args, **kwargs):`, which is different from BinaryPb's `def update(self, keys: Iterable[str], values: Iterable[bytes], *args, **kwargs)`

Delete was the same, removed it from DBMS

cristianmtr:
@nan-wang 

I don't see the point of your suggestion

jina/executors/indexers/query/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        _metas: Optional[Dict] = None,
+        _requests: Optional[Dict] = None,
+        fill_in_metas: bool = True,
+    ) -> None:
+        super()._post_init_wrapper(_metas, _requests, fill_in_metas)
+        self.dump_path = _metas.get('dump_path')
+        # TODO fix this
+        # don't do this for Compounds, as the _components
+        # are not yet set at this stage.
+        # for Compound we use a `_post_components`
```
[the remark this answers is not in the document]
cristianmtr: clarified

jina/executors/indexers/dbms/keyvalue.py
```diff
@@ -0,0 +1,83 @@
...
+            shards=shards,
+            size=self.size,
+            data=self._get_generator(ids),
+        )
+        self.query_handler.close()
+        self.handler_mutex = False
+        # noinspection PyPropertyAccess
+        del self.query_handler
+
+    def add(self, ids, vecs, metas, *args, **kwargs):
```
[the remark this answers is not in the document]
cristianmtr: added

jina/executors/dump.py
```diff
@@ -0,0 +1,122 @@
+import os
```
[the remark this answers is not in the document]
cristianmtr: moved

maximilianwerk (approved):
LGTM

nan-wang (approved):
LGTM

Issue #2228, opened 2021-03-24
Rename DocumentSet -> DocumentList


maximilianwerk:
**Describe the feature**
The `jina.types.sets.document.DocumentSet` is acutally a list. The same is true for all other modules/classes in `jina.types.sets`. They should be renamed consistently to `...List` including the package name `jina.types.lists`. While doing this, the old names should still stay until Jina 2.0.0 and be deprecated via the deprecated decorator.

**Subtasks**

- [ ] move all modules inside `jina.types.sets` to `jina.types.lists`
- [ ] Prerequisite for the next: add a new deprecate decorator for classes (see `deprecated_alias` to get an idea what should be logged)
- [ ] add the old names back to `jina.types.sets` by using the moved modules/classes from `jina.types.lists` and deprecate them directly with the removing announcement with `Jina 2.0.0`
- [ ] Fix all mentions inside our documentations
- [ ] Follow-up: switch all modules in the Hub to the new naming
- [ ] Follow-up: switch all examples to the new naming

FionnD:
Good issue for new hire

PR #2266, opened 2021-03-31
refactor: refactor set types

mkhilai:


Yongxuanzhang (commented):
Hi @mkhilai, thanks for your contribution! Just want to make sure that you want to work on https://github.com/jina-ai/jina/issues/2228 right?

JoanFM:
Hey @mkhilai ,

Thank you very much for the PR; It is very helpful.

Could you keep inside a sets module a set of aliases with the old names with some deprecation notice? This way we can have a smooth deprecation process for users to not have breaking changes so drastically applied on them?

maximilianwerk:
recheckcla

jina/types/sets/document.py
```diff
@@ -26,6 +26,7 @@
 __all__ = ['DocumentSet']
 
 
+@deprecated_class(new_classname="DocumentList")
 class DocumentSet(TraversableSequence, MutableSequence):
```
maximilianwerk, line 30: I would prefer removing all the inner code of this class and just type it as
```
@deprecated_class(...)
class DocumentSet(DocumentList):
    pass
```
Same holds true for the other deprecated classes.

jina/types/lists/chunk.py
```diff
@@ -0,0 +1,64 @@
...
+
+        chunk.set_attrs(
+            parent_id=self._ref_doc.id, granularity=self.granularity, **kwargs
+        )
+
+        if not chunk.mime_type:
+            chunk.mime_type = self._ref_doc.mime_type
+        chunk.update_content_hash()
+        return chunk
+
```
maximilianwerk, line 50: The changed `extend` function is missing. Please re-add it with the sibblings count.

maximilianwerk (changes requested):
Thanks a lot for your contribution. This is a great refactoring.

I would prefer not keeping the old code, but rather linking the old classes to the new ones. As a consequence the old tests can also be removed.

Furthermore, the siblings attribute is used for some ranking in the Hub. Please re-add them.

JoanFM (changes requested):
Please make sure u do not change files under hub module

jina/helper.py
```diff
@@ -128,6 +129,48 @@ def wrapper(*args, **kwargs):
     return deco
 
 
+def deprecated_class(dep_class=None, new_class=None):
+    """
+    Usage, new_class is a new class.
```
maximilianwerk, line 134: It took me some playing around to get what this decorator does. Really nice thing. Anyhow, I'd like to improve the docstring, since it is not obvious, that during runtime the old class will not even be there and replaced completely by the new one. Perhaps something like
```
"""
After applying `deprecated_class` the `dep_class` will only be an alias for the `new_class`.
Any code inside of `dep_class` will be completely overwritten.
"""
```

jina/types/lists/doc_groundtruth.py
```diff
@@ -0,0 +1,13 @@
+from typing import Sequence
+from .traversable import TraversableSequence
+
+if False:
+    from ..document.helper import DocGroundtruthPair
+
+
+class DocumentGroundtruthSequence(TraversableSequence):
```
maximilianwerk, line 8: The old implementation of this is not yet deprecated.

jina/types/lists/traversable.py
```diff
@@ -0,0 +1,87 @@
+from typing import Iterable
+import itertools
+
+if False:
+    from ..document.traversable import Traversable
+
+
+class TraversableSequence:
```
maximilianwerk, line 8: The old implementation of this is not yet deprecated.

maximilianwerk (changes requested):
Great progress. I have two more requests, then we are ready to merge:

1. Please deprecate the not yet deprecated classes, I mention below.
2. Please remove the tests for the old classes (e.g. all files in `tests/unit/types/sets/*`.

Thanks again for this refactoring. This is a great contribution!

maximilianwerk:
Hey @mkhilai ,

There are some minor cleanup tasks needed and a rebase to/merge from current master is also needed. Would you mind me taking your branch and doing it?

Have a great day!
Maximilian

mkhilai:
Hey @maximilianwerk sure, go ahead!

Regards
Mykyta

maximilianwerk:
I merged via the other PR. Thanks a lot again @mkhilai for implementing the `deprecate_class` decorator and making the initial effort. Great work, happy to have you around in the community.

PR #2271, opened 2021-04-01
fix: import warnings verbose off by default

davidbp:
This PR:

- Avoids warning users of external packages with specific arrays that are allowed in documents
- Refactors `update_ndarray` code making it more compact
- fixes `protbuff_updater` mispelled variable names to `protbuf_updater`

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -474,7 +474,7 @@ class Document(ProtoTypeMixin, Traversable):
             proto=getattr(self._pb_body, k),
         ).value = v
 
-    def _check_sparse_arrays_installed(self):
+    def _check_installed_array_packages(self):
         from ... import JINA_GLOBAL
 
         if JINA_GLOBAL.scipy_installed is None:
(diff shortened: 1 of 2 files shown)


jina/types/document/__init__.py
```diff
@@ -498,6 +501,14 @@ def embedding(self, value: Union['np.ndarray', 'jina_pb2.NdArrayProto', 'NdArray
         """
         self._update_ndarray('embedding', value)
 
+    def _update_sparse_ndarray(self, k, v, sparse_cls):
+        protbuf_updater = NdArray(
```
JoanFM, line 505: To avoid naming conflict, we can directly do

```suggestion
       NdArray(
            is_sparse=True,
            sparse_cls=sparse_cls,
            proto=getattr(self._pb_body, k),
        ).value = v
```

JoanFM (changes requested):

jina/types/document/__init__.py
davidbp: Updated removing the local variable

JoanFM (approved):
LGTM

hanxiao (changes requested):
the whole codebase is designed in a very conservative way on header logic, no heavylifting/importing is written directly in the module, they should be hided inside function/class scope

davidbp:
> the whole codebase is designed in a very conservative way on header logic, no heavylifting/importing is written directly in the module, they should be hided inside function/class scope

Commit `fix: reformat update ndarray to improve load time` should handle this issue.

jina/types/document/__init__.py
```diff
@@ -516,20 +506,42 @@ def _update_ndarray(self, k, v):
...
             NdArray(getattr(self._pb_body, k)).is_sparse = v.is_sparse
             NdArray(getattr(self._pb_body, k)).value = v.value
-        elif scipy_installed and scipy.sparse.issparse(v):
-            from ..ndarray.sparse.scipy import SparseNdArray
 
-            self._update_sparse_ndarray(k=k, v=v, sparse_cls=SparseNdArray)
-        elif tensorflow_installed and isinstance(v, tensorflow.SparseTensor):
-            from ..ndarray.sparse.tensorflow import SparseNdArray
+        else:
+            from ... import JINA_GLOBAL
```
JoanFM, line 511: can we refactor this into a private function?

JoanFM (changes requested):

jina/types/document/__init__.py
davidbp: Do you mean the whole logic from line 511 up to line 541 ?

JoanFM: yes, if possible?

davidbp: Done

PR #2275, opened 2021-04-02
revert(helper): revert get ip

hanxiao:


bwanglzu:
might need to rebase master back to the feature branch

PR #2276, opened 2021-04-02
test: fix slow tests in ci

bwanglzu:
Take out proto from `test_eval_collect_driver`, faster, and remove unused file.


tests/unit/drivers/test_eval_collect_driver.py
```diff
@@ -14,7 +15,7 @@ def input_function():
     ev2 = doc2.evaluations.add()
     ev2.value = 2
     ev2.op_name = 'op2'
-    return [doc1, doc2]
+    return [Document(doc1), Document(doc2)]
```
JoanFM, line 18: remove the need for DocumentProto

JoanFM: at the instantiation time

JoanFM (changes requested):

tests/unit/drivers/test_eval_collect_driver.py
Yongxuanzhang: yes `doc1 = DocumentProto()` can be replaced by `doc1 = Document()`

JoanFM (approved):

PR #2284, opened 2021-04-07
Fix batching hub executors 2157

Kelton8Z:


JoanFM (commented):
Why is this code duplicated? there was another branch with this code

PR #2291, opened 2021-04-08
fix: pea order pod

florian-hoenicke:


JoanFM (commented):
Why is this wanted? is it about the closure order?

florian-hoenicke:
> Why is this wanted? is it about the closure order?

Yes closing order and starting order in case of `noblock_on_start = False`.
The head pea should start after all the other peas started.
In the compound pod case, the head pea of a pod will be a dealer at the same time.
Means on startup it will send an idle request to the compound pod head pea. The compound pod head pea might then immediately return a data request.
In case the other peas of the pod are not started, the requests can not be processed.

JoanFM:
> > Why is this wanted? is it about the closure order?
> 
> Yes closing order and starting order in case of `noblock_on_start = False`.
> The head pea should start after all the other peas started.
> In the compound pod case, the head pea of a pod will be a dealer at the same time.
> Means on startup it will send an idle request to the compound pod head pea. The compound pod head pea might then immediately return a data request.
> In case the other peas of the pod are not started, the requests can not be processed.

Are we sure about this? When a Dealer Pea starts, sends the IDLE message to the `router`. In this case it will not be started right?

florian-hoenicke:
@JoanFM I thought this might solve a problem I had with the replicas where some requests were hanging. But it was not the reason.

PR #2296, opened 2021-04-09
refactor: clarify annotation types to support sparse

JoanFM:


jina/drivers/index.py
```diff
@@ -39,8 +39,7 @@ class VectorIndexDriver(BaseIndexDriver):
     If `method` is not 'delete', documents without content are filtered out.
     """
 
-    @staticmethod
-    def _get_documents_embeddings(docs: 'DocumentSet'):
+    def _get_documents_embeddings(self, docs: 'DocumentSet'):
```
cristianmtr, line 42: Why not static anymore?

jina/drivers/index.py
```diff
@@ -54,9 +53,18 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
...
     """An alias to have coherent naming with the required SparseVectorSearchDriver """
 
-    @staticmethod
-    def _get_documents_embeddings(docs: 'DocumentSet'):
-        return docs.all_sparse_embeddings
+    @property
+    def exec_sparse_cls_type(self) -> str:
+        return self.exec.sparse_cls_type
+
+    def _get_documents_embeddings(self, docs: 'DocumentSet'):
```
cristianmtr, line 60: oh I see

cristianmtr: But should the driver be storing this spare class type? Maybe it should still be static and just pass the type as another argument? I don't see it as being part of the Driver's expected knowledge

jina/drivers/search.py
```diff
@@ -20,12 +20,12 @@ class BaseSearchDriver(BaseExecutableDriver):
     """Drivers inherited from this Driver will bind :meth:`query` by default """
 
     def __init__(
-        self,
-        executor: Optional[str] = None,
-        method: str = 'query',
-        traversal_paths: Tuple[str] = ('r', 'c'),
-        *args,
-        **kwargs,
+            self,
```
cristianmtr, line 23: The formatting change is odd. Is that the result of black? It shouldn't happen anymore

jina/drivers/index.py
JoanFM: because the inherited function needs instance attributes

jina/drivers/index.py
JoanFM: oh yes, that would make sense too. We could keep it static like this

jina/drivers/search.py
JoanFM: Did not pass black check yesterday.

jina/drivers/index.py
cristianmtr: default argument being dense representation class?

jina/types/sets/document.py
```diff
@@ -163,21 +195,47 @@ def all_embeddings(self) -> Tuple['np.ndarray', 'DocumentSet']:
...
+    def get_all_sparse_embeddings(
+        self, sparse_cls_type: str, scipy_cls_type: Optional[str]
+    ) -> Tuple['SparseEmbeddingType', 'DocumentSet']:
+        # TODO: proper type hinting and docstrings and warn when scipy type BSR and CSC
+        """Return all embeddings from every document in this set as a sparse array
 
         :return: The corresponding documents in a :class:`DocumentSet`,
                 and the documents have no embedding in a :class:`DocumentSet`.
-        :rtype: A tuple of embedding in :class:`np.ndarray`
+        :rtype: A tuple of embedding amd DocumentSet as sparse arrays
```
cristianmtr, line 206:
```suggestion
        :rtype: A tuple of embedding and DocumentSet as sparse arrays
```

tests/unit/drivers/test_vector_index_driver.py
```diff
@@ -11,6 +11,9 @@
 from jina.types.document import Document
 
 
+# TODO: Add tests for sparse vectors index driver.
```
cristianmtr, line 14: just a reminder

tests/unit/drivers/test_vector_search_driver.py
```diff
@@ -7,6 +7,7 @@
 from jina.drivers.search import VectorSearchDriver
 from jina.executors.indexers import BaseVectorIndexer
 
+# TODO: Add tests for sparse vectors search driver.
```
cristianmtr, line 10: just a reminder

jina/drivers/index.py
JoanFM: It would not be used in the Dense case. It is handy to access dense embbedding via a property

jina/drivers/search.py
```diff
@@ -162,16 +161,31 @@ class SparseVectorSearchDriver(VectorSearchDriver):
...
     """
 
-    @staticmethod
-    def _get_documents_embeddings(docs: 'DocumentSet'):
-        return docs.all_sparse_embeddings
+    @property
+    def exec_sparse_cls_type(self) -> str:
+        """Get the sparse type from executor.
+
+        :return: Sparse matrix type, default value is `scipy_coo`.
```
JoanFM, line 168: do not put this note here as it is hard to mantain. The default value will be in the base class of executor, not here

jina/executors/indexers/__init__.py
```diff
@@ -270,7 +289,7 @@ def add(self, keys: Iterable[str], vectors: 'np.ndarray', *args, **kwargs) -> No
         raise NotImplementedError
 
     def query(
-        self, vectors: 'np.ndarray', top_k: int, *args, **kwargs
+        self, vectors: 'EncodingType', top_k: int, *args, **kwargs
```
JoanFM, line 292: would it be more clear if we creatr anotjher type var saying EncodeBatchType that is just an alias?

tests/unit/drivers/test_vector_index_driver.py
JoanFM: @bwanglzu I have done changes to avoid having extra `Driver` classes, so that it is easier for the user to use.

Still the `VectorIndexDriver` and `VectorSearchDriver` need to be unit tested

tests/unit/drivers/test_vector_search_driver.py
JoanFM: @bwanglzu I have done changes to avoid having extra Driver classes, so that it is easier for the user to use.

Still the VectorIndexDriver and VectorSearchDriver need to be unit tested

bwanglzu (approved):
LGTM!

PR #2297, opened 2021-04-09
feat: add driver and changes in document needed for sparse pipeline

JoanFM:


jina/drivers/encode.py
```diff
@@ -34,3 +34,22 @@ def _apply_all(self, docs: 'DocumentSet', *args, **kwargs) -> None:
                 raise LengthMismatchException(msg)
             for doc, embedding in zip(docs_pts, embeds):
                 doc.embedding = embedding
+
+
+class ScipySparseEncodeDriver(FlatRecursiveMixin, BaseEncodeDriver):
```
JoanFM, line 39: is there a unified way to work with this with other sparse vectors? Isn't it better to detect the type in `EncodeDriver`? There may be no need for an extra `encodedriver`?

davidbp: We could do this in EncodeDriver adding a generator that allow us to iterate over the rows of different 'tensor types' with the same unified interface:

```

            for doc, embedding in zip(docs_pts, embeds):
                doc.embedding = embedding
```
by
```
            for doc, embedding in zip(docs_pts, matrix_iterator(embeds)):
                doc.embedding = embedding
```

where 

```
def matrix_iterator(X):
    n_rows = X.shape[0]
    for i in range(n_rows):
        if scipy.sparse.issparse(X):
            yield X.getrow(i)
        else:
            yield X[i]
```

JoanFM: Something like this, but we do not even need to changr the iterator, we can do the if else one level above (just be careful on how to check dependencies)

JoanFM: have a private `assign_embeddings` method that works differently for numpy arrays than for sparse arrays

tests/integration/sparse_pipeline/test_sparse_pipeline.py
```diff
@@ -34,23 +34,23 @@ def encode(self, data: Any, *args, **kwargs) -> Any:
...
 
-
-    def query(
-        self, vectors: 'scipy.sparse.coo_matrix', top_k: int, *args, **kwargs
-    ):
+    def query(self, vectors: 'scipy.sparse.coo_matrix', top_k: int, *args, **kwargs):
         assert isinstance(vectors, sparse.coo_matrix)
         distances = [item for item in range(0, min(top_k, len(self.keys)))]
-        return np.array(self.keys[:top_k]), np.array(distances)
+        return np.array(self.keys[:top_k]), np.array(
```
davidbp, line 51: I think this was a bug, distances needs to be an array of arrays (for each vector I want the closest distances)

JoanFM: query is not done in batches

tests/integration/sparse_pipeline/test_sparse_pipeline.py
```diff
@@ -66,10 +65,14 @@ def get_query_handler(self):
 
 def test_sparse_pipeline(mocker, docs_to_index):
     def validate(response):
-        print(response)
+        assert len(response.docs) == 10
```
JoanFM, line 68: we should assert the length of the matches, and the type of embedding returned in the matches

tests/integration/sparse_pipeline/test_sparse_pipeline.py
```diff
@@ -66,10 +65,14 @@ def get_query_handler(self):
...
 
     f = Flow().add(uses=DummySparseEncoder).add(uses=DummyCSRSparseIndexer)
 
+    error_mock = mocker.Mock()
+
     with f:
         f.index(inputs=docs_to_index)
-        f.search(inputs=docs_to_index, on_done=validate)
+        f.search(inputs=docs_to_index, on_done=validate, on_error=error_mock)
```
JoanFM, line 76: use mock instead of directly validate, check other tests how they do

JoanFM (commented):
We are missing unittests for new methods, for instancr as `all_sparse_embeddings` and some of the new sparse drivers

bwanglzu (approved):
LGTM!

Issue #1868, opened 2021-02-04
Track workspace_id in workspace store during POST flow/pod/pea 


deepankarm:
**Describe the bug**
During `POST /flows`, `POST /pods`, `POST /peas`, workspace store doesn't track the workspace_id that created the flow/pod/pea. This should be tracked for later operations (GET/DELETE)

Additional info: 
- We keep `WorkspaceStore` updated everytime there's interaction with `POST /workspaces` and `DELETE /workspaces` endpoints. 
- When a Flow/Pod/Pea gets created via POST endpoint or gets terminated via DELETE endpoint, there are updates to different workspaces, but these don't get tracked in `WorkspaceStore`.

FionnD:
@deepankarm is this still valid?

deepankarm:
Fixed in #2473

PR #2302, opened 2021-04-12
fix: track workspace_id

n0acar:
- Fixes issue: #1868 
- When a Flow/Pod/Pea gets created via POST endpoint or gets terminated via DELETE endpoint, there are updates to different workspaces and their workspace_id get tracked in WorkspaceStore.

### DIFF
--- a/daemon/stores/workunit.py
+++ b/daemon/stores/workunit.py
@@ -1,0 +1,43 @@
+import uuid
+from pathlib import Path
+from typing import Union
+
+from jina.helper import colored
+from .base import BaseStore
+
+class WorkunitStore(BaseStore):
+    def __init__(self):
+        super().__init__()
+        from . import workspace_store
+        self._workspace_store = workspace_store
+
+    def add(self, *args, **kwargs) -> 'uuid.UUID':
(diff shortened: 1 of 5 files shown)


daemon/stores/workunit.py
```diff
@@ -0,0 +1,45 @@
+import uuid
+from pathlib import Path
+from typing import Union
+
+from jina.helper import colored
+from .base import BaseStore
+
+
+class WorkunitStore(BaseStore):
```
deepankarm, line 9: What is a `WorkunitStore`? How is it different from a `WorkspaceStore`?

deepankarm (changes requested):
Thank you so much for your PR 🙏
Could you please add tests for this?

daemon/stores/workunit.py
n0acar: WorkunitStore is an abstract class that FlowStore, PeaStore and PodStore extends. Its main function is to hold workspace_store instance and encapsulate delete function from WorkspaceStore. When a DELETE request comes to Flow/Pea/Pod, both itself and workspace should be deleted from corresponding stores. WorkunitStore provides this behavior.

deepankarm:
Superseded by #2473

PR #2307, opened 2021-04-13
fix: benchmark dump test

cristianmtr:
Fix dumprequest not being propagated to the send ctrl message

Baseline:
```
### baseline - query time with 1 on 100000 docs ...	         Client@86294[S]:connected to the gateway at 0.0.0.0:55803!
### baseline - query time with 1 on 100000 docs takes 1 second (1.57s)
### baseline - indexing: 100000 docs ...	         Client@86294[S]:connected to the gateway at 0.0.0.0:37613!
### baseline - indexing: 100000 docs takes 9 seconds (9.75s)
```

Dump / Reload

```
### indexing 100000 docs takes 17 seconds (17.56s)
### dumping 100000 docs takes 3 seconds (3.27s)
### dump path size: 112.75556 MBs
### reloading 100000 takes 2 seconds (2.50s)
```

The reason indexing is slower for the DBMSBinaryPb vs Baseline BinaryPb is because we need to pickle both the vector and the metadata before storing them in the underlying KV data structure the BinaryPb uses:

```
    def add(
        self, ids: List[str], vecs: List[np.array], metas: List[bytes], *args, **kwargs
    ):
        """Add to the DBMS Indexer, both vectors and metadata

        :param ids: the ids of the documents
        :param vecs: the vectors
        :param metas: the metadata, in binary format
        :param args: not used
        :param kwargs: not used
        """
        if not any(ids):
            return

        vecs_metas = [pickle.dumps((vec, meta)) for vec, meta in zip(vecs, metas)]
        self._add(ids, vecs_metas)
```

### DIFF
--- a/jina/types/message/__init__.py
+++ b/jina/types/message/__init__.py
@@ -126,7 +126,9 @@ class Message:
 
         :return: boolean which states if data is requested
         """
-        return self.envelope.request_type != 'ControlRequest' or self.request.propagate
+        return (
+            self.envelope.request_type != 'ControlRequest' or self.request.propagate
+        ) and self.envelope.request_type != 'DumpRequest'
 
     def _add_envelope(
         self,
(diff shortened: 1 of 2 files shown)


maximilianwerk (approved):
LGTM

PR #2313, opened 2021-04-18
refactor(driver): add the extract-apply-update pattern

hanxiao:
# Background

As presented on April 7 2021 Engineering Show & Tell, and private discussions with @bwanglzu @JoanFM and @florian-hoenicke we aim to solve **the usability problems when writing a new `Executor`**. To recap, the usability problem can be specified as three subproblems when a user tries to create a new `Executor`:
1. User has to find out which `Executor` to inherit.
2. User has to know which method to override.
3. User has to know what do the arguments of the overridden method mean.

# Scope of this PR

This PR solves **Problem 3**, Problem 2 is solved for free (_we have it supported before this PR_). It also sheds a light on Problem 1. While reviewing this PR, please focus on Problem 3, as it is my major & original intention.

This PR presents a new Driver mixin class `DocsExtractUpdateMixin` that summarizes the pattern of extracting -> applying -> updating. Specifically, 
- It leverages **Python reflection & type hint** to inspect executors `exec_fn`'s arguments;
- It extracts the same name attributes from `Document` types;
- It feeds the extraction to `exec_fn`;
- It updates `Document` with the the result returned from `exec_fn`.

# Understanding Problem 3

Take `BaseEncoder` as the example, the "core" logic function `encode` has the following specification:
```python
    def encode(self, data: Any, *args, **kwargs) -> 'EncodingType':
```

It is impossible to tell what `data` argument is by just looking at this signature. The same problem also appears in `BaseClassifier`, `BaseIndexer`, etc.

```python
def predict(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
```

```python
def add(
        self, keys: Iterable[str], values: Iterable[bytes], *args, **kwargs
    ) -> None:
```

```python
def query(self, key: str, *args, **kwargs)
```

It is impossible for users to know what `data`, `key`, `values`, `keys` actually mean.

## Reason behind Problem 3

The reason behind is that:
- `Driver` describes what those arguments in the executor means. The extraction logic is written inside the `Driver`.
- `Driver` has much less visibility than `Executor` to users.
- Hence from `Executor` perspective, one can hardly guess the argument without reading the code of the `Driver`.

The following figure describes this problem.

image

# General Idea of the Solution

As presented on April 7 2021 Engineering Show & Tell, and private discussions with @bwanglzu @JoanFM and @florian-hoenicke, the solution is to implement a contract between the `Document` attributes and `Executor`'s logic function; and removing the customized extraction logic from the Driver. Simply put, the executor's logic function must name arguments as valid `Document` attributes. (At the current version, this means the name of the arguments must come from `{'parent_id', 'content_type', 'siblings', 'mime_type', 'chunks', 'granularity', 'buffer', 'tags', 'level_name', 'proto', 'offset', 'modality', 'text', 'uri', 'matches', 'evaluations', 'content_hash', 'blob', 'weight', 'id', 'adjacency', 'location', 'embedding', 'score', 'non_empty_fields', 'content'}`. Note that this list is auto-derived at runtime, not hardcoded.)

Comparing to previous where the driver gives the executor whatever data structure built inside the driver, now the driver gives what executor's `exec_fn` explicitly asks for. From the Executor's developer aspect, he now gets a feeling of "controlling" what he want.

image

There are some smart logics implemented inside `DocsExtractUpdateMixin` to improve usability. In general, the principles on the new user experience of Executor's logic function are:
- Learn `Document` type once, write all;
- Readability matters;
- Throw helpful error message.

## Affects & Aftermath

This PR currently covers the following drivers and executors:
- `EncodeDriver` <-> `Encoder`
- `CraftDriver` <-> `Crafter`
- `SegmentDriver` <-> `Segmenter`
- `PredictDriver` <-> `Classifier`

`IndexDriver` & `SearchDriver` & `Indexer` are in the scope as they fall into the same pattern. But refactoring them maybe overshot in this PR.

### Aftermath on Hub Executors

Note that this PR only requires simple follow-up changes on hub executors: i.e. changing their `exec_fn` signature to comply with Document types. As a hub developer, I would very much happy for this change as it greatly improves the readability of the executor.

[A temp patch is added to the chatbot demo](https://github.com/jina-ai/jina/pull/2313/files#diff-38e1b2dbc86f8dbb056decbebc2fdf488f5a671283c679da92224ff63062c7afR46-R59), this should be reverted after Hub executors got updated.

# Code Examples

The encoder's `encode` now can be written as:

```python
class MyEncoder(BaseEncoder):
  def encode(content: 'np.ndarray'):
    pass
```

`content` is an attribute defined in `Document` type.

### Accept arbitrary `Document` attributes

`encode` function now can also accepts multiple & arbitrary document types, e.g.

```python
class MyEncoder(BaseEncoder):
  def encode(content: 'np.ndarray', uri: str, mime_type: str, id: str):
    pass
```

A new document property `raw` is added, hence the executor can directly access the Document.

```python
class MyEncoder(BaseEncoder):
  def encode(raw: 'Document')
    pass
```

### Type annotation matters

Type hint an argument with `np.ndarray` or `ndarray` will tell the driver to stack data into Numpy Ndarray. For example,

```python
class MyExecutor(BaseEncoder):
    def encode(self, id, embedding):
        pass
```

gives:
```
[['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'], [array([0.5488135 , 0.71518937, 0.60276338, 0.54488318, 0.4236548 ,
       0.64589411, 0.43758721, 0.891773  , 0.96366276, 0.38344152]), array([0.15896958, 0.11037514, 0.65632959, 0.13818295, 0.19658236,
       0.36872517, 0.82099323, 0.09710128, 0.83794491, 0.09609841]),...]]
```
---

```python
class MyExecutor(BaseEncoder):
    def encode(self, id: 'ndarray', embedding):
        pass
```

gives:
```
([array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'], dtype='<U1'), [array([0.5488135 , 0.71518937, 0.60276338, 0.54488318, 0.4236548 ,
       0.64589411, 0.43758721, 0.891773  , 0.96366276, 0.38344152]), array([0.15896958, 0.11037514, 0.65632959, 0.13818295, 0.19658236,
       0.36872517, 0.82099323, 0.09710128, 0.83794491, 0.09609841]),...]]
```
---

```python
class MyExecutor(BaseEncoder):
    def encode(self, id: 'ndarray', embedding: 'ndarray'):
        pass
```

gives:
```
([array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'], dtype='<U1'), array([[0.5488135 , 0.71518937, 0.60276338, 0.54488318, 0.4236548 ,
        0.64589411, 0.43758721, 0.891773  , 0.96366276, 0.38344152],
       [0.15896958, 0.11037514, 0.65632959, 0.13818295, 0.19658236,
        0.36872517, 0.82099323, 0.09710128, 0.83794491, 0.09609841], ...]
```

# Problem 2

Though not the major focus of this PR, it is obvious that Problem 2 can be solved by setting `method` args for the driver. For example,

```python
class MyEncoder(BaseEncoder):
  def foo(content: 'np.ndarray'):
     pass
```

To bind `foo` with `EncodeDriver`, one only need to set `method='foo'`, e.g. in YAML via
```yaml
jtype: MyEncoder
requests:
  on:
     IndexRequest:
       - jtype: EncodeDriver
         with:
            method: foo
```

Or in Python (often used in tests), 

```python
exec = MyEncoder()
bd = EncodeDriver(method='foo')
bd.attach(exec, runtime=None)
```

# Relation to Problem 1

After Problem 2 & 3 are solved, the foundation of solving Problem 1 is there. 

To recap, the emerging of `DocsExtractUpdateMixin` divides the all `ExecutorDriver` classes into two categories: 
- The ones that fit to "extract -> apply -> update pattern", e.g. `Encode`, `Craft`, `Segment`, `Index`, `Search`
- The ones do not, e.g. `Rank`, `Evaluate`.

The ones that do not fit the `DocsExtractUpdateMixin` pattern are often due to the complicated extraction and extra data preprocessing before calling `exec_fn`. In other way, if `Document` class can provide powerful attributes and `doc.complex_attribute` can be used as a property, then they can also fit with `DocsExtractUpdateMixin`.

That is almost to say: there is **no need to have polymorphism on the `ExecutorDriver`**, which translates to the following YAML specification:

```yaml
...
requests:
  on:
    IndexRequest:
           - GenericExecutorDriver
              with:
                  ...
    SearchRequest:
           - GenericExecutorDriver
              with:
                  ...
```

The only thing that makes multiple `GenericExecutorDriver` different is the method that it binds to. 

## Code Example

For the sake of clarity, the code snippet is presented first, explanation on it comes after. This PR enables the following code example:

```python
import numpy as np

from jina import Flow, Document, Executor
from jina.executors.decorators import requests


class MyExecutor(Executor):

    @requests
    def foo(self, id, embedding):
        print(id)
        print(embedding)


f = Flow().add(uses=MyExecutor)

with f:
    f.index(Document(embedding=np.array([1, 2, 3])))
```

---

One can also put all logic functions in one class and use `@requests(on=)` to route them.

```python
class MyExecutor(Executor):
    @requests
    def foo(self, id):
        return [{'embedding': np.array([1, 2, 3])}] * len(id)

    @requests(on='SearchRequest')
    def bar(self, id):
        return [{'embedding': np.array([4, 5, 6])}] * len(id)

    @requests(on='UpdateRequest')
    def bar2(self, id):
        return [{'embedding': np.array([10, 11, 12])}] * len(id)


# test code
def validate_index_resp(req):
    np.testing.assert_equal(req.docs[0].embedding, np.array([1, 2, 3]))

def validate_search_resp(req):
    np.testing.assert_equal(req.docs[0].embedding, np.array([4, 5, 6]))

def validate_update_resp(req):
    np.testing.assert_equal(req.docs[0].embedding, np.array([10, 11, 12]))

f = Flow().add(uses=MyExecutor)

with f:
    f.index(Document(), on_done=validate_index_resp)
    f.search(Document(), on_done=validate_search_resp)
    f.update(Document(), on_done=validate_update_resp)
```


JoanFM (commented):
We had a chat with @maximilianwerk and he came with a proposal that may solve many of these aspects.

He would like to present it soon and it may make this change not needed. If possible, let's see if we can discuss that before merging this.

hanxiao:
are you talking about a proposal or a working PR?

JoanFM:
> are you talking about a proposal or a working PR? I don't see any ongoing work on this direction. If we are talking about a rough idea without concrete implementation then I'm afraid I can't wait

an idea proposal.

maximilianwerk:
> are you talking about a proposal or a working PR? I don't see any ongoing work on this direction. If we are talking about a rough idea without concrete implementation then I'm afraid I can't wait.
> 
> I will merge it after tmr fireside gathering, to save time i will only answer question you raise as i think the PR description is concrete enough

It is more then just a rough idea. Anyhow, it does not conflict this PR but rather a further advancement. Let me show concrete code to you in tomorrows fireside.

Apart from that: I like the binding of the function signature to the extracted data a lot. Great idea!

jina/drivers/__init__.py
```diff
@@ -397,6 +399,162 @@ def _apply_all(
         """
 
 
+class DocsExtractUpdateMixin:
```
JoanFM, line 402: This assumes a lot of parameters of the drivers, maybe it should be directly a Driver itself? And then they can directly inherit from it.

JoanFM:
> > are you talking about a proposal or a working PR? I don't see any ongoing work on this direction. If we are talking about a rough idea without concrete implementation then I'm afraid I can't wait.
> > I will merge it after tmr fireside gathering, to save time i will only answer question you raise as i think the PR description is concrete enough
> 
> It is more then just a rough idea. Anyhow, it does not conflict this PR but rather a further advancement. Let me show concrete code to you in tomorrows fireside.
> 
> Apart from that: I like the binding of the function signature to the extracted data a lot. Great idea!

well to me the point would be more about to have less changes and refactors

hanxiao:
> have less changes and refactors

Note that this PR only requires simple follow-up changes on hub executors: i.e. changing their `exec_fn` signature to comply with Document types.

jina/drivers/__init__.py
hanxiao: > maybe it should be directly a Driver itself?

i thought about that, but the side-effect on making it default behavior is not clear to me and maybe an overshoot for this PR.

JoanFM:
> > have less changes and refactors
> 
> Note that this PR only requires simple follow-up changes on hub executors: i.e. changing their `exec_fn` signature to comply with Document types. As a hub developer, I would very much happy for this change as it greatly improves the readability of the executor.

I meant changes for people with custom executors, or is it backcompatible?

hanxiao:
> I meant changes for people with custom executors, or is it backcompatible?

It will be a breaking change for Hub & executor developers, but it is a good & **necessary breaking change that leads executor to better readability and maintainability.

The old function signature is not usable anymore. [There is already a detailed error message to guide people to migrate.](https://github.com/jina-ai/jina/pull/2313/files#diff-fd4fcfd9dd90764a7ae665fee46f8c554d362f87b495268750bcdaa92955b1e9R540)

JoanFM:
While is handy, I am not a big fan of coupling the `data` to be extracted with the `class` instead of the `instance`. 

If I  `craft` text into something different, why should I not be able to `craft` other `fields` of a `Document`

PR #2316, opened 2021-04-19
test: add tests for sparsity

JoanFM:


cristianmtr (approved):

Yongxuanzhang:
There are errors in https://github.com/jina-ai/jina/pull/2316/checks?check_run_id=2380856819

Yongxuanzhang (commented):
tests/unit/drivers/test_vector_index_driver.py::test_vector_index_driver_add_bad_docs[tf] FAILED

jina/types/sets/document.py
```diff
@@ -219,7 +219,7 @@ def stack_embeddings(embeddings):
             elif sparse_cls_type == 'tf':
                 import tensorflow as tf
 
-                return tf.sparse.concat(axis=0, sp_inputs=embeddings)
+                return embeddings
```
Yongxuanzhang, line 222: how does this work? Is there still a need to import tf?

JoanFM: oh good point

jina/drivers/search.py
```diff
@@ -139,6 +139,8 @@ def _get_documents_embeddings(self, docs: 'DocumentSet'):
                 scipy_cls_type = embedding_cls_type.split('_')[1]
                 embedding_cls_type = 'scipy'
 
+            print(f'scipy_cls_type {scipy_cls_type} ')
```
Yongxuanzhang, line 142: should we use print or logger?

JoanFM: oh, my bad

jina/drivers/search.py
```diff
@@ -155,12 +155,12 @@ def _fill_matches(self, doc, op_name, topks, scores, topk_embed):
         else:
             for idx, (numpy_match_id, score) in enumerate(zip(topks, scores)):
                 vector = None
-                if topk_embed is not None:
+                if topk_embed[idx] is not None:
```
Yongxuanzhang, line 158: I wonder if this could work if topk_embed is `none`?
Will it get this?
`TypeError: 'NoneType' object is not subscriptable`

JoanFM: it seems not to be the case, as it is not considerrd in the case with dense embeddings

Yongxuanzhang: So this won't happen? topk_embed won't be None but the elements could be None

JoanFM: yes, see the caller

Yongxuanzhang (approved):
LGTM

PR #2321, opened 2021-04-20
Fix jina readme

slettner:


hanxiao:
it is actually named as `*.jaml` on purpose as we basically define a subset on top of YAML. you don't want to override your default YAML intellisense, therefore either `*.jaml` or `*.jina.yml`, but not `*.yml` nor `*.yaml`

slettner:
That makes a lot of sense! Thanks for clarification

Issue #2320, opened 2021-04-20
File seems to be closed when trying to index


JoanFM:
Yes @JoanFM, using the prefetch argument in the flow-config solved the original problem. I tried to pass the batch_size and the request_size like you, but I still get the same error. I have some additional logs, which might indicate what causes the problem. As I said, the first few batches are indexed successfully, but at some point I get this error:

```
        encoder@63[I]:no update since 2021-04-10 07:38:17, will not save. If you really want to save it, call "touch()" before "save()" to force saving
        encoder@63[I]:#sent: 3 #recv: 4 sent_size: 13.8 MB recv_size: 1.2 MB
      chunk_idx@66[I]:recv IndexRequest  from gateway▸crafter/ZEDRuntime▸encoder/ZEDRuntime▸chunk_idx/ZEDRuntime▸⚐
        encoder@63[I]:recv IndexRequest  from gateway▸crafter/ZEDRuntime▸encoder/ZEDRuntime▸⚐
   NumpyIndexer@66[S]:artifacts of this executor (wrap_idx) is persisted to /var/huforce/fs/workspace/wrap_idx-0/wrap_idx.bin
BinaryPbIndexer@66[S]:artifacts of this executor (chunk_idx) is persisted to /var/huforce/fs/workspace/chunk_idx-0/chunk_idx.bin
      chunk_idx@66[I]:no update since 2021-04-10 07:38:17, will not save. If you really want to save it, call "touch()" before "save()" to force saving
      chunk_idx@66[I]:#sent: 3 #recv: 4 sent_size: 1.9 MB recv_size: 15.6 MB
        doc_idx@67[I]:recv IndexRequest  from gateway▸crafter/ZEDRuntime▸encoder/ZEDRuntime▸chunk_idx/ZEDRuntime▸doc_idx/ZEDRuntime▸⚐
DEBUG | Message: Successfully indexed 4 documents!
        doc_idx@67[S]:artifacts of this executor (doc_idx_file) is persisted to /var/huforce/fs/workspace/doc_idx_file-0/doc_idx_file.bin
        doc_idx@67[I]:#sent: 3 #recv: 4 sent_size: 6.3 KB recv_size: 2.2 MB
       | 📃     16 ⏱️ 262.9s 🐎 0.1/s      4   requests        gateway@72[I]:send: 5 recv: 4 pending: 1
        crafter@60[I]:recv IndexRequest  from gateway▸crafter/ZEDRuntime▸⚐
[...]

PR #2322, opened 2021-04-20
fix: proper context management of handlers

JoanFM:
Try to have a better and more robust implementation of `BinaryPbIndexer` and the access to the `Writing` and `ReadHandlers`.

I believe this can be a cause of #2320


nan-wang (changes requested):
Is it possible to add a test to reproduce the issue? It seems this fix will slow down the `adding` by frequent closing and opening the files

jina/executors/indexers/keyvalue.py
```diff
@@ -50,29 +63,39 @@ class _ReadHandler:
...
-            self.header = {
-                r[0]: None
-                if np.array_equal((r[1], r[2], r[3]), HEADER_NONE_ENTRY)
-                else (r[1], r[2], r[3])
-                for r in tmp
-            }
-        self._body = open(path, 'r+b')
+        self.path = path
+        self.header = {}
+        if os.path.exists(self.path + '.head'):
```
cristianmtr, line 68: add an `else` with a raise

cristianmtr (changes requested):
maybe add a benchmark comparison with `master`. I have a feeling this will slow things a bit

jina/executors/indexers/keyvalue.py
JoanFM: no raise, this happens when there is no data in the shard. This was hidden because before file was always created just bu creating the WriteHandler

JoanFM:
I am not sure now about the `delete_on_dump` and the general closing of the executor

cristianmtr:
> I am not sure now about the `delete_on_dump` and the general closing of the executor

Can you perhaps try to apply the fix without the refactoring first? See if we can isolate it

JoanFM:
> > I am not sure now about the `delete_on_dump` and the general closing of the executor
> 
> Can you perhaps try to apply the fix without the refactoring first? See if we can isolate it

I am not having a fix, It is just a feeling that the structure is not robust and may cause it

JoanFM:
Fixes #2320

cristianmtr (changes requested):
can you run a benchmark? When I tried to do it with ctx mger, it slowed it down a bit. I think we should know

JoanFM:
> can you run a benchmark? When I tried to do it with ctx mger, it slowed it down a bit. I think we should know. Use some of the crud tests or the benchmark in `test_dump_dbms`

Run on master:

``
### baseline - query time with 1 on 100000 docs takes 2 seconds (2.67s)
### baseline - indexing: 100000 docs takes 11 seconds (11.65s)
### indexing 100000 docs takes 19 seconds (19.82s)
### dumping 100000 docs takes 5 seconds (5.33s)
### reloading 100000 takes 3 seconds (3.75s)
```

on this branch:

```
### baseline - query time with 1 on 100000 docs takes 2 seconds (2.66s)
### baseline - indexing: 100000 docs takes 12 seconds (12.13s)
### indexing 100000 docs takes 19 seconds (19.61s)
### dumping 100000 docs takes 5 seconds (5.16s)
### reloading 100000 takes 4 seconds (4.11s)
```

File: jina/executors/indexers/keyvalue.py [hunk 2]
```diff
@@ -159,23 +179,27 @@ def delete(self, keys: Iterable[str], *args, **kwargs) -> None:
...
+                            ('', np.int64),
+                        ],
+                    ).tobytes()
+                )
+                self._size -= 1
 
     def _query(self, key):
         pos_info = self.query_handler.header.get(key, None)
-        if pos_info is not None:
+        if (
```
cristianmtr, line 199:
```suggestion
        if (pos_info and getattr(self.query_handler, 'body', None))
```

no?

JoanFM: Well, the real check is to check when we assign `pos_info` if `header` exists. But now I will raise an exception if the file in `_ReadHandler` does not exist, and then this will not be needed anymore

jina/executors/indexers/keyvalue.py
```diff
@@ -23,8 +23,20 @@ class _WriteHandler:
     """
 
     def __init__(self, path, mode):
-        self.body = open(path, mode)
-        self.header = open(path + '.head', mode)
+        self.path = path
+        self.mode = mode
+        self.body = open(self.path, self.mode)
+        self.header = open(self.path + '.head', self.mode)
```
cristianmtr, line 29: Why are we opening the files in `__init__`? Shouldn't we only open them in `__enter__`?

jina/executors/indexers/keyvalue.py
```diff
@@ -23,8 +23,20 @@ class _WriteHandler:
...
+
+    def __enter__(self):
+        if self.body.closed:
+            self.body = open(self.path, self.mode)
+        if self.header.closed:
+            self.header = open(self.path + '.head', self.mode)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.flush()
```
cristianmtr, line 39: Why don't we `close()` on exit?

cristianmtr (approved):

cristianmtr (changes requested):
2 questions

jina/executors/indexers/keyvalue.py
JoanFM: Because we wanted to avoid the need to reopen all the time, but at least we assume the data is flushed

jina/executors/indexers/keyvalue.py
JoanFM: To avoid the concerns from @nan-wang , we avoid opening and closing at everybatch which could damage performance. But we still check the need for opening

jina/executors/indexers/keyvalue.py
nan-wang: Closing the file after every 'add' will be quite expensive.

nan-wang (approved):
LGTM

jina/executors/indexers/keyvalue.py
cristianmtr: I see. So we're kinda using "half" of what a ctx mger should do (open and close within in the context, right)?

jina/executors/indexers/keyvalue.py
cristianmtr: Yes, got it. That's what I tried in the previous approach. And we're kinda doing "half" of the ctx management. Makes sense

cristianmtr (approved):

PR #2329, opened 2021-04-21
refactor: small changes to compound pod

JoanFM:
I think these changes make CompoundPods semantically more similar to Pod. The host and the port now refer to the headPea which is what the Flow may be interested in


bwanglzu (dismissed):
LGTM!

jacobowitz (approved):
lgtm, as far as I can tell ;)

nan-wang (approved):
LGTM👍

PR #2331, opened 2021-04-21
test: replicas with dump-reload

cristianmtr:
Todo

- [x] stress test with jinad and threads
- [x] async reload/query
- [x] use only REST API
- [x] remote flow on `jinad`

added `jinad` func. for Dump/Rolling Update triggering

Part of https://github.com/jina-ai/internal-tasks/issues/40

Benchmark

```
### baseline - query time with 3 on 100000 docs ...	         
### baseline - query time with 3 on 100000 docs takes 2 seconds (2.59s)
### baseline - indexing: 100000 docs takes 7 seconds (7.63s)
### indexing 100000 docs takes 12 seconds (12.72s)
### dumping 100000 docs takes 2 seconds (2.73s)
### dump path size: 112.75556 MBs
### rolling update on 100000 takes 2 seconds (2.03s)
### reloading 100000 takes 0 seconds (0.97s)
```


tests/integration/dump/test_dump_dbms.py
```diff
@@ -149,15 +149,31 @@ def error_callback(resp):
...
+            with TimeContext(f'### dumping {len(docs)} docs'):
+                flow_dbms.dump('indexer_dbms', dump_path, shards=shards, timeout=-1)
 
-        with TimeContext(f'### dumping {len(docs)} docs'):
-            flow_dbms.dump('indexer_dbms', dump_path, shards=shards, timeout=-1)
+            dir_size = path_size(dump_path)
+            print(f'### dump path size: {dir_size} MBs')
+
+            # TODO add param to dump path
+            flow_query.rolling_update('indexer_query')
```
JoanFM, line 166: In order to pass a parameter, we need to refactor how `executors` are being reloaded at __init__ time.

JoanFM (commented):
The good test would be to have a client deattached from `flow` and see that the `queries` are still served while `reloading`

tests/integration/dump/test_dump_dbms.py
```diff
@@ -149,15 +152,33 @@ def error_callback(resp):
...
+                docs[:nr_search],
+                on_done=_validate_results_nonempty,
+                on_error=error_callback,
+            )
 
-        dir_size = path_size(dump_path)
-        print(f'### dump path size: {dir_size} MBs')
+            # TODO dump again new data
+            # repeat rolling update search
+            # TODO maybe async
```
JoanFM, line 180: It has to be done separating the processes where the `client` sends data and the main processes where the `flow` objects exist. Otherwise we can't test that we do query_while_indexing in a non-blocking way

JoanFM (commented):
I think the way `replica_args` are created need to be refactored

jina/peapods/pods/__init__.py
```diff
@@ -133,8 +133,8 @@ def _parse_base_pod_args(
             self._set_after_to_pass(args)
             self.is_head_router = True
             self.is_tail_router = True
-            parsed_args['head'] = BasePod._copy_to_head_args(args, polling_type.is_push)
-            parsed_args['tail'] = BasePod._copy_to_tail_args(args, polling_type.is_push)
+            parsed_args['head'] = BasePod._copy_to_head_args(args, polling_type)
+            parsed_args['tail'] = BasePod._copy_to_tail_args(args, polling_type)
             parsed_args[attribute] = BasePod._set_peas_args(
```
JoanFM, line 138: when `attribute` is `replica` is kind of weird that the `args["replica"]` that should be `Pod` arguments are actuall `Pea` arguments.

cristianmtr:
> I think the way `replica_args` are created need to be refactored

Not going to happen in this PR. The focus is to just get it to work at this point. But I agree there are a lot of things that need improving here.

tests/integration/dump/test_dump_dbms.py
cristianmtr: Yes, good point. I will write it in that approach

jina/peapods/pods/__init__.py
cristianmtr: Yes, this args processing is kinda hard to track right now. Will refactor after we see that everything at least works in the MVP

tests/integration/dump/test_dump_dbms.py
```diff
@@ -183,6 +198,76 @@ def error_callback(resp):
...
+                docs[:nr_search],
+                on_done=_validate_results_empty,
+                on_error=error_callback,
+            )
+            t.join()
+
+            # done with both -- we should have matches now
+            flow_query.search(
+                docs[:nr_search],
+                on_done=_validate_results_nonempty,
```
JoanFM, line 266: use mockers to ensure callbacks are indeed called

tests/integration/dump/test_dump_dbms.py
```diff
@@ -149,27 +154,126 @@ def error_callback(resp):
...
+            #     flow_query.rolling_update('indexer_query', dump_path)
+            #
+            # flow_query.search(
+            #     docs[:nr_search],
+            #     on_done=_validate_results_nonempty,
+            #     on_error=error_callback,
+            # )
+
+            # test with query while reloading async.
+            t = Thread(
```
JoanFM, line 180: I am not sure if the `multithreading` here is real. Since `flow_query.search` is blocking, the GIL may not be free and parallelism may not be real

JoanFM: to be looked at

tests/integration/dump/test_dump_dbms.py
cristianmtr: done

tests/integration/dump/test_dump_dbms.py
cristianmtr: added print statement and an assert on their order

jina/clients/__init__.py
```diff
@@ -202,6 +202,53 @@ def reload(
             **kwargs,
         )
 
+    def dump(
+        self,
+        targets: Union[str, List[str]],
+        on_done: CallbackFnType = None,
+        on_error: CallbackFnType = None,
+        on_always: CallbackFnType = None,
+        dump_path: str = None,
```
JoanFM, line 211: according to signature, dump path and shards should be Optional..

However I think it would be more handy to have them as the first arguments and maybe have no default value

jina/clients/__init__.py
```diff
@@ -202,6 +202,53 @@ def reload(
...
+        :param targets: the regex string or list of regex strings to match the pea/pod names.
+        :param on_done: the function to be called when the :class:`Request` object is resolved.
+        :param on_error: the function to be called when the :class:`Request` object is rejected.
+        :param on_always: the function to be called when the :class:`Request` object is  is either resolved or rejected.
+        :param kwargs: additional parameters
+        :return: None
+        """
+
+        if shards is None:
+            raise ValueError('`shards` cannot be None')
```
JoanFM, line 228: under the above comments, this should not be necessary

jina/drivers/control.py
```diff
@@ -86,9 +86,8 @@ def __call__(self, *args, **kwargs):
         elif self.req.command == 'CANCEL':
             pass
         elif self.req.command == 'DUMP':
-            self.logger.warning(
-                'DUMP command cannot be processed by ControlReqDriver. Use DumpDriver'
-            )
+            print(f'### DUMP in Driver')
+            self._dump()
```
JoanFM, line 90: The ControlDriver will all have dump?

jina/drivers/control.py
```diff
@@ -104,6 +103,22 @@ def __call__(self, *args, **kwargs):
...
+            patterns = self.req.targets
+            if isinstance(patterns, str):
+                patterns = [patterns]
+            for p in patterns:
+                if re.match(p, self.runtime.name):
+                    self.logger.info(
+                        f'Dumping from Executor `{self.runtime._executor.name}` in `{self.runtime.name}`'
+                    )
+                    req_dict = dict(self.req.args)
+                    self.runtime._executor.dump(
```
JoanFM, line 117: code smell accessing the private attribute of runtime.

If dump needs access to Executor, maybe we need to rethink if it should be a ControlRequest

JoanFM (changes requested):

jina/drivers/control.py
JoanFM: or have ControlReqDriver implement BaseExecutableDriver interface

cristianmtr: Initially it was a different Request, but it does make sense that it's just another ControlRequest, as it doesn't behave in an inherently different way. It just has a bunch of args (which ControlReq always had but were never used). It would make sense that the Driver is attached to an executor directly. Not sure what the vision is for this change Han/Max mentioned @maximilianwerk ?

JoanFM: what is clear is that this Driver needs to implement BaseExecutableDriver. this changes challenges the design of ControlDriver right now.

jina/drivers/control.py
```diff
@@ -86,9 +86,8 @@ def __call__(self, *args, **kwargs):
         elif self.req.command == 'CANCEL':
             pass
         elif self.req.command == 'DUMP':
-            self.logger.warning(
-                'DUMP command cannot be processed by ControlReqDriver. Use DumpDriver'
-            )
+            print(f'### DUMP in Driver')
```
maximilianwerk, line 89: Just a reminder: needs to be removed before merging

jina/drivers/control.py
maximilianwerk: To be honest, I don't yet have a clear view of the `ControlDriver` after the API refactoring. It might just not be much affected. It might also be joined with the `GenericDriver`. In any case, I am OK with this hack until we understand the design of the new Executor handling in detail.

jina/types/request/control.py
```diff
@@ -6,4 +8,14 @@
...
+    def args(self):
+        """struct args
+
+
+        .. #noqa: DAR201"""
+        return self.proto._args
+
+    @args.setter
+    def args(self, value: Dict):
+        self.proto.args = value
```
maximilianwerk, line 21: I thought this should be `self.proto.args.update(value)`!? If we call the update outside, I'd rather implement a 
```
def update_args(self, value: Dict):
    self.proto.args.update(value)
```
to not manipulate state outside of this object.

jina/flow/base.py
```diff
@@ -974,27 +974,17 @@ def identity(self, value: str):
...
-    def dump(self, pod_name, path: str, shards: int, timeout=-1):
-        """Emit a Dump request to a specific Pod
-
-        :param shards: the nr of shards in the dump
-        :param path: the path to which to dump
-        :param pod_name: the name of the pod
-        :param timeout: time to wait (seconds)
-        """
-        pod = self._pod_nodes[pod_name]
-        pod.dump(path, shards, timeout)
```
maximilianwerk, line 986: While I agree, that in general this should be done via a separate client, do we really want to cut the ability to do it via the Flow object?

maximilianwerk (commented):
Haven't deeply followed all tests. Apart from that: I like it.

jina/drivers/control.py
JoanFM: so lets make a huge TODO and clarify that is a hack to be checked later

daemon/models/__init__.py
```diff
@@ -1,5 +1,17 @@
...
+
+
+class UpdateOperationEnum(Enum):
+    """Represents the type of operation to perform in the update
+
+    We consider these an `update` operation since they **change** the underlying state
+    """
+
+    rolling_update = 1
+    dump = 1
```
JoanFM, line 17: 2

jina/peapods/pods/__init__.py
```diff
@@ -569,3 +570,26 @@ def _parse_base_pod_args(self, args):
...
+    def dump(self, pod_name, dump_path, shards, timeout):
+        """Emit a Dump request to its Peas
+
+        :param pod_name: the pod to target
+        :param shards: the nr of shards in the dump
+        :param dump_path: the path to which to dump
+        :param timeout: time to wait (seconds)
+        """
+        for pea in self.peas:
+            if 'head' not in pea.name and 'tail' not in pea.name:
```
JoanFM, line 583: I think there may be a better way to access peas without head and tail, or be based on some other logic than the name

JoanFM (changes requested):

jina/drivers/control.py
cristianmtr: added a TODO there

daemon/api/endpoints/flow.py
```diff
@@ -78,3 +79,18 @@ async def _status(
 @router.on_event('shutdown')
 def _shutdown():
     store.reset()
+
+
+@router.put(
+    path='/{id}',
+    summary='Run an update operation on the Flow object',
+    description='Types supported: "rolling_update" and "dump"',
```
JoanFM, line 87: I am not sure if we should use a better naming, it is confusing to have `update` for a thing that can be `rolling_update` or `dump`.

JoanFM (commented):
The `PodStore` should also offer this `endpoint` in `jinad`.

And it should be tested how a `local` flow could send DUMP or ROLLING_UPDATE command to remote instances.

I guess not part of here but important to keep in mind.

@cristianmtr @FionnD @florian-hoenicke @maximilianwerk

daemon/api/endpoints/flow.py
cristianmtr: > to have `update` for a thing

What do you mean here? We are using the `PUT` method on the `/flows` endpoint. Is that what you mean by `update`? Not sure what the best alternative would be. I wouldn't add another endpoint, like `/flows/rolling_update` etc., as semantically it makes more sense to only have `/entity/ITS_ID`

jina/executors/indexers/query/__init__.py
```diff
@@ -57,3 +57,13 @@ def delete(self, keys: Iterable[str], *args, **kwargs) -> None:
         .. # noqa: DAR101
         """
         self._log_warn()
+
+    def is_exist(self) -> bool:
+        """
```
JoanFM, line 62: Since all the boilerplate and the extra code in the Executor will be removed, explicitly overwrite how the write handler and query handler are computed. Or do some alternative naming.

Just overdiding is_exist to False is too hacky. It invalidates the semantics

cristianmtr: This is not used. I simply get the write handler

jina/peapods/pods/__init__.py
cristianmtr: Any suggestions?

jina/types/request/control.py
cristianmtr: Fixed

jina/drivers/control.py
cristianmtr: Yes. We are now just sending a control req to the Pod/Peas, and it will be received by the ControlDriver

jina/executors/indexers/dbms/keyvalue.py
```diff
@@ -16,6 +17,7 @@ def _get_generator(
         for id_ in ids:
             vecs_metas_list_bytes = super()._query([id_])
             vec, meta = pickle.loads(vecs_metas_list_bytes[0])
+            assert Document(meta).tags['tag_field'] is not None
```
JoanFM, line 20: why is this?

jina/peapods/pods/__init__.py
JoanFM: have a private property for that?

tests/conftest.py
```diff
@@ -42,9 +42,10 @@ def docker_compose(request):
     )
     time.sleep(5)
     yield
-    os.system(
-        f"docker-compose -f {request.param} --project-directory . down --remove-orphans"
-    )
+    # os.system(
+    #     f"docker-compose -f {request.param} --project-directory . down --remove-orphans"
+    # )
+    os.system(f"docker-compose -f {request.param} --project-directory . down")
```
JoanFM, line 48: keep remove orphans

JoanFM (changes requested):

jina/executors/indexers/dbms/keyvalue.py
cristianmtr: was used during debugging an issue. removed

tests/conftest.py
cristianmtr: yes, thanks. was used during debugging

cristianmtr:
> https://github.com/jina-ai/jina/pull/2331#discussion_r625137941

@JoanFM Added a property

jina/executors/indexers/dump.py
```diff
@@ -29,17 +34,25 @@ def _handle_dump(
...
+        extra = size % shards
+        shard_range = list(range(shards))
+        for shard_id in shard_range:
+            if shard_id == shard_range[-1]:
+                size_this_shard = size_per_shard + extra
+            else:
+                size_this_shard = size_per_shard
+            _write_shard_data(data, path, shard_id, size_this_shard)
+    else:
+        print(os.listdir(path))
```
JoanFM, line 52: remove print

JoanFM (changes requested):

jina/executors/indexers/dump.py
cristianmtr: done

tests/unit/peapods/pods/test_compound_pods.py
```diff
@@ -118,21 +118,33 @@ def test_pod_naming_with_parallel(runtime):
 
         assert bp.replicas[0].name == 'pod/0'
         assert bp.replicas[0].peas[0].name == 'pod/0/head'
+        assert bp.replicas[0].peas[0].inner is False
```
JoanFM, line 121: this should go in another test or the test needs to be renamed

JoanFM (changes requested):

daemon/stores/base.py
```diff
@@ -24,7 +26,17 @@ def _init_stats(self):
...
     def add(self, *args, **kwargs) -> 'uuid.UUID':
-        """Add a new element to the store. This method needs to be overridden by the subclass"""
+        """Add a new element to the store. This method needs to be overridden by the subclass
+
+
+        .. #noqa: DAR101"""
+        raise NotImplementedError
+
+    def update(self, *args, **kwargs) -> 'uuid.UUID':
+        """Add a new element to the store. This method needs to be overridden by the subclass
```
JoanFM, line 36: fix docstring

jina/drivers/control.py
```diff
@@ -112,6 +100,38 @@ def __call__(self, *args, **kwargs):
         else:
             raise UnknownControlCommand(f'don\'t know how to handle {self.req.command}')
 
+    def _reload(self):
+        # TODO should this be removed, since we now have proper rolling update?
```
JoanFM, line 104: Let's wait, I think some users rely on this for now

jina/drivers/search.py
```diff
@@ -1,6 +1,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
+import sys
```
JoanFM, line 4: not needed

jina/peapods/pods/__init__.py
```diff
@@ -181,6 +182,10 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
         super().__exit__(exc_type, exc_val, exc_tb)
         self.join()
 
+    @staticmethod
+    def _set_after_to_pass(args):
```
JoanFM, line 186: remove this, not needed

jina/peapods/pods/compoundpod.py
```diff
@@ -232,9 +225,10 @@ def _set_replica_args(
             result.append(_args)
         return result
 
-    def rolling_update(self):
-        """
-        Update all pods of this compound pod.
+    def rolling_update(self, dump_path):
```
JoanFM, line 228: type hinting, this should be an `optional` parameter defaulting to `None`. This way we can also test without `dump_path` (Better split of infra/functional code)

JoanFM (changes requested):

jina/peapods/pods/compoundpod.py
cristianmtr: I actually don't see why it should be done here, as it kinda implies you could do rolling update without a path. It should be passed a `None` in tests where we don't care that it's correct. But by default, it should not allow users to start a query flow without a dump paht

JoanFM: if in tests will be passed to None is because it is accepting.

The concern that this should not work for Query Flows is handled somewhere else.

Here the CompoundPod is just asked to restart its replicas.

daemon/stores/base.py
cristianmtr: What about it?

JoanFM: it does not add a new element to the store right?

jina/peapods/pods/__init__.py
cristianmtr: done

jina/drivers/search.py
cristianmtr: yes, cleaned

jina/drivers/control.py
cristianmtr: sure

jina/peapods/pods/compoundpod.py
cristianmtr: It works to just pass `None` in the rolling update test, so no problem. It works there because we don't care about whatever dummy indexer receives it and doesn't use it.

It should not be "invisible", though, as users might think it is truly optional, when it isn't.

daemon/stores/base.py
cristianmtr: ah yes, corrected

jina/peapods/pods/compoundpod.py
JoanFM: I would even generallize and have a set of params to update for the Pod. So that this is less linked to the actual functionality.

But no big deal, we can change later. But since the parametrr accepts None, the thpe hint should be there.

The point is to keep a distinction about the functionality it offers and the infrastructure used for it.

tests/unit/peapods/pods/test_compound_pods.py
cristianmtr: I don't think that's really useful. Because you need to know which ones to assert are `inner` or not, so you need to keep track of the names anyway. I'd keep them together

JoanFM (approved):

jina/peapods/pods/compoundpod.py
cristianmtr: Added type annotations, but I think we should keep it as is otherwise. The distinction is only there in the test. You wouldn't use it without a proper dump to reload.

tests/integration/dump/test_dump_dbms.py
```diff
@@ -111,65 +116,247 @@ def path_size(dump_path):
...
+
+@pytest.mark.repeat(5)
+@pytest.mark.parametrize('nr_docs', [700])
+@pytest.mark.parametrize('emb_size', [10])
+def test_threading_query_while_reloading(tmpdir, nr_docs, emb_size, mocker):
+    global operations
+
+    # TODO better way to test async procedure call order
+    # patch
+    def _rolling_update(self, dump_path):
```
JoanFM, line 259: is this needed?

cristianmtr: You mean the test in general, or the patched method? 

If the former, maybe not anymore, since we've added the stress test.

If the latter, I patch it in order to log and then make sure the operations are run in a specific order

JoanFM: the patch, ssems weird, but okey.

cristianmtr: Maybe it is better to remove it, as it is quite fragile. I wrote it before we added the more comprehensive jinad threading stress test, to show that is possible to query while reloading.

PR #2338, opened 2021-04-22
feat: hosts for replicas

florian-hoenicke:
closes https://github.com/jina-ai/jina/issues/2339

### DIFF
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -121,8 +121,6 @@ class BasePod(ExitStack):
         self,
         args,
         attribute,
-        id_attribute_name,
-        role_type,
         repetition_attribute,
         polling_type,
     ):
@@ -133,14 +131,10 @@ class BasePod(ExitStack):
             self._set_after_to_pass(args)
             self.is_head_router = True
             self.is_tail_router = True
-            parsed_args['head'] = BasePod._copy_to_head_args(args, polling_type.is_push)
-            parsed_args['tail'] = BasePod._copy_to_tail_args(args, polling_type.is_push)
-            parsed_args[attribute] = BasePod._set_peas_args(
+            parsed_args['head'] = BasePod._copy_to_head_args(args, polling_type)
+            parsed_args['tail'] = BasePod._copy_to_tail_args(args, polling_type)
+            parsed_args[attribute] = self._set_middle_args(
                 args,
-                role_type=role_type,
-                repetition_attribute=repetition_attribute,
-                id_attribute_name=id_attribute_name,
-                polling_type=polling_type,
                 head_args=parsed_args['head'],
                 tail_args=parsed_args['tail'],
             )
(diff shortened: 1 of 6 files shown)


jina/peapods/pods/__init__.py
```diff
@@ -150,20 +144,12 @@ def _parse_base_pod_args(
...
-                )
+                parsed_args['head'] = self._copy_to_head_args(args, polling_type)
             if getattr(args, 'uses_after', None):
                 self.is_tail_router = True
-                parsed_args['tail'] = self._copy_to_tail_args(
-                    args, polling_type.is_push
-                )
-            parsed_args[attribute] = self._set_peas_args(
+                parsed_args['tail'] = self._copy_to_tail_args(args, polling_type)
+            parsed_args[attribute] = self._set_middle_args(
```
JoanFM, line 151: do not change the name yet, I would like to see what is needed here to share between Pea and Pod

jina/peapods/pods/__init__.py
```diff
@@ -613,3 +531,64 @@ def dump(self, path, shards, timeout):
                     DumpMessage(path=path, shards=shards),
                     timeout=timeout,
                 )
+
+    @staticmethod
```
JoanFM, line 535: If u separate the code for Pod and CompoundPod, this should be kept as set pea args

JoanFM (changes requested):
I think to have hosts for different Pods in the CompoundPod we need another layer of arguments.

We could start by adding replica hosts, to allow every replica to be on one host, and then maybe consider different hosts for every Pea of every replica

florian-hoenicke:
>We could start by adding replica hosts, to allow every replica to be on one host, and then maybe consider different hosts for every Pea of every replica

With this implementation, every pea can have its own host.

JoanFM:
> > We could start by adding replica hosts, to allow every replica to be on one host, and then maybe consider different hosts for every Pea of every replica
> 
> With this implementation, every pea can have its own host.

but what about different Peas of different Replicas?

florian-hoenicke:
> but what about different Peas of different Replicas?
yes, all the peas generated by parallel * replicas can run on different hosts.

JoanFM:
> > but what about different Peas of different Replicas?
> > yes, all the peas generated by parallel * replicas can run on different hosts.

My biggest concern is, that the new pattern of havinf CompoundPod just a composition of Pods is not so well defined.

jina/peapods/pods/__init__.py
florian-hoenicke: I thought `_set_middle_args` would be a good method name for everything which is between head and tail.
This method is then be used in the compound pod and in the pod.
In the compound pod the middle args are for the replicas and in the pod the middle args are as we used to for the peas.
I can also change it back if you want.

florian-hoenicke:
Interesting - a timeout again. I will investigate what it is caused by.

jina/peapods/pods/__init__.py
JoanFM: since it is not in the common interface and is private I think it is better to have self explanatory

florian-hoenicke: It is actually the same function_name used for pods and compound pods.

florian-hoenicke: I think we will refactor this part anyways in the close future.

JoanFM: but not declared in interface, different names make the usage description explicit

florian-hoenicke: okay I will do it

florian-hoenicke: I added the base method as abstract method.

JoanFM: Private methods should not be part of the interface. I think specialized names are better. Set peas args for Pod and set pod args for CompoundPod would be the more explainable ones.

After all, you are seeing that any BasePod needs to implement this method, which no external component should care about

florian-hoenicke: okay, changed thanks

florian-hoenicke:
@JoanFM is there anything left to do for me?

tests/unit/peapods/pods/test_compound_pods.py
```diff
@@ -162,3 +162,54 @@ def test_pod_naming_with_parallel(runtime):
...
+            '--parallel',
+            '2',
+            '--replicas',
+            '3',
+            '--peas-hosts',
+            *[f'0.0.0.{i+1}' for i in range(num_hosts)],
+            '--runtime-backend',
+            'process',
+        ]
+    )
```
cristianmtr, line 210: Can you try to convert one of the functional integration tests (CRUD etc.) to use this approach, to see that the ops work properly in this new setup?

JoanFM (approved):

cristianmtr (changes requested):
one request

PR #2341, opened 2021-04-23
ci: test with python 3 dot 9

cristianmtr:


cristianmtr:
Seems TF2 does not support Python 3.9, but they are working on it. Let's circle back to this when it's ready https://github.com/tensorflow/tensorflow/issues/44485

PR #2346, opened 2021-04-24
refactor: abstract the construction of different pods

JoanFM:
Allow CompoundPod instantiation from CLI


JoanFM:
Need to add unit tests

niuzs-alan (approved):
LGTM

I ran the `intergration/sharding` test which throws `Timeout > 360s` locally.
It worked out and took 239s.

Maybe I should increase the timeout.

---
I ran it again locally, the 2nd one took 596s.
there were both no other tasks consuming large computer resource when these 2 test ran locally.

florian-hoenicke (approved):

PR #2347, opened 2021-04-25
test: compound pod sockets

florian-hoenicke:
testing the socket configuration of the compound pod
closes https://github.com/jina-ai/jina/issues/2348


tests/unit/peapods/pods/test_compound_pods.py
```diff
@@ -162,3 +163,65 @@ def test_pod_naming_with_parallel(runtime):
...
+    assert head.socket_in == SocketType.PULL_BIND
+    assert head.socket_out == SocketType.ROUTER_BIND
+    assert head.scheduling == SchedulerType.LOAD_BALANCE
+    tail = replica_args['tail']
+    assert tail.socket_in == SocketType.PULL_BIND
+    assert tail.socket_out == SocketType.PUSH_BIND
+    assert tail.scheduling == SchedulerType.LOAD_BALANCE
+    replicas = replica_args['replicas']
+    for pod_args in replicas:
+        pod = Pod(pod_args)
```
JoanFM, line 220: I see why this is needed, but this is a code smell. We should not need to add logic in a test. We should be able to instantiate a CompoundPod and test what's inside their Pods and Peas.

I think what the test is showing is the need to change the way we initialize our layers.

We should have at `__init__` time:

```
CompoundPod instantiates inner Pods
Pod instantiates inner Peas
Peas instantiate inner Runtime (Already so)
```

At `__enter__` time:
```
CompoundPod enters the context of Pod
Pod enters the context of Pea
Pea starts runtime
```

JoanFM (changes requested):

tests/unit/peapods/pods/test_compound_pods.py
florian-hoenicke: I see your point. There should be a differentiation between the configuration and starting of compound pods, pods and peas.
The refactoring is not trivial, though.
It does involve refactoring the `noblock_on_start` attribute used by compound pod, pod, and pea.
I would suggest doing this in a different pr

JoanFM: The noblock on start can be set at the Init also, it only affects the initialization right? or is a new attribute to keep.

if u want to do in another PR I would suggest to do before this test. 

Because this test has too much logic in it

JoanFM:
Also it would be cool to `test` port and host configurations

tests/unit/peapods/pods/test_compound_pods.py
JoanFM: PR with proposal in #2360

JoanFM (commented):
Would love to see these adapted with also different `hosts`

JoanFM:
@florian-hoenicke , can u make sure to realign with master and pass the tests?

tests/unit/peapods/pods/test_compound_pods.py
```diff
@@ -259,22 +259,21 @@ def test_sockets(polling, parallel, pea_scheduling, pea_socket_in, pea_socket_ou
...
+        tail = compound_pod.tail_args
         assert tail.socket_in == SocketType.PULL_BIND
         assert tail.socket_out == SocketType.PUSH_BIND
         assert tail.scheduling == SchedulerType.LOAD_BALANCE
-        replicas = compound_pod.replica_list
+        replicas = compound_pod.replicas
         for replica in replicas:
             if parallel > 1:
-                assert replica.args.polling == polling_type
+                # assert replica.args.polling == polling_type
```
JoanFM, line 273: remove or correct

florian-hoenicke: thanks, deleted

JoanFM (approved):

PR #2349, opened 2021-04-26
fix: replace None return type of gRPC request handlers

### LINKED ISSUE #2063: run_async returns future - not None
crud functions in `jina/clients/__init__.py` have the None annotation. However, they return a future.

please make sure the return type is correct

Kelton8Z:
close #2063

### DIFF
--- a/jina/clients/__init__.py
+++ b/jina/clients/__init__.py
@@ -2,7 +2,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-from typing import Union, List, Awaitable
+from typing import Union, List
 
 from . import request
 from .base import BaseClient, CallbackFnType, InputType, InputDeleteType
--- a/jina/clients/__init__.py
+++ b/jina/clients/__init__.py
@@ -2,7 +2,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-from typing import Union, List
+from typing import Union, List, Awaitable
 
 from . import request
 from .base import BaseClient, CallbackFnType, InputType, InputDeleteType
@@ -69,7 +69,7 @@ class Client(BaseClient):
         on_error: CallbackFnType = None,
         on_always: CallbackFnType = None,
         **kwargs
-    ) -> None:
+    ) -> asyncio.Future:
         """Issue 'search' request to the Flow.
 
         :param inputs: input data which can be an Iterable, a function which returns an Iterable, or a single Document
--- a/jina/clients/__init__.py
+++ b/jina/clients/__init__.py
@@ -41,7 +41,7 @@ class Client(BaseClient):
         on_error: CallbackFnType = None,
         on_always: CallbackFnType = None,
         **kwargs
-    ) -> None:
+    ) -> Awaitable:
         """Issue 'train' request to the Flow.
 
         :param inputs: input data which can be an Iterable, a function which returns an Iterable, or a single Document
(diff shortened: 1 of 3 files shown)


jina/clients/__init__.py
```diff
@@ -41,7 +41,7 @@ def train(
         on_error: CallbackFnType = None,
         on_always: CallbackFnType = None,
         **kwargs
-    ) -> None:
+    ) -> Awaitable:
```
florian-hoenicke, line 44: you can define the type of the `Awaitable`. like `Awaitable[...]`

jina/clients/__init__.py
```diff
@@ -69,7 +69,7 @@ def search(
...
```
florian-hoenicke, line 72: you can check if other return types are possible as well.

florian-hoenicke (changes requested):

florian-hoenicke:
@Kelton8Z There are other places as well where the type annotation is needed `clients/asyncio`, `jina/flow/mixin/crud.py ...

jina/clients/__init__.py
```diff
@@ -98,7 +98,7 @@ def index(
...
```
florian-hoenicke, line 101: @Kelton8Z Is `Awaitalbe` the correct type? That is the sync client - not the async variation.

florian-hoenicke (changes requested):

jina/clients/__init__.py
Kelton8Z: is index a sync client in the sense that the flow needs to finish indexing before moving on to query or other process?

in that case i also need to change the code within the index function ?

JoanFM (changes requested):
Lets separate the 2 PR. they are different stories

JoanFM (changes requested):
@Kelton8Z watch out because u have changes in the hub submodule

jina/clients/__init__.py
```diff
@@ -126,15 +126,15 @@ def update(
         on_error: CallbackFnType = None,
         on_always: CallbackFnType = None,
         **kwargs
-    ) -> Awaitable:
+    ) -> Awaitable[list]:
```
JoanFM, line 129: capital Letter. Do we know whay is the list of? is a list of Documents?

Kelton8Z: oh yeah, a list of responses

nan-wang (approved):
LGTM👍

Kelton8Z:
> @Kelton8Z watch out because u have changes in the hub submodule

"The behaviour of git commit -a seems to be to include changes to submodules, if they have new commits inside them." is probably how I introduced them accidentally.

JoanFM:
> > @Kelton8Z watch out because u have changes in the hub submodule
> 
> "The behaviour of git commit -a seems to be to include changes to submodules, if they have new commits inside them." is probably how I introduced them accidentally.

do not use -a, just select what u want to add

jina/clients/asyncio.py
```diff
@@ -93,7 +94,7 @@ async def search(
         on_error: CallbackFnType = None,
         on_always: CallbackFnType = None,
         **kwargs
-    ) -> None:
+    ) -> AsyncGenerator:
```
JoanFM, line 97: and here?

jina/clients/__init__.py
```diff
@@ -126,7 +126,7 @@ def update(
         on_error: CallbackFnType = None,
         on_always: CallbackFnType = None,
         **kwargs
-    ) -> None:
+    ) -> Awaitable:
```
JoanFM, line 129: Awaiteble of something?

nan-wang:
duplicated with #2386

PR #2359, opened 2021-04-28
fix: return all default values in rest

### LINKED ISSUE #2103: REST API - search - missing attribute
When calling the REST API for search where we don't get matches, the matches attribute is not included in the response. 
`
with flow: \#restful == True
    res = requests.post('http://0.0.0.0:{flow.port_expose}/api/search', documents)
`
UAC
- all attributes of a document are included in the response - even if they are empty lists or dictionaries

JoanFM:
Fixes #2103


jacobowitz (dismissed):
lgtm

deepankarm (approved):

PR #2362, opened 2021-04-29
fix: fixed double close

cristianmtr:
Investigate and fix hanging in some tests

### DIFF
--- a/jina/peapods/runtimes/zmq/zed.py
+++ b/jina/peapods/runtimes/zmq/zed.py
@@ -53,6 +53,7 @@ class ZEDRuntime(ZMQRuntime):
 
     def teardown(self):
         """Close the `ZmqStreamlet` and `Executor`."""
+        print(f'### ZEDRuntime {self.name} calling _zmqlet.close()')
         self._zmqlet.close()
         self._executor.close()
         super().teardown()
--- a/jina/peapods/runtimes/zmq/zed.py
+++ b/jina/peapods/runtimes/zmq/zed.py
@@ -53,7 +53,6 @@ class ZEDRuntime(ZMQRuntime):
 
     def teardown(self):
         """Close the `ZmqStreamlet` and `Executor`."""
-        print(f'### ZEDRuntime {self.name} calling _zmqlet.close()')
         self._zmqlet.close()
         self._executor.close()
         super().teardown()
--- a/jina/peapods/zmq/__init__.py
+++ b/jina/peapods/zmq/__init__.py
@@ -208,16 +208,17 @@ class Zmqlet:
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
-    def close(self, *args, **kwargs):
+    def close(self, force_close=False, *args, **kwargs):
         """Close all sockets and shutdown the ZMQ context associated to this `Zmqlet`.
 
         .. note::
             This method is idempotent.
 
+        :param force_close: forces a close, even if `is_closed` is already set to True
         :param args: Extra positional arguments
         :param kwargs: Extra key-value arguments
         """
-        if not self.is_closed:
+        if not self.is_closed or force_close:
             self.is_closed = True
             self._close_sockets()
             if hasattr(self, 'ctx'):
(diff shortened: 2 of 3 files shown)


jina/peapods/zmq/__init__.py
```diff
@@ -208,16 +208,17 @@ def __enter__(self):
...
 
         .. note::
             This method is idempotent.
 
+        :param force_close: forces a close, even if `is_closed` is already set to True
         :param args: Extra positional arguments
         :param kwargs: Extra key-value arguments
         """
-        if not self.is_closed:
+        if not self.is_closed or force_close:
```
JoanFM, line 221: I guess is just for investiagtion. But if it ends up being the solution, it would be a smell. Then it would mean that the `semantics` of `is_closed` would be false

jina/peapods/zmq/__init__.py
```diff
@@ -495,7 +507,11 @@ def send_message(
     num_bytes = 0
     try:
         _prep_send_socket(sock, timeout)
-        sock.send_multipart(msg.dump())
+        if 'flags' in kwargs:
+            kwargs['flags'] |= zmq.NOBLOCK
+        else:
+            kwargs['flags'] = zmq.NOBLOCK
+        sock.send_multipart(msg.dump(), **kwargs)
```
maximilianwerk, line 514: What else could be in `kwargs`? It was not used before, so why use it now?

jina/peapods/zmq/__init__.py
maximilianwerk: No it is actually needed. The child class checks `self.is_closed` and sets it, if it was not closed beforehand. But the parent class should still be shut down.

JoanFM: then the naming or semantics need to change.

jina/peapods/zmq/__init__.py
cristianmtr: It contained the default flags. We needed to add this specific one

jina/peapods/zmq/__init__.py
cristianmtr: Not sure it's actually needed, after we realized the change was in the flags. Will get back to this when I have some time

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -53,6 +53,7 @@ def setup(self):
 
     def teardown(self):
         """Close the `ZmqStreamlet` and `Executor`."""
+        print(f'### ZEDRuntime {self.name} calling _zmqlet.close()')
```
JoanFM, line 56: remove print

jina/peapods/zmq/__init__.py
JoanFM: Let's put this as draft

JoanFM (changes requested):

jina/peapods/zmq/__init__.py
```diff
@@ -385,29 +382,24 @@ def close(self, flush: bool = True, *args, **kwargs):
         :param kwargs: Extra key-value arguments
         """
         if not self.is_closed:
-            print(f'### is_closed was False {self.name}')
             self.is_closed = True
```
maximilianwerk, line 385: If we remove the `force_close`, we need to remove this one as well.

jina/peapods/runtimes/zmq/zed.py
cristianmtr: done

jina/peapods/zmq/__init__.py
cristianmtr: Why?

jina/peapods/zmq/__init__.py
```diff
@@ -495,7 +499,11 @@ def send_message(
     num_bytes = 0
     try:
         _prep_send_socket(sock, timeout)
-        sock.send_multipart(msg.dump())
+        if 'flags' in kwargs:
```
JoanFM, line 502: if we ever receive flags in kwargs, lets put it as Optional parametrr with default to None

JoanFM (changes requested):

jina/peapods/zmq/__init__.py
cristianmtr: Not really easy to do, since the kwargs are passed as part of other kwargs, from up the call stack

maximilianwerk:
I close this, since we decided to not go the NOBLOCK way.

PR #2366, opened 2021-04-29
feat: ranker trainer

bwanglzu:
- [x] executor interface
- [x] driver
- [x] integration test
- [x] docstrings and minor updates


jina/executors/rankers/trainer.py
```diff
@@ -0,0 +1,50 @@
+__copyright__ = "Copyright (c) 2021 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+import abc
+
+from . import BaseRanker
```
JoanFM, line 6: For now, no need to add `ranker`  or anything as parameter.

For now, let's have a `RankerTrainer` as a class that trains something than then the `RankerExecutor` can understand.

But no direct link between the 2 classes

JoanFM (changes requested):

jina/executors/rankers/trainer.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class RankerTrainer(abc.ABC):
```
JoanFM, line 9: the `RankerTrainer` needs to be an `executer`. Or this is just a `mixin`?

tests/integration/ranker_training/offline/yaml/trainer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self, params: dict = None, weights_shape: float = (2, 2)):
+        super().__init__()
+        self._params = params
+        self._weights_shape = weights_shape
+        self._weights = np.random.uniform(
+            -self.WEIGHT_LIMIT, self.WEIGHT_LIMIT, size=weights_shape
+        )
+
+    def train(self, *args, **kwargs):
```
JoanFM, line 24: it would be nice to see how the `trainer` actually receives data about matches and trains accordingly

JoanFM (changes requested):

tests/integration/ranker_training/offline/test_offline_training.py
```diff
@@ -0,0 +1,87 @@
...
+    # Before Ranker Trainer, the feature is completely rely on `price` tag, `size` can be seen as a bias.
+
+    from sklearn.linear_model import LinearRegression
+
+    model = LinearRegression()
+
+    X = [[1, 1], [2, 1], [4, 1], [8, 1], [16, 1]]
+    y = [1, 2, 3, 4, 5]
+    model.fit(X, y)
+    path = '/Users/bo/Documents/work/jina'
```
JoanFM, line 69: make sure it works with `tmpdir`

tests/integration/ranker_training/offline/test_offline_training.py
```diff
@@ -0,0 +1,87 @@
...
+        doc.matches.add(match)
+    return doc
+
+
+def test_train_offline(documents_to_train, doc_to_query):
+    def validate_ranking_by_price(req):
+        pred = set()
+        for match in req.docs[0].matches:
+            pred.add(match.score.value)
+        assert len(pred) == 1  # since price tag never changes, all scores are the same.
```
JoanFM, line 50: do not understand this? What does it prove?

JoanFM (changes requested):

tests/integration/ranker_training/offline/test_offline_training.py
bwanglzu: @JoanFM we're training a model based on 2 features, `size` and `price`, in this case, the model was trained on `price` feature, and our `doc_to_query` has a fixed price value. So the predicted results are always the same

JoanFM: ok, make a long documentation about this

JoanFM (dismissed):

bwanglzu:
@hanxiao can you take a look at this one? I touched `extra-requirements.txt`

hanxiao (approved):

Issue #2367, opened 2021-04-29
Websocket client blocks on empty index requests


jacobowitz:
The _WebSocketClient_  can get stuck on index requests with generators as input. 
This happens when the generator is already exhausted and never yields any documents. In this case the  _WebSocketClient_ does (correctly) not send any actual request to Jina, but it still waits for a response to be send. This response is never received and the client blocks waiting on this non existing response.

jacobowitz:
I noticed this issue while working on features related to https://github.com/jina-ai/jina/issues/2082

PR #2368, opened 2021-04-29
fix(websocket-client): fix blocking on empty requests

jacobowitz:
This fixes https://github.com/jina-ai/jina/issues/2367

I am basically just checking if we enter the for loop which is generating the actual requests to Jina. If this is not the case I will gracefully disconnect (close the websocket) again. This stops the waiting for a server response and the client unblocks.

There are no tests for our clients unfortunately, but I checked the new code manually with a local stress test and it behaves as expected: If a caller passes in an exhausted input generator, the client will just disconnect without any error and the caller can continue its business. Before this change we blocked the caller indefinetely.

### DIFF
--- a/jina/clients/websocket.py
+++ b/jina/clients/websocket.py
@@ -76,7 +76,7 @@ class WebSocketClientMixin(BaseClient, ABC):
                         await websocket.send(bytes(True))
                     else:
                         # There is nothing to send, disconnect gracefully
-                        await websocket.close(reason="No data to send")
+                        await websocket.close(reason='No data to send')
 
                 with ProgressBar(task_name=tname) as p_bar, TimeContext(tname):
                     # Unlike gRPC, any arbitrary function (generator) cannot be passed via websockets.
--- a/jina/clients/websocket.py
+++ b/jina/clients/websocket.py
@@ -64,13 +64,19 @@ class WebSocketClientMixin(BaseClient, ABC):
                 self.num_responses = 0
 
                 async def _send_requests(request_iterator):
+                    next_request = None
                     for next_request in request_iterator:
                         await websocket.send(next_request.SerializeToString())
                         self.num_requests += 1
-                    # Server has no way of knowing when to stop the await on sending response back to the client
-                    # We send one last message to say `request_iterator` is completed.
-                    # On the client side, this :meth:`send` doesn't need to be awaited with a :meth:`recv`
-                    await websocket.send(bytes(True))
+                    # Check if there was any request generated
+                    if next_request is not None:
+                        # Server has no way of knowing when to stop the await on sending response back to the client
+                        # We send one last message to say `request_iterator` is completed.
+                        # On the client side, this :meth:`send` doesn't need to be awaited with a :meth:`recv`
+                        await websocket.send(bytes(True))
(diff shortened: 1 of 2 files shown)


jina/clients/websocket.py
```diff
@@ -64,13 +64,19 @@ async def _get_results(
...
-                    await websocket.send(bytes(True))
+                    # Check if there was any request generated
+                    if next_request is not None:
+                        # Server has no way of knowing when to stop the await on sending response back to the client
+                        # We send one last message to say `request_iterator` is completed.
+                        # On the client side, this :meth:`send` doesn't need to be awaited with a :meth:`recv`
+                        await websocket.send(bytes(True))
+                    else:
+                        # There is nothing to send, disconnect gracefully
+                        await websocket.close(reason="No data to send")
```
JoanFM, line 79: let's use single quote as per style

JoanFM (changes requested):
Good catch!! Could we add a unittest for this? or an integration one?

jacobowitz:
A test here would be a fair amount of work due to the communication and async functionality here. I guess some kind of integration test would be the least amount of work, because I could use a real server and dont have to mock everything away myself. At least my case and some regular good case should be easy to reproduce in an integration test

JoanFM:
> A test here would be a fair amount of work due to the communication and async functionality here. I guess some kind of integration test would be the least amount of work, because I could use a real server and dont have to mock everything away myself. At least my case and some regular good case should be easy to reproduce in an integration test

let's go for an integration test, but something needs to test this

jacobowitz:
I actually found some unit tests doing some tests around clients. I reused that scheme and added a test there which tests that a simple good case works and the case I am fixing here. With the old code my test fails. With the new code it should pass.

One thing I am unsure about are the callbacks (on_always): I am intentionally not calling the on_always callback here, because there was no request and no response in my case. But its not 100% clear that the client should behave like this I guess. The callbacks rely on the actual server response though, which dont exist in this case so the callbacks dont make much sense here.

tests/unit/clients/python/test_client.py
```diff
@@ -26,6 +26,11 @@ def flow_with_rest_api_enabled():
     return Flow(rest_api=True).add()
 
 
+@pytest.fixture(scope='function')
+def websocket_client():
```
JoanFM, line 30: where is this `ficture` used?

tests/unit/clients/python/test_client.py
```diff
@@ -111,3 +116,29 @@ def test_client_csv(restful, mocker, func_name):
...
+                ]
+            )
+        )
+        # Test that a regular index request triggers the callback
+        client.index(iter([Document()]), request_size=1, on_always=mock)
+        mock.assert_called_once()
+
+        mock = mocker.Mock()
+        # Test that an empty index request does not trigger the callback and does not time out
+        client.index(iter([()]), request_size=1, on_always=mock)
```
JoanFM, line 143: I would also add a mock in the `on_error` to ensure is not called either?

JoanFM (changes requested):

JoanFM (approved):
LGTM!

Issue #2363, opened 2021-04-29
fix naming of pods and replicas in logs


cristianmtr:
See tests.unit.peapods.pods.test_compound_pods.test_pod_naming_with_parallel

Pod should be `replica`

JoanFM:
The current naming is:

`CompoundPod` as a `Pod` receives the `name` from input `args`.

The inner Pods inherit the `name` plus add an `idx` depending on their replica-id. What would be the naming you would expect?

In the `test`, u see Pod because the --name argument says so.

cristianmtr:
The idea is to make the logs look like 

```
        replica-0/pea-1@109904[I]:starting jina.peapods.runtimes.zmq.zed.ZEDRuntime...
        replica-0/pea-1@109904[I]:input tcp://0.0.0.0:39103 (DEALER_CONNECT) output tcp://0.0.0.0:57387 
```

Maybe it's not a good first issue. I haven't looked at where specifically the changes would need to be done. I was only thinking that a PR for this would be fairly detached from the core, and wouldn't require a lot of overhead for understanding the internals

FionnD:
Ah yes I see your logic, that's fair enough. An experienced developer might take it.

properGrammar:
Hi All, I've been digging into this, and believe I've nailed down how to make this happen. Before I prepare the PR, I have a few questions:

1 - The easiest solution here is not simply a matter of tweaking logging settings, but involves altering the object names. Is this an acceptable tradeoff to resolve this issue? (I'm assuming it is, as the current solution is also implemented this way, but just want to make sure nothing else depends on the current naming convention.)
2 - Currently, in the case where the object name does not yet exist, it is just assigned the index number for a name. Would you like "replica-" and "pea-" to be prepended in these cases or no?
3 - Do you want "pea-" to be prepended when the Pea name is "head" or "tail"? (I'm assuming not, but just checking.)

I should be able to have something whipped up fairly quickly after hearing back from you on these questions. Cheers!

properGrammar:
As you can see, I went about with the PR; I hope I followed the commit style correctly.

I assumed that the answer to the above questions was: 1. Yes 2. No 3. No. I look forward to any followup herein. Be Well

PR #2382, opened 2021-05-06
fix(test_compound_pods): fixed naming of pods and replicas in logs

properGrammar:
Intended to fix jina-ai/jina#2363

Hi All, I've been digging into this, and believe I've nailed down how to make this happen. Before I prepare the PR, I have a few questions:

1 - The easiest solution here is not simply a matter of tweaking logging settings, but involves altering the object names. Is this an acceptable tradeoff to resolve this issue? (I'm assuming it is, as the current solution is also implemented this way, but just want to make sure nothing else depends on the current naming convention.)
2 - Currently, in the case where the object name does not yet exist, it is just assigned the index number for a name. Would you like "replica-" and "pea-" to be prepended in these cases or no?
3 - Do you want "pea-" to be prepended when the Pea name is "head" or "tail"? (I'm assuming not, but just checking.)

I should be able to have something whipped up fairly quickly after hearing back from you on these questions. Cheers!

As you can see, I went forward with the PR; I hope I followed the commit style correctly.

I assumed that the answers to the above questions were: 1. Yes 2. No 3. No. I look forward to any followup herein. Be Well


cristianmtr:
> Intended to fix #2363

Hey! Thanks for the contribution.

1. I see that a test is failing. Can you investigate?
2. Make sure to rebase to `master` (or merge) in order to get the docstring and black check to pass. (I'm going to assume you did set up the pre-commit hooks as described in [CONTRIBUTING](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)
3. Can you perhaps try to add a test that captures the logging output and assert the names of the components? You can use https://docs.pytest.org/en/stable/logging.html#caplog-fixture . Not sure how easy / if at all possible this would be, but I think it would be worth adding. You should add this to an integration test that actually inserts some data

jina/peapods/pods/__init__.py
```diff
@@ -593,4 +593,4 @@ def _parse_base_pod_args(self, args):
             parsed_args['peas'] = [args]
 
         # note that peas_args['peas'][0] exist either way and carries the original property
-        return parsed_args
+        return parsed_args
```
maximilianwerk, line 596: Please don't remove end of file newlines.

JoanFM (changes requested):
I do not think this is the solution to the issue. What we would like to see is the replica and pea prefix added before their respective index.

We do not have any control of the name of the Pod itself. 

So we should aim for 

`pod_name/replica-{idx}/pea-{idx}` to the name of the Pea.

Therefore the name of the Pod inside the CompounsPod would be.

`pod_name/replica-{idx}`

jina/peapods/pods/__init__.py
properGrammar: Sorry, that was unintentional. I have fixed this in the most recent commit.

properGrammar:
Note to all: There were some notes regarding my thinking on jina-ai/jina#2363 that I have copied to the top comment on this page. They may provide some context for this discussion.

@cristianmtr 
1. The tests that failed were the two runs of `test_pod_naming_with_parallel` in `test_pods.py`. The assert statements weren't expecting the updated naming convention for Peas (see my notes at top.) I have updated them for the most recent commit.
2. I rebased to master and was told that the branch is up to date. Should I merge? I did follow the directions in CONTRIBUTING, but I may have overlooked a step. Where exactly are the pre-commit hooks I should be setting up?
3. I will be happy to look into this once it is determined that the fixes I have proposed are the way forward (see more details below.)

@JoanFM 
Here is what Cristian asked for in #2363:
> The idea is to make the logs look like
> replica-0/pea-1@109904[I]:starting jina.peapods.runtimes.zmq.zed.ZEDRuntime...
> replica-0/pea-1@109904[I]:input tcp://0.0.0.0:39103 (DEALER_CONNECT) outpu...

Here is what I'm seeing locally after my fixes:
> replica-0/pea-1@49675[I]:recv ControlRequest  from ctl▸replica-0/pea-1/ZEDRuntime▸⚐
> replica-0/pea-1@49675[I]:#sent: 1 #recv: 1 sent_size: 255 Bytes recv_size: 138 Bytes

As you said in #2363, the Pod name is set via the `--name` argument in `test_compound_pods::test_pod_naming_with_parallel`. The way I was able to get the above results (again, as you noted we would have to do in 2363) was by naming the Pod "replica".

Getting the change you have suggested is easy enough by prepending a hardcoded "replica" into the solution I've already created and reverting the Pod name back to "pod" (or something else if y'all would prefer.) Please let me know how you'd like me to proceed on this point.

properGrammar:
Unless I hear otherwise soon, I'm going to assume @JoanFM 's preferences here take precedence over @cristianmtr 's and institute the changes he has asked for.

JoanFM:
recheckcla

properGrammar:
> recheckcla

@JoanFM Could you please give me a bit more detail on how to do this? I Googled it a bit, but wasn't able to come up with much. (This is one of my first experiences in submitting a PR to an open source project, as is probably obvious lol. Sorry, too, for any missteps I may have made thus far.)

I'm also not able to submit new commits in the wake of resolving the conflicts, so I don't know how to push a commit that passes `lint` and `black` at this point... Please let me know if you need me to do anything more here, thanks.

JoanFM:
Hey @properGrammar, I think the quickest solution would be to open a new PR in a new branch copying the current changes and do a new commit with the short message.

Also make sure to apply the black formatting.

Thank you very much

properGrammar:
Done; see #2418 

It's funny, I had this exact idea myself. Was happy to do it. Fingers crossed I did everything correctly!

properGrammar:
#2508 resolves this

PR #2386, opened 2021-05-07
fix: change async client function's return type

Kelton8Z:


Kelton8Z:
close #2063 without hub submodule conflict

nan-wang (approved):
LGTM👍

PR #2389, opened 2021-05-10
feat: inverted boolean expression indexer

JoanFM:


hanxiao:
why are we still adding new feature to 1.x while 2.0 is lacking hands?

hanxiao:
in 2.0 the whole querylang has been removed. even from the protobuf level

JoanFM:
> in 2.0 the whole querylang has been removed. even from the protobuf level

I saw it, I am thinking how we should proceed. And I think we should make our QueryLang core to the Doc/DocArray type.

PR #2399, opened 2021-05-11
fix: ensure same precision in dump

cristianmtr:
Found issue while working on https://github.com/jina-ai/examples/pull/573

### DIFF
--- a/jina/executors/indexers/dump.py
+++ b/jina/executors/indexers/dump.py
@@ -7,6 +7,7 @@ import numpy as np
 from jina.logging import JinaLogger
 
 BYTE_PADDING = 4
+DUMP_DTYPE = np.float64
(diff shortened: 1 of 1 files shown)


jacobowitz (approved):
lgtm :)
If we think this `dtype` is ever gonna change we could also store it alongside the dump and use the value found there, but I think as far as we know this is a fixed value?

cristianmtr:
> lgtm :)
> If we think this `dtype` is ever gonna change we could also store it alongside the dump and use the value found there, but I think as far as we know this is a fixed value?

Yes, that would be a good idea. Just thought about it and it would require a breaking change to pass the dtype to the `dump_streaming` func. Let's keep those to after `2.0`

PR #2401, opened 2021-05-11
refactor: clarify controlmessage args

JoanFM:
**Changes introduced**
The way the `ControlMessage` is used seems weird.

The `pod_name` is set to `ctl` to determine that the message is sent from `ctl` and the `target_pod_name` should be another one.


hanxiao:
`ctl` is just a placeholder for naming, could be `GodPod`, `User`, `Keyboard`, it has no other purpose then set routes.

PR #2403, opened 2021-05-12
feat: recover evaluation for fashion helloworld

JoanFM:


jina/helloworld/fashion/executors.py
```diff
@@ -97,3 +113,58 @@ def _norm(A):
...
+    def _recall(self, actual, desired):
+        if self.eval_at == 0:
+            return 0.0
+        actual_at_k = actual[: self.eval_at] if self.eval_at else actual
+        ret = len(set(actual_at_k).intersection(set(desired)))
+        return ret / len(desired)
+
+    @requests(on='/index')
+    def skip(self, **kwargs):
+        pass
```
hanxiao, line 151:
```suggestion
    @requests
    def _skip(self, **kwargs):
        pass
```

`@requests` with no `on` will catch all https://github.com/jina-ai/jina/blob/devel-2.0/.github/2.0/COOKBOOK-Executor.md#default-binding-requests-without-on

PR #2405, opened 2021-05-12
refactor: remove unneeded decorators

JoanFM:


jina/executors/decorators.py
```diff
@@ -136,292 +74,13 @@ def arg_wrapper(self, *args, **kwargs):
...
-    if isinstance(final_result[0], np.ndarray):
-        if len(final_result[0].shape) > 1:
-            final_result = np.concatenate(final_result, merge_over_axis)
-    elif isinstance(final_result[0], list) and flatten:
-        final_result = list(chain.from_iterable(final_result))
-
-    return final_result
-
-
-def batching(
```
Yongxuanzhang, line 184: seems batching is still imported in some tests, should they be removed/modified?

JoanFM: in which tests? yes they should

PR #2416, opened 2021-05-12
feat: hello multimodal 2.0

bwanglzu:
multimodal helloworld for jina 2.0

- [x] executors
- [x] flow

to be fixed:

- [x] bathing in `ImageEncoder` (decorator has been removed)
- [x] weighted ranker `score` function
- [x] test index flow
- [x] test search flow
- [x] test the example
- [x] fix system test


jina/helloworld/multimodal/flow.yml
```diff
@@ -0,0 +1,48 @@
+jtype: Flow
+version: '1.0'
```
JoanFM, line 2: remember to add the bifurcations with needs

bwanglzu:
with conversion:
```python
    # DocVectorIndexer
    @requests(on='/index')
    def index(self, docs: 'DocumentArray', **kwargs):
        for chunk in docs.traverse_flat(['c']):
            if chunk.mime_type == 'image/jpeg':
                chunk.convert_blob_to_buffer()
                chunk.convert_buffer_to_uri()
                chunk.convert_uri_to_datauri()
            self._docs.append(chunk)
```
the image is not showing up yet
image

JoanFM (approved):

hanxiao:
but it's not yet working right?

hanxiao:
This level of refactoring is not what I expected, it's completely restricted by the 1.x mindset.  Why we need to maintain 1:1 flow as 1.x? **We have to move out from there**. This is the one of the first three examples users will see. If we teach them an inefficient pattern, they will learn this bad pattern, or even worse, they will drop the learning.

What I expect: we know what we want to present, so let's drop all 1.x code & flow and start from scratch with 2.0 mindset.

PR #2424, opened 2021-05-14
feat: test nested matches of doc in a flow

Kelton8Z:


tests/integration/high_order_matches/test_document.py
```diff
@@ -0,0 +1,63 @@
...
+    def index(self, docs, **kwargs):
+        self.doc = docs[0]
+        for match in self.doc.matches:
+            for i in range(5):
+                match.matches.append(Document())
+
+        return self.doc
+
+
+def test_high_order_matches(mocker):
```
JoanFM, line 34: Good job, lets keep only the second test since it includes also the first

JoanFM (changes requested):

tests/integration/high_order_matches/test_document.py
Kelton8Z: Alright

JoanFM (approved):

PR #2431, opened 2021-05-17
feat: added dump type

maximilianwerk:
This allows dump/reloading compatibility between indexers. Reasons for putting it into core:

- It does not belong to one Indexers but is a shared component between all of them
- It must be in sync for all the Indexers


hanxiao:
this can not go to the core at this moment, this looks very executor-specific (in this case, indexer specific). Any executor-specific logic should be either go to the hub or maintained privately

PR #2441, opened 2021-05-19
minor improvement and typo fix to the cookbook of Executor

nan-wang:
minor improvement and typo fix to the cookbook of Executor

### DIFF
--- a/.github/2.0/cookbooks/Executor.md
+++ b/.github/2.0/cookbooks/Executor.md
@@ -65,6 +65,31 @@ with f:
 
 ### With YAML
 
+```text
+.
+├── __init__.py
+├── app.py
+├── my.yml
+└── foo.py
+
+```
+
+`foo.py`:
+
+```python
+from jina import Executor
+
+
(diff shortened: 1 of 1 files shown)


.github/2.0/cookbooks/Executor.md
```diff
@@ -234,24 +256,25 @@ If you don't need some arguments, you can suppress it into `**kwargs`. For examp
...
-    bar(kwargs['docs_matrix'])
+def foo_using_no_arg(**kwargs):
+    # the args are suppressed into kwargs
+    print(kwargs['docs_matrix'])
 ```
 
 ### Method Returns
 
-Method decorated with `@request` can return `Optional[DocumentSet]`. If not `None`, then the current `Request.docs` will
+Method decorated with `@request` can return `Optional[DocumentArray]`. If not `None`, then the current `Request.docs` will
```
JoanFM, line 277: I would here maybe emphasize that the changes done in the `docs` affect the `Request.docs`. (That `Document` is just a memory view)

PR #2445, opened 2021-05-19
fix: support generator

JoanFM:


hanxiao:
reject, this will affect iterator as well, not only generator

JoanFM:
> is this `next()` generator-specific, otherwise this will affect iterator as well, not only generator

maybe there is a safer place

hanxiao (approved):

PR #2451, opened 2021-05-20
fix: fixed hanging pod termination

maximilianwerk:
This PR fixed the hanging pods bug. It also reintroduces the integration test for dump reload and demonstrate its usage.

### DIFF
--- a/jina/peapods/zmq/__init__.py
+++ b/jina/peapods/zmq/__init__.py
@@ -495,11 +495,7 @@ def send_message(
     num_bytes = 0
     try:
         _prep_send_socket(sock, timeout)
-        if 'flags' in kwargs:
-            kwargs['flags'] |= zmq.NOBLOCK
-        else:
-            kwargs['flags'] = zmq.NOBLOCK
-        sock.send_multipart(msg.dump(), **kwargs)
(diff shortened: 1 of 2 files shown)


jina/peapods/zmq/__init__.py
```diff
@@ -495,7 +495,11 @@ def send_message(
     num_bytes = 0
     try:
         _prep_send_socket(sock, timeout)
-        sock.send_multipart(msg.dump())
+        if 'flags' in kwargs:
```
JoanFM, line 498: we still have no reason for this right?

maximilianwerk: @florian-hoenicke said there were some issues, when not setting it. I can remove it and we see.

nan-wang:
reference: https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.send

nan-wang (approved):
LGTM👍

hanxiao:
Im reverting this PR as im not convinced by adding NOBLOCK flag, seems without solid reason. Playing network parameter in this way is dangerous and should be avoided.

PR #2454, opened 2021-05-20
fix: exclusivity doccontent

jacobowitz:
**Problem:**
There is an inconsistency in the document _cookbook_ and potentially in the implementation of the `Document` constructor.

The cookbook explains that the document content fields are mutually exclusive (buffer/text/blob/uri) and that they are all `content`.
So a `Document` can only have one content, not a `uri` AND a `text` etc.
In the current example of the _cookbook_ this is actually the case: It creates a `Document` which has a content and a different uri. That does not happen with the actual implementation, it will always take the last content parameter.

This behavior of the code is also debatable. It is not really incorrect, but it is a bit confusing that the implementation just takes the last value in the `kwargs` list. I think it might be better to raise a `ValueError` in this case, because it is unclear what to do in case these **mutually exclusive** fields are provided by the caller.

**This PR changes the following:**

1. Change the cookbook by remove the additional content parameter from the Document constructor and change the debug output accordingly
2. Raise a ValueError if a Document is constructed with conflicting doc content fields
3. A test for the Error cases

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -83,6 +83,8 @@ DocumentSourceType = TypeVar(
 
 _all_mime_types = set(mimetypes.types_map.values())
 
+_all_doc_content_keys = ('content', 'uri', 'blob', 'text', 'buffer')
+
 
 class Document(ProtoTypeMixin):
     """
@@ -262,6 +264,11 @@ class Document(ProtoTypeMixin):
         if self._pb_body.id is None or not self._pb_body.id:
             self.id = random_identity(use_uuid1=True)
 
+        # check if there are mutually exclusive content fields
+        if _contains_conflicting_content(**kwargs):
+            raise ValueError(
+                f'Document content fields are mutually exclusive, please provide only one of {_all_doc_content_keys}'
+            )
         self.set_attributes(**kwargs)
         self._mermaid_id = random_identity()  #: for mermaid visualize id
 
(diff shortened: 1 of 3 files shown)


tests/unit/types/document/test_document.py
```diff
@@ -791,3 +792,25 @@ def test_evaluations():
...
+
+@contextmanager
+def does_not_raise():
+    yield
+
+
+@pytest.mark.parametrize(
+    'doccontent, expectation',
+    [
+        ({'content': 'hello', 'uri': 'https://jina.ai'}, pytest.raises(Exception)),
```
JoanFM, line 805: you could be more explicit with `ValueError`?

JoanFM (commented):
Like it a lot!

jina/types/document/__init__.py
```diff
@@ -262,6 +264,11 @@ def __init__(
         if self._pb_body.id is None or not self._pb_body.id:
             self.id = random_identity(use_uuid1=True)
 
+        # check if there are mutually exclusive content fields
+        if _contains_conflicting_content(**kwargs):
+            raise ValueError(
+                f'Document content fields are mutually exclusive, please provide only one of {_all_doc_content_keys}'
```
bwanglzu, line 270: i think `content` and other keys are note mutually exclusive? e.g. we can have `content` and `uri` together

bwanglzu (commented):
Great, left some comment

jina/types/document/__init__.py
JoanFM: uri is content now

jacobowitz: hmm not really? From my understand text/uri/blob/buffer are specific contents.
So if you set some string like 'hello world' as the content of a document it will have both text and context set to 'hello world'. Thus it does not really make sense to provide both content and text at the same time because it is in fact the same.

bwanglzu: you're right, ignore my comment.

PR #2459, opened 2021-05-21
revert: revert default value fields

hanxiao:


JoanFM (commented):
Why do we need to remove this? I think it is an important feature that the user can expect everything to be in the response

hanxiao:
default values and miss

PR #2462, opened 2021-05-22
feat: add score value shortcut

hanxiao:
enable the shortcut such as `doc.score = 1.2`


JoanFM:
Just food for thought, would it make sense to have Match and Chunk as classes the same way MatchArray and ChunkArray have specific functionality? this way u can always add a score only in the scope of a match?

This may be a way to enforce correctness to the user?

PR #2465, opened 2021-05-22
refactor: meth update for document primitive type

bwanglzu:
1. make `update` easier to use & understand, get rid of `included_fields` and `exclude_fields` and all the complex logic with `fields`, if user specify `fields`, then only updated specified fields. Otherwise update all fields.
2. Resolve a bug when calling `update`, e.g. `doc1.update(doc2)`, in current implementation will call `CopyFrom`, `CopyFrom` will clear everything in `doc1` and assign `doc2` to `doc1`. This is incorrect when `doc1` has more attributes than `doc2`, the extra attributes in `doc1` will gone. We only want to replace the existent fields.


hanxiao:
test broken

tests/unit/types/document/test_document.py
```diff
@@ -828,3 +732,105 @@ def test_doc_match_score_assign():
...
+
+
+def test_doc_update_given_nested_fields_and_attributes_identical(test_docs):
+    # doc1 and doc2 has the same fields, id, content, tags, embedding and chunks.
+    doc1, doc2 = test_docs
+    # After update, only specified nested fields are updated.
+    doc1.update(source=doc2, fields=['tags', 'embedding', 'chunks'])
+    assert doc1.id != doc2.id
+    assert doc1.content != doc2.content  # None was updated by source's content.
+    assert doc1.tags == doc2.tags
```
nan-wang, line 799: As for the `tags`, it is better to stay consistent with the [python `update` method](https://docs.python.org/3/library/stdtypes.html?highlight=update#dict.update). To be specific, if the key exists, replace the old value with the new one. Otherwise, append the key-value. 

```suggestion
    target = {'a': 'b'}
    target.update({'c': 'd'})
    assert doc1.tags == target
```

nan-wang (changes requested):

nan-wang:
> test broken

The test is fixed by #2471

tests/unit/types/document/test_document.py
bwanglzu: @nan-wang code updated

hanxiao:
very good observation and good PR

jina/types/document/__init__.py
```diff
@@ -354,120 +355,73 @@ def tags(self, value: Dict):
...
-                replace_message_field=True,
-            )
-
-            # recover exclude fields
-            destination.proto.MergeFrom(_dest)
+            if fields_to_preserve:
+                for field in fields_to_preserve:
+                    setattr(destination, field, getattr(dest_copy, field))
+        # For the tags, stay consistent with the python update method.
+        self._pb_body.tags.update(dest_copy.tags)
```
hanxiao, line 403: so all other fields are overridden during the update, but `tags` are expanded? I'm not really convinced by this inconsistency.

tests/unit/types/document/test_document.py
hanxiao: > As for the tags, it is better to stay consistent with the python update method.
im not really convinced by this logic, see my comment below

jina/types/document/__init__.py
bwanglzu: true, I'll think about a better solution

jina/types/document/__init__.py
```diff
@@ -354,120 +354,60 @@ def tags(self, value: Dict):
...
+        fields: Optional[List[str]] = None,
     ) -> None:
-        """Updates fields specified in ``include_fields`` from the source to current Document.
+        """Updates fields specified in ``fields`` from the source to current Document.
 
         :param source: source :class:`Document` object.
-        :param exclude_fields: a tuple of field names that excluded from the current document,
-                when not given the non-empty fields of the current document is considered as ``exclude_fields``
-        :param include_fields: a tuple of field names that included from the source document
+        :param fields: a list of field names that included from the current document,
```
JoanFM, line 399: this wording is weird. These fields are from `self` or from `source`?

JoanFM (commented):
I like it better actually. What do you think @nan-wang?

jina/types/document/__init__.py
bwanglzu: @nan-wang how do you think? shall we replace the tags or expand it?

nan-wang (approved):
LGTM👍

jina/types/document/__init__.py
nan-wang: > so all other fields are overridden during the update
@hanxiao During `update`, the existing fields are overridden and the non-existing fields are added. The same logic goes for `tags` because `tags` is a dictionary by itself. The existing key/value pairs in the tags are overridden and only the non-existing ones are added.

PR #2467, opened 2021-05-24
feat: enable context variable to overwrite the envvar

nan-wang:
When overwriting the variables in the YAML, the `context` in the local function should have a higher priority than the environment variables because setting `context` has a smaller scope than the envvar.


nan-wang:
reopen at #2469

PR #2469, opened 2021-05-24
feat(jaml): enable the context to overwrite env var

nan-wang:
enable the context to overwrite env var


hanxiao (changes requested):
this does not enable anything, this only moves the existing code and changes the priority of variables replacement logic. My question is, 
- why is the priority in this PR makes more sense than the old one?
- where is this priority documented?

Without being clear about these questions, the value of this PR is limited.

PR #2480, opened 2021-05-25
feat: add example for executor using tensorflow

jacobowitz:
This adds another example Executor to the cookbook. It is using Keras/Tensorflow and built in MobileNet.

In a real world case one would pass certain parameters as args to the Executor (like image dimensions). 

Also I am assuming here that the content is stored in the _buffer_ field, but one could also argue that should be _blob_. I am using _buffer_ because `DocumentArray.from_files()` will set _buffer_ and not _blob_.

The example was tested with this flow:

```
if __name__ == '__main__':
    from jina import Flow, DocumentArray
    f = Flow().add(uses=TfMobileNetEncoder)

    with f:
        f.post(on='/index', inputs=DocumentArray(DocumentArray.from_files('/some/path/*.png', read_mode='rb')))
```

### DIFF
--- a/.github/2.0/cookbooks/Executor.md
+++ b/.github/2.0/cookbooks/Executor.md
@@ -514,5 +514,44 @@ class PaddleMwuExecutor(Executor):
             doc.embedding = np.array(_output)  # assign the encoding results to ``embedding``
 ```
 
+### Tensorflow
+
+```python
+import numpy as np
+import tensorflow as tf
+from keras.applications.mobilenet_v2 import MobileNetV2, preprocess_input
+from tensorflow.python.framework.errors_impl import InvalidArgumentError
+
+from jina import Executor, requests
+
+
+class TfMobileNetEncoder(Executor):
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.image_dim = 224
+        self.model = MobileNetV2(pooling='avg', input_shape=(self.image_dim, self.image_dim, 3))
+
+    @requests
+    def encode(self, docs, **kwargs):
+        buffers, docs = docs.get_attributes_with_docs('buffer')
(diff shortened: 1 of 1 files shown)


.github/2.0/cookbooks/Executor.md
```diff
@@ -514,5 +514,44 @@ class PaddleMwuExecutor(Executor):
...
+        self.model = MobileNetV2(pooling='avg', input_shape=(self.image_dim, self.image_dim, 3))
+
+    @requests
+    def encode(self, docs, **kwargs):
+        buffers, docs = docs.get_attributes_with_docs('buffer')
+
+        tensors = [tf.io.decode_image(contents=b, channels=3) for b in buffers]
+        resized_tensors = preprocess_input(np.array(self._resize_images(tensors)))
+
+        embeds = self.model.predict(np.stack(resized_tensors))
```
JoanFM, line 541: this prediction will not be the embedding but the label right? Not sure

jacobowitz: hmm I am not sure what we really want as the embedding here, but I think it should be good enough as it is here?
`predict(batch)` will basically return a numpy array describing the scores for all 1000 classes in this model.

To actually get proper labels one would then continue and use some like this https://www.tensorflow.org/api_docs/python/tf/keras/applications/mobilenet/decode_predictions

JoanFM: No, it is good. I am not sure if we should point to the user that these features do not tend to be good for search. But anyway, same happens to FastAI, What do you think @nan-wang ?

hanxiao: it's good

PR #2484, opened 2021-05-26
feat: Add description for tf executor example

jacobowitz:
This adds a small description to the Tensorflow Executor in the cookbook. One could add potentially how that can or should not be used in a search context, but I think its a bit too much detail at this point.


davidbp (dismissed):
LGTM!

JoanFM (approved):

PR #2485, opened 2021-05-26
feat: add test for store serialization

jacobowitz:
This adds a simple test for serialization of the `BaseStore`. The test just stores and loads some dummy store (inheriting from `BaseStore`) and checks if the stores before and after are equal.

The PR also makes two code changes:

1. For the empty dict one should specify the factory I believe and not use the mutable `{}`
2. In `set_size()` I am checking for the existence of 'items'. Without this check my test fails as the function gets called once without items in the dict. Not sure why this happens.

### DIFF
--- a/daemon/models/base.py
+++ b/daemon/models/base.py
@@ -15,9 +15,12 @@ class StoreStatus(BaseModel):
     time_updated: datetime = Field(default_factory=datetime.now)
     num_add: int = 0
     num_del: int = 0
-    items: Dict[DaemonID, StoreItem] = {}
+    items: Dict[DaemonID, StoreItem] = Field(default_factory=dict)
 
     @root_validator(pre=False)
     def set_size(cls, values):
-        values['size'] = len(values['items'])
+        if 'items' in values:
+            values['size'] = len(values['items'])
+        elif 'size' not in values:
+            values['size'] = 0
         return values
--- a/daemon/models/base.py
+++ b/daemon/models/base.py
@@ -19,8 +19,5 @@ class StoreStatus(BaseModel):
 
     @root_validator(pre=False)
     def set_size(cls, values):
-        if 'items' in values:
-            values['size'] = len(values['items'])
-        elif 'size' not in values:
-            values['size'] = 0
+        values['size'] = len(values['items']) if 'items' in values else 0
         return values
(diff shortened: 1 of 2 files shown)


tests/daemon/unit/stores/test_basestore.py
```diff
@@ -1,14 +1,18 @@
...
 from daemon.stores.base import BaseStore
 from jina.helper import random_identity, random_uuid
+from daemon import __root_workspace__
 
-keys = [uuid.UUID(random_identity()) for _ in range(3)]
+keys = [DaemonID(f'jflow-{uuid.UUID(random_identity())}') for _ in range(3)]
 
 
-store_items = {keys[0]: {'object': 'abc'}, keys[1]: {'object': 'hij'}, keys[2]: {}}
+store_items = {keys[0]: StoreItem(), keys[1]: StoreItem(), keys[2]: StoreItem()}
```
deepankarm, line 15:
```suggestion
store_items = {DaemonID('jflow'): StoreItem(), DaemonID('jflow'): StoreItem(), DaemonID('jflow'): StoreItem()}
```

daemon/models/base.py
```diff
@@ -15,9 +15,12 @@ class StoreStatus(BaseModel):
...
-    items: Dict[DaemonID, StoreItem] = {}
+    items: Dict[DaemonID, StoreItem] = Field(default_factory=dict)
 
     @root_validator(pre=False)
     def set_size(cls, values):
-        values['size'] = len(values['items'])
+        if 'items' in values:
+            values['size'] = len(values['items'])
+        elif 'size' not in values:
+            values['size'] = 0
```
deepankarm, line 25:
```suggestion
        values['size'] = len(values['items']) if 'items' in values else 0
```

tests/daemon/unit/stores/test_basestore.py
```diff
@@ -1,14 +1,18 @@
...
 
 
 def test_base_store_clear():
```
deepankarm, line 18: `test_base_store_clear` & `test_base_store_del` don't make sense anymore. Can you update/delete them?

deepankarm (changes requested):
Minor comments

deepankarm (approved):

PR #2486, opened 2021-05-26
refactor: remove module from typename

jakob1111996:
The following problems occurred when running examples in the 2.0 API:
When indexing or searching by yml file Flow.load_config(...), the index files are stored in
```
workspace/jinahub.DocVectorIndexer/0/file.json
```
When I do the same using the Python syntax Flow().add(...), then the index files are stored in
```
workspace/executors.DocVectorIndexer/0/file.json
```
This causes problems when you want to index via yml file, but search using the Python only interface.
This PR adresses this problem by removing the module name from the folder name.


jina/helper.py
```diff
@@ -949,7 +949,7 @@ def typename(obj):
     if not isinstance(obj, type):
         obj = obj.__class__
     try:
-        return f'{obj.__module__}.{obj.__name__}'
```
JoanFM, line 952: I think typename should not be touched. If the name is taken from the `typename` is some downstream problem. But typename may be better if it returns all.

If the code is only used to identify the executor, then it would need to change its name this method

JoanFM (changes requested):

jina/helper.py
jakob1111996: Do you like this better? https://github.com/jina-ai/jina/pull/2486/commits/c8ccfd7706b3c6aa3d15a52b428e5e87347c98ea

JoanFM (changes requested):
may u add a test for this?

jakob1111996:
Oh sorry, did a rebase to master -  will fix this in a second

tests/unit/executors/test_set_metas.py
```diff
@@ -48,3 +49,30 @@ def test_default_args_from_python():
...
+    jaml_typename = typename(jaml_executor)
+
+    # Second way of importing directly via path in python
+    from .metas_executors import TestExecutor
+
+    py_executor = TestExecutor()
+    py_metas_name = py_executor.metas.name
+    py_typename = typename(py_executor)
+
+    assert jaml_metas_name == py_metas_name
```
JoanFM, line 77: and what is the expected value for each of these?

jakob1111996: https://github.com/jina-ai/jina/pull/2486/commits/e42c536831e118a7e5232501ce87de24b2472fae

JoanFM (approved):

JoanFM:
make sure u rebase with master to pass checks

jakob1111996:
> make sure u rebase with master to pass checks

I don't think the failing checks are caused by the rebase. I just rebased and the problem is still there. 
Could you check the error logs, seems like there is some weird error? @JoanFM

JoanFM (approved):

PR #2489, opened 2021-05-26
ci: only trigger ci when update py yml and requirements

bwanglzu:
if we only update markdown/readme or files not important, we should not wait too long for CI. This change will create a filter to only trigger CI if we change:
1. python files
2. yml (CI/CD) files
3. requirements


bwanglzu:
close this seems not working

PR #2491, opened 2021-05-26
test: fix chatbot codecov missing

### LINKED ISSUE #2481: chatbot helloworld is tested, not covered in codecov
we have system test for chatbot, while codecov ignored chatbot completely (we lose 141 lines of code coverage):

image

bwanglzu:
this PR close #2481 .

it's because we tested it using a subprocess as a package, instead of running the `hello_world` inside test case.


tests/system/chatbot/test_chatbot.py
```diff
@@ -27,20 +30,13 @@ def expected_result():
...
-    xprocess.ensure("server", Starter)
-    yield
-    xprocess.getinfo("server").terminate()
-
-
-def test_chatbot(payload, post_uri, expected_result):
-    """Regression test for chatbot example."""
+def test_multimodal(helloworld_args, expected_result, payload, post_uri):
+    """Regression test for helloworld example."""
+    thread = mp.Process(target=hello_world, args=(helloworld_args,))
```
JoanFM, line 35: let's not call a thread what is a `process`

JoanFM (changes requested):

nan-wang (dismissed):
LGTM👍

bwanglzu:
@nan-wang do not merge yet, sometimes it pass sometimes fail in CI, I need to take a look

tests/system/chatbot/test_chatbot.py
```diff
@@ -27,22 +30,12 @@ def expected_result():
...
     """Regression test for chatbot example."""
+    p = mp.Process(target=hello_world, args=(helloworld_args,))
+    p.start()
+    p.join(30)
     resp = requests.post(post_uri, json=payload)
     assert resp.status_code == 200
-    print('-------------------------1111111----')
-    print(resp.text)
     assert expected_result in resp.text
+    p.kill()
```
JoanFM, line 41: why is there this need?

tests/system/chatbot/test_chatbot.py
```diff
@@ -1,48 +1,70 @@
...
+        }
+    }
 
-    xprocess.ensure("server", Starter)
-    yield
-    xprocess.getinfo("server").terminate()
+    mock_on_done = mocker.Mock()
+    mock_on_fail = mocker.Mock()
 
+    download_data(
```
JoanFM, line 52: I think we do not need to `download` again. We can `index` some set of `text` and `query` with something, not necessarily the same data as in the `example`? It would take too long otherwise

JoanFM (changes requested):

tests/system/chatbot/test_chatbot.py
```diff
@@ -1,48 +1,67 @@
...
-        args = ["jina", "hello", "chatbot"]
-        max_read_lines = 10000
-
-    xprocess.ensure("server", Starter)
-    yield
-    xprocess.getinfo("server").terminate()
+    mock_on_done = mocker.Mock()
+    mock_on_fail = mocker.Mock()
 
+    flow = (
```
JoanFM, line 43: to make sure that we use the `exact` same `flow`, we could add a function in the helloworld that creates the `chatbot flow`? Like this we can make sure that changes in the Flow are properly tested

JoanFM (changes requested):

JoanFM (approved):

PR #2503, opened 2021-05-28
feat: support url read in converters

bwanglzu:
since url is `oneof` uri.  This changes allows us to do:

```python
document = Document(uri='https://jina.ai/assets/images/search_logo.svg')
doc.convert_image_uri_to_blob() # with a real image url
```


bwanglzu:
close this PR, I should use `convert_image_datauri_to_blob` instead.

PR #2505, opened 2021-05-28
fix(daemon): handle empty requirements

jacobowitz:
sorry for the large diff, the black prehook did the formatting.

relevant changes are:

- https://github.com/jina-ai/jina/pull/2505/files#diff-37ce282b4c82b8fba0f0fa251dcd66a6800e19984e046661f523ccacc95dbe91R185
- https://github.com/jina-ai/jina/pull/2505/files#diff-37ce282b4c82b8fba0f0fa251dcd66a6800e19984e046661f523ccacc95dbe91R206
- https://github.com/jina-ai/jina/pull/2505/files#diff-eb700ccf4382124005490360e31674bb156310e0a5bf05c607777b2f0f00e72dR86
- https://github.com/jina-ai/jina/pull/2505/files#diff-eb700ccf4382124005490360e31674bb156310e0a5bf05c607777b2f0f00e72dR98
- 

What this PR actually changes:

- fix empty requirements (before it failed because we expect a list in the pydantic model but it was space separated string)
- fix some metadata for the workflow store (size)
- add some very basic test

### DIFF
--- a/daemon/stores/base.py
+++ b/daemon/stores/base.py
@@ -83,7 +83,6 @@ class BaseStore(MutableMapping):
         """
         self.status.items[key] = value
         self.status.num_add += 1
-        self.status.size += 1
         self.status.time_updated = datetime.now()
 
     def __delitem__(self, key: DaemonID) -> None:
@@ -95,7 +94,6 @@ class BaseStore(MutableMapping):
         .. #noqa: DAR201"""
         self.status.items.pop(key)
         self.status.num_del += 1
-        self.status.size -= 1
         self.status.time_updated = datetime.now()
 
     def __setstate__(self, state: Dict):
@@ -144,3 +142,6 @@ class BaseStore(MutableMapping):
         """Calling :meth:`clear` and reset all stats """
         self.clear()
         self.status = StoreStatus()
+
+    def __len__(self):
+        return len(self.items())
--- a/daemon/stores/base.py
+++ b/daemon/stores/base.py
@@ -62,7 +62,9 @@ class BaseStore(MutableMapping):
     def values(self) -> Sequence[Union['WorkspaceItem', 'ContainerItem']]:
         return self.status.items.values()
 
-    def items(self) -> Sequence[Tuple['DaemonID', Union['WorkspaceItem', 'ContainerItem']]]:
+    def items(
+        self,
+    ) -> Sequence[Tuple['DaemonID', Union['WorkspaceItem', 'ContainerItem']]]:
         return self.status.items.items()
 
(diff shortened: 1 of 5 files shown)


daemon/stores/base.py
```diff
@@ -92,6 +95,7 @@ def __delitem__(self, key: DaemonID) -> None:
         .. #noqa: DAR201"""
         self.status.items.pop(key)
         self.status.num_del += 1
+        self.status.size -= 1
```
deepankarm, line 98: `size` should always be handled in the pydantic model using the validator. If it is not correctly updated, fix should go in `StoreStatus`

daemon/stores/workspaces.py
```diff
@@ -161,26 +165,25 @@ def bob_the_builder(self, id: DaemonID, files: List[UploadFile]):
...
-                    [f.filename for f in files]
-                )
+                self[id].arguments.files.extend([f.filename for f in files])
             self[id].arguments.jinad = {
                 'build': daemon_file.build,
                 'dockerfile': daemon_file.dockerfile,
             }
             self[id].metadata.image_id = image_id
-            self[id].arguments.requirements = daemon_file.requirements
+            self[id].arguments.requirements = list(
```
deepankarm, line 184: Please handle `requirements` in `DaemonFile` property

tests/daemon/unit/stores/test_workspacestore.py
```diff
@@ -0,0 +1,29 @@
...
+    return temp_filepath
+
+
+def test_workspace_store(filepath):
+    store = WorkspaceStore()
+    workspace_id = store.add(files=[UploadFile(filepath)])
+
+    assert store.status.size == 1
+    assert store.status.num_add == 1
+    assert len(list(store.status.items.values())[0].arguments.files) == 1
```
deepankarm, line 23: BTW `store.status.items.values()` is the same as `store.values()`

daemon/stores/workspaces.py
deepankarm: `requirements` is one of the build-args in devel.dockerfile. We need to pass the args as space-separated so that docker build-args can understand it (it doesn't understand a python list). Hence the fix should be in the pydantic model.

deepankarm (changes requested):

daemon/stores/base.py
deepankarm: Now that I think of it, `size` is just the length of items, it can just be `len(store)`. I guess, we can remove size?

deepankarm (approved):

PR #2506, opened 2021-05-29
fix: fail to import on armv8 

### LINKED ISSUE #2447: Unable to run Jina on Raspberry Pi 4
I installed Jina using the installation instructions in the docs and I was unable to run even the Jina command without the error

```
free(): invalid pointer
Aborted
```

I have tried installing Jina from master branch, 1.2.4, and from pip. Yet, the problem persists 

---

**Environment**
Can't actually run this command. 

Latest version of Raspbian OS (aarm64), on a RPi4. Python 3.7.3

**Screenshots**

image

hanxiao:
It's a strange bug caused `import grpc`, yet i'm not sure it is really the upstream problem.

As mentioned in #2447 , running `jina -v` on armv8 throws the exception immediately:
```
free(): invalid pointer
Aborted
```

After some inspections, I found the cause is `import grpc`. To reproduce, **(only on armv8)**, 

-  adding `import grpc` _before_ these lines of `jina/__init__.py`
  ```python
from jina.types.document import Document
from jina.types.arrays.document import DocumentArray
  ``` 
passes.

-  adding `import grpc` _after_ these lines of `jina/__init__.py`
  ```python
from jina.types.document import Document
from jina.types.arrays.document import DocumentArray
  ``` 
fails.

Unfortunately, I don't have time to dig in on this, but this PR at least figured out this issue and fix #2447

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -177,7 +177,7 @@ jobs:
         with:
           context: .
           file: Dockerfiles/debianx.Dockerfile
-          platforms: linux/armhf,linux/amd64,linux/arm/v6,linux/arm/v8
+          platforms: linux/armhf,linux/amd64,linux/arm/v6,linux/arm64
           push: true
           tags: jinaai/jina:master${{ matrix.tag_pyversion }}${{ matrix.tag_stage }}, ${{env.TAG_ALIAS}}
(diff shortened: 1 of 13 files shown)


alexcg1:
I'll check if this works on RaspPi Zero now

PR #2512, opened 2021-05-30
fix: use update instead of MergeFrom

bwanglzu:
with #2465 merged, we switch to `update`.


hanxiao:
im not sure if our tests cover it? could you run `jina hello multimodal` to see if pictures are showed?

PR #2517, opened 2021-05-31
feat: allow adding external Pods to Flow

JoanFM:
- add --freeze-network-settings in the argparser of ZEDRuntime
- add --external in the argparser of Pod
basically,

Pod that labels with freeze-network-settings wont change the external Pod's socket type and port number.
when something must be changed in order to work, but  _build_flow can not change it (e.g. both first and second are freeze-network-settings), then _build_flow should throw an exception. This is inevitable.

Now:

--external=True means mandatory host_in/out, port_in/out
--external=True means setting --freeze-network-settings=True
--external=True is adding a new Pod arg to the Flow, but not really start that Pod in Flow.start , neither close it in Flow.close; external means not controlled by this Flow's context.


hanxiao:
awesome! 👍 ⚡

hanxiao (dismissed):

jina/parsers/peapods/runtimes/zed.py
```diff
@@ -120,3 +120,13 @@ def mixin_zed_runtime_parser(parser):
         if _SHOW_ALL_ARGS
         else argparse.SUPPRESS,
     )
+
+    gp.add_argument(
+        '--freeze-network-settings',
+        action='store_true',
+        default=False,
+        help='if set to true, the host-in/host-out and the port-in/port-out parameters must be properly set, '
```
hanxiao, line 128:
```suggestion
        help='If set, then `host-in`, `host-out`, `port-in`, `port-out`, `socket-in`, `socket-out` network-related parameters can not be changed by the Flow during build time.'
```

PR #2524, opened 2021-05-31
feat: add pretty print for embedding and blob

JoanFM:
**Changes introcduced**
Try to have pretty dict printing for `Document`. At the same time we need to be able to build the `Document` back from a pretty dict like this.

Same would apply for `json`.

image


hanxiao:
maybe better adding `__dict__` to `jina.types.ndarray`?

JoanFM:
> maybe better adding `__dict__` to `jina.types.ndarray`?

Wouldn't it be weird to have this for `ndarray`? I think what more people may do is to check their `document` or to send via rest somehow?

JoanFM:
> maybe better adding `__dict__` to `jina.types.ndarray`?

What we can do is add a method to `DocumentArray` that prints an array of `json` or an array of `dicts`

hanxiao (approved):

Issue #2479, opened 2021-05-25
Move integration tests to 2.0 


FionnD:
**Describe the feature**
As we move towards 2.0, we need to redo each of our current integration tests and update them to the new codebase. 

Integration tests in `1.x`

When moving tests from `issues` to somewhere else, keep the issue ID in the docstring of the test.

TO WORK ON THIS

1. Put your name on a task you pick
1. Open a PR
1. Add the PR next to the checkbox
1. Tick the checkbox once the PR is done

(check first item below)

* [x] crud. Basic: https://github.com/jina-ai/jina/pull/2501 cristianmtr 
  * [x] test workspace usage in CRUD (serialization) https://github.com/jina-ai/jina/pull/2521
  * [x] rename to just `crud` https://github.com/jina-ai/jina/pull/2521
  * [x] test chunks in CRUD https://github.com/jina-ai/jina/pull/2521
* [x] **rest usage in `integration/flow`** cristianmtr https://github.com/jina-ai/jina/pull/2525
* [x] crud_corrupted_docs
  * [x] to be removed
  * [x] **mime type detection under tests/docs** cristianmtr  https://github.com/jina-ai/jina/pull/2531
* ~[ ] debug -- to be moved to hub~
* [x] doccache -- only on 'r' level https://github.com/jina-ai/executor-indexers/pull/21
* [x] dump -- moved to hub integration https://github.com/jina-ai/indexers/pull/8 cristianmtr 
* [x] eval_flow. To be kept. Tests hanging pods for evaluation (name has changed: `test_inspect_pods_flow.py`)
* [x] evaluation. 
  * [x] one part move to Hub, about the Evaluator @JoanFM 
  * [x] groundtruth test in core, check to see if you can pass GroundTruths to inputs #2520 maximilianwerk 
* ~[ ] executors_batching~. We don't have batching anymore. Responsibility of Executors
* ~[ ] executors_generic~. Tested in other places.
* [x] gateway_non_blocking -- added
* [x] high_order_matches -- added
* [x] hub_usage -- added
* [x] incremental_indexing -- to be added to DocCache in Hub https://github.com/jina-ai/executor-indexers/pull/21
* **issues**
  * ~[ ] github_1072~
  * [x] **check if filter is tested on DA** cristianmtr https://github.com/jina-ai/jina/pull/2535
  * [x] check if parameters are passed to f.post #2520 maximilianwerk
  * ~[ ] github_1229~
[...]

PR #2525, opened 2021-05-31
test: crud rest

cristianmtr:
Part of https://github.com/jina-ai/jina/issues/2479

### DIFF
--- a/tests/integration/crud/flow.yml
+++ b/tests/integration/crud/flow.yml
@@ -1,6 +1,7 @@
 jtype: Flow
 with:
   return_results: True
+  restful: $RESTFUL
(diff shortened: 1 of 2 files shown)


maximilianwerk (approved):
My first reaction was: That should be two separate tests.

But actually, I really like, that for each step you directly see how a rest request and a grpc requests relate. Really nice!

cristianmtr:
> My first reaction was: That should be two separate tests.
> 
> But actually, I really like, that for each step you directly see how a rest request and a grpc requests relate. Really nice!

Yeah, I thought it would be useful to have it in one place, for comparison

PR #2532, opened 2021-06-01
test: added test for startup exception

maximilianwerk:
This tests previous issues/github_1861 which checked, that a Flow is not hanging, when an Executor crashes during startup.


jacobowitz (approved):
lgtm :+1:

PR #2537, opened 2021-06-01
test: added a test for docmatrix in the tail pea

maximilianwerk:
This is a test showcasing, how a `MatchMerger` in a TailPea could look like. Beware, that replica/sharding for `(1,2)` and `(2,1)` are broken. That needs to be fixed before 2.0 release.

### DIFF
--- a/scripts/black.sh
+++ b/scripts/black.sh
@@ -5,6 +5,7 @@ echo we ignore non-*.py files and files generated from protobuf
 excluded_files=(
    jina/proto/jina_pb2.py
    jina/proto/jina_pb2_grpc.py
+   tests/
 )
 for changed_file in $CHANGED_FILES; do
   if [[ ${changed_file} == *.py ]] && ! [[ " ${excluded_files[@]} " =~ " ${changed_file} " ]]; then
--- a/scripts/black.sh
+++ b/scripts/black.sh
@@ -5,7 +5,6 @@ echo we ignore non-*.py files and files generated from protobuf
 excluded_files=(
    jina/proto/jina_pb2.py
    jina/proto/jina_pb2_grpc.py
-   tests/
 )
 for changed_file in $CHANGED_FILES; do
   if [[ ${changed_file} == *.py ]] && ! [[ " ${excluded_files[@]} " =~ " ${changed_file} " ]]; then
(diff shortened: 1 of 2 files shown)


cristianmtr (approved):

JoanFM:
(1, 2) should work now.

What I am not sure about with `replicas`. Right now the `HeadPea` or `TailPea` of a `CompoundPod` are not equipped with anything (which is an assumption of the polling types)

tests/integration/v2_api/test_docs_matrix_tail_pea.py
```diff
@@ -0,0 +1,77 @@
+import pytest
+from collections import OrderedDict
+from jina import Document, DocumentArray, Executor, Flow, requests
+
+
+@pytest.mark.timeout(5)
+@pytest.mark.parametrize('replicas_shards', [(1, 1), (2, 2)])
```
JoanFM, line 7:
```suggestion
@pytest.mark.parametrize('replicas, shards', [(1, 1), (2, 2)])
def test_sharding_tail_pea(nu,m_replicas, num_shards):
```

I think this works better and u do not need to separate

JoanFM (changes requested):

scripts/black.sh
```diff
@@ -5,6 +5,7 @@ echo we ignore non-*.py files and files generated from protobuf
 excluded_files=(
    jina/proto/jina_pb2.py
    jina/proto/jina_pb2_grpc.py
+   tests/
```
maximilianwerk, line 8: Why?

JoanFM: I agree with Max, I like having them beautified

tests/integration/v2_api/test_docs_matrix_tail_pea.py
```diff
@@ -0,0 +1,94 @@
...
+                    results[doc.id].chunks.extend(doc.chunks)
+                else:
+                    results[doc.id] = doc
+        return DocumentArray(results.values())
+
+
+@pytest.mark.timeout(5)
+@pytest.mark.parametrize('num_replicas, num_shards', [(1, 1), (2, 2)])
+def test_sharding_tail_pea(num_replicas, num_shards):
+    """TODO(Maximilian): Make (1, 2) and (2, 1) also workable"""
```
cristianmtr, line 54: opened an issue for this https://github.com/jina-ai/jina/issues/2547

cristianmtr (approved):

PR #2545, opened 2021-06-02
test: slow shutdown

cristianmtr:
Part of https://github.com/jina-ai/jina/issues/2479


maximilianwerk:
I will change the 360 to something like 10 seconds. I don't see anything that could happen in the 350 seconds in between, which would justify making our unit tests run forever.

If you disagree, we need to think about another way of running this test. putting it into a separate integration test or so.

maximilianwerk (approved):
LGTM

PR #2546, opened 2021-06-02
test: check distinct executors in roll update

cristianmtr:


tests/integration/rolling_update/test_rolling_update.py
```diff
@@ -294,3 +295,40 @@ def test_num_peas(config):
             + 1  # compound pod tail
             + 1  # gateway
         )
+
+
+@pytest.mark.repeat(5)
+@pytest.mark.timeout(30)
+def test_distinct_executor_ids(config, docs, mocker, reraise):
```
JoanFM, line 302: isn't this tested in `test_normal`? I agree we can retest with multithreading, but we can then use the `num_replicas`, `num_shards` logic?

cristianmtr: Wasn't sure it was working properly, as there were more shards. Wanted to have the minimum nr to assert on

maximilianwerk (changes requested):
The added test does fail quite often and is flaky. Should not be added before we get the root cause solved.

nan-wang:
> The added test does fail quite often and is flaky. Should not be added before we get the root cause solved.

Let's close this PR for the time being.

PR #2551, opened 2021-06-02
feat: draft of dynamic routing

maximilianwerk:
Implementation of dynamic routing for jina. See #2572


JoanFM:
I think the key here would be to be able to show that u can do this supporting the `bifurcation pattern` (currently covered by `PUB-SUB`) in an `async fashion`

jina/peapods/peas/dynamic_head_pea.py
```diff
@@ -0,0 +1,134 @@
...
+        self.receiver = self.context.socket(zmq.ROUTER)
+        self.receiver.bind(self._in_socket)
+
+    def run(self):
+        while True:
+            self.receiver.recv()
+            message = self.receiver.recv_json()
+
+            self.do_work(message)
+            self.post(message)
```
hanxiao, line 36: if i understand it correctly, these lines here basically summarizes head pea, worker pea and tail pea into one, for demo purpose.

Roughly head pea:
```python
self.receiver.recv()
message = self.receiver.recv_json()
```

Roughly worker pea
```python
self.do_work(message)
```

Roughly tail pea:
```python
self.post(message)
```

Am I right?


This code is for demo purpose, but in practice, the following lines should be implemented with `ZmqStreamlet` or `AsyncZmqlet` as we currently use them in ZEDRuntime:
```
self.receiver.recv()
...
self.post(message)
```

jina/peapods/peas/dynamic_head_pea.py
```diff
@@ -0,0 +1,134 @@
...
+        for next_target, downstream in next_routes.items():
+            if next_target not in self.routes:
+                new_socket = self.context.socket(zmq.DEALER)
+                new_socket.connect(next_target)
+                self.routes[next_target] = new_socket
+
+            next_message = copy.deepcopy(message)
+            next_message['routes'] = downstream
+
+            self.routes[next_target].send_json(next_message)
```
hanxiao, line 60: in practice, use `AsyncZmqlet.send`

jina/peapods/peas/dynamic_head_pea.py
hanxiao: am I also right about the affection here?
image

maximilianwerk: Yes, all of you observation are right. Only the following three communication types are affected:

- Gateway -> HeadPea
- TailPea->HeadPea 
- TailPea -> Gateway

If we treat Gateway as a combination of a HeadPea and a TailPea (from a networking perspective) it boils down to `TailPea -> HeadPea`.

jina/peapods/peas/dynamic_head_pea.py
maximilianwerk: :+1:

jina/peapods/peas/dynamic_head_pea.py
JoanFM: So talking in current `naming` it seems that every HeadPea will be a `ROUTER server (bind) (ROUTER_BIND)` and every TailPea a `DEALER client (connect) (DEALER_CONNECT)`, right?

JoanFM: https://stackoverflow.com/questions/19070140/zeromq-dealer-client-to-multiple-servers-router

this may be good to take into account

JoanFM:
should we have a `rough` estimation on how it differs the performance from having this `multisocket` router pattern in comparision to a `publisher` `subscriber` one?

Maybe we can have an example with both patterns and check the amount of messages processed at a time?

jina/peapods/zmq/__init__.py
```diff
@@ -16,9 +16,11 @@
 from ...importer import ImportExtensions
 from ...logging.predefined import default_logger
 from ...logging.logger import JinaLogger
+from ...proto import jina_pb2
```
JoanFM, line 19: I guess this is leftover

tests/unit/test_echostream.py
```diff
@@ -55,6 +56,66 @@ def test_simple_zmqlet():
...
+            'ROUTER_BIND',
+            '--socket-out',
+            'PUSH_BIND',
+            '--timeout-ctrl',
+            '-1',
+        ]
+    )
+
+    logger = logging.getLogger('zmq-test')
+    with BasePea(args2), Zmqlet(args, logger) as z:
```
JoanFM, line 97: what is z in thse 2 context managed? why is the Pea needed and how is it validated?

jina/peapods/zmq/__init__.py
```diff
@@ -166,31 +174,29 @@ def _init_sockets(self) -> Tuple:
...
+                    self.args.socket_out,
+                    self.identity,
+                    ssh_server=self.args.ssh_server,
+                    ssh_keyfile=self.args.ssh_keyfile,
+                    ssh_password=self.args.ssh_password,
+                )
+                self.logger.debug(
+                    f'output {self.args.host_out}:{colored(self.args.port_out, "yellow")}'
+                )
+            else:
```
JoanFM, line 191: I like better to have `out_sock, out_addr = None, None` above as part of initialization better than in an `else`

jina/peapods/zmq/__init__.py
```diff
@@ -305,10 +358,23 @@ async def send_message(self, msg: 'Message', sleep: float = 0, **kwargs):
...
+            await self.send_message_dynamic(msg)
+            return
+        else:
+            self._send_message_via(self.out_sock, msg)
+
+    async def send_message_dynamic(self, msg: 'Message'):
+        for routing_graph, out_sock in self.get_dynamic_next_routes(msg):
+            new_message_proto = msg.proto
+            new_message_proto.envelope.targets.CopyFrom(routing_graph.proto)
+            await self._send_message_via(
```
JoanFM, line 371: Just to clarify my understanding. This is `asynchronous` but until the `_send_message_via` returns the next one cannot be sent right? 

Does it need to be awaited? Can't all of these results put in different futures and await them after the `for-loop`?

maximilianwerk: Yes, that might be an optimization in the future.

deepankarm: `asyncio.create_task(blah())` should work for fire & forget.

JoanFM: just want to make sure if we are aware that this `for loop` will actually be executed sequentially

jina/flow/base.py
```diff
@@ -393,6 +398,76 @@ def gather_inspect(
...
+                'expected_parts': 0,
+            },
+        )
+
+    def _update_port_type(self, pod):
+        pod.head_args.socket_in = SocketType.ROUTER_BIND
+        pod.tail_args.dynamic_out_routing = True
+        pod.tail_args.socket_out = SocketType.DEALER_CONNECT
+
+    def _get_routing_graph(self):
```
JoanFM, line 416: I think this can be `unified` with the code that is done now in `Flow.build` where a `map` from starting Pods to outgoing Pods is built by name.

This could be the input of this way of building the graph

maximilianwerk: No the other can just be removed. Not needed anymore. As I wrote in slack: cleanup is pending...

jina/peapods/zmq/__init__.py
```diff
@@ -61,16 +63,25 @@ def __init__(
...
+        else:
+            self.out_sock_type = zmq.DEALER
+        self.ctrl_sock_type = self.ctrl_sock.type
         self._register_pollin()
+        self.opened_socks.extend([self.in_sock, self.ctrl_sock])
+        if self.out_sock is not None:
+            self.opened_socks.append(self.out_sock)
 
-        self.opened_socks.extend([self.in_sock, self.out_sock, self.ctrl_sock])
+        self.out_sockets = {}
```
maximilianwerk, line 77: I wonder if all sockets insight the current `Zmqlet` should go into another `Sockets` class, which cares about opening/closing (aka context managing) sockets, while the `Zmqlet` cares about the business logic.

jina/peapods/zmq/__init__.py
maximilianwerk: But this is the functional approach: Set each variable only once and don't overwrite it. Why would you initialize the variable and then overwrite them? I guess it boils down to C-style code vs functional programming (Scala?) style code.

JoanFM: ok, I would consider it part of rhe initialization but fair enough

jina/flow/base.py
```diff
@@ -605,6 +616,83 @@ def gather_inspect(
...
+                {
+                    'host': pod.head_host,
+                    'port': pod.head_port_in,
+                    'expected_parts': 0,
+                },
+            )
+            for pod_id, pod in self._pod_nodes.items()
+            if pod_id != GATEWAY_NAME
+        ]
+        target_pods = (
```
JoanFM, line 648: why do not unify in the line above?

jina/flow/base.py
```diff
@@ -605,6 +616,83 @@ def gather_inspect(
...
+            for start in pod.needs:
+                if start == GATEWAY_NAME:
+                    start = f'start-{GATEWAY_NAME}'
+
+                if start not in pods_id_to_position:
+                    raise FlowMissingPodError(
+                        f'{start} is not in this flow, misspelled name?'
+                    )
+
+                end_pod = pods_id_to_position[end]
```
JoanFM, line 674: can this be extracted in 2 meaningful names?

jina/flow/base.py
```diff
@@ -605,6 +616,83 @@ def gather_inspect(
...
+                end_pod[1]['expected_parts'] += 1
+                start_pos = pods_id_to_position[start][0]
+                end_pos = end_pod[0]
+                edges[start_pos].append(end_pos)
+
+        targets = [pod[1] for pod in target_pods]
```
JoanFM, line 680: same here

jina/flow/base.py
```diff
@@ -605,6 +616,83 @@ def gather_inspect(
...
+        parsed_edges = {key: {'targets': value} for key, value in edges.items()}
+        graph = {'active_pod_index': 0, 'pods': targets, 'edges': parsed_edges}
+
+        routing_pb = jina_pb2.RoutingGraphProto()
+        json_format.ParseDict(graph, routing_pb)
+        return RoutingGraph(routing_pb)
```
JoanFM, line 686: Doesn't it make more sense to hide this `Parsing` inside the `RoutingGraph` construction? So that it is better testable

jina/flow/base.py
```diff
@@ -605,6 +616,83 @@ def gather_inspect(
...
+
+    def _init_dynamic_routing(self):
+        routing_graph = self._get_routing_graph()
+        if not routing_graph.is_acyclic():
+            raise RoutingGraphCyclicError(
+                "The routing graph has a cycle. This would result in an infinite loop. Fix your Flow setup."
```
JoanFM, line 692: single quotes

jina/flow/base.py
```diff
@@ -605,6 +616,83 @@ def gather_inspect(
...
+
+        targets = [pod[1] for pod in target_pods]
+        parsed_edges = {key: {'targets': value} for key, value in edges.items()}
+        graph = {'active_pod_index': 0, 'pods': targets, 'edges': parsed_edges}
+
+        routing_pb = jina_pb2.RoutingGraphProto()
+        json_format.ParseDict(graph, routing_pb)
+        return RoutingGraph(routing_pb)
+
+    def _init_dynamic_routing(self):
```
JoanFM, line 688: I would name `_set_initial_dynamic_routing_graph`. This naming seems that is an initialization that will be changed.

Why I guess it is the `graph` that will be used by default by the `Gateway` right?

jina/peapods/pods/factory.py
```diff
@@ -21,6 +26,13 @@ def build_pod(args: 'Namespace', needs: Optional[Set[str]] = None) -> BasePod:
         :return: the created BasePod
         """
         if getattr(args, 'replicas', 1) > 1:
-            return CompoundPod(args, needs=needs)
+            pod = CompoundPod(args, needs=needs)
         else:
-            return Pod(args, needs=needs)
+            pod = Pod(args, needs=needs)
+
+        if setup_dynamic_routing:
```
JoanFM, line 33: this should be in the `constructor` of `Pod` and `CompoundPod` themselves

jina/peapods/pods/gateway.py
```diff
@@ -0,0 +1,13 @@
+from .. import Pod
+from ...types.routing.graph import RoutingGraph
+
+
+class GatewayPod(Pod):
+    """A GatewayPod is a wrapper around the gateway pea.
```
JoanFM, line 6: not a wrapper, a context manager.

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -361,7 +362,14 @@ def expect_parts(self) -> int:
 
         :return: expected number of partial messages
         """
-        return self.args.num_part if self.message.is_data_request else 1
+        if self.message.is_data_request:
+            if self.args.socket_in == SocketType.ROUTER_BIND:
```
JoanFM, line 366: I think this logic should be based on the presence or not of `routing_graph` and not on the type of socket

jina/peapods/zmq/__init__.py
JoanFM: I would vote for this

jina/peapods/zmq/__init__.py
```diff
@@ -231,22 +240,73 @@ def print_stats(self):
...
+            ssh_server=self.args.ssh_server,
+            ssh_keyfile=self.args.ssh_keyfile,
+            ssh_password=self.args.ssh_password,
+        )
+        self.logger.debug(
+            f'output {self.args.host_out}:{colored(self.args.port_out, "yellow")}'
+        )
+        return out_sock
+
+    def _get_dynamic_out_socket(self, target_pod, as_streaming=False):
```
JoanFM, line 259: what is this `as_streaming` and where is it used?

JoanFM (commented):
Looks really good

jina/flow/base.py
maximilianwerk: I changed the naming of the above variable and it should be better now.

jina/flow/base.py
maximilianwerk: Great suggestion!

jina/peapods/runtimes/zmq/zed.py
maximilianwerk: Haha, on this is more subtle. You always have a routing graph. But only the the HeadPeas should respect the `expected_parts` coming from the routing graph. TailPeas after sharding for example still need to check `self.args.num_part`.

I could at an argument to HeadPeas, similar to `dynamic_out_routing` to be `dynamic_in_routing`. Would you prefer that?

jina/peapods/zmq/__init__.py
maximilianwerk: It is used in the `ZmqStreamlet`.

jina/peapods/pods/factory.py
maximilianwerk: This will create conflicts with the ContainerRuntime. If we do this, I would still need to pass the `setup_dynamic_routing`. Are you fine with it being forwarded?

jina/peapods/runtimes/zmq/zed.py
JoanFM: I would do, I think is less error-prone and more explicit. 

I can imagine int he future for some reason setting a `ROUTER_BIND` in a very different context and having a very bad surprise

jina/peapods/pods/factory.py
maximilianwerk: Imagine the following test:
```
def test_flow_with_replica_container_ext_yaml(docker_image_built):
    f = Flow().add(
        name='dummyEncoder3',
        uses=f'docker://{img_name}',
        uses_internal=os.path.join(cur_dir, '../../../mwu-encoder/mwu_encoder_ext.yml'),
        parallel=3,
    )

    with f:
        f.index(inputs=random_docs(10))
        f.index(inputs=random_docs(10))
        f.index(inputs=random_docs(10))
```
This should spawn three docker instances insight a CompoundPod. Inside each of the docker instances, you spawn another Pod, which should NOT get the dynamic routing setup. Anyhow, it is spawned via `jina pod` CLI and thus call the `PodFactory.build_pod`. Thus, we need a flip for the `PodFactory.build_pod` method, whether the build pod should use dynamic routing. 

Thinking about it, this again should be handled via the `args` Namespace and potentially only on the actual Pea level. I will create a solution and let you double-check it again.

jina/peapods/zmq/__init__.py
```diff
@@ -305,10 +365,24 @@ async def send_message(self, msg: 'Message', sleep: float = 0, **kwargs):
...
+
+    async def _send_message_dynamic(self, msg: 'Message'):
+        tasks = []
+        for routing_graph, out_sock in self.get_dynamic_next_routes(msg):
+            new_message_proto = msg.proto
+            new_message_proto.envelope.routing_graph.CopyFrom(routing_graph.proto)
+            tasks.append(
+                self._send_message_via(out_sock, Message.from_proto(new_message_proto))
             )
+        await asyncio.gather(*tasks)
```
maximilianwerk, line 381: Can you please double-check, that this is the Jina conform way of doing it? @deepankarm @JoanFM Thanks!

jina/peapods/zmq/__init__.py
maximilianwerk: Changed it. Please double check, if it as you imagine.

jina/peapods/zmq/__init__.py
JoanFM: I think so, but @deepankarm  is the asyncio master

jina/peapods/networking.py
```diff
@@ -0,0 +1,57 @@
...
+    ) is not None and connect_args.uses.startswith('docker://')
+
+    # is BIND & CONNECT all on the same remote?
+    bind_conn_same_remote = (
+        not bind_local and not conn_local and (bind_host == connect_args.host)
+    )
+
+    if platform in ('linux', 'linux2'):
+        local_host = __default_host__
+    else:
```
cristianmtr, line 40: Should we add an if here to catch other OSs?

jina/proto/jina.proto
```diff
@@ -164,6 +164,21 @@ message RouteProto {
...
 
+message TargetPodProto{
+    string host = 1; // the host HeadPea of the BasePod
+    uint32 port = 2; // the port HeadPea of the BasePod
+    uint32 expected_parts = 3; // the number of parts, the pod should expect
+    repeated string out_edges = 4;
+}
+
+message RoutingGraphProto {
+    // Pods that get visited during a Flow. Gateway should be both, the first and the last entry.
```
cristianmtr, line 175:
```suggestion
    // Pods that get visited during a Flow. Gateway should be both the first and the last entry.
```

jina/proto/jina.proto
```diff
@@ -164,6 +164,21 @@ message RouteProto {
     StatusProto status = 5; // the status of the execution
 }
 
+message TargetPodProto{
+    string host = 1; // the host HeadPea of the BasePod
+    uint32 port = 2; // the port HeadPea of the BasePod
+    uint32 expected_parts = 3; // the number of parts, the pod should expect
```
cristianmtr, line 170:
```suggestion
    uint32 expected_parts = 3; // the number of parts the pod should expect
```

jina/proto/jina.proto
```diff
@@ -164,6 +164,21 @@ message RouteProto {
...
+    repeated string out_edges = 4;
```
cristianmtr, line 171: Can we add a short docstring on this?

jina/types/routing/graph.py
```diff
@@ -0,0 +1,192 @@
...
+
+        :return: port as an int
+        """
+        return self.proto.port
+
+    @property
+    def host(self):
+        """Returns the `host` field of this TargetPod
+
+        :return: host as string
```
cristianmtr, line 33: move the `as <type>` to type annotations of the property?

tests/distributed/test_join_local_from_remote/flow.yml
```diff
@@ -3,6 +3,7 @@ version: '1'
 with:
   port_expose: 45678
   show_exc_info: true
+  host: flow
```
cristianmtr, line 6: Is using 'flow' as `host` the way to specify it as localhost?

maximilianwerk: No, this is coupled with the `docker-compose` naming scheme of docker pods.

jina/peapods/networking.py
maximilianwerk: This is exactly the same code as before, just in a different location. I would not touch it, since I am not a Mac expert.

JoanFM (approved):

PR #2556, opened 2021-06-03
feat: wrap flow/pea/pod in mini JinaD

jacobowitz:
Please dont review this yet, its not ready. I am using it for tracking status and some questions

TODO:

- [x] Replace direct jina entrypoint in Docker with "mini" jinad, which starts pod/pead/flow as necessary
- [x] Start the correct API from 'mini' jinad
- [ ] Forward flow/pod/pea questions from main jinad to correct mini jinad

QUESTIONS:
- How do we do the mapping for the request forwarding. There will be some orginal request like /flow/rolling_update and it needs to mapped to something like `/flows/<id>/rolling_update `?
- Atm I am adding a new entrypoint for the 'mini' jinad. That works but may confuse people as its parts of the complete Jina component. We might want to hide it more as its a very specialized use case?
- Do I need to do anything special for the log/fluentd config?

### DIFF
--- a/daemon/dockerize.py
+++ b/daemon/dockerize.py
@@ -57,14 +57,33 @@ class Dockerizer:
         if workspace_id in cls.networks:
             network = cls.client.networks.get(network_id=workspace_id)
         else:
-            network: 'Network' = cls.client.networks.create(name=workspace_id, driver='bridge')
+            from .stores import workspace_store
+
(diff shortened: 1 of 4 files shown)


jacobowitz:
https://github.com/jina-ai/jina/pull/2558

PR #2565, opened 2021-06-05
refactor: merge grpc and rest client

bwanglzu:
This PR aims at simplify C/S mode (Flow as service) Jina Python client usage with a uniform facade, before and after:

```python
# before
from jina import Client
from jina.clients import WebsocketClient

client = Client('localhost', 12345) # connect to grpc gateway
client = WebsocketClient('localhost', 12345) # connect to restful gateway

# now
from jina import Client
client = Client('localhost', 12345)  # connect to grpc gateway
client = Client('localhost', 12345, restful=True)  # connect to restful gateway

# it uses the same semantics as what we have defined for the flow:
f = Flow()
f = Flow(restful=True)
```


JoanFM (approved):
I like the change

jina/flow/base.py
```diff
@@ -912,7 +914,7 @@ def use_grpc_gateway(self, port: Optional[int] = None):
 
     def _switch_gateway(self, gateway: str, port: int):
         restful = gateway == 'RESTRuntime'
-        client = WebSocketClient if gateway == 'RESTRuntime' else Client
+        client = WebSocketClient if gateway == 'RESTRuntime' else GrpcClient
```
deepankarm, line 917: Why can we not use `Client` here passing restful?

bwanglzu: @deepankarm we can, while it does not make a lot difference since `Client` is actually initialising `GrpcClient` and `WebSocketClient`

JoanFM: then lets have a single poijt of entry, change it to Client with the proper argument, u can use partial functions if needed

bwanglzu: done

deepankarm: I meant, not using `GrpcClient` & `WebSocketClient` at all in `flow/base.py` and using `Client` directly. Not important though.

bwanglzu: gave a try, and I broke a lot of tests. I'll look into if I can optimise this part in a future PR.

jina/clients/__init__.py
```diff
@@ -4,25 +4,42 @@
...
+    """Jina Python client.
+
+    :param host: Host address of the flow.
+    :param port_expose: Port number of the flow.
+    :param restful: If connect to a Restful gateway, default is ``False``, connect to GrpcGateway.
+    :return: An instance of :class:`GrpcClient` or :class:`WebSocketClient`.
+    """
+    args = set_client_cli_parser().parse_args(
+        ['--host', host, '--port-expose', str(port_expose)]
+    )
```
hanxiao, line 22: when written in this way, all other client arguments are not passed. better use `**kwargs` and use `ArgNamespace.kwargs2list()`
```
usage: jina client [-h] [--request-size] [--continue-on-error]
                   [--show-progress] [--return-results] [--host]
                   [--port-expose] [--proxy]

Start a Python client that connects to a remote Jina gateway

⚙️  Optional Arguments

  -h, --help           show this help message and exit

⚙️  Client Arguments

  --request-size       type: int; default: 100

                       The number of Documents in each Request.

  --continue-on-error  default: disabled, use "--continue-on-error" to
                       enable it

                       If set, a Request that causes error will be logged only
                       without blocking the further requests.

  --show-progress      default: disabled, use "--show-progress" to enable
                       it

                       If set, client will show a progress bar on receiving
                       every request.

  --return-results     default: disabled, use "--return-results" to enable
                       it

                       This feature is only used for AsyncClient.

                       If set, the results of all Requests will be returned as
                       a list. This is useful when one wants process Responses
                       in bulk instead of using callback.

⚙️  Expose Arguments

  --host               type: str; default: 0.0.0.0

                       The host address of the runtime, by default it is
                       0.0.0.0.

  --port-expose        type: int; default: 58570

                       The port of the host exposed to the public

  --proxy              default: disabled, use "--proxy" to enable it

                       If set, respect the http_proxy and https_proxy
                       environment variables. otherwise, it will unset these
                       proxy variables before start. gRPC seems to prefer no
                       proxy
```

jina/flow/base.py
```diff
@@ -169,7 +172,7 @@ def _add_gateway(self, needs, **kwargs):
                 name=pod_name,
                 ctrl_with_ipc=True,  # otherwise ctrl port would be conflicted
                 runtime_cls='GRPCRuntime'
-                if self._cls_client == Client
+                if self._cls_client == GrpcClient
```
hanxiao, line 175: better uniform the case, `GRPCRuntime` then `GRPCClient` or `GrpcRuntime` and `GrpcClient`

jina/flow/base.py
hanxiao: > and I broke a lot of tests

I believe the reason was because your `Client()` function does not accept the full arguments that CLI defined. This oversimplification causes incomplete setup of the client and yields the broken test.

https://github.com/jina-ai/jina/pull/2565#discussion_r647896285

hanxiao:
Could you inject `jina client` CLI to `def Client`? 
- uses `scripts/inject-cli-as-overload.py`
- follows the syntax as in #2598 via `# overload_inject_start_client` & `# overload_inject_end_client`

hanxiao:
note that, one has to run `scripts/inject-cli-as-overload.py` manually and `black -s` it

bwanglzu:
applied some changes to `scripts/inject-cli-as-overload.py `, method defined inside / outside class should have different indentation.

scripts/inject-cli-as-overload.py
```diff
@@ -76,37 +87,66 @@ def fill_overload(
     return_type,
     filepath,
     overload_fn,
+    class_method,
     indent=' ' * 4,
```
hanxiao, line 91: wait, isn't the modification the same as `fill_overload(..., indent=' '*2)`?

bwanglzu: yes you're right, working on it now

bwanglzu: @hanxiao it's not only about `indent`, the script will also insert a `self` inside the method etc..this is what happened after i change `indent = ' ' * 2`:

C48D9157-14A7-40FA-88AF-8367A1A16BC8

hanxiao (approved):

PR #2575, opened 2021-06-07
refactor: refactor optimizer parameter

JoanFM:


jina/optimizers/parameters.py
```diff
@@ -19,6 +23,16 @@ def __init__(
         else:
             self.jaml_variable = jaml_variable
 
+    @abstractmethod
```
bwanglzu, line 26: have you set abc as meta class?

JoanFM: no, is it needed?

bwanglzu: yes otherwise it won't work

JoanFM: done!

bwanglzu (approved):
lgtm!

PR #2583, opened 2021-06-08
fix: fix _mermaid_id to be id

JoanFM:


JoanFM:
BREAKS THE PLOT

JoanFM:
Another option will be considered

PR #2587, opened 2021-06-08
docs(readme-banner): github banner for readme

gitcommitshow:
Before readers get into details, can we help users visualize what they will be able to do using Jina?
This is the question this banner answers.


JoanFM (approved):

alexcg1:
See Slack - creative team are already designing new animated banner. So I'm afraid we'll have to close this :(

PR #2597, opened 2021-06-09
feat: put graph as one of content

JoanFM:
**Changes introduced**
Move `graph_info` to `content` in a `Document`.

**Benefits**
As suggested by @maximilianwerk this enforces the design ot a root `Document` **BEING** a graph and not only containing graph information

**Drawbacks**
Potentially there may be cases where one would also like to have some content in the root graph (I am not sure this is true as I did not find any case for this)

**Open questions**
Should we add the ability to traverse on `nodes` as synonim for `chunks`? Should this be only for `GraphDocument`?

### DIFF
--- a/jina/proto/jina.proto
+++ b/jina/proto/jina.proto
@@ -106,6 +106,9 @@ message DocumentProto {
 
         // a uri of the document could be: a local file path, a remote url starts with http or https or data URI scheme
         string uri = 9;
+
+        // Related information to be used when a Document represents a graph with its chunks as nodes
+        GraphProto graph_info = 27;
     }
 
     // list of the sub-documents of this document (recursive structure)
@@ -140,9 +143,6 @@ message DocumentProto {
 
     // List of evaluations performed on the document, each element corresponds to a metric
     repeated NamedScoreProto evaluations = 23;
-
-    // Related information to be used when a Document represents a graph with its chunks as nodes
-    GraphProto graph_info = 26;
 }
 
 /**
--- a/jina/proto/jina.proto
+++ b/jina/proto/jina.proto
@@ -108,7 +108,7 @@ message DocumentProto {
         string uri = 9;
 
         // Related information to be used when a Document represents a graph with its chunks as nodes
-        GraphProto graph_info = 27;
+        GraphProto graph = 27;
     }
 
     // list of the sub-documents of this document (recursive structure)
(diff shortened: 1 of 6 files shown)


jina/proto/jina.proto
```diff
@@ -106,6 +106,9 @@ message DocumentProto {
 
         // a uri of the document could be: a local file path, a remote url starts with http or https or data URI scheme
         string uri = 9;
+
+        // Related information to be used when a Document represents a graph with its chunks as nodes
+        GraphProto graph_info = 27;
```
hanxiao, line 111: how about just `graph`

JoanFM: yes, may make sense.

What about the `traversal-path` I comment in the `Description`, any opinion?

hanxiao:
oh btw, we tend to keep field `mime_type` standardized, and if there is a "graph"-like type approved by IANA, https://www.iana.org/assignments/media-types/media-types.xhtml please just use it

JoanFM:
> oh btw, we tend to keep field `mime_type` standardized, and if there is a "graph"-like type approved by IANA, https://www.iana.org/assignments/media-types/media-types.xhtml please just use it

It does not seem to have a general graph mime type

hanxiao:
then we probably should not create one, what is this for btw? we already can tell doc content type via `doc.content_type`

JoanFM:
> then we probably should not create one, what is this for btw? we already can tell doc content type via `doc.content_type`

Ok, just thought it may make sense to call `graph` a new `mime_type` but we can remove

jina/types/document/graph.py
```diff
@@ -48,7 +48,9 @@ def __init__(
         super().__init__(document=document, copy=copy, **kwargs)
         self._node_id_to_offset = {
             node.id: offset for offset, node in enumerate(self.nodes)
-        }  # dangerous because document is stateless, try to work only with proto,
+        }  # dangerous because document is stateless, try to work only with proto
+
+        self.mime_type = 'graph'
```
JoanFM, line 53:
```suggestion
```

tests/unit/types/document/test_graph_document.py
```diff
@@ -22,6 +22,7 @@ def graph():
 
 
 def validate_graph(graph):
+    assert graph.mime_type == 'graph'
```
JoanFM, line 25:
```suggestion
```

hanxiao (approved):

PR #2598, opened 2021-06-09
feat: inject cli args as overload function 

hanxiao:
- [background: `@overload`](https://www.python.org/dev/peps/pep-3124/)
- [usage in Python](https://github.com/search?l=Python&q=org%3Apython+%27%40overload%27&type=Code)
- adding `@overload` method gives nice code-completion and intellisense, help users to onboard much quicker, see images below.
- `Flow.__init__` and `Flow.add()` are two popular user-facing methods; also `Client.__init__` (but this is blocked by #2565 )

#### Overload enables autocompletion

image

#### Overload enables docstring hint
image

### DIFF
--- a/scripts/inject-cli-as-overload.py
+++ b/scripts/inject-cli-as-overload.py
@@ -1,0 +1,117 @@
+import re
+
+from cli.export import api_to_dict
+
+
+def _cli_to_schema(
+    api_dict,
+    target,
+):
+    pod_api = None
+
+    for d in api_dict['methods']:
+        if d['name'] == target:
+            pod_api = d['options']
+            break
+
+    _schema = {
+        'properties': {},
+        'required': [],
+    }
+
+    for p in pod_api:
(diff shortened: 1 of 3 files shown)


scripts/inject-cli-as-overload.py
```diff
@@ -0,0 +1,117 @@
...
+        re.DOTALL,
+    )
+
+    with open(filepath, 'w') as fp:
+        fp.write(final_code)
+
+
+if __name__ == '__main__':
+    for d in entries:
+        fill_overload(**d)
```
JoanFM, line 117: is this going to align in the CI all the arguments into this `overload`?

hanxiao: no, one has to run this manually. and then `black -S` it

JoanFM: would it make sense to make it part of our ci?

PR #2599, opened 2021-06-09
refactor: change the name of the parameter

JoanFM:


bwanglzu (approved):
LGTM!

florian-hoenicke (approved):
Makes sense to use the executor name convention instead 👍

PR #2606, opened 2021-06-09
feat: daemon custom container

jacobowitz:
Draft PR for showcasing custom container creation on workspace creation.
With this PR jinad will start a container with workspace creation if a jinad file is used with a run command.

Missing in this PR:
- [ ] Lifecycle management (recreate on updates)
- [x] Stop container on delete (also add option to only stop the container and not delete workspace)
- [x] tests

### DIFF
--- a/daemon/api/endpoints/workspaces.py
+++ b/daemon/api/endpoints/workspaces.py
@@ -31,11 +31,18 @@ async def _clear_all():
     path='/{id}',
     summary='Deleting an existing Workspace',
 )
-async def _delete(id: DaemonID):
+async def _delete(id: DaemonID, container_only: bool = False):
     try:
-        store.delete(id=id, everything=True)
+        if container_only:
+            return store.delete_container_only(id=id)
+        else:
+            return store.delete(id=id)
     except KeyError:
         raise HTTPException(status_code=404, detail=f'{id} not found in {store!r}')
+    except ValueError:
+        raise HTTPException(
+            status_code=404, detail=f'There is no container to kill in {store!r}'
+        )
 
 
 @router.post(
--- a/daemon/api/endpoints/workspaces.py
+++ b/daemon/api/endpoints/workspaces.py
@@ -31,9 +31,21 @@ async def _clear_all():
     path='/{id}',
     summary='Deleting an existing Workspace',
 )
-async def _delete(id: DaemonID, container: bool = False):
+async def _delete(
+    id: DaemonID,
+    container: bool = False,
+    network: bool = False,
+    files: bool = False,
+    everything: bool = True,
+):
     try:
-        return store.delete(id=id, container=container)
+        return store.delete(
+            id=id,
+            container=container,
+            network=network,
+            files=files,
+            everything=everything,
+        )
     except KeyError:
         raise HTTPException(status_code=404, detail=f'{id} not found in {store!r}')
     except ValueError:
--- a/daemon/api/endpoints/workspaces.py
+++ b/daemon/api/endpoints/workspaces.py
@@ -31,12 +31,9 @@ async def _clear_all():
     path='/{id}',
     summary='Deleting an existing Workspace',
 )
-async def _delete(id: DaemonID, container_only: bool = False):
+async def _delete(id: DaemonID, container: bool = False):
     try:
-        if container_only:
-            return store.delete_container_only(id=id)
-        else:
-            return store.delete(id=id)
+        return store.delete(id=id, container=container)
     except KeyError:
         raise HTTPException(status_code=404, detail=f'{id} not found in {store!r}')
     except ValueError:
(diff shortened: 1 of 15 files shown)


daemon/stores/workspaces.py
```diff
@@ -60,3 +60,21 @@ def delete(self, id: DaemonID, **kwargs):
         Dockerizer.rm_network(id=self[id].metadata.network)
         del self[id]
         self._logger.success(f'{colored(str(id), "cyan")} is released from the store.')
+        return id
+
+    @BaseStore.dump
+    def delete_container_only(self, id: DaemonID):
```
deepankarm, line 66: I'd keep this logic under `delete` itself with additional args.

tests/distributed/test_against_external_daemon/test_single_instance.py
```diff
@@ -126,3 +129,24 @@ def test_create_pea_timeout(parallels):
     )
     with f:
         f.index(inputs=random_docs(10))
+
+
+def test_create_custom_container():
```
deepankarm, line 134: Let's move this to a different file?

tests/daemon/unit/models/good_ws_multiple_files/.jinad
```diff
@@ -1,3 +1,4 @@
 build = devel
 python = 3.7
 run = "echo Hello"
+ports = 12345,123456
```
deepankarm, line 4: could you add tests where
- `ports` is not passed
- `ports` has a value same as Flow's `port_expose` (there would be a port conflict then, we should probably have a different error for this).

daemon/dockerize.py
```diff
@@ -182,6 +195,7 @@ def run(
                 ports=ports,
                 detach=True,
                 command=command,
+                entrypoint=entrypoint,
```
deepankarm, line 198: Wouldn't `entrypoint=None` invalidate the default entrypoint for mini-jinad based containers?

daemon/api/endpoints/workspaces.py
```diff
@@ -31,11 +31,18 @@ async def _clear_all():
     path='/{id}',
     summary='Deleting an existing Workspace',
 )
-async def _delete(id: DaemonID):
+async def _delete(id: DaemonID, container_only: bool = False):
```
deepankarm, line 34: The list of supported args here can be `network`, `container`, `files`, `everything`. We'd call `store.delete` for all cases, which would validate before removing.

deepankarm (changes requested):

tests/daemon/unit/models/good_ws_multiple_files/.jinad
jacobowitz: I think empty is tested already as part of this parameterize:
`('good_ws_emptyfile', ('devel', '3.8', '', [])),`

The port conflict is a bit more complicated as this happens at two different times? The custom container gets started on workspace creation. The flow will be started later by a separat POST to /flows endpoint? And the same goes probably for all exposed ports between all entities in a workspace?

daemon/dockerize.py
jacobowitz: I think nothing will happen if that is None and the regular entrypoint will be used

daemon/api/endpoints/workspaces.py
jacobowitz: I added the parameters you mentioned, but files/network won't do anything for now. I think we should add this functionality in a separate PR?

container = True -> Delete container
everything = True -> Delete everything including workspace

daemon/stores/workspaces.py
```diff
@@ -53,10 +53,62 @@ def update(
...
+            Dockerizer.rm_network(id=self[id].metadata.network)
+            del self[id]
+            self._logger.success(
+                f'{colored(str(id), "cyan")} is released from the store.'
+            )
+            deleted_entities.append(id)
+        return deleted_entities
+
+    @BaseStore.dump
+    def delete_container_only(self, id: DaemonID):
```
deepankarm, line 100: Is this used anywhere?

daemon/stores/workspaces.py
```diff
@@ -53,10 +53,62 @@ def update(
...
+            Dockerizer.rm_container(container_id)
+            self._logger.success(
+                f'{colored(container_id, "cyan")} is killed and removed from the store.'
+            )
+            self[id].metadata.container_id = None
+            deleted_entities.append(container_id)
+
+        # TODO: add network and file deletion
+        if everything:
+            Dockerizer.rm_image(id=self[id].metadata.image_id)
```
deepankarm, line 90: Would need to check if container is deleted here.

tests/daemon/unit/models/good_ws_multiple_files/.jinad
deepankarm: yah. This is a bigger problem that needs to be handled in several places.

daemon/stores/workspaces.py
jacobowitz: The container gets deleted above? If we reach this point it should always have been deleted. In case of errors it should have thrown before. So I think it should be safe here to delete the image and the network (deleting the network we may actually want to move out of here though as it can be done separately now).

We need to extend the delete logic anyway as also file deletion is missing here and flows etc are also not stopped/deleted

PR #2609, opened 2021-06-10
refactor(hub): remove all hub cli related codes and tests

numb3r3:


florian-hoenicke:
@numb3r3 
Is there some test logic we could reuse?

numb3r3:
> Is there some test logic we could reuse?

@florian-hoenicke I will revert some test logic if necessary.

PR #2614, opened 2021-06-10
fix: graph iteration in case it has no edges

davidbp:
This PR Fixes iteration over graphs with no edges

### DIFF
--- a/tests/unit/types/document/test_graph_document.py
+++ b/tests/unit/types/document/test_graph_document.py
@@ -178,6 +178,8 @@ def test_from_dgl_graph(graph):
     assert (graph.adjacency.col == jina_graph.adjacency.col).all()
     assert (graph.adjacency.col == jina_graph.adjacency.col).all()
 
+
+def test_from_dgl_graph_without_edges():
     from dgl import DGLGraph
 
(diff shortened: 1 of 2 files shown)


JoanFM (commented):
Please add a test that shows it does not crash

JoanFM (changes requested):
tests to be added

davidbp:
> tests to be added

added a simple test to verify iteration is done without breaking on grah without edges

JoanFM (changes requested):
Add test to create `DGLGraph` from graph with no edges

tests/unit/types/document/test_graph_document.py
```diff
@@ -178,6 +178,13 @@ def test_from_dgl_graph(graph):
     assert (graph.adjacency.col == jina_graph.adjacency.col).all()
     assert (graph.adjacency.col == jina_graph.adjacency.col).all()
 
+    from dgl import DGLGraph
```
JoanFM, line 181: can this be an individual test with a meaningful test (test_convert_no_edges_graph) or something?

JoanFM (approved):

PR #2621, opened 2021-06-11
feat: add hub push cli

numb3r3:
Usage:

```bash
$ jina hub push [--private/--public] [--force] [--secret] [PATH]
```

Now, you can work with local hubble system via http://localhost:3001 (default host is: https://hubble.jina.ai (TBD)):
```bash
$ JINA_HUBBLE_REGISTRY=http://localhost:3001 jina hub push /path/to/executor --public
```

or you can run unit-test directly
```bash
$ pytest -v tests/unit/hubble
```


jina/hubble/helper.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def archive_package(package_folder: 'Path') -> 'io.BytesIO':
+    """
+    Archives the given folder in zip format and return a data stream.
+    :param package_folder: the folder path of the package
+    :return: the data stream of zip content
+    """
+    import os
+    import zipfile
```
hanxiao, line 14: standard python library can be imported at header, as they are usually very fast to load

jina/hubble/helper.py
```diff
@@ -0,0 +1,59 @@
...
+    ignored_file_specs = ['**/*.pyc']
+    ignored_dir_specs = [
+        '__MACOSX',
+        '.DS_Store',
+        '__pycache__',
+        '.eggs',
+        '*.egg-info',
+        '.git',
+        '.vscode',
+    ]
```
hanxiao, line 28: how about:
- put this file https://github.com/github/gitignore/blob/master/Python.gitignore under `jina/resources`
- load it via
```python
from jina import __resources_path__
with open(os.path.join(__resources_path__, 'Python.gitignore')) as fp:
   ignored_dir_specs = fp.readlines()
```

jina/hubble/hubio.py
```diff
@@ -0,0 +1,97 @@
...
+    def push(self) -> None:
+        """Push the executor pacakge to Jina Hub."""
+
+        import requests
+        import hashlib
+
+        is_public = True
+        if self.args.private:
+            is_public = False
+            self.args.public = False
```
hanxiao, line 44: this looks redundant, we only need `--public` we don't need `--private`,

when not given `--public` (by default), then `self.args.public` will be False, so it is private.

so default is `private`

jina/hubble/hubio.py
```diff
@@ -0,0 +1,97 @@
...
+                'meta': json.dumps(meta),
+                'env': json.dumps(env),
+                'is_public': is_public,
+                'md5sum': md5_digest,
+                'force': self.args.force,
+                'secret': self.args.secret,
+            }
+
+            # upload the archived executor to Jina Hub
+            upload_url = HUBBLE_REGISTRY + '/v1/executors/push'
```
hanxiao, line 77: put it in the header after `HUBBLE_REGISTRY` for better visibility and maintainbility

jina/hubble/hubio.py
```diff
@@ -0,0 +1,97 @@
...
+                    f'There is some errors while pushing executor "{self.args.path}"'
+                )
+
+        except Exception as e:  # IO related errors
+            self.logger.error(
+                f'Error when trying to push the executor at {self.args.path}: {e!r}'
+            )
+            raise e
+
+        self.logger.success(f'🎉 The executor located at {pkg_path} is now published!')
```
hanxiao, line 97: i think this success msg is wrongly placed as even an error will lead to this line.

jina/parsers/hubble/push.py
```diff
@@ -0,0 +1,41 @@
...
+The content source to be shipped into a Jina Hub executor. It can one of the followings:
+- a directory containing Dockerfile, manifest.yml, README.md, zero or more yaml config, zero or more Python file.
+''',
+    )
+    gp.add_argument(
+        '--private',
+        action='store_true',
+        default=False,
+        help='If set, the published executor is invisible to public',
+    )
```
hanxiao, line 23:
```suggestion

```

jina/parsers/hubble/push.py
```diff
@@ -0,0 +1,41 @@
...
+    gp.add_argument(
+        '--public',
+        action='store_true',
+        default=True,
```
hanxiao, line 27:
```suggestion
        default=False,
```

jina/parsers/hubble/push.py
```diff
@@ -0,0 +1,41 @@
...
+        help='If set, the published executor is visible to public',
+    )
+    gp.add_argument(
+        '--force',
+        type=str,
+        default='',
```
hanxiao, line 33:
```suggestion

```
no need

jina/parsers/hubble/push.py
```diff
@@ -0,0 +1,41 @@
...
+        help='To overwrite the executor identified as UUID8',
+    )
+    gp.add_argument(
+        '--secret',
+        type=str,
+        default='',
```
hanxiao, line 39:
```suggestion
```
no need

hanxiao (changes requested):
please change accordingly

jina/hubble/hubio.py
numb3r3: done

hanxiao (approved):

jina/hubble/hubio.py
mapleeit: @hanxiao I think we should keep the `--public` and `--private`. We may use them with the `--force` option.

We people do `jina hub push --public --force UUID8 --secret TOKEN .`. They may want to change the visibility of this executor. From private to public, or from public to private.

mapleeit: And why the default is `private`? As far as I can see, if we want more public shared executors in jina hub, public default value could be better.

PR #2624, opened 2021-06-11
fix: workspace defaults from metas

cristianmtr:
Workspace is set as a `metas`, not `runtime_args`

Required for https://github.com/jina-ai/indexers/pull/13


JoanFM (changes requested):
add some tests that show the problem maybe?

cristianmtr:
> add some tests that show the problem maybe?

Ah, nevermind. Not worth the trouble to rewrite all the tests. I will just do the same in my integration test. So from now on, the standard for setting workspace will be via `executors`

PR #2627, opened 2021-06-11
feat: undirected graph

davidbp:
Add user-friendly undirected graph support


jina/types/document/graph.py
```diff
@@ -195,6 +209,9 @@ def remove_edge(self, doc1: 'Document', doc2: 'Document'):
             if row.item() == offset1 and col.item() == offset2:
                 self._remove_edge_id(edge_id, f'{doc1.id}-{doc2.id}')
 
+            if self.undirected:
```
JoanFM, line 212: I think this is very dangerous. U would be removing the same edge twice?

JoanFM (changes requested):

jina/types/document/graph.py
davidbp: I think there are two things we can do (both of which are reasonable):


- we only work with halve the adjacency matrix and we convert `i->j` to `j->i` when needed 

- we consider every undirected edge `a-b`  is in bijection with `a->b` and ` b->a` and therefore, removing `a-b` would mean removing `a->b` and `b->a`.

JoanFM: I think that since there is no computational capacity in our side ( no operations are run in this Graph), I would maybe just store half of it, and flag it as undirected. Does this make sense?

davidbp: Solved in the last commit: yes we essentially store only one edge (ordered lexicographically).

tests/unit/types/document/test_graph_document.py
```diff
@@ -198,3 +198,57 @@ def test_edge_features_getter(graph):
...
+        == dgl_undirected_adj_coo.row
+    ).all()
+    assert (
+        np.concatenate((undirected_graph.adjacency.col, undirected_graph.adjacency.row))
+        == dgl_undirected_adj_coo.col
+    ).all()
+
+
+@pytest.mark.parametrize(
+    "graph, expected_output",
```
JoanFM, line 243: single quotes

JoanFM (commented):
We need more tests that show that `edges` are only added once for both `directed` and `undirected`

tests/unit/types/document/test_graph_document.py
```diff
@@ -198,3 +198,119 @@ def test_edge_features_getter(graph):
...
+    graph.add_edge(doc0, doc1)
+
+    assert graph.num_edges == expected_output
+
+
+@pytest.mark.parametrize(
+    'graph, expected_output',
+    [(GraphDocument(force_undirected=True), 1), (GraphDocument(), 1)],
+)
+def test_added_edges_in_edge_features(graph, expected_output):
```
JoanFM, line 291: assert the edge feature is None

JoanFM (changes requested):

JoanFM (approved):

hanxiao (dismissed):
please also fix the visualization for undirected graph, instead of using `-->`, use `---`

jina/types/document/graph.py
```diff
@@ -215,12 +250,21 @@ def edge_features(self):
     @property
     def adjacency(self):
```
hanxiao, line 251: type hint on return?

jina/types/document/graph.py
```diff
@@ -201,7 +235,8 @@ def remove_edge(self, doc1: 'Document', doc2: 'Document'):
             zip(self.adjacency.row, self.adjacency.col)
         ):
             if row.item() == offset1 and col.item() == offset2:
-                self._remove_edge_id(edge_id, f'{doc1.id}-{doc2.id}')
+                edge_key = self._get_edge_key(doc1, doc2)
+                self._remove_edge_id(edge_id, edge_key)
 
     @property
     def edge_features(self):
```
hanxiao, line 242: type hint on return?

PR #2629, opened 2021-06-11
fix: get workspace from either metas or runtime args

cristianmtr:


JoanFM (dismissed):
a test to clarify this behavior would be good

tests/unit/executors/test_executor.py
```diff
@@ -69,7 +69,37 @@ def test_metas_workspace_replica_peas(tmpdir, replica_id, pea_id):
...
     assert executor.workspace == os.path.abspath(
-        test_metas_workspace_simple['workspace']
+        os.path.join(
+            test_metas_workspace_simple['workspace'],
+            test_metas_workspace_simple['name'],
+        )
+    )
+
+
+def test_executor_workspace_simple_workspace(tmpdir):
```
cristianmtr, line 79: @JoanFM this ok?

JoanFM: what happens if metas workspace and runtime args workspace have different values?

cristianmtr: Added. Metas has priority

jina/executors/__init__.py
```diff
@@ -195,19 +195,18 @@ def workspace(self) -> str:
 
         :return: returns the workspace of the shard of this Executor.
         """
-        if getattr(self.runtime_args, 'workspace', None):
-            complete_workspace = os.path.join(
-                self.runtime_args.workspace, self.metas.name
-            )
+        workspace = getattr(self.metas, 'workspace') or getattr(
```
JoanFM, line 198:
```suggestion
        workspace = getattr(self.metas, 'workspace', or getattr(self.runtime_args, 'workspace', None))
```

JoanFM (approved):

jina/executors/__init__.py
cristianmtr: Good idea, but seems that it doesn't work.

The default workspace in metas is `''` which breaks the `getattr`

See https://github.com/jina-ai/jina/blob/c7e45999cc40fb220cea5fe3774d4156049ec67e/jina/executors/metas.py#L15

Not sure if this choice was intentional

PR #2630, opened 2021-06-11
feat: add multiple scores fields and pythonic evaluations

JoanFM:
**Changes introduced**
This PR deals with 2 things in the same way:

- It turns `scores` into a `repeated field` in the `DocumentProto`. This makes `scores` exactly the same as `evaluations`.
- To keep the look and feel of `score` property of `Document` I mantain it so that `doc.score` always points to the first value of the `scores` list.
- Then to insert different `values` to `scores` or `evaluations` provide the interface to insert by `name` (op_name).

To be able to extract these values via `dunder` syntax, added the consideration of `lists` into the `dunder_get` method

### DIFF
--- a/.github/2.0/cookbooks/Document.md
+++ b/.github/2.0/cookbooks/Document.md
@@ -536,8 +536,11 @@ d0.plot()  # simply `d0` on JupyterLab
 
 |     |     |
 | --- | --- |
-| `doc.score` | The relevance information of this Document |
+| `doc.scores` | The relevance information of this Document |
 | `doc.evaluations` | The evaluation information of this Document |
+
+Jina Document supports multiple relevance fields in the same instance to support, however for simple cases it also
+offers the interface to access these values as if there was a single relevance field.
 
 You can add a relevance score to a `Document` object via:
 
@@ -545,13 +548,13 @@ You can add a relevance score to a `Document` object via:
 from jina import Document
 
 d = Document()
-d.score.value = 0.96
+d.score = 0.96
 d.score.description = 'cosine similarity'
 d.score.op_name = 'cosine()'
 ```
 
 ```text
-<jina.types.document.Document id=0a986c50-aeff-11eb-84c1-1e008a366d48 score={'value': 0.96, 'opName': 'cosine()', 'description': 'cosine similarity'} at 6281686928>
+<jina.types.document.Document id=0a986c50-aeff-11eb-84c1-1e008a366d48 scores=[{'value': 0.96, 'opName': 'cosine()', 'description': 'cosine similarity'}] at 6281686928>
 ```
 
 Score information is often used jointly with `matches`. For example, you often see the indexer adding `matches` as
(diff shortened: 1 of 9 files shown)


jina/types/document/__init__.py
```diff
@@ -954,16 +997,122 @@ def score(
...
+        :param name: the operation name of the evaluation to be added
+        :param value: the value to set the evaluation of the Document from
+        """
+        idx = self._evaluations_offset_map.get(name, len(self.evaluations))
+        if idx == len(self.evaluations):
+            self._pb_body.evaluations.add()
+        self._set_evaluation(idx, value)
+        self._pb_body.evaluations[idx].op_name = name
+
+    def get_evaluation(self, name: str) -> Optional['NamedScore']:
```
bwanglzu, line 1104: why not set as property?

.github/2.0/cookbooks/Document.md
```diff
@@ -564,10 +567,30 @@ from jina import Document
...
+<jina.types.document.Document id=184049a9-cac1-11eb-8d87-e86a64801cb1 scores=[{'value': 0.96, 'op_name': 'score', 'ref_id': '184049a8-cac1-11eb-8d87-e86a64801cb1'}] adjacency=1 at 140519623882704>
+```
+
+You can also set `score` or `evaluation` with a `metric` name by calling:
+```python
+from jina import Document
+
+d = Document()
+d.set_score('cosine distance', 0.5)
+d.set_score('relevance', 10.0)
```
bwanglzu, line 584: why `cosine similarity` is represented as a score, not a field in tag?

bwanglzu (commented):
left some comments

.github/2.0/cookbooks/Document.md
```diff
@@ -800,6 +823,23 @@ doc.tags__dimensions__weight
...
 
+This can also be used to access nested attributes of repeated `Document` fields such as `scores` or `evaluations`.
+
+```python
+from jina import Document
+
+doc = Document()
+doc.scores.add()
+doc.scores[0].value = 50
+doc.scores.add()
```
bwanglzu, line 834: this `doc.scores.add()` looks very `protobuf`, the syntax is a bit strange in python

jina/types/document/__init__.py
JoanFM: There is a `property` to access the `first` element, which is kind of syntactic sugar. Let me explain better in the `description`

bwanglzu (commented):
i'm not sure about `doc.scores.add()` as well

.github/2.0/cookbooks/Document.md
JoanFM: Yes, I agree, but to explain the access of repeated fields is the best way I found

.github/2.0/cookbooks/Document.md
JoanFM: `score` is a `special` field in `Document`

bwanglzu:
if we decide to replace `score` with `scores`, I would suggest we do not maintain `score` as a property anymore. It's meaningless. We could maintain a `getter` of `scores`, no `setter`. Something like this:

```python
@property
def scores(self):
    return ... # List[NamedScoreTuple]
```

# we could refactor NamedScore a bit to allow user to assign score (like a view of dict)
```python
d = Document()
d.scores['cosine_similarity'] = 0.6
d.scores['ndcg'] = 2.4 # same for evaluation
print(d.scores['ndcg'] # getter instead of `get_score(attribuet_name)`
>>> 2.4
```

how do you think?

JoanFM:
> if we decide to replace `score` with `scores`, I would suggest we do not maintain `score` as a property anymore. It's meaningless. We could maintain a `getter` of `scores`, no `setter`. Something like this:
> 
> ```python
> @property
> def scores(self):
>     return ... # List[NamedScoreTuple]
> ```
> 
> we could refactor NamedScore a bit to allow user to assign score like a `dict`
> 
> ```python
> d = Document()
> d.scores['cosine_similarity'] = 0.6
> d.scores['ndcg'] = 2.4 # same for evaluation
> print(d.scores['ndcg'] # dict like access instead of `get_score(attribuet_name)`
> >>> 2.4
> ```
> 
> how do you think?

It makes a lot of sense, specially if this is done the first time. I am not sure is since people may be used to use `doc.score` as simple property what could be the best

PR #2634, opened 2021-06-12
refactor: response casting

JoanFM:


alanos95 (approved):

jina/types/request/__init__.py
```diff
@@ -230,10 +234,11 @@ def as_response(self):
...
         Return a weak reference of this object but as :class:`Response` object. It gives a more
         consistent semantics on the client.
+
+        :return: `self` as an instance of `Response`
         """
-        base_cls = self.__class__
-        base_cls_name = self.__class__.__name__
-        self.__class__ = type(base_cls_name, (base_cls, Response), {})
+        print(f' as_respons')
+        return Response(self)
```
maximilianwerk, line 241: You are not passing `envelope`.

What brings me to the question: Why is `envelope` anyhow a field of a request? It should only be part of a message. Strange :thinking:

jina/clients/base.py
```diff
@@ -161,7 +161,7 @@ async def _get_results(
                 with cm1 as p_bar, cm2:
                     async for resp in stub.Call(req_iter):
                         resp.as_typed_request(resp.request_type)
-                        resp.as_response()
+                        resp = resp.as_response()
```
maximilianwerk, line 164: This feels like the new `resp` should be a completely new object and not just a week reference. I smell problems here in the future. Anyhow, good for now.

maximilianwerk (changes requested):
Haven't checked, whether you changed all occurrences of the `as_response` call. If so: LGTM (apart from the print statement)

jina/types/request/__init__.py
JoanFM: I just realized, it is there to extract compression algorithm, so changed to pass `compression` algorithm instead

jina/clients/base.py
JoanFM: For performance reasons, all these `types` objects are views to the same `proto` object

maximilianwerk (approved):
That escalated quickly. If all tests pass: LGTM

hanxiao:
why this PR becomes so complicated? I think just `class Response(Request)` would be enough, if the idea is to allow IDE's intellisense

JoanFM:
> why this PR becomes so complicated? I think just `class Response(Request)` one line would be enough, if the idea is to allow IDE's intellisense

I am trying to avoid having so much magic but If I do not succeed soon I will drop it

JoanFM:
I surrender

PR #2636, opened 2021-06-13
feat: info loglevel rest

florian-hoenicke:
When creating custom REST endpoints, we get not any logs. Also, exceptions are silently ignored.

Having the logs helps debugging the application.
Example:

```
INFO:     127.0.0.1:51113 - "POST /search_docs HTTP/1.1" 200 OK
```

### DIFF
--- a/jina/peapods/runtimes/asyncio/rest/__init__.py
+++ b/jina/peapods/runtimes/asyncio/rest/__init__.py
@@ -51,6 +51,7 @@ class RESTRuntime(AsyncNewLoopRuntime):
         # as 'wsproto' is less performant and adds another dependency.
 
         from .....helper import extend_rest_interface
+        import os
 
         self._server = UviServer(
             config=Config(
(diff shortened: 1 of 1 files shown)


jina/peapods/runtimes/asyncio/rest/__init__.py
```diff
@@ -58,7 +58,7 @@ async def serve(self, sockets=None):
                 host=self.args.host,
                 port=self.args.port_expose,
                 ws='wsproto',
-                log_level='critical',
+                log_level='info',
```
hanxiao, line 61: let's make it configurable via `os.environ['JINA_LOG_LEVEL']`, which is the env that globally appreciated by Jina.
in general, we dont want to make our log too verbose. That gives new users a headache

hanxiao: `os.environ.get('JINA_LOG_LEVEL', 'critical')`

florian-hoenicke: thanks, changed.
I used `error` as the fallback as discussed.

PR #2639, opened 2021-06-13
feat: custom endpoints

florian-hoenicke:
Currently, it is possible to have custom endpoints by using the `extend_rest_interface` method.
However, it is not possible to setup endpoints which have custom logic but sending requests through the flow as well.
This implementation aims to solve this by providing `send_data` method in the `extend_rest_interface`.
It works now, but I have the feeling there must be a better way of doing it.

custom request and response model.
image

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -11,16 +11,15 @@ import time
 import uuid
 import warnings
 from argparse import ArgumentParser, Namespace
-from contextlib import contextmanager
 from datetime import datetime
 from itertools import islice
-from pathlib import Path
 from types import SimpleNamespace
 from typing import (
     Tuple,
     Optional,
     Iterator,
     Any,
+    Callable,
     Union,
     List,
     Dict,
@@ -1163,12 +1162,18 @@ def dunder_get(_dict: Any, key: str) -> Any:
 
 if False:
     from fastapi import FastAPI
+    from jina.peapods.runtimes.asyncio.rest.models import JinaRequestModel
+    from starlette.responses import StreamingResponse
 
 
-def extend_rest_interface(app: 'FastAPI') -> 'FastAPI':
+def extend_rest_interface(
+    app: 'FastAPI',
+    client: Optional[Callable[['JinaRequestModel', str], 'StreamingResponse']] = None,
+) -> 'FastAPI':
     """Extend Jina built-in FastAPI instance with customized APIs, routing, etc.
 
     :param app: the built-in FastAPI instance given by Jina
+    :param client: function to send requests to the Flow
     :return: the extended FastAPI instance
 
     .. highlight:: python
--- a/jina/parsers/peapods/runtimes/remote.py
+++ b/jina/parsers/peapods/runtimes/remote.py
@@ -74,6 +74,12 @@ def mixin_rest_server_parser(parser=None):
         help='If set, use this to specify float field valid digits.',
     )
 
+    gp.add_argument(
+        'use_default_endpoints',
+        default=True,
+        help='If set, use the default endpoints for index, update, delete, search',
+    )
+
 
 def mixin_grpc_server_parser(parser=None):
     """Add the options for gRPC
--- a/jina/peapods/runtimes/asyncio/rest/__init__.py
+++ b/jina/peapods/runtimes/asyncio/rest/__init__.py
@@ -1,4 +1,4 @@
-from .app import get_fastapi_app
+from .app import AppWrapper
 from ..base import AsyncNewLoopRuntime
 from .....importer import ImportExtensions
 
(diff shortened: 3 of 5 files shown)


jina/helper.py
```diff
@@ -1163,12 +1162,18 @@ def dunder_get(_dict: Any, key: str) -> Any:
 
 if False:
     from fastapi import FastAPI
+    from jina.peapods.runtimes.asyncio.rest.models import JinaRequestModel
```
JoanFM, line 1165: as per style, relative import

jina/helper.py
```diff
@@ -1163,12 +1162,18 @@ def dunder_get(_dict: Any, key: str) -> Any:
...
+    from starlette.responses import StreamingResponse
 
 
-def extend_rest_interface(app: 'FastAPI') -> 'FastAPI':
+def extend_rest_interface(
+    app: 'FastAPI',
+    client: Optional[Callable[['JinaRequestModel', str], 'StreamingResponse']] = None,
```
JoanFM, line 1171: I think `client` does not seem like an appropiate name.

Also the `type` hint seems like `missing` Unions?

jina/parsers/peapods/runtimes/remote.py
```diff
@@ -74,6 +74,12 @@ def mixin_rest_server_parser(parser=None):
         help='If set, use this to specify float field valid digits.',
     )
 
+    gp.add_argument(
+        'use_default_endpoints',
```
JoanFM, line 78: `action='store_true'`

jina/peapods/runtimes/asyncio/rest/__init__.py
```diff
@@ -53,9 +53,19 @@ async def serve(self, sockets=None):
         from .....helper import extend_rest_interface
         import os
 
+        app_wrapper = AppWrapper(self.args, self.logger)
+        fastapi_app = app_wrapper.get_fastapi_app()
+        print(
```
JoanFM, line 58: remove printline

JoanFM (changes requested):

jina/peapods/runtimes/asyncio/rest/__init__.py
```diff
@@ -53,9 +53,19 @@ async def serve(self, sockets=None):
...
+            '# extend_rest_interface.__code__.co_argcount',
+            extend_rest_interface.__code__.co_argcount,
+        )
+        if extend_rest_interface.__code__.co_argcount == 2:
+            fastapi_app = extend_rest_interface(fastapi_app, app_wrapper.send_request)
```
deepankarm, line 63: How do we generalize that the 2nd arg would always be `send_request`?

tests/unit/test_extend_fastapi.py
```diff
@@ -19,3 +24,46 @@ async def foo():
...
+    def map_request(request):
+        return JinaRequestModel(data=[request.custom_request_attribute])
+
+    async def map_response(search_resp):
+        async for x in search_resp:
+            yield json.dumps(
+                CustomResponseModel(
+                    custom_response_attribute=x.data.docs[0].text
+                ).dict()
+            )
```
deepankarm, line 45:
```suggestion
            yield CustomResponseModel(
                    custom_response_attribute=x.data.docs[0].text
                ).json()
```

deepankarm (commented):
Is the requirement here to enable `servicer.Call()` for any custom endpoint? The current way of adding arg to `extend_rest_interface` is not generalizable. We can use [Dependencies](https://fastapi.tiangolo.com/tutorial/dependencies/) here to execute `servicer.Call` and expose that along with the app.

hanxiao:
> However, it is not possible to setup endpoints which have custom logic but sending requests through the flow as well.

From your test code, do you mean you need an interface like the following:

```python
@app.post(
    path='/customized_endpoint',
    response_model=JinaRequestModel,
)
async def index_api(body: JinaIndexRequestModel):
    bd = body.dict()
    # < --- Your customized code here --- >
    return StreamingResponse(
        result_in_stream(request_generator(**bd)), media_type='application/json'
    )
```

florian-hoenicke:
> > However, it is not possible to setup endpoints which have custom logic but sending requests through the flow as well.
> 
> From your test code, do you mean you need an interface like the following:
> 
> ```python
> @app.post(
>     path='/customized_endpoint',
>     response_model=JinaRequestModel,
> )
> async def index_api(body: JinaIndexRequestModel):
>     bd = body.dict()
>     # < --- Your customized code here --- >
>     return StreamingResponse(
>         result_in_stream(request_generator(**bd)), media_type='application/json'
>     )
> ```

Almost.
We need to use our own pydantic models for the requests and the responses.
In the snippet you wrote, there is always a `JinaIndexRequestModel` used for the request and a `JinaRequestModel` returned.

jina/peapods/runtimes/asyncio/rest/__init__.py
florian-hoenicke: good question. I'm not so happy with the implementation I did there. It was just the only thing coming to my mind. I'm sure there must be a better solution.

florian-hoenicke:
This feature is now supported and this pr can be closed

PR #2645, opened 2021-06-14
fix: Response class as alias of Request class

hanxiao:
fix #2634 

image

### DIFF
--- a/jina/types/request/__init__.py
+++ b/jina/types/request/__init__.py
@@ -253,7 +253,7 @@ class Request(ProtoTypeMixin):
         self._pb_body.parameters.update(value)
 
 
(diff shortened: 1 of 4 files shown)


jina/types/request/__init__.py
```diff
@@ -253,7 +253,7 @@ def parameters(self, value: Dict):
         self._pb_body.parameters.update(value)
 
 
-class Response:
+class Response(Request):
```
JoanFM, line 256: I would add the Doc and GroundTruth mixin

hanxiao: yep, you can do them in another pr

PR #2648, opened 2021-06-14
feat: add scores and evaluations as maps

JoanFM:
**Changes introduced**
It turns `scores` and `evaluations` inside `DocumentProto` into a `Map-like` struct to have the following experience:

```python
from jina import Document
d = Document()
d.scores['cosine'] = 1
d.scores['euclidean'] = 0.6
d.evaluations['precision'] = 0.5
d.evaluations['recall'] = 0.7
```


bwanglzu:
one question, do we still need `NamedProto`?

JoanFM:
> one question, do we still need `NamedProto`?

maybe we could remove, but it still convenient and not damaging I think

jina/types/score/map.py
```diff
@@ -0,0 +1,102 @@
...
+from ..score import NamedScore
+from ..mixin import ProtoTypeMixin
+from ...excepts import BadNamedScoreType
+from ...helper import typename
+from ...proto import jina_pb2
+
+__all__ = ['MappedNamedScore']
+
+
+class MappedNamedScore(ProtoTypeMixin):
```
bwanglzu, line 14: suggestion, refine the name to `NameScoreMapping`? Python community often use this kind of name  such as `Mapping`, `MutableMapping` etc..

jina/types/score/map.py
```diff
@@ -0,0 +1,102 @@
...
+                raise ValueError(f'{typename(scores)} is not recognizable')
+        except Exception as ex:
+            raise BadNamedScoreType(
+                f'fail to construct a MappedNamedScore from {scores}'
+            ) from ex
+
+    def __setitem__(
+        self,
+        key: str,
+        value: Union[jina_pb2.NamedScoreProto, NamedScore, float, np.generic],
```
bwanglzu, line 69: i think it's safe to remove `NamedScore` unless we really want to keep the fields other than `op_name` and `value`

JoanFM: so far we want to keep them

bwanglzu (commented):
left some comments

bwanglzu:
and we might be able to use `collections.abc`, same as what we've used in `DocumentArray` (inherit from `MutableSequence`)

jina/types/score/map.py
bwanglzu: okay

jina/types/score/map.py
JoanFM: Good point!, Added

JoanFM:
> and we might be able to use `collections.abc`, same as what we've used in `DocumentArray` (inherit from `MutableSequence`)， maybe [MutableMapping](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping)

Added

bwanglzu (dismissed):
Looks great to me! let's see of other people have some ideas!

JoanFM:
Wait @bwanglzu I am gonna change the `cookbooks`

florian-hoenicke (approved):

hanxiao (approved):

PR #2657, opened 2021-06-15
feat(core): add wrapper for dump requests

jacobowitz:
This adds a small wrapper around the dump requests. It also adds a small test case to showcase its usage. It mainly depends on dump functionality provided by users in a custom executor.

We want to use this wrapper in jinad when exposing this via a REST api to the user.


hanxiao (changes requested):
tend to reject.
- flow should not implement any communication interface, they are all implemented in `jina/clients/mixin.py::PostMixin`/`AsyncPostMixin` and then carried to `Flow` class via mixin.
- this `dump` interface is very executor-specific, I don't support adding it to the core.

jacobowitz:
@hanxiao 
True the `dump` interface is very executor specific, only few will actually support it.
We want to expose `dump`/`rolling_update` semantics via `PUT` in **jinad** though. If we do this, there is no way around defining `dump` somewhere in code? Even though the logic itself is executor specific? It may be true though that it is better to only do this in jinad and not the actual core itself.

We have an existing case showcasing this logic:
https://github.com/jina-ai/jina/blob/master/tests/distributed/test_remote_flow_dump_rolling_update/test_dump_dbms_remote.py#L147

cristianmtr:
> tend to reject.
> 
> * flow should not implement any communication interface, they are all implemented in `jina/clients/mixin.py::PostMixin`/`AsyncPostMixin` and then carried to `Flow` class via mixin.
> * this `dump` interface is very executor-specific, I don't support adding it to the core.

It should be in the core, since the `rolling_update` is in the core and it implies the existence of a "dump" (via its `dump_path`). 

Besides, it is better if we reserve these keywords for methods ourselves, to have better clarity in documentation. Yes, users can implement any endpoints and call them whatever they like, instead of `dump`. But whenever we refer to it and present it, it will be these two phrases - `dump - rolling_update`

jacobowitz:
closing this one for now, lets discuss this again when we merge daemon-2.0

PR #2663, opened 2021-06-15
fix: set sparse content

JoanFM:


jina/types/document/__init__.py
```diff
@@ -71,10 +71,12 @@
         tf.SparseTensor,
     )
 
+
 __all__ = ['Document', 'DocumentContentType', 'DocumentSourceType']
 DIGEST_SIZE = 8
 
-DocumentContentType = TypeVar('DocumentContentType', bytes, str, np.ndarray)
+# This list is not exhaustive because we cannot add the `sparse` types without adding the `dependencies`
```
JoanFM, line 78:
```suggestion
# This list is not exhaustive because we cannot add the `sparse` types without adding the `dependencies`
```

JoanFM, line 78:
```suggestion
```

bwanglzu:
maybe add a unit test case as the screenshot?

JoanFM:
> maybe add a unit test case as the screenshot?

the test where the `parameter` is added already tests that

bwanglzu:
@JoanFM but the value you set is not a sparse matrix

JoanFM:
> @JoanFM but the value you set is not a sparse matrix

oh yes my bad I was confused. May u add a commit?

hanxiao (approved):

PR #2666, opened 2021-06-15
feat: add sparse example support

davidbp:
Add  in the cookbook examples that show how sparse data can be used as embeddings.


JoanFM (changes requested):
I would rather keep a new section with Sparse arrays for both embedding and blob.

I would keep this section for the regular dense array, so that the user is not overwhelmed

PR #2668, opened 2021-06-16
feat: check if jina is running in ci

hanxiao:


tests/unit/test_helper.py
```diff
@@ -310,3 +311,10 @@ def test_find_request_binding():
     assert r['index'] == 'bar'
     assert r['search'] == 'bar2'
     assert 'foo2' not in r.values()
+
+
+@pytest.mark.skipif(
+    'GITHUB_WORKFLOW' not in os.environ,
+)
+def test_ci_vendor():
+    assert get_ci_vendor() == 'GITHUB_ACTIONS'
```
maximilianwerk, line 320: As a consequence, we could change for all tests
```
@pytest.mark.skipif('GITHUB_WORKFLOW' not in os.environ)
```
to
```
@pytest.mark.skipif(get_ci_vendor() == 'GITHUB_ACTIONS')
```
? I like!

maximilianwerk (approved):
LGTM

PR #2673, opened 2021-06-16
fix: document creates content hash at init

cristianmtr:
Document should update its content hash at init time

This is especially important if the user employs the DocCache, which relies on the `content_hash`. If it's not automatically updated, the cache will not be able to verify the document properly. The user shouldn't need to manually call `.update_content_hash()`, as it's an internal aspect.


jina/types/document/__init__.py
```diff
@@ -287,6 +287,7 @@ def _update_doc(d: Dict):
             )
         self.set_attributes(**kwargs)
         self._mermaid_id = random_identity()  #: for mermaid visualize id
+        self.update_content_hash()
```
JoanFM, line 290: This is very dangerous, why if the `content_hash` is already computed and properly stored in the `proto`? Then u do not want to update

JoanFM (changes requested):

hanxiao (approved):

jina/types/document/__init__.py
cristianmtr: Handled by using a flag, that is set to False at iteration

JoanFM (approved):

PR #2675, opened 2021-06-16
refactor: refactor DocumentArray

JoanFM:
**Changes proposed**
Have `DocumentArray` be a `view` over a `List or RepeatedContainer of DocumentProro` instead of over an `iterable` of `Document`.

But from the user perspective return `Document` instead of `DocumentProto` when iterating or accessing elements in it.

**Benefits**
We had to revert #2648 because `DocumentArray` and `MatchArray` when instantiated from `query.matches` where behaving very differently because the internal `self._doc_views` can hold multiple types


JoanFM:
Makes no sense, `DocumentArrayProto` is not used

jina/types/arrays/document.py
```diff
@@ -98,24 +107,48 @@ class DocumentArray(
...
+                self._pb_body = docs._pb_body
+            elif isinstance(docs, list):
+                # This would happen in the client
+                for doc in docs:
+                    if isinstance(doc, Document):
+                        self._pb_body.append(doc.proto)
+                    elif isinstance(doc, jina_pb2.DocumentProto):
+                        self._pb_body.append(doc)
+                    else:
+                        raise ValueError(f'Unexpected element in an input list')
```
hanxiao, line 140: so this introduces extra-copying, we should then reduce the overall usage of this constructor

jina/types/arrays/document.py
```diff
@@ -98,24 +107,48 @@ class DocumentArray(
...
 
-    :param doc_views: A list of :class:`Document`
-    :type doc_views: Optional[Union['RepeatedContainer', Iterable['Document']]]
+    It is supposed to act as a view containing a pointer to a `RepeatedContainer` of `DocumentProto` while offering `Document` Jina native types
+    when getting items or iterating over it
+
+    :param docs: the document array to construct from. One can also give `DocumentArrayProto` directly, then depending on the ``copy``,
+                it builds a view or a copy from it. It also can accept a List
+    :param copy: when ``docs`` is given as a :class:`DocumentArrayProto` object, build a
+                view (i.e. weak reference) from it or a deep copy from it.
```
hanxiao, line 116:
```suggestion
```
i dont think this is implemented anywhere

hanxiao: if copy feature is not needed anywhere, then i suggest remove this docstring and ready to merge

JoanFM: oh yes, leftover

hanxiao (approved):

PR #2681, opened 2021-06-16
fix(daemon): fix/extend unit tests

jacobowitz:
fixing some unit tests and writing new tests for the partial stores/endpoints


tests/conftest.py
```diff
@@ -35,6 +39,37 @@ def fastapi_client():
...
+def get_partial_client(mode, monkeypatch):
+    monkeypatch.setattr(os, "kill", lambda *args, **kwargs: None)
+    from daemon import __root_workspace__
+
+    pathlib.Path(__root_workspace__).mkdir(parents=True, exist_ok=True)
+    from daemon import _get_app
+    from daemon.models.enums import PartialDaemonModes
+    from daemon import jinad_args
+
+    jinad_args.mode = PartialDaemonModes.get_mode(mode)
```
deepankarm, line 66: We don't need `get_mode`, I think. `PartialDaemonModes(mode)` would work too.

tests/daemon/unit/api/endpoints/partial/test_flow.py
```diff
@@ -0,0 +1,49 @@
...
+            'filename': f'{cur_dir}/../good_flow.yml',
+            'id': DaemonID('jflow'),
+            'port_expose': 1234,
+        },
+    )
+    assert response
+
+    response = partial_flow_client.get(api)
+    assert response
+    assert response.json()['arguments']['port_expose'] == 1234
```
deepankarm, line 23: Can we add a simple indexing here? using `Client`?

tests/conftest.py
```diff
@@ -35,6 +39,37 @@ def fastapi_client():
...
+    yield from get_partial_client(mode='flow', monkeypatch=monkeypatch)
+
+
+@pytest.fixture(scope='function', autouse=False)
+def partial_pod_client(monkeypatch):
+    yield from get_partial_client(mode='pod', monkeypatch=monkeypatch)
+
+
+@pytest.fixture(scope='function', autouse=False)
+def partial_pea_client(monkeypatch):
```
deepankarm, line 53: Do we need `monkeypatch` here?

deepankarm (changes requested):

PR #2690, opened 2021-06-17
feat: response casting

JoanFM:


JoanFM:
@hanxiao lazyness is mantained. I will think in another PR maybe how we could make sure of that for a test

hanxiao (approved):

PR #2694, opened 2021-06-17
fix: empty document array not none

florian-hoenicke:


jina/peapods/runtimes/zmq/zed.py
```diff
@@ -389,10 +389,12 @@ def _get_docs(self, field: str) -> 'DocumentArray':
...
             result = getattr(self.request, field)
 
-        # to unify all length=0 DocumentArray (or any other results) will simply considered as None
-        # otherwise the executor has to handle DocArray(0)
-        if len(result):
-            return result
+        # to unify all length=0 DocumentArray (or any other results) will simply considered as []
+        # otherwise the executor has to handle None values
+
+        if result is None:
```
JoanFM, line 395: I would do it on client side. it is good to have the option to have docs as None (as Dump) or whatever, and it is what someone would expect if None is in inpus

JoanFM (changes requested):

jina/peapods/runtimes/zmq/zed.py
JoanFM: Otherwise the condition of empty batch vs on purpose None inputs cannot be separated

florian-hoenicke: I what file would you put it?

JoanFM: Look for this function:

```python
U need to fix most likely this function:
 def _add_docs_groundtruths(req, batch, data_type, _kwargs):
     for content in batch:
         if isinstance(content, tuple) and len(content) == 2:
             # content comes in pair,  will take the first as the input and the second as the ground truth
 
             # note how data_type is cached
             d, data_type = _new_doc_from_data(content[0], data_type, **_kwargs)
             gt, _ = _new_doc_from_data(content[1], data_type, **_kwargs)
             req.docs.append(d)
             req.groundtruths.append(gt)
         else:
             d, data_type = _new_doc_from_data(content, data_type, **_kwargs)
             req.docs.append(d)
```

JoanFM: I think u need to consider when batch is empty

JoanFM: not 100% sure

florian-hoenicke: I think the logic of adding the groundtruth should be separated from `None` or [] decisions

Gamol123 (approved):

jina/peapods/runtimes/zmq/zed.py
JoanFM: Also it poses the problem that is cheating the user. This new DocArray p

Kelton8Z: maybe instruct users to do `bool(docArray)` to check if the document array is empty ?

nan-wang:
https://jinaai.slack.com/archives/C018F60RBL5/p1626335308267700

PR #2714, opened 2021-06-21
feat: pass executor specific params

JoanFM:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200498158412714) by [Unito](https://www.unito.io)


jina/peapods/runtimes/zmq/zed.py
```diff
@@ -171,6 +171,15 @@ def _post_hook(self, msg: 'Message') -> 'ZEDRuntime':
         msg.update_timestamp()
         return self
 
+    @staticmethod
+    def _parse_params(parameters: Dict, executor_name: str):
+        parsed_params = parameters
+        specific_parameters = parameters.get(executor_name, None)
+        if specific_parameters:
+            parsed_params.update(**specific_parameters)
+            del parsed_params[executor_name]
```
maximilianwerk, line 180: Why deleting?

maximilianwerk (commented):
I like it. At some point we have to think about what happens, if we call the same Executor several times in one Flow in different positions with potentially different arguments. That is what dynamic routing allows now.

jina/peapods/runtimes/zmq/zed.py
JoanFM: I did not have a clear view. I found it more clear to have a single source of truth, but can remove

JoanFM:
> I like it. At some point we have to think about what happens, if we call the same Executor several times in one Flow in different positions with potentially different arguments. That is what dynamic routing allows now.

It is true. Bur rhe normal usage will be that 2 different Flows use it, and in that case would still work

jina/peapods/runtimes/zmq/zed.py
JoanFM: Should I change this @maximilianwerk , it may be more conservative yes

maximilianwerk: Yes, please don't delete it.

JoanFM: Done!

tests/integration/override_executor_specific_params/test_specific_params.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    mock = mocker.Mock()
+    error_mock = mocker.Mock()
+
+    with f:
+        f.index(
+            inputs=DocumentArray([Document()]),
+            parameters={'param1': 50, 'param2': 60, 'exec_name': {'param1': 'changed'}},
+            on_done=mock,
```
cristianmtr, line 36: can u use `return_results=True`? it's easier to read

cristianmtr: and do the asserts on the returned valued.

```python
resp = f.index...
validate(resp)
```

cristianmtr (approved):

PR #2720, opened 2021-06-21
fix: have set_routing_graph at interface level

JoanFM:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200498812131103) by [Unito](https://www.unito.io)

### DIFF
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -271,7 +271,7 @@ class BasePod(ExitFIFO):
         .. # noqa: DAR101
         """
         raise NotImplementedError(
-            'Set routing graph should only be called for `GateWayPod`'
+            'Set routing graph should only be called for `Pod`'
         )
 
 
--- a/jina/peapods/pods/__init__.py
+++ b/jina/peapods/pods/__init__.py
@@ -265,6 +265,15 @@ class BasePod(ExitFIFO):
         """
         ...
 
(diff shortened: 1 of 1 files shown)


hanxiao (changes requested):
there is no "GatewayPod"

jina/peapods/pods/__init__.py
```diff
@@ -265,6 +265,15 @@ def is_singleton(self) -> bool:
         """
         ...
 
+    def set_routing_graph(self, *args, **kwargs):
+        """
+        .. # noqa: DAR102
+        .. # noqa: DAR101
+        """
+        raise NotImplementedError(
+            'Set routing graph should only be called for `GateWayPod`'
```
JoanFM, line 274:
```suggestion
            'Set routing graph should only be called for `Pod`'
```

hanxiao:
IDE complains of what?

hanxiao:
simply `self._pod_nodes = OrderedDict()  # type: Dict[str, Pod]`

JoanFM:
> IDE complains of what?

that BasePod does not have a set routing Graph method. Strictly speaking a CompoundPod instance could be part of the Flow and does not have it implemented.

hanxiao:
this function has been removed completely in https://github.com/jina-ai/jina/pull/2726 and move to inline

PR #2724, opened 2021-06-21
feat: add helper for getting parameter from req to executor

cristianmtr:
Helper method to retrieve either param or executor.metas.name.param from a parameters Dict in an endpoint

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200500170751991) by [Unito](https://www.unito.io)


JoanFM (commented):
This is not needed I would say  This is implicitly done in ZEDRuntime. So executor should only care about the root Dict

PR #2726, opened 2021-06-22
fix: make args.routing_graph serializable

hanxiao:
- make `.routing_graph` serializable via JSON str
- remove redundant message copy by using envelope
- replace the usage of `genent` with `socket` to resolve host name
- `RouteGraph` class refactoring, add `ProtoTypeMixin` to it
- rename `RouteGraph` to `RouteTable`

closes #2720


JoanFM (commented):
I think a way to fix the pickling issue is to delete from the Pod args once the socket types are set for the Peas

PR #2473, opened 2021-05-25
feat(daemon): containerized dependency management

deepankarm:
## JinaD 2.0

The primary goal of this item is to containerize every object creation under JinaD, thereby separating the environment for each object, allowing custom build & non-conflicting networks. [Usage of JinaD is documented here.](https://github.com/jina-ai/jina/blob/daemon-2.0/.github/2.0/cookbooks/Daemon.md)

Items achieved under this PR:

### Store:

- [x] Refactor Workspace Store & enable docker-build.
- [x] Define a Container Store & enable docker-run.
- [x] Enable (de)serialization with stores.
- [ ] Add thread locking to store object
- [ ] Add validation to the store metadata.
- [ ] Optionally enable building store from all local images/containers.

###### Runtime:

- [x] Enable default bash runtime with fluentd. 
- [x] Add build context for actual run & devel. 
- [x] Enable a different runtime with jinad jr (another fastapi layer) inside docker.
- [x] Setting Env vars during run

###### Network:

- [x] Create docker network during workspace creation.
- [x] Handle port-mapping during docker run ?? 

### Configuration:

- [x] Define a .jinad syntax (possibly ini) & create dockerfile from it.
- [x] requirements.txt
- [ ] setup commands
- [x] entrypoint command for projects

### Request Handling:

- [x] Build web-queue-worker architecture.
- [x] Enable endpoint for tracking requests

### Endpoints:

- [x] Ensure logging works inside containers.
- [x] Handle `DELETE` methods.
- [x] Add `Depends` to endpoints to handle params/validation etc.
- [ ] Support environment variables for all endpoints.
- [x] Separate POST vs PUT for all endpoints.
- [x] Enable endpoint for `/projects` ?? (tobi)
- [ ] Enable file reading endpoint inside workspaces. Store file metadata.
- [ ] Add a `/stats` endpoint to execute docker stats (Tobi)

### Housekeeping:

- [x] Add custom-type (DaemonID) for validation str+uuid

### Bugs:

- [x] DELETE flows during DELETE workspaces
- [x] Logging during docker run/build
- [x] Logging connection doesn't drop even when client has dropped
- [x] Change log endpoint to accept only id (fetch workspace_id from id if required)

##### Pending items:

- Currently, JinaD builds all images from `test-daemon` which is only available during development, Once this PR is merged, we'll add support to pulling from Docker Hub in a new PR. 
[...]

PR #2729, opened 2021-06-22
fix(daemon): fix control host setting for docker

jacobowitz:
Fix the control host setting for docker cases (0.0.0.0 should be host.docker.internal)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200504551981657) by [Unito](https://www.unito.io)


deepankarm:
This might have side effects, cannot be merged as-is. Let's work on #2473 directly.

PR #2730, opened 2021-06-22
feat: enforce trav paths being iterable of str

cristianmtr:
Consequence of discussion [here](https://jinaai.slack.com/archives/C0183GCN8AK/p1624358262177400)

We want to enforce `traversal_paths` always being `List[str]` and not `Union[List[str],str]`.

1. Easier to reason and catch
1. Easier to teach. New users don't need to understand both cases of traversing.
1. The type annotation is never actually checked, and it might fail silently, because you can traverse characters in a `str`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200504913520864) by [Unito](https://www.unito.io)

### DIFF
--- a/jina/types/arrays/traversable.py
+++ b/jina/types/arrays/traversable.py
@@ -2,6 +2,11 @@ import itertools
 from typing import Iterable
 
 
+def _check_traversal_path_type(traversal_paths):
+    if isinstance(traversal_paths, str):
+        raise ValueError('traversal_paths needs to be an Iterable[str]')
+
+
 class TraversableSequence:
     """
     A mixin used for traversing a `Sequence[Traversable]`.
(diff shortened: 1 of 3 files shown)


JoanFM (changes requested):
Is it updated in Document cookbook?

maximilianwerk (dismissed):

jina/types/arrays/traversable.py
```diff
@@ -2,6 +2,11 @@
 from typing import Iterable
 
 
+def _check_traversal_path_type(traversal_paths):
```
maximilianwerk, line 5: I prefer calling it `_assert_traversal_paths_type`. That is completely a matter of taste and I am good if you ignore this comment.

maximilianwerk (approved):
LGTM

JoanFM:
Ok! Added issue to update cookbook then #2731

JoanFM (approved):

PR #2734, opened 2021-06-22
fix(daemon): remove dump specific logic from flow

jacobowitz:
This removes the dump specific logic from the flow and jinad as we dont want to force the user into any specific dump mechanism.

Dump can still be issues manually directly via HTTP POST request like any other data request.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200506186791219) by [Unito](https://www.unito.io)


deepankarm:
@jacobowitz can you please rebase & fix the `test_remote_flow_dump_rolling_update` test? We can merge after thar

PR #2756, opened 2021-06-24
fix(hubble): put `metas` and `envs` in request header

numb3r3:
* fix 1: put `metas` and `envs` in request header
* fix 2: add request_id in request header
* fix 3: use local gitignore pattern

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200515138427558) by [Unito](https://www.unito.io)

### DIFF
--- a/jina/resources/Python.gitignore
+++ b/jina/resources/Python.gitignore
@@ -54,7 +54,7 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
-cover/
+#cover/
 
 # Translations
 *.mo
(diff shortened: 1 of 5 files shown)


jina/resources/Python.gitignore
```diff
@@ -78,7 +78,7 @@ docs/_build/
 
 # PyBuilder
 .pybuilder/
-target/
+# target/
```
hanxiao, line 81: i suggest we keep the original ones

hanxiao (approved):

jina/resources/Python.gitignore
numb3r3: I really think `target` directory is a general one. If we ignore this directory, it will cause some problems.

Issue #2500, opened 2021-05-28
Gateway information in routes list can be wrong


jacobowitz:
**Context**
As part of of the `Message` envelope we keep track of all peas touching a message as it passes through a `Flow`. This is done in the [routes field](https://github.com/jina-ai/jina/blob/c7e45999cc40fb220cea5fe3774d4156049ec67e/jina/types/message/__init__.py#L323) of the Envelope.
After a request has completed in the Gateway, those routes are copied to the response to the client. A simplified example of such a response with routes can be seen in the [Flow cookbook](https://github.com/jina-ai/jina/blob/c4a93181a0756d2b157b1dd888bf26b4245158da/.github/2.0/cookbooks/Flow.md#curl-with-http-request).

**Describe the bug**
The current code assumes that the combination of pod/pod_id are unique as they are used as keys in some intermediate processing dict:
[`routes = {(r.pod + r.pod_id): r for m in msgs for r in m.envelope.routes}`](https://github.com/jina-ai/jina/blob/c7e45999cc40fb220cea5fe3774d4156049ec67e/jina/types/message/__init__.py#L426)

Unfortunately this assumption is wrong for the Gateway pod because it can be part of a Flow multiple times when we have [intra parallel executions](https://github.com/jina-ai/jina/blob/master/.github/2.0/cookbooks/Flow.md#intra-parallelism-via-needs) (e.g. document and vector indexing parts of a Flow). In this case the processing can branch at the Gateway and one message is send to each parallel execution branch of the Flow. This means the Gateway will be added multiple times to the routes list, each time overwriting the previous entry.
Especially in a distributed setting the execution order of parallel branches of a Flow is not guaranteed.

Lets consider this simplified example:
```
Flow()
    .add(name='a1', needs='gateway')
    .add(name='a2')
    .add(name='b1', needs='gateway')
```
The execution order can happen like this (_tx_ are timestamps):
_client - gateway (t0) - a1 (t1) - a2  (t2) - gateway (t3) - b1 (t4) - gateway (t5) - client_

The correct route file would look like this
```
{'pod': 'gateway', 'pod_id': '37f4610e-7e81-4b08-8581-a6d7ec3ed915', 'start_time': 't0', 'end_time': 't5'}
{'pod': 'a1', 'pod_id': 'd6bd5907-2980-436c-bbd7-2cb9a6f0de23', 'start_time': 't1', 'end_time': 't1'}
[...]

PR #2760, opened 2021-06-24
fix(core): add complex test for routing list

jacobowitz:
This adds a test to verify that https://github.com/jina-ai/jina/issues/2500 got fixed as a side effect of dynamic routing (https://github.com/jina-ai/jina/pull/2551).
Not sure we need this test actually as part of the test suite, but I was writing it anyway to verify the behaviour and I think there are no other tests doing this.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200517178620445) by [Unito](https://www.unito.io)

### DIFF
--- a/tests/unit/flow/test_flow.py
+++ b/tests/unit/flow/test_flow.py
@@ -1,4 +1,6 @@
+import datetime
 import inspect
+import json
 import os
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


hanxiao:
actually i belive #2500 is fixed due to https://github.com/jina-ai/jina/pull/2750/files#diff-d9002011181954eeec0b9b94a9d08f7bc8028338275a35bea7823aa07d08041dL33

tests/unit/flow/test_flow.py
```diff
@@ -747,3 +749,71 @@ def test_flow_common_kwargs():
 def test_flow_set_asyncio_switch_post(is_async):
     f = Flow(asyncio=is_async)
     assert inspect.isasyncgenfunction(f.post) == is_async
+
+
```
bwanglzu, line 753: can you mark it as a slow test?
```suggestion
pytest.mark.slow
```

bwanglzu (commented):
can you benchmark these two test if slow mark it as `pytest.mark.slow`?

jacobowitz:
> can you benchmark these two test if slow mark it as `pytest.mark.slow`?

I checked its 3 seconds. Lets not mark it as slow as its rather nice to execute locally (no docker etc involved)

bwanglzu (approved):
LGTM!

Issue #2735, opened 2021-06-22
Possible bug when using the axis parameter from_ndarray document generator 


abellimz:
**Describe your proposal/problem**
Axis parameter in this `from_ndarray` function here does not seem to work as documented, and seems like it is not used in the function. 
https://github.com/jina-ai/jina/blob/010789331ce896ffcaea897402d698944f1a9d40/jina/types/document/generators.py#L15-L40

The axis parameter does not seemed to be used here when iterating through the array
https://github.com/jina-ai/jina/blob/010789331ce896ffcaea897402d698944f1a9d40/jina/types/document/generators.py#L36

similarly, when shuffle=true, axis parameter is not used to generate permutation
https://github.com/jina-ai/jina/blob/010789331ce896ffcaea897402d698944f1a9d40/jina/types/document/generators.py#L32-L34

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200506197672475) by [Unito](https://www.unito.io)

PR #2788, opened 2021-06-28
fix: issue #2735 unused axis argument in from_ndarray generator

valemore:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200533970296212) by [Unito](https://www.unito.io)

### DIFF
--- a/jina/types/document/generators.py
+++ b/jina/types/document/generators.py
@@ -31,9 +31,9 @@ def from_ndarray(
 
     if shuffle:
         # shuffle for random query
-        array = np.take(array, np.random.permutation(array.shape[0]), axis=axis)
+        array = np.take(array, np.random.permutation(array.shape[axis]), axis=axis)
     d = 0
-    for r in array:
+    for r in np.moveaxis(array, axis, 0):
         yield Document(content=r)
         d += 1
         if size is not None and d >= size:
--- a/tests/unit/test_generators.py
+++ b/tests/unit/test_generators.py
@@ -1,0 +1,17 @@
+import numpy as np
+
+from jina.types.document.generators import from_ndarray
+
+def test_from_ndarray():
+    test_array = np.random.random((10, 64))
+    # yield using from_ndarray along axis 0
+    test_generator = from_ndarray(test_array)
+    assert next(test_generator).blob.shape == np.array([64])
+
+    # yield using from_ndarray along axis 1
+    test_generator = from_ndarray(test_array, axis=1)
+    assert next(test_generator).blob.shape == np.array([10])
+
+    # yield using from_ndarray along axis 1, with shuffle
+    test_generator = from_ndarray(test_array, axis=1, shuffle=True)
+    assert next(test_generator).blob.shape == np.array([10])


maximilianwerk:
Dear @valemore,

thanks a lot for your contribution. could you add a test in order to showcase, what you are solving? Thanks a lot :)

valemore:
> Dear @valemore,
> 
> thanks a lot for your contribution. could you add a test in order to showcase, what you are solving? Thanks a lot :)

Hi @maximilianwerk, I added a unit test that previously would have failed because the axis argument to from_ndarray has no effect.

tests/unit/test_generators.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+
+from jina.types.document.generators import from_ndarray
+
+def test_from_ndarray():
+    test_array = np.random.random((10, 64))
+    # yield using from_ndarray along axis 0
+    test_generator = from_ndarray(test_array)
+    assert next(iter(test_generator)).blob.shape == np.array([64])
+
```
JoanFM, line 10: why is iter needed?

tests/unit/test_generators.py
```diff
@@ -0,0 +1,18 @@
...
+    # yield using from_ndarray along axis 1
+    test_array = test_array.T
+    test_generator = from_ndarray(test_array, axis=1)
```
JoanFM, line 13: shouldnt here the shape be 10?

JoanFM: I see now, the transpose. May u parametrize better the input and the expected shape? this way the test would be more readable?

tests/unit/test_generators.py
valemore: Not needed indeed, it's redundant :)

tests/unit/test_generators.py
valemore: yes, the transpose can be missed. Parametrized like this it's maybe more explicit.

valemore (commented):
updated as per @JoanFM 's feedback

nan-wang:
@valemore I'm afraid you need to rebase your codes to pass the CI. https://github.com/jina-ai/jina/pull/2788/checks?check_run_id=2988606249#step:4:9

```
git checkout master
git pull 
git checkout fix-generators-2735
git reset --soft <last_commit_your_branch_share_with_master>
git commit -m 'fix: fix the issue 2735'
git rebase master
git push -f origin fix-targepeapod-2910
```

jina/types/document/generators.py
```diff
@@ -31,9 +31,9 @@ def from_ndarray(
 
     if shuffle:
         # shuffle for random query
-        array = np.take(array, np.random.permutation(array.shape[0]), axis=axis)
+        array = np.take(array, np.random.permutation(array.shape[axis]), axis=axis)
```
bwanglzu, line 34: maybe it's even better to:
```suggestion
        array = np.random.shuffle(array) 
```

and parameterize the unit test

bwanglzu (changes requested):

jina/types/document/generators.py
bwanglzu: and remove the `axis`, i don't see why it's needed

bwanglzu:
AFA2C3B2-4FA0-47EC-BDAF-6292B094E2C2
if we decided to go with `np.take`, please ignore my comments above, but i found `shuffle` more efficient.

PR #2790, opened 2021-06-29
feat(hubble): ignore uuid and secret while push again

numb3r3:
Now, everytime the user want to **update** the executor, the `--force` and `--secret` options is required:

```bash
$ jina hub push <executor_folder> --force <UUID8> --secret <SECRET>
```

I don't think it is a good UX experience. I think, the `UUID8` and `SECRET` info can be tracked in a local file, so that the user can directly update executor in hubble without typing `UUID` repeatly.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200534404801610) by [Unito](https://www.unito.io)


hanxiao (changes requested):
- this does not considered multiple uuid will have different secret
- this PR should be targeted at `refactor-hub-cli`

numb3r3:
>  this does not considered multiple uuid will have different secret

@hanxiao  I don't understand this problem? Now, each folder only contains one executor, and so only one uuid should be considered.

hanxiao (approved):

PR #2793, opened 2021-06-29
fix: set cors for helloworld-chatbot

rumbarum:
I have tried simple hello-world chatbot command and this did not work. 
I found CORS was not set.

I also find CORS error on helloworld-mulit-modal. It works different from chat, I could not fix it. I think this also need set CORS true. (Is it possible to set cors on yml?)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200535300693002) by [Unito](https://www.unito.io)

### DIFF
--- a/jina/helloworld/chatbot/app.py
+++ b/jina/helloworld/chatbot/app.py
@@ -19,7 +19,7 @@ else:
 def _get_flow(args):
     """Ensure the same flow is used in hello world example and system test."""
     return (
-        Flow()
+        Flow(cors=True)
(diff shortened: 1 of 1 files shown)


bwanglzu:
hi @rumbarum yes!
if you want to set parameter in YAML:

```python
jtype: Flow
version: '1'
with:
  [PARAMETER-NAME]:  [PARAMETER-VALUE]
  workspace: . # for example, set `worskapce` to the current directory
```

Issue #2768, opened 2021-06-25
Persistent root workspace fails to store uploaded files in JinaD


jacobowitz:
JinaD uploads files nessary for a Flow to the workspace it creates. Workspaces/files are stored in the root workspace of JinaD.
If run inside Docker, this works if the root workspace is some temporary directory on the host like /tmp/jinad.
I tried to run JinaD with mounting some persistent folder as the root workspace and uploaded files are no longer stored correctly.

In my example I started JinaD like this:
`docker run --add-host=host.docker.internal:host-gateway --name jinad -v /var/run/docker.sock:/var/run/docker.sock -v /home/tobias/some_persistent_folder:/tmp/jinad --network=host -d jinaai/jina:test-daemon --store`

This makes the test test_upload_simple in test_remote_workspaces.py fail.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200521995795659) by [Unito](https://www.unito.io)

PR #2799, opened 2021-06-29
fix(daemon): fix flexible root workspace mount in docker

jacobowitz:
This fixes https://github.com/jina-ai/jina/issues/2768

It also fixes an issue with CI setting for jinad. We need to set some environment variable `JINA_DAEMON_BUILD` to `DEVEL` to make sure CI builds docker from the code in the PR and not from dockerhub release version. Before this was incorrect, because appending to the github env file only takes effect in the next step. We actually need it in the current step there.

JinaD now basically checks the host directory used for mounting its root workspace and will use the correct one when creating child containers (flow/peads/pods/.jinad). Before we were always mounting /tmp/jinad, which can be wrong if the user changes it.

I've added a test for this using docker compose, because I cant use the regular jinad instance as I need to control the volume mounting in other to test the correct thing here. This feels a bit akward there, but the test itself is okay and workling locally.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200536052886982) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,7 +236,7 @@ jobs:
         run: |
           if [[ "${{ matrix.test-path }}" =~ ^tests/distributed/* || "${{ matrix.test-path }}" =~ ^tests/daemon/* ]]; then
             # Build daemon for all daemon/distributed tests
-            echo "JINA_DAEMON_BUILD=DEVEL" >> $GITHUB_ENV
+            export JINA_DAEMON_BUILD=DEVEL
             docker build -f Dockerfiles/debianx.Dockerfile -t jinaai/jina:test-daemon .
             if [[ "${{ matrix.test-path }}" == "tests/distributed/test_against_external_daemon/" ]]; then
               docker run --add-host host.docker.internal:host-gateway --name jinad --env JINA_DAEMON_BUILD=DEVEL -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/jinad:/tmp/jinad -p 8000:8000 -d jinaai/jina:test-daemon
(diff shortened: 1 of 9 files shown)


deepankarm (approved):

.github/workflows/ci.yml
```diff
@@ -157,6 +157,8 @@ jobs:
   core-test:
     needs: [prep-testbed, commit-lint, lint-flake-8]
     runs-on: ubuntu-20.04
+    env:
```
deepankarm, line 160: Please remember to make this change in cd.yml too.

jacobowitz:
rebased against https://github.com/jina-ai/jina/pull/2824 to check for test_common ci test failure

jacobowitz:
rebased against master now as https://github.com/jina-ai/jina/pull/2824 got merged

jacobowitz:
I think tests are fine, the only failure is flaky  tests/integration/rolling_update/

deepankarm (changes requested):
Please remove `echo "JINA_DAEMON_BUILD=DEVEL" >> $GITHUB_ENV` from cd.yml:L246

deepankarm (approved):

JoanFM (approved):

hanxiao (approved):

soljarjb13 (commented):
?

soljarjb13 (commented):
?

soljarjb13 (commented):
?

soljarjb13 (commented):
?

soljarjb13 (commented):
?

PR #2806, opened 2021-06-29
fix(logs): info logs input, out, ctrl port

JoanFM:
**Changes introduced**
This log must be helpful for external pods, also it gives less the feeling of a blocked executor

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200537650891463) by [Unito](https://www.unito.io)

### DIFF
--- a/jina/peapods/zmq/__init__.py
+++ b/jina/peapods/zmq/__init__.py
@@ -195,7 +195,7 @@ class Zmqlet:
             else:
                 out_sock, out_addr = None, None
 
-            self.logger.debug(
+            self.logger.info(
                 f'input {colored(in_addr, "yellow")} ({self.args.socket_in.name}) '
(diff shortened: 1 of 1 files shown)


niuzs-alan (approved):
LGTM

niuzs-alan:
Some error happend in the CI.
I read the details, and I don't think the update of log level could cause these errors.
Let me re-run the test firstly.
And maybe we need to do some more research for the automatic CI tasks' schedule.

hanxiao (changes requested):
"ready and listening" msg serves the purpose already

JoanFM:
> "ready and listening" msg serves the purpose already

But if I want to have an external Pod. I may want to have a log to know its input port so that I can configure my other Flows

PR #2824, opened 2021-07-01
fix(core): prevent peas from hanging on close

jacobowitz:
We noticed that this [test](https://github.com/jina-ai/jina/blob/master/tests/daemon/unit/api/endpoints/test_common.py#L159) was failing a lot (but randomly) in [ci](https://github.com/jina-ai/jina/pull/2799/checks?check_run_id=2950566069)

I was running this test locally a lot and saw it less often, but it happened as well occasionally. The reason for the failure is that the flow closing hangs, which does so because the Pea closing hangs.
After discussion with @JoanFM we decided to remove the alive check the process as it seems to be unnecessary and was just added recently and may be the reason for the issue.
Another possible workaround was a `time.sleep() `after terminating a Pea, but we would rather try to avoid this.

I was running this changed version locally hundreds of time and did not encounter the issue anymore.

### DIFF
--- a/jina/peapods/peas/__init__.py
+++ b/jina/peapods/peas/__init__.py
@@ -115,8 +115,7 @@ class BasePea:
         :param args: extra positional arguments to pass to join
         :param kwargs: extra keyword arguments to pass to join
         """
-        if self.worker.is_alive():
-            self.worker.join(*args, **kwargs)
+        self.worker.join(*args, **kwargs)
 
     def terminate(self):
         """Terminate the Pea.


JoanFM (approved):

jacobowitz:
hangs again :( https://github.com/jina-ai/jina/pull/2824/checks?check_run_id=2960786051

Will add the sleep, but try again reproducing locally as well

jacobowitz:
> hangs again :( https://github.com/jina-ai/jina/pull/2824/checks?check_run_id=2960786051
> 
> Will add the sleep, but try again reproducing locally as well

actually we may use an old docker image here due to https://github.com/jina-ai/jina/pull/2799/commits/7978a39dd40074970c45895b3b51a8fb074f33d1

I will rebase the other pr and check what happens

jacobowitz:
> > hangs again :( https://github.com/jina-ai/jina/pull/2824/checks?check_run_id=2960786051
> > Will add the sleep, but try again reproducing locally as well
> 
> actually we may use an old docker image here due to [7978a39](https://github.com/jina-ai/jina/commit/7978a39dd40074970c45895b3b51a8fb074f33d1)
> 
> I will rebase the other pr and check what happens

passing: https://github.com/jina-ai/jina/pull/2799/checks?check_run_id=2960905711

jacobowitz:
this should be merged, despite the failing tests:
* test_common fails because we test against old jinad version due to ci bug (fixed in https://github.com/jina-ai/jina/pull/2799). I reran ci with the fixed version in the other PR and it fixed the bug we are fixing here.
* the rolling_update integration test failure is some other flaky test failing often in ci, not related to this change (we should fix that one as well though)

hanxiao (approved):

PR #2825, opened 2021-07-01
refactor: close from start error

JoanFM:


jacobowitz (approved):
definitely cleaner for the pod :+1:

PR #2829, opened 2021-07-01
changed tag for pip install, from standard to http

longwuyuan:
standard tag will work after a new release.

So maybe ok to let this drop as new release maybe coming shortly.
But as of the time of submitting this PR, the standard tag results in error. Details in this slack discussion

https://jina-ai.slack.com/archives/CTH1CMP8V/p1625148905234200


longwuyuan:
new release works with standard tag hence closing

PR #2833, opened 2021-07-02
refactor: refactor is ready

JoanFM:
**Changes introduced**
We are currently setting `is_ready` before the runtime is actually really ready to handle traffic. This PR gives the responsability to set `is_ready` flag to the `runtime` and tries to make sure that `is_ready` is more meaningful.

It also allows us to avoid the need to call `terminate` and have a potentially more grateful shutdown

### DIFF
--- a/jina/excepts.py
+++ b/jina/excepts.py
@@ -118,3 +118,7 @@ class SocketTypeError(Exception):
 
 class RoutingTableCyclicError(Exception):
     """Raised when the routing graph has cycles."""
+
+
+class RuntimeRunForeverEarlyError(Exception):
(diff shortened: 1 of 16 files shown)


jacobowitz (dismissed):
lgtm, we may add some wrapper later in the base runtime to hide the event implementation from the concrete runtimes, but that does not change too much.

jacobowitz (commented):
Actually there are some tests failing which look legit: https://github.com/jina-ai/jina/pull/2833/checks?check_run_id=2969742801

jacobowitz (approved):

soljarjb13 (commented):
b

PR #2836, opened 2021-07-02
added a kubernetes demo installer folder

longwuyuan:
I tried a demo of jina in k8s. Not sure even if it makes sense. I will understand if it gets rejected


longwuyuan:
moving this to cloud-ops repo

PR #2839, opened 2021-07-02
test: make unit tests success local

bwanglzu:
the purpose of this pr is to make all unit tests, able to successfully run locally. Also identify slow tests for future CI improvement.

expection: once user install jina with `pip install jina[cicd, test]`, he/she should be able to run all the unit tests successfully without any exceptions, same behavior as CI/CD:

```shell
pytest -v tests/unit/  --ignore tests/unit/peapods/runtimes/container/ --ignore tests/unit/peapods/runtimes/test_runtimes.py
```

Or run the `not slow` tests (fast run) with:

```shell
pytest -v tests/unit/ -m "not slow" --ignore tests/unit/peapods/runtimes/container/ --ignore tests/unit/peapods/runtimes/test_runtimes.py
```

Fast run is expected to finish in 3 minutes (slow in the beginning much faster after 35%), this allow dev's run tests locally without triggering CI to see the results.  Trade-off between dev's time and system resource usage.

This is what you get with fast run （828/1066, 78% tests covered with fast run in 3.5 minutes）:

0782EDC3-7CCB-44C1-BEFB-C8C7BE61EE34

Currently ignored `tests/unit/peapods/runtimes` since needs to build `test-pip` docker image.

Some test cases has been fixed:

1. Remove unused `JINA_LOG_LEVEL` environment variable in `test_load_balance` which impacted `test_logging.py`, if you run `test_logging` independently it works fails run all tests. Created a global autoused fixture `JINA_LOG_LEVEL` as debug in `conftest.py`. 
2. Renamed `MyExecutor` from `test_dynamic_routing`, another `MyExecutor` is being used while running all tests. (we have too many `MyExecutor` in tests, and might bring issue when didn't correctly specify scope)
3. Removed `yield_fixture` with `fixture`.
4. fixed import error in hubble test.
5. Removed super long log test `logger.info('x' * 65536)`, it's actually failing with `OSError`.
6. `test_topological_sorting` is not working properly, `topological_sorting` is getting random results overtime, this has been captured by running all tests.


JoanFM (dismissed):

tests/unit/test_loadbalance.py
```diff
@@ -6,8 +6,6 @@
 from jina.enums import SchedulerType
 from tests import random_docs
 
-os.environ['JINA_LOG_LEVEL'] = 'DEBUG'
```
JoanFM, line 8: you would need to change this back to info with a fixture.

However, I would like to have DEBUG logs when testing for the tests? what do u think? we could have a general fixture for that?

bwanglzu: agreed, done! @JoanFM

tests/unit/types/test_routing_graph.py
```diff
@@ -110,7 +110,8 @@ def test_topological_sorting():
     assert topological_sorting[2] in ['pod1', 'pod2', 'pod3']
     assert topological_sorting[3] in ['pod2', 'pod3']
     assert topological_sorting[4] == 'pod4'
-    assert topological_sorting == ['pod0', 'pod2', 'pod1', 'pod3', 'pod4']
+    # TODO FIX THIS, topological_sorting get random result each time!
+    # assert topological_sorting == ['pod0', 'pod2', 'pod1', 'pod3', 'pod4']
```
JoanFM, line 114: then assert the sets equality

JoanFM (changes requested):

tests/unit/types/test_routing_graph.py
bwanglzu: but we're testing the `sorting`, it should be ordered, right? I'd say let's make it as todo and debug in a seperate ticket

JoanFM: ok

JoanFM: now I see the test it is not coherent with the rest of the asseets it only makes sense if u assert the set

JoanFM (changes requested):

jacobowitz (approved):
great initiative! 

Minor comment: Maybe you could define some convenience entry points for running the tests as you described in your PR description? Otherwise people always have to look it up somewhere as its quite lengthy.

JoanFM (approved):

PR #2837, opened 2021-07-02
fix(tests): storm client test case

numb3r3:
Start flow
```bash
$ python python tests/storm/echo_flow.py [port] [--parallel] [--protocal] [--polling]
```

Then, run the storm test:
```bash
$ python tests/storm/storm_client.py [port] [--protocal] [--concurrent]
```

**Note**: all of the arguments have default value, so you can start the testing without providing any options.

**NEW UPDATES**
```
$ pytest  tests/integration/concurrent_clients/
```

PR #2843, opened 2021-07-02
fix(core): fix request/response missmatch while prefetching

jacobowitz:
This fixes the issue surfaced here: https://github.com/jina-ai/jina/pull/2837

I added a message buffer during prefetching, which is used for matching the incoming responses from the flow to the correct requests from the clients. The clients are still waiting for their messages to be received from the Flow, which is done by waiting for the futures in the message buffer.
As prefetching now is stateful, I removed the mixin design of the prefetcher.

I also added a very simple unit test. The test sends an unexpected response (simulating other clients running concurrently) before sending the expected response. In current master this test fails, as we would incorrectly return the unexpected message. With the new functionality this message is dropped and the correct message is returned later.


jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -16,6 +17,21 @@
 class PrefetchMixin(ABC):
     """JinaRPCServicer """
 
+    def __init__(self):
```
JoanFM, line 20: a mixin should not have `state` if possible. This should be some kind of abstract property

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -16,6 +17,21 @@
...
+        self._message_buffer: Dict[str, Future[Message]] = dict()
+        asyncio.create_task(self._receive())
```
JoanFM, line 22: I do think this `_receive` should be scheduled in the `fetch_to.append()` call

JoanFM (changes requested):

jina/peapods/runtimes/gateway/prefetch.py
jacobowitz: I removed the mixin now as its no longer possible actually to have this as a mixin I think. I adapted the class hierarchy slightly to reflect this.

jina/peapods/runtimes/gateway/prefetch.py
jacobowitz: We need to run the receive coroutine permanently in the background I think as we don't know when our responses will arrive.
One should probably also add some timeout mechanism to recover from disappearing messages, but this is not part of the PR at the moment (as it did not exist before as well).

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -1,22 +1,57 @@
...
+                        f'Discarding unexpected message with request id {message.request_id}'
+                    )
+        except asyncio.CancelledError:
+            pass
+
+    async def close(self):
+        """
+        Stop receiving messages
+        """
+        self._is_running = False
```
maximilianwerk, line 51: This is pretty un-forgiving. Should we give the Gateway some seconds after a close, to collect missing messages? Or do we say: You wanted to close it, so you should expect missing messages? (Which would be implemented in the `_receive` function.)

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -1,22 +1,57 @@
...
+        self._is_running = True
+        self._receive_task = asyncio.create_task(self._receive())
+
+    async def _receive(self):
+        try:
+            while self._is_running:
+                message = await self.zmqlet.recv_message(callback=lambda x: x.response)
+                if message.request_id in self._message_buffer:
+                    self._message_buffer[message.request_id].set_result(message)
+                    del self._message_buffer[message.request_id]
```
maximilianwerk, line 39: I think the more pythonic syntax would be:
```
future = self._message_buffer.pop(message.request_id)
future.set_result(message)
```

jina/peapods/runtimes/gateway/grpc/__init__.py
```diff
@@ -1,25 +1,24 @@
...
-from .....logging.logger import JinaLogger
 from .....proto import jina_pb2_grpc
 
 __all__ = ['GRPCRuntime']
 
+from .....types.message import Message
 
-class GRPCPrefetchCall(PrefetchMixin, jina_pb2_grpc.JinaRPCServicer):
+
+class GRPCPrefetchCall(PrefetchCaller, jina_pb2_grpc.JinaRPCServicer):
```
maximilianwerk, line 16: Why inheriting it and not wrapping it? I am OK with doing, but would like to see a good argument for it.

maximilianwerk: Ah I think I got the argument. We don't have to handle down the `close` call. Nevermind.

tests/unit/peapods/runtimes/gateway/test_prefetch.py
```diff
@@ -0,0 +1,49 @@
...
+    args = ArgNamespace.kwargs2namespace({}, set_gateway_parser())
+    mock_zmqlet = ZmqletMock()
+    servicer = PrefetchCaller(args, mock_zmqlet)
+
+    request = _generate_request()
+
+    response = servicer.send(iter([request]))
+
+    async for r in response:
+        assert r == request
```
maximilianwerk, line 47: Shouldn't we send a big bunch of requests for testing the beforehand failing behavior?

maximilianwerk (changes requested):
I would love to see the original test from @numb3r3 included in our CI/CD pipeline.

tests/unit/peapods/runtimes/gateway/test_prefetch.py
jacobowitz: I am simulating this behaviour with the mock here: https://github.com/jina-ai/jina/pull/2843/files#diff-4836b2c1886183887896c31a1d41c18e6e6a3f8f371e10ea5bd04d7563cf2c8bR31

This basically will respond to a request first with some random unrelated response and then with the correct one. This fails with current master.

Sending actually a lot of other messages is in the test a bit complicated and I would rather avoid in this unit test (I would need to create threads etc). This scenario is also rather well tested with the real end to end test of Felix.

jina/peapods/runtimes/gateway/grpc/__init__.py
```diff
@@ -42,16 +41,16 @@ async def async_setup(self):
...
+        self._prefetcher = GRPCPrefetchCall(self.args, self.zmqlet)
+        jina_pb2_grpc.add_JinaRPCServicer_to_server(self._prefetcher, self.server)
         bind_addr = f'{self.args.host}:{self.args.port_expose}'
         self.server.add_insecure_port(bind_addr)
         await self.server.start()
 
     async def async_cancel(self):
         """The async method to stop server."""
         await self.server.stop(0)
+        await self._prefetcher.servicer.close()
```
deepankarm, line 53:
```suggestion
        await self._prefetcher.close()
```
There's no `servicer` object in `_prefetcher` right?

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -46,16 +81,16 @@ async def prefetch_req(num_req, fetch_to):
                         raise TypeError(
                             f'{typename(request_iterator)} does not have `__anext__` or `__next__`'
                         )
+
+                    future = Future()
```
deepankarm, line 85:
```suggestion
                    future = get_or_reuse_loop().create_future()
```
[Official docs](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future) recommend using `loop.create_future()`. Not important right notw but it might be required as we use uvloop.

deepankarm (changes requested):

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -1,22 +1,57 @@
...
+        :param zmqlet: zeromq object
+        """
+        self.args = args
+        self.zmqlet = zmqlet
+        self.name = args.name or self.__class__.__name__
+
+        self.logger = JinaLogger(self.name, **vars(args))
+        self._message_buffer: Dict[str, Future[Message]] = dict()
+        self._is_running = True
+        self._receive_task = get_or_reuse_loop().create_task(self._receive())
```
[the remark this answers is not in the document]
jacobowitz: yes, some test was failing because there was no loop

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -1,22 +1,57 @@
...
+        :param args: args from CLI
+        :param zmqlet: zeromq object
+        """
+        self.args = args
+        self.zmqlet = zmqlet
+        self.name = args.name or self.__class__.__name__
+
+        self.logger = JinaLogger(self.name, **vars(args))
+        self._message_buffer: Dict[str, Future[Message]] = dict()
+        self._is_running = True
```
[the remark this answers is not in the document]
jacobowitz: You want to wait on the event in the receive loop instead of checking the bool? Is there an advantage in doing this?

jina/peapods/runtimes/gateway/prefetch.py
jacobowitz: I think we should not really wait for messages to come back? We cant really know for how long to wait and if you close this, you probably close the Flow and it should expect incomplete messages?

But we should probably at least cancel also the outstanding futures, which I didnt do

jina/peapods/runtimes/gateway/prefetch.py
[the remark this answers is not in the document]
deepankarm: Mainly cause `Events` should be used for synchronization. The receive loop spends most of its time awaiting on `recv_message`. When `close` sets `_is_running` to False, there's no assurance that receive can read that. The current logic works (primarily) because we `cancel` the `receive` task during close. 

I think we should use an Event for `is_running` or not use anything at all (use `task.cancel()` directly). Not sure of the best practice here.

JoanFM (changes requested):

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -1,22 +1,63 @@
...
 if False:
     from ...zmq import AsyncZmqlet
 
 
-class PrefetchMixin(ABC):
-    """JinaRPCServicer """
+class PrefetchCaller:
+    """An async zmq request sender to be used in the Gateway"""
+
+    def __init__(self, args: argparse.Namespace, zmqlet: 'AsyncZmqlet'):
```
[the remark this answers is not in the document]
jacobowitz: hmm but its semantically the same isnt it? The `GRPCPrefetchCall` is just having another name for the send method because we need to adhere to some RPC interface.
If we do composition we end up with more boilerplate as we also have to forward the close() call

jina/peapods/runtimes/gateway/prefetch.py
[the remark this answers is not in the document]
jacobowitz: ok sure, Joan mentioned the same about `is_running` and I agree that its not really needed, the canceling should always work.

jina/peapods/runtimes/gateway/prefetch.py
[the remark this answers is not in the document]
JoanFM: But what I mean is that in Python is kind of dangerous to have multiple inheritance because is not so explicit which parent u are initializing in super().__init__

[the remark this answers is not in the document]
jacobowitz: ok sure, I guess if I assign the methods directly its also just one additional line ;)

so now it looks like this

```
class GRPCPrefetchCall(jina_pb2_grpc.JinaRPCServicer):
    """JinaRPCServicer """

    def __init__(self, args, zmqlet):
        super().__init__()
        self._servicer = PrefetchCaller(args, zmqlet)
        self.Call = self._servicer.send
        self.close = self._servicer.close
```

jacobowitz:
I merged https://github.com/jina-ai/jina/pull/2837 into this branch.

I removed the manual app for testing (https://github.com/jina-ai/jina/pull/2837/files#diff-ec88489890bdade23d4a2d2704d1e6b317e0fdd93298de42a695448fa26f6f86) as I think it doesnt belong in the test folder.

The PyTest is still there: https://github.com/jina-ai/jina/pull/2843/files#diff-a1efda043579da8eecc8b8aa338e7358decb195883ef1c1804167f606c7c79a2

jacobowitz:
Storm test works now: https://github.com/jina-ai/jina/pull/2843/checks?check_run_id=2997162540
It failed before: https://github.com/jina-ai/jina/pull/2837/checks?check_run_id=2986555682

numb3r3:
> Storm test works now: https://github.com/jina-ai/jina/pull/2843/checks?check_run_id=2997162540
> It failed before: https://github.com/jina-ai/jina/pull/2837/checks?check_run_id=2986555682

@jacobowitz wired, some errors are not captured by `pytest`, the following is the error log
Screen Shot 2021-07-07 at 2 30 05 PM

or maybe, this error log is normal one we expected. I think you should double check them.

tests/integration/concurrent_clients/test_concurrent_clients.py
```diff
@@ -0,0 +1,52 @@
...
+
+@pytest.mark.parametrize('protocal', ['http', 'grpc'])
+@pytest.mark.parametrize('parallel', [10])
+@pytest.mark.parametrize('polling', ['ANY', 'ALL'])
+@pytest.mark.parametrize('prefetch', [1, 10])
+@pytest.mark.parametrize('concurrent', [15])
+def test_concurrent_clients(concurrent, protocal, parallel, polling, prefetch, reraise):
+    def pong(peer_hash, resp: Response):
+        for d in resp.docs:
+            with reraise:
```
JoanFM, line 25: the `reraise` context can be put on the line above

jacobowitz:
> > Storm test works now: https://github.com/jina-ai/jina/pull/2843/checks?check_run_id=2997162540
> > It failed before: https://github.com/jina-ai/jina/pull/2837/checks?check_run_id=2986555682
> 
> @jacobowitz wired, some errors are not captured by `pytest`, the following is the error log for `grpc` protocol
> Screen Shot 2021-07-07 at 2 30 05 PM
> 
> or maybe, this error log is normal one we expected. I think you should double check them.

@numb3r3 
I think this error is a multi threading issue of the GRPC client, maybe one could work around it in the test somehow but I dont think its worth it: https://github.com/grpc/grpc/issues/25364

JoanFM:
> > > Storm test works now: https://github.com/jina-ai/jina/pull/2843/checks?check_run_id=2997162540
> > > It failed before: https://github.com/jina-ai/jina/pull/2837/checks?check_run_id=2986555682
> > 
> > 
> > @jacobowitz wired, some errors are not captured by `pytest`, the following is the error log for `grpc` protocol
> > Screen Shot 2021-07-07 at 2 30 05 PM
> > or maybe, this error log is normal one we expected. I think you should double check them.
> 
> @numb3r3
> I think this error is a multi threading issue of the GRPC client, maybe one could work around it in the test somehow but I dont think its worth it: [grpc/grpc#25364](https://github.com/grpc/grpc/issues/25364)

It is indeed an upstream error seen in many tests. Would not pay attention in this context

JoanFM (approved):

maximilianwerk (approved):
LGTM. I wondered, whether the `PrefetchCaller` should be a parent class of the `GRPCPrefectchCall`, but I think I like the current implementation more.

PR #2847, opened 2021-07-04
fix: add readme snippet to CI

hanxiao:


jina/types/score/map.py
```diff
@@ -44,59 +42,44 @@ class NamedScoreMapping(ProtoTypeMixin, MutableMapping):
 
     def __init__(
         self,
-        scores: Optional[jina_pb2.NamedScoreMappingProto] = None,
-        copy: bool = False,
-        **kwargs,
+        scores,
```
JoanFM, line 45: what are the accepted types?

JoanFM (changes requested):

PR #2852, opened 2021-07-05
fix: avoid gateway creation using jinad

deepankarm:
A flow config like below tries to create a remote gateway, which is not supported by jinad. This PR fixes the if clause to avoid that. 

Pending: adding test.

```yaml
jtype: Flow
version: '1'
with:
  host: jina-gateway
  port_expose: 55500
pods:
  - host: jina-pod-1
    port_expose: 55500
```

### DIFF
--- a/jina/peapods/peas/__init__.py
+++ b/jina/peapods/peas/__init__.py
@@ -65,7 +65,7 @@ class BasePea:
         # arguments needed to create `runtime` and communicate with it in the `run` in the stack of the new process
         # or thread. Control address from Zmqlet has some randomness and therefore we need to make sure Pea knows
         # control address of runtime
-        self.runtime_cls, self._is_remote_controlled = self._get_runtime_cls()
+        self.runtime_cls = self._get_runtime_cls()
         self._timeout_ctrl = self.args.timeout_ctrl
         self._set_ctrl_adrr()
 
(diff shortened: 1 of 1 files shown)


jina/peapods/peas/__init__.py
```diff
@@ -311,13 +311,20 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
     def _get_runtime_cls(self) -> Tuple[Any, bool]:
-        is_remote_controlled = False
-        if self.args.host != __default_host__ and not self.args.disable_remote:
+        gateway_runtime_dict = {
+            GatewayProtocolType.GRPC: 'GRPCRuntime',
+            GatewayProtocolType.WEBSOCKET: 'WebSocketRuntime',
+            GatewayProtocolType.HTTP: 'HTTPRuntime',
+        }
```
maximilianwerk, line 318: This should be a constant outside of this function, right?

jina/peapods/peas/__init__.py
```diff
@@ -311,13 +311,20 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        if (
```
maximilianwerk, line 319: This is a highly non-obvious fix of the bug. Why does a simple
```
disable_remote = self.args.hasattr('disable_remote') and self.args.disable_remote
if self.args.host != __default_host__ and not disable_remote:
    ...
```
not work?

maximilianwerk (commented):
This fix is non-obvious from my point of view. But I am also not deep down in this code.

jina/peapods/peas/__init__.py
deepankarm: Because it is not just about `disable_remote`. The fact that, `disable_remote` is _not set only_ for gateway peas, lies in a different part of the code.  Even if that is checked properly, we should explicitly avoid setting `JinadRuntime` for gateway peas.

maximilianwerk (approved):
LGTM

PR #2857, opened 2021-07-05
fix: fix rolling update

JoanFM:
**Changes introduced**
The way the Pod and Pea closure is implemented leads to deadlocks when rolling update happens.

The closure of a `Dealer` should be handled by the `Pea` runtime as the only guarantee that no `data` request is locked causing a deadlock


jina/peapods/runtimes/zmq/zed.py
```diff
@@ -205,6 +206,7 @@ def _handle(self) -> 'ZEDRuntime':
             return self
 
         # migrated from the previously RouteDriver logic
+        self.logger.debug(f' I have the following idle dealers {self._idle_dealer_ids}')
```
JoanFM, line 209: to be removed

Issue #2854, opened 2021-07-05
cache invalidation in documentarray


bwanglzu:
`__contains__` logic in `DocumentArray` is heavily affected by `cached_property`. In this ticket, want to raise the concern about the issue. `cached_property` has affected everything, such as `insert`, `add`, `iadd`, `del`, `append`, `extend` and `clear`. This makes the `contains` logic fragile.

needs to add `@mutating` decorator and wrap those functions that would invalidate the cache

PR #2859, opened 2021-07-05
feat: cache invalidator

bwanglzu:
#2854  implement cache invalidation to deal with inconsistency in `DocumentArray`.


jina/helper.py
```diff
@@ -909,6 +909,48 @@ def __delete__(self, obj):
...
+        is the name to invalidate.
+    """
+
+    def __init__(self, func, func_to_invalidate: str):
+        self.func = func
+        self.func_to_invalidate = func_to_invalidate
+
+    def __call__(self, *args, **kwargs):
+        obj = args[0]
+        cached_key = f'CACHED_{self.func_to_invalidate}'
```
JoanFM, line 928: what is the impact of having this extra `CACHED` that would add an extra entry to the dict. I think this can be removed from cached_property as well as it was used previously in the `executor` dump and is not used anymore

jina/helper.py
```diff
@@ -909,6 +909,48 @@ def __delete__(self, obj):
...
+    """Class for cache invalidation, remove strategy.
+
+    :param func: func to wrap as a decorator.
+    :param func_to_invalidate: String as the function name to invalidate cached
+        data. E.g. in :class:`cached_property` we cache data inside the class obj
+        with the `key`: `CACHED_{func.__name__}`, the func name in `cached_property`
+        is the name to invalidate.
+    """
+
+    def __init__(self, func, func_to_invalidate: str):
```
JoanFM, line 922: func_to_invalidate or `property` or `attribute` right? I think `attribute` is more generic

JoanFM (changes requested):

jina/helper.py
bwanglzu: @JoanFM you mean renamed the variable?

jina/helper.py
bwanglzu: @JoanFM i didn't get your idea,  how to we cache if we remove `CACHE` from `cached_property`?

jina/helper.py
JoanFM: yes

jina/helper.py
JoanFM: do we need to add the CACHED prefix? why do not we use the original attribute?

bwanglzu: it is a design already exist in `class:cached_property`, so I did kept it consistent here

JoanFM: I know, I think is not needed there either

bwanglzu: okay clear @hanxiao shall we remove the `CACHED` prefix here, can you give some feedback?

maximilianwerk: You need some prefix, otherwise you might overwrite the function definition itself. you can try removing the prefix, but I believe it will not work. Can you test it?

bwanglzu: i did remove the `CACHED_` prefix and runed all primitive types unit test, it worked actually

jina/helper.py
```diff
@@ -909,6 +909,48 @@ def __delete__(self, obj):
             del obj.__dict__[f'CACHED_{self.func.__name__}']
 
 
+class _cache_invalidate(object):
```
hanxiao, line 912:
```suggestion
class _cache_invalidate:
```

JoanFM (approved):

PR #2864, opened 2021-07-06
fix(logging): change debug color to grey

jacobowitz:
This changes the debug color from black to grey. It says white here in the PR but the result looks actually grey for me.

see top right
Screenshot from 2021-07-06 11-52-48


bwanglzu (dismissed):
LGTM!

bwanglzu (approved):
LGTM!

PR #2865, opened 2021-07-06
feat: add request id to datareq log

JoanFM:
**Changes introduced**
Adding the request id helps in debugging networking issues

### DIFF
--- a/jina/peapods/runtimes/zmq/zed.py
+++ b/jina/peapods/runtimes/zmq/zed.py
@@ -138,7 +138,7 @@ class ZEDRuntime(ZMQRuntime):
 
         info_msg = f'recv {msg.envelope.request_type} '
         if self.request_type == 'DataRequest':
-            info_msg += f'({self.envelope.header.exec_endpoint}) '
+            info_msg += f'({self.envelope.header.exec_endpoint}) - ({self.envelope.request_id}) '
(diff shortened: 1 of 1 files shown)


jina/peapods/runtimes/zmq/zed.py
```diff
@@ -138,7 +138,7 @@ def _pre_hook(self, msg: 'Message') -> 'ZEDRuntime':
 
         info_msg = f'recv {msg.envelope.request_type} '
         if self.request_type == 'DataRequest':
-            info_msg += f'({self.envelope.header.exec_endpoint}) '
+            info_msg += f'({self.envelope.header.exec_endpoint}) - ({self.envelope.request_id}) '
```
jacobowitz, line 141: Building this `info_msg` seems rather expensive here? It looks ugly but it may be worth to check if `self.logger` would actually log the message and only construct it then? I think doing string operations on every data request for debugging purposes is not the best idea?

jacobowitz (commented):
Question regarding the debug log handling, not sure about it, may require some more thinking?

jina/peapods/runtimes/zmq/zed.py
JoanFM: that makes sense @jacobowitz, may u do a PR to fix this?

PR #2866, opened 2021-07-07
feat(hubble): add filelock while installing executor from hubble

numb3r3:
add a file lock to make the `pulling` thread-safe


JoanFM (dismissed):

nan-wang (approved):
LGTM👍

PR #2874, opened 2021-07-07
feat: enabled latency-tracking CI job on push & PR

maateen:
**Changes:**
- enabled latency-tracking CI job on push & PR

### DIFF
--- a/.github/workflows/latency-tracking.yml
+++ b/.github/workflows/latency-tracking.yml
@@ -1,13 +1,8 @@
-name: CI
+name: Latency Tracking
 
-# on:
-#  pull_request_target:
-# on:
-#   push:
-#     branches-ignore:
-#       - '**'  # temporally ignore all
 on:
-  workflow_dispatch:
+  - push
+  - pull_request
 
 jobs:
   latency-tracking:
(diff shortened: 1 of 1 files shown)


.github/workflows/latency-tracking.yml
```diff
@@ -16,8 +10,8 @@ jobs:
       NUM_LAST_RELEASE: 1
     steps:
       - uses: actions/checkout@v2
-#        with:
-#          submodules: true
+        with:
```
deepankarm, line 13: This should be commented out, as hub is not a submodule anymore.

.github/workflows/latency-tracking.yml
```diff
@@ -1,13 +1,7 @@
...
 
-# on:
-#  pull_request_target:
-# on:
-#   push:
-#     branches-ignore:
-#       - '**'  # temporally ignore all
 on:
-  workflow_dispatch:
+  - pull_request
```
deepankarm, line 4: I think it was `pull_request_target` as we wanted the ability to _update_ a comment. Not sure if that's still required.

deepankarm (changes requested):

.github/workflows/latency-tracking.yml
maateen: I think `pull_request` will do the job while `pull_request_target` might be [dangerous](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target) in some cases.

deepankarm (approved):

PR #2878, opened 2021-07-07
refactor: remove unused exception

bwanglzu:


JoanFM (commented):
Please close, I think some executors rely on this. Which may be wrong actually. But lets first make sure of this

PR #2881, opened 2021-07-08
fix(hubble): forcely ignore jina in requirements.txt 

numb3r3:
avoid to install `jina` when pull executor with `--install-requirements` option

### DIFF
--- a/jina/hubble/helper.py
+++ b/jina/hubble/helper.py
@@ -4,10 +4,13 @@ import hashlib
 import io
 import json
 import os
+import sys
+import subprocess
 import zipfile
+import pkg_resources
 from functools import lru_cache
 from pathlib import Path
-from typing import Tuple, Optional, Dict
+from typing import Tuple, Optional, Dict, List
 from urllib.parse import urlparse, urljoin
 from urllib.request import Request, urlopen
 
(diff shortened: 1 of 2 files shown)


jina/hubble/helper.py
```diff
@@ -250,3 +253,33 @@ def upload_file(
...
+    """
+
+    with requirements_file.open() as requirements_txt:
+        install_reqs = [
+            str(req)
+            for req in pkg_resources.parse_requirements(requirements_txt)
+            if req.project_name not in excludes or len(req.extras) > 0
+        ]
+
+    subprocess.check_call(
```
JoanFM, line 275: isn't there a `pip` python client?

nan-wang:
I suggest not providing this option. For the official images, we have control over this. If the user indeed wants to fix the Jina version, we should not keep them from doing this.

jina/hubble/helper.py
numb3r3: this is officially recommended way https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program 

Since pip v10, some api has been moved into different places.

CatStark:
Closing this because no longer needed

PR #2902, opened 2021-07-09
fix: wait for server to stop in cancel

JoanFM:


jina/peapods/runtimes/gateway/http/__init__.py
```diff
@@ -45,6 +51,14 @@ async def serve(self, sockets=None):
                 """
                 await self.main_loop()
                 await self.shutdown(sockets=sockets)
+                self._exit = True
+
+            async def stop(self):
```
jacobowitz, line 56:
```
 while not self._exit:
    await asyncio.sleep(0.1)
```

jina/peapods/runtimes/gateway/websocket/__init__.py
```diff
@@ -45,6 +51,14 @@ async def serve(self, sockets=None):
                 """
                 await self.main_loop()
                 await self.shutdown(sockets=sockets)
+                self._exit = True
+
+            async def stop(self):
+                while True:
```
jacobowitz, line 57: same as above

jacobowitz (changes requested):
just some minor suggestion for the while loop, I think its clearer with the short version?

jacobowitz (approved):

PR #2909, opened 2021-07-09
fix: only flush incoming

JoanFM:


jina/peapods/zmq/__init__.py
```diff
@@ -538,7 +538,7 @@ def close(self, flush: bool = True, *args, **kwargs):
             time.sleep(0.01)
             if flush:
                 for s in self.opened_socks:
-                    s.flush()
+                    s.flush(flag=zmq.POLLIN)
```
JoanFM, line 541: I do not like this, because what about the `cancel` message, maybe is not yet sent?

PR #2911, opened 2021-07-11
fix: fix #2910

### LINKED ISSUE #2910: Multi-path Flow breaks when one path skips the request and `parallel>1`
**Describe the bug**
When using `parallel>1`  and the `target_peapod` is not the one with multiple `parallel`, it keeps hanging.

```python
from jina import Document, Executor, Flow, requests
class UpExecutor(Executor):
    @requests
    def foo(self, **kwargs):
        print(f'hello, up')
class DownExecutor(Executor):
    @requests
    def foo(self, **kwargs):
        print(f'hello, down')
f = (Flow()
     .add(uses=UpExecutor, parallel=3, name='up')
     .add(uses=DownExecutor, needs='gateway', name='down')
     .join(needs=['up', 'down']))
with f:
    f.post(on='/foo', target_peapod='down', inputs=Document())
```

**Describe how you solve it**

---

**Environment**

```shell
- jina 2.0.6
- jina-proto 0.0.84
- jina-vcs-tag (unset)
- libzmq 4.3.4
- pyzmq 1.20.3
- protobuf 3.17.0
- proto-backend cpp
- grpcio 1.37.1
- pyyaml 5.4.1
- python 3.8.5
- platform Darwin
- platform-release 20.3.0
- platform-version Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021;
  root:xnu-7195.81.3~1/RELEASE_X86_64
- architecture x86_64
- processor i386
- uid 154505279077031
- session-id 55bd87ac-e185-11eb-b4a4-8c8590467aa7
- uptime 2021-07-10T21:47:14.809177
- ci-vendor (unset)
* JINA_ARRAY_QUANT (unset)
* JINA_CONTROL_PORT (unset)
* JINA_DEFAULT_HOST (unset)
* JINA_DISABLE_UVLOOP (unset)
* JINA_FULL_CLI (unset)
* JINA_HUBBLE_REGISTRY (unset)
* JINA_HUB_CACHE_DIR (unset)
* JINA_HUB_ROOT (unset)
* JINA_LOG_CONFIG (unset)
* JINA_LOG_ID (unset)
* JINA_LOG_LEVEL DEBUG
* JINA_LOG_NO_COLOR (unset)
* JINA_LOG_WORKSPACE (unset)
* JINA_OPTIMIZER_TRIAL_WORKSPACE(unset)
* JINA_POD_NAME (unset)
* JINA_RANDOM_PORT_MAX (unset)
* JINA_RANDOM_PORT_MIN (unset)
* JINA_VCS_VERSION (unset)
```

**Screenshots**

nan-wang:
- resolved #2910 
- fixed a bug in matching `target_peapod`

### DIFF
--- a/jina/peapods/runtimes/zmq/zed.py
+++ b/jina/peapods/runtimes/zmq/zed.py
@@ -204,14 +204,13 @@ class ZEDRuntime(ZMQRuntime):
             Handle does not handle explicitly message because it may wait for different messages when different parts are expected
         :return: ZEDRuntime procedure.
         """
-
-        if (
-            not re.match(self.envelope.header.target_peapod, self.name)
-            or self.request_type != 'DataRequest'
-        ):
+        # skip executor for non-DataRequest
+        if self.request_type != 'DataRequest':
+            self.logger.debug(f'skip executor: not data request')
             return self
 
         # migrated from the previously RouteDriver logic
+        # set dealer id
         if self._idle_dealer_ids:
             dealer_id = self._idle_dealer_ids.pop()
             self.envelope.receiver_id = dealer_id
@@ -219,11 +218,25 @@ class ZEDRuntime(ZMQRuntime):
             # when no available dealer, pause the pollin from upstream
             if not self._idle_dealer_ids:
                 self._zmqstreamlet.pause_pollin()
+            self.logger.debug(
+                f'using route, set receiver_id: {self.envelope.receiver_id}'
+            )
 
+        # skip executor if target_peapod mismatch
+        if not re.match(self.envelope.header.target_peapod, self.name):
+            self.logger.debug(
+                f'skip executor: mismatch target, target: {self.envelope.header.target_peapod}, name: {self.name}'
+            )
+            return self
+
+        # skip executor if endpoints mismatch
         if (
             self.envelope.header.exec_endpoint not in self._executor.requests
             and __default_endpoint__ not in self._executor.requests
         ):
+            self.logger.info(
+                f'skip executor: mismatch request, exec_endpoint: {self.envelope.header.exec_endpoint}, requests: {self._executor.requests}'
+            )
             return self
 
         params = self._parse_params(self.request.parameters, self._executor.metas.name)
--- a/jina/peapods/runtimes/zmq/zed.py
+++ b/jina/peapods/runtimes/zmq/zed.py
@@ -223,7 +223,10 @@ class ZEDRuntime(ZMQRuntime):
             )
 
         # skip executor if target_peapod mismatch
-        if not re.match(self.envelope.header.target_peapod, self.name):
+        if (
+            self.envelope.header.target_peapod
+            and self.envelope.header.target_peapod != self.name.split('/')[0]
+        ):
             self.logger.debug(
                 f'skip executor: mismatch target, target: {self.envelope.header.target_peapod}, name: {self.name}'
             )
--- a/jina/peapods/runtimes/zmq/zed.py
+++ b/jina/peapods/runtimes/zmq/zed.py
@@ -237,7 +237,7 @@ class ZEDRuntime(ZMQRuntime):
             self.envelope.header.exec_endpoint not in self._executor.requests
             and __default_endpoint__ not in self._executor.requests
         ):
-            self.logger.info(
+            self.logger.debug(
                 f'skip executor: mismatch request, exec_endpoint: {self.envelope.header.exec_endpoint}, requests: {self._executor.requests}'
             )
             return self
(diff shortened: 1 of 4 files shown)


jina/peapods/runtimes/zmq/zed.py
```diff
@@ -204,26 +204,39 @@ def _handle(self) -> 'ZEDRuntime':
...
 
             # when no available dealer, pause the pollin from upstream
             if not self._idle_dealer_ids:
                 self._zmqstreamlet.pause_pollin()
+            self.logger.debug(
+                f'using route, set receiver_id: {self.envelope.receiver_id}'
+            )
 
+        # skip executor if target_peapod mismatch
+        if not re.match(self.envelope.header.target_peapod, self.name):
```
JoanFM, line 226: equality check may be more efficient than regex match

JoanFM (changes requested):

jina/peapods/runtimes/zmq/zed.py
nan-wang: We only want to compare the beginning of the string. Regex seems slightly better.


```python
import random
import string
import re

pat = ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
foo_str = [pat + '/ZEDRuntime'] * 10000 + ['foo/ZEDRuntime'] * 10000

from datetime import datetime

beg = datetime.now()
for foo in foo_str:
    result = re.match(pat, foo)
end = datetime.now()
print(f're cost: {(beg-end).microseconds} microsecond')

beg = datetime.now()
for foo in foo_str:
    result = foo.startswith(pat)
end = datetime.now()
print(f'string cost: {(beg-end).microseconds} microsecond')

# re cost: 980502 microsecond
# string cost: 996321 microsecond
```

nan-wang: hmmm, using regex match actually introduces a potential bug. I've switched to the plain string matching

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -204,26 +204,42 @@ def _handle(self) -> 'ZEDRuntime':
...
         :return: ZEDRuntime procedure.
         """
-
-        if (
-            not re.match(self.envelope.header.target_peapod, self.name)
-            or self.request_type != 'DataRequest'
-        ):
+        # skip executor for non-DataRequest
+        if self.request_type != 'DataRequest':
+            self.logger.debug(f'skip executor: not data request')
```
hanxiao, line 209:
```suggestion
```

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -204,26 +204,42 @@ def _handle(self) -> 'ZEDRuntime':
...
             if not self._idle_dealer_ids:
                 self._zmqstreamlet.pause_pollin()
+            self.logger.debug(
+                f'using route, set receiver_id: {self.envelope.receiver_id}'
+            )
 
+        # skip executor if target_peapod mismatch
+        if (
+            self.envelope.header.target_peapod
+            and self.envelope.header.target_peapod != self.name.split('/')[0]
```
hanxiao, line 228: behavior change without reason and explain to user. please revert back to `re.match(self.envelope.header.target_peapod, self.name)`

jina/peapods/runtimes/zmq/zed.py
hanxiao: - where is "We only want to compare the beginning of the string" come from?
- please revert back to `re.match`

nan-wang: > where is "We only want to compare the beginning of the string" come from?
This is what the old codes have implemented. I've added a test to show the issue. Check it out at https://github.com/jina-ai/jina/blob/72f3a6db6bfac2d5ca7efbfb6d721fcbf493946d/tests/integration/v2_api/test_func_routing.py#L191

jina/peapods/runtimes/zmq/zed.py
nan-wang: After checking the [docstring](https://github.com/jina-ai/jina/blob/master/.github/2.0/cookbooks/Flow.md#function-signature), I see your point. However, I need to make it clear there will be an unexpected behavior, `target_peapod='foo'` will match both pods with the name `foo` and `foo_what_ever_follows`

jina/peapods/runtimes/zmq/zed.py
nan-wang: reverted

jina/peapods/runtimes/zmq/zed.py
nan-wang: I've added a warning message in the docstring at jina/clients/mixin.py

JoanFM (approved):

PR #2925, opened 2021-07-12
docs readme event banner

alexcg1:
@hanxiao isn't here so asking you to review Nan! This will go up each month, a week before our EAH. After that it'll be taken down until a week before the next EAH

- docs(readme): add eah banner
- chore: polish wording

### DIFF
--- a/.github/images/eah.svg
+++ b/.github/images/eah.svg
@@ -21,7 +21,7 @@
             </style>
             <p>
                 <a href="https://www.meetup.com/jina-community-meetup/events/279298872/">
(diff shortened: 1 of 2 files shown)


alexcg1:
I've confirmed with @BingHo1013 that we can move forwards with this

alexcg1:
I implemented this as css-styled HTML wrapped in an SVG, since GitHub natively doesn't allow CSS styling (or much of any other styling) in its markdown format.

nan-wang (approved):
LGTM👍

PR #2929, opened 2021-07-12
ci: adding rich to daemon reqs

deepankarm:


deepankarm:
@nan-wang This needs your approval as it is changing extra-req file.

nan-wang (approved):
LGTM👍

PR #2935, opened 2021-07-13
feat: support offline executors

alaeddine-13:
- default to disk cache when fetching hubble executor meta information if there's no internet connection
- skip pulling the docker image if there's no internet connection and the docker image was already pulled

### DIFF
--- a/jina/hubble/helper.py
+++ b/jina/hubble/helper.py
@@ -4,10 +4,11 @@ import hashlib
 import io
 import json
 import os
+import shelve
 import zipfile
-from functools import lru_cache
+from functools import lru_cache, wraps
 from pathlib import Path
-from typing import Tuple, Optional, Dict
+from typing import Tuple, Optional, Dict, Iterable
 from urllib.parse import urlparse, urljoin
 from urllib.request import Request, urlopen
 
@@ -250,3 +251,33 @@ def upload_file(
     response = getattr(requests, method)(url, data=data, headers=headers, stream=stream)
 
     return response
+
+
+def disk_cache(exceptions: Iterable[Exception], cache_file: str = 'disk_cache.db'):
+    """
+    Decorator which caches a function in disk and uses cache when an exception is raised
+
+    :param exceptions: exceptions used to trigger using the cache
+    :param cache_file: the cache file
+
+    :return: function decorator
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            call_hash = f'{func.__name__}({", ".join(args)})'
+            with shelve.open(cache_file) as cache_db:
+                try:
+                    result = func(*args, **kwargs)
+                    cache_db[call_hash] = result
+                except exceptions:
+                    if call_hash in cache_db:
+                        return cache_db[call_hash]
+                    else:
+                        raise
+            return result
+
+        return wrapper
+
+    return decorator
--- a/jina/hubble/helper.py
+++ b/jina/hubble/helper.py
@@ -5,6 +5,7 @@ import io
 import json
 import os
 import shelve
+import urllib
 import zipfile
 from functools import lru_cache, wraps
 from pathlib import Path
@@ -253,15 +254,13 @@ def upload_file(
     return response
 
 
-def disk_cache(
-    exceptions: Iterable[Exception],
+def disk_cache_offline(
     cache_file: str = 'disk_cache.db',
     message: str = 'Calling {func_name} failed, using cached results',
 ):
     """
-    Decorator which caches a function in disk and uses cache when an exception is raised
+    Decorator which caches a function in disk and uses cache when a urllib.error.URLError exception is raised
 
-    :param exceptions: exceptions used to trigger using the cache
     :param cache_file: the cache file
     :param message: the warning message shown when defaulting to cache. Use "{func_name}" if you want to print
         the function name
@@ -277,7 +276,7 @@ def disk_cache(
                 try:
                     result = func(*args, **kwargs)
                     cache_db[call_hash] = result
-                except exceptions:
+                except urllib.error.URLError:
                     if call_hash in cache_db:
                         default_logger.warning(message.format(func_name=func.__name__))
                         return cache_db[call_hash]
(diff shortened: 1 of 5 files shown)


tests/unit/hubble/test_helper.py
```diff
@@ -44,19 +44,31 @@ def test_unpack_package(tmpdir, dummy_zip_file):
...
 
 
-first_time = True
+first_time = False
 
 
-def test_disk_cache():
-    @disk_cache((Exception,))
+def test_disk_cache(tmpdir):
+    global first_time
```
JoanFM, line 51: what is `first_time` needed? Isn't there a better way? `globals` are scary

JoanFM (changes requested):

jina/hubble/helper.py
```diff
@@ -250,3 +251,40 @@ def upload_file(
     response = getattr(requests, method)(url, data=data, headers=headers, stream=stream)
 
     return response
+
+
+def disk_cache(
+    exceptions: Iterable[Exception],
+    cache_file: str = 'disk_cache.db',
+    message: str = "Calling {func_name} failed, using cached results",
```
JoanFM, line 259: single quote

jina/hubble/helper.py
```diff
@@ -250,3 +251,40 @@ def upload_file(
     response = getattr(requests, method)(url, data=data, headers=headers, stream=stream)
 
     return response
+
+
+def disk_cache(
```
JoanFM, line 256: Let's do this one more explicit and do not let add `Exceptions` as one wants.

Handling `exceptions` as normal behavior is a little dangerous and while it makes sense for this case, I would prefer to discourage it and specialize better this decorator

tests/unit/hubble/test_helper.py
```diff
@@ -41,3 +42,34 @@ def test_archive_package(tmpdir):
...
 def test_unpack_package(tmpdir, dummy_zip_file):
     helper.unpack_package(dummy_zip_file, tmpdir / 'dummp_executor')
+
+
+first_time = False
+
+
+def test_disk_cache(tmpdir):
+    global first_time
+    tmpfile = f"jina_test_{next(tempfile._get_candidate_names())}.db"
```
JoanFM, line 52: please single quote

tests/unit/hubble/test_helper.py
```diff
@@ -41,3 +42,34 @@ def test_archive_package(tmpdir):
...
+
+    class _Exception(Exception):
+        pass
+
+    @disk_cache((_Exception,), cache_file=str(tmpdir / tmpfile))
+    def _myfunc() -> bool:
+        global first_time
+        if not first_time:
+            raise _Exception("Failing")
```
JoanFM, line 61: single quote

tests/unit/hubble/test_helper.py
```diff
@@ -41,3 +42,34 @@ def test_archive_package(tmpdir):
...
+        else:
+            first_time = False
+            return True
+
+    # test fails
+    with pytest.raises(_Exception) as info:
+        _myfunc()
+    assert "Failing" in str(info.value)
```
JoanFM, line 69: single quote

tests/unit/hubble/test_helper.py
```diff
@@ -41,3 +42,34 @@ def test_archive_package(tmpdir):
 
 def test_unpack_package(tmpdir, dummy_zip_file):
     helper.unpack_package(dummy_zip_file, tmpdir / 'dummp_executor')
+
+
+first_time = False
```
JoanFM, line 47: avoid adding a `global` variable

JoanFM (changes requested):

tests/unit/hubble/test_helper.py
JoanFM: but u do not need a `global`. U can have a local variable inside the `test` and manipulate it there

alaeddine-13: Can't access it : UnboundLocalError: local variable 'first_time' referenced before assignment

alaeddine-13: never mind I know now why that shows up

jina/hubble/helper.py
alaeddine-13: So this means hardcoding the exception right ?
Should I rename the decorator ?

tests/unit/hubble/test_helper.py
alaeddine-13: done

jina/hubble/hubio.py
```diff
@@ -332,7 +341,16 @@ def pull(self) -> str:
...
-                    self._client.images.pull(executor.image_name)
+                    try:
+                        self._client.images.pull(executor.image_name)
+                    except docker.errors.APIError:
+                        img_not_found = False
+                        try:
+                            self._client.images.get(executor.image_name)
+                        except docker.errors.ImageNotFound:
+                            img_not_found = True
+                        if img_not_found:
```
JoanFM, line 352: this seems weird, why not just raising and avoid this extra boolean?

JoanFM (changes requested):

jina/hubble/hubio.py
alaeddine-13: Actually I need to re-raise APIError.
Maybe I can create a helper function `docker_image_exists(image_name)` because the same logic is also used in other places (dockerize.py, container/__init__.py).
I'm worried about the usage of docker (if I'll put it in a helper function) since docker is not installed in all tags.
What do you think ?

JoanFM: anyway, u can do except APIError as apiex, and then raise apiex. And u would not need this variable

JoanFM (changes requested):

jina/hubble/hubio.py
alaeddine-13: done

jina/hubble/helper.py
alaeddine-13: done

JoanFM (approved):

Issue #2930, opened 2021-07-12
Dont send responses for some control messages


jacobowitz:
At the moment we are sending response to all control and data requests, which is done in the msg_callback in the ZEDRuntime:
https://github.com/jina-ai/jina/blob/master/jina/peapods/runtimes/zmq/zed.py#L303

For some control requests this is wrong and potentially dangerous (at least IDLE/CLOSE) as the other side of the communication may never expect a response and could already be dead by the time we try to respond. This is one of the reasons for hanging CI at the moment when a head pea tries to respond to IDLE requests of its already closed peas.

We should write a test to capture the cases where we accept and dont accept responses to control requests.

Discovered with @JoanFM

PR #2940, opened 2021-07-13
fix(core): dont respond to cancel and idle

jacobowitz:
Skip sending responses to IDLE and CANCEL requests.

The test only tests IDLE unfortunately, but CANCEL actually impacts the tested pod so I could not use that one.

Fixes https://github.com/jina-ai/jina/issues/2930

### DIFF
--- a/tests/unit/peapods/peas/test_pea.py
+++ b/tests/unit/peapods/peas/test_pea.py
@@ -244,15 +244,24 @@ def test_gateway_args(protocol, expected):
 
 @pytest.mark.timeout(30)
 @pytest.mark.slow
-def test_idle_does_not_create_response():
+@pytest.mark.parametrize(
+    'command, response_expected',
+    [
+        ('IDLE', 0),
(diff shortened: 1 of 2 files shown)


JoanFM (changes requested):

tests/unit/peapods/peas/test_pea.py
```diff
@@ -238,3 +240,19 @@ def test_gateway_args(protocol, expected):
...
     assert p.runtime_cls.__name__ == expected
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.slow
+def test_idle_does_not_create_response():
+    args = set_pea_parser().parse_args([])
+
+    with Pea(args) as p:
+        msg = ControlMessage('IDLE', pod_name='fake_pod')
```
JoanFM, line 251: parametrize in control messages?

JoanFM (changes requested):

JoanFM (approved):

PR #2942, opened 2021-07-13
feat: add match function

davidbp:
Add matching function to Document array

The following script can be used to benchmark the updated distance computation
```python
import numpy as np
import matplotlib.pyplot as plt
import pandas as pd

def _get_ones(x, y):
    return np.ones((x, y))

def _ext_A(A):
    nA, dim = A.shape
    A_ext = _get_ones(nA, dim * 3)
    A_ext[:, dim : 2 * dim] = A
    A_ext[:, 2 * dim :] = A ** 2
    return A_ext

def _ext_B(B):
    nB, dim = B.shape
    B_ext = _get_ones(dim * 3, nB)
    B_ext[:dim] = (B ** 2).T
    B_ext[dim : 2 * dim] = -2.0 * B.T
    del B
    return B_ext

def _euclidean(A_ext, B_ext):
    sqdist = A_ext.dot(B_ext).clip(min=0)
    return np.sqrt(sqdist)

def _euclidean_squared(A_ext, B_ext):
    sqdist = A_ext.dot(B_ext).clip(min=0)
    return sqdist

def _norm(A):
    return A / np.linalg.norm(A, ord=2, axis=1, keepdims=True)

def _cosine(A_norm_ext, B_norm_ext):
    return A_norm_ext.dot(B_norm_ext).clip(min=0) / 2

def prev_euclidean(X,Y):
    X_ext = _ext_A(X)
    Y_ext = _ext_B(Y)
    return _euclidean(X_ext, Y_ext)

def prev_euclidean_squared(X,Y):
    X_ext = _ext_A(X)
    Y_ext = _ext_B(Y)
    return _euclidean_squared(X_ext, Y_ext)

def new_euclidean_squared(X, Y):
    dists = np.sum(Y**2, axis=1) + np.sum(X**2, axis=1)[:, np.newaxis] -2*np.dot(X, Y.T)
    return np.sqrt(dists)

n_X_list = [1, 8, 32, 128, 256]
n_Y_list = [100, 1000, 5000, 10000, 20000]
n_Feat = [100, 250, 500, 1000]
old_times = []
prev_times = []
nXYD = []

for n_features in n_Feat:
    for n_X in n_X_list:
        for n_Y in n_Y_list:
            X = np.random.random((n_X, n_features)).astype(np.float32)
            Y = np.random.random((n_Y, n_features)).astype(np.float32)
            nXYD.append((n_X, n_Y,n_features))
            
            new_time = %timeit -o new_euclidean_squared(X, Y)
            new_times.append(new_time)
            
            prev_time =  %timeit -o prev_euclidean_squared(X, Y)
            prev_times.append(prev_time)

H = [x.average for x in prev_times]
B = [x.average for x in new_times]

df = pd.DataFrame({'Old':H, 'New':B, 'nXnYnD':nXYD})

improvement_X = np.array(H)/np.array(B) 

######### plot 1 ###############
df = pd.DataFrame({ 'nXnYnD':nXYD, 'improvement_over_old':improvement_X})
df.set_index('nXnYnD').plot(kind='bar',figsize=(30,5))
#df.set_index('nXnYnD').plot(kind='barh', figsize=(5,30))
plt.ylabel('X times faster')
plt.xlabel('(Examples X, Examples Y, Embedding size)')
plt.title('Time improvement')
plt.tight_layout()
plt.savefig('results_x_times.pdf')

######## Plot 2 #############
df = pd.DataFrame({ 'nXnYnD':nXYD, 'improvement_over_old':improvement_X})
df.set_index('nXnYnD').plot(figsize=(10,5))
#df.set_index('nXnYnD').plot(kind='barh', figsize=(5,30))
plt.ylabel('X times faster')
plt.xlabel('(Examples X, Examples Y, Embedding size)')
plt.title('Time improvement')
plt.tight_layout()
plt.savefig('results_line.pdf')
```

Which should generate the plots

image

image


jina/types/arrays/document.py
```diff
@@ -3,6 +3,7 @@
 from abc import abstractmethod
 from collections.abc import MutableSequence, Iterable as Itr
 from contextlib import nullcontext
+import numpy as np
```
JoanFM, line 6: `numpy` import should come after the system imports

JoanFM (changes requested):

jina/types/arrays/document.py
```diff
@@ -315,20 +329,30 @@ def match(
...
-        for _q, _ids, _dists in zip(docs, idx, dist):
+        for _q, _ids, _dists in zip(self, idx, dist):
             for _id, _dist in zip(_ids, _dists):
-                d = Document(self.docs[int(_id)], copy=True)
-                d.scores['metric'] = _dist
+                d = Document(darray[int(_id)], copy=True)
+                if metric == 'cosine':
+                    d.scores[metric] = 1 - _dist
+                else:
+                    d.scores[metric] = _dist
```
JoanFM, line 355: parametrize to return `distance` or `similarity`. In case of `euclidean`, similarity can be considered with `1/1+dist`

JoanFM (changes requested):

jina/types/arrays/document.py
```diff
@@ -299,6 +300,77 @@ def overriden_key(proto):
...
+    def match(
+        self,
+        darray: 'DocumentArray',
+        metric: str = 'cosine',
+        limit: Optional[int] = None,
+    ) -> None:
+        """Compute embedding based nearest neighbour in `another` for each Document in `self`,
+        and store results in `matches`.
+
+        Note:
```
JoanFM, line 333: this needs to be parametrizable

jina/types/arrays/document.py
```diff
@@ -299,6 +300,77 @@ def overriden_key(proto):
...
+            - IF metric is 'euclidean' it uses the euclidean **distance**.
+
+        :param darray: the other DocumentArray to match against
+        :param metric: the distance metric
+        :param limit: the maximum number of matches, when not given
+                      all Documents in `another` are considered as matches
+        """
+        X = np.stack(self.get_attributes('embedding'))
+        Y = np.stack(darray.get_attributes('embedding'))
+        if limit is None:
```
JoanFM, line 344:
```suggestion
limit = limit or len(darray)
```

JoanFM (changes requested):

jina/types/arrays/document.py
davidbp: updated in a single line with ` limit = min(limit, len(darray))
`

jina/types/arrays/document.py
davidbp: done

jina/types/arrays/document.py
davidbp: good point. Done

jina/types/arrays/document.py
davidbp: done

jina/types/arrays/document.py
```diff
@@ -299,6 +302,113 @@ def overriden_key(proto):
         else:
             self._pb_body.sort(*args, **kwargs)
 
+    def compute_distances(
+        self, X: 'np.ndarray', Y: 'np.ndarray', metric: str
+    ) -> 'np.ndarray':
```
hanxiao, line 307: looks like a private & out-class method

davidbp: good point, I will leave it ouside the class

jina/types/arrays/document.py
```diff
@@ -21,6 +21,9 @@
 from ...helper import typename, cached_property, cache_invalidate
 from ...proto import jina_pb2
 
+import numpy as np
+from math import inf
```
JoanFM, line 25: math is a system dependency, so it goes with the first group.

We do:

```python
# system
import os
import typing
import math

#external
import numpy as np
import torch

#jina dependencies
import ..types.
```

jina/types/arrays/document.py
```diff
@@ -299,6 +302,85 @@ def overriden_key(proto):
...
+            if metric == 'cosine':
+                if is_distance:
+                    X = X
+                else:
+                    X = 1 - X
+            if metric == 'euclidean_squared' or metric == 'euclidean':
+                if is_distance:
+                    X = X
+                else:
+                    X = 1 / X
```
JoanFM, line 346: we need to avoid exceptions with `division by zero`
```suggestion
                    X =  (1 / X + 1)
```

JoanFM (changes requested):

jina/types/arrays/document.py
davidbp: ok

tests/unit/types/arrays/test_neuraloperations.py
```diff
@@ -0,0 +1,235 @@
...
+    d5_m = Document(embedding=np.array([0, 0, 3]))
+
+    D1 = DocumentArray([d1, d2, d3, d4])
+    D2 = DocumentArray([d1_m, d2_m, d3_m, d4_m, d5_m])
+    return D1, D2
+
+
+@pytest.fixture
+def docarrays_for_embedding_distance_computation():
+    return get_docarrays_for_embedding_distance_computation()
```
JoanFM, line 32: Why do not put this directly in the `fixture`?

JoanFM (changes requested):

tests/unit/types/arrays/test_neuraloperations.py
davidbp: I've refactored this as a fixture

jina/types/arrays/document.py
```diff
@@ -16,7 +16,9 @@
     TypeVar,
 )
 
+
 from .traversable import TraversableSequence
+from .neuraloperations import DocumentArrayNeuralOperationsMixin
```
hanxiao, line 21:
```suggestion
from .neural_ops import DocumentArrayNeuralOperationsMixin
```

jina/types/arrays/neuraloperations.py
```diff
@@ -0,0 +1,140 @@
...
+        :param top_k: number of values to retrieve
+        :return: Lowest values in dists and the indices of those values
+        """
+        if top_k >= dists.shape[1]:
+            indices = dists.argsort(axis=1)[:, :top_k]
+            dists = np.take_along_axis(dists, indices, axis=1)
+        else:
+            indices_ps = dists.argpartition(kth=top_k, axis=1)[:, :top_k]
+            dists = np.take_along_axis(dists, indices_ps, axis=1)
+            indices_fs = dists.argsort(axis=1)
```
tadejsv, line 88: Can you explain what is the point of this? You are doing partial sort first, and full sort later- which means you are loosing any performance gain that partial sort would bring. 

I would either just use the full sort, or re-write this to use partial sort only (if it's possible).

jina/types/arrays/neuraloperations.py
```diff
@@ -0,0 +1,140 @@
...
+            indices = np.take_along_axis(indices_ps, indices_fs, axis=1)
+            dists = np.take_along_axis(dists, indices_fs, axis=1)
+
+        return indices, dists
+
+
+def compute_distances(X: 'np.ndarray', Y: 'np.ndarray', metric: str) -> 'np.ndarray':
```
tadejsv, line 95: I agree with your comment (in slack) that we should have a `distances` module somewhere. We could also add some batching options there, to economize on the memory requirement

tests/unit/types/arrays/test_neuraloperations.py
```diff
@@ -0,0 +1,177 @@
...
+    return D1, D2
+
+
+@pytest.fixture
+def numpy_array():
+    return np.array([[1, 0, 0], [2, 0, 0], [3, 0, 0]])
+
+
+@pytest.fixture
+def numpy_array_query():
```
tadejsv, line 38: Same as above - so perhaps `embedding_query`?

tests/unit/types/arrays/test_neuraloperations.py
```diff
@@ -0,0 +1,177 @@
...
+    d4_m = Document(embedding=np.array([0, 0, 2]))
+    d5_m = Document(embedding=np.array([0, 0, 3]))
+
+    D1 = DocumentArray([d1, d2, d3, d4])
+    D2 = DocumentArray([d1_m, d2_m, d3_m, d4_m, d5_m])
+    return D1, D2
+
+
+@pytest.fixture
+def numpy_array():
```
tadejsv, line 33: Can you give this a more descriptive name - like `embeddings_keys` or something?

tests/unit/types/arrays/test_neuraloperations.py
```diff
@@ -0,0 +1,177 @@
...
+    return np.array([[1, 0, 0], [2, 0, 0], [3, 0, 0]])
+
+
+@pytest.fixture
+def numpy_array_query():
+    return np.array([[1, 0, 0]])
+
+
+def test_new_distances_equal_previous_distances():
```
tadejsv, line 42: How about we test this not against our own (previous?) home-cooked functions, but against stable libraries (`scipy` or `scikit-learn`)? Would it be possible to do that, and add them as a test-only dependency (if they are not already a base dpendency)?

jina/types/arrays/neuraloperations.py
tadejsv: Ah never mind, I see you change `dists`.

tests/unit/types/arrays/test_neuraloperations.py
davidbp: I already compared with scipy cdist library and the results are the same, from our own previous codebase and the new proposal.

jina/types/arrays/neuraloperations.py
davidbp: Agree with batching or more general methods (priority queue) for gathering the top_k results, but I think this optimizations might come in a later ticket.

tests/unit/types/arrays/test_neuraloperations.py
tadejsv: Looking at [here](https://github.com/jina-ai/jina/blob/master/extra-requirements.txt) it seems there is already a way to add a test-only dependency (the `cicd` tag), and scipy is in fact already there

jina/types/arrays/document.py
davidbp: updated

tests/unit/types/arrays/test_neuraloperations.py
davidbp: good point, updated to  `embedding_query`

tests/unit/types/arrays/test_neural_ops.py
```diff
@@ -0,0 +1,177 @@
...
+def embeddings():
+    return np.array([[1, 0, 0], [2, 0, 0], [3, 0, 0]])
+
+
+@pytest.fixture
+def embedding_query():
+    return np.array([[1, 0, 0]])
+
+
+def test_new_distances_equal_previous_distances():
```
tadejsv, line 42: I would still propose that this be replaced by comparison against [scipy.distance](https://docs.scipy.org/doc/scipy/reference/spatial.distance.html) module. The dependency is already installed at test time.

Right now it might make sense to compare against "previous distances", but by jina 3.0 nobody will know what these are and why we need to compare against them. However, your call.

tadejsv (approved):

tests/unit/types/arrays/test_neural_ops.py
davidbp: Will add the test right away and merge then

JoanFM (dismissed):

tadejsv (approved):
LGTM

JoanFM (approved):

PR #2948, opened 2021-07-14
feat(daemon): async interaction with children

### LINKED ISSUE #2746: Move from requests to aiohttp in JinaD
**Describe the bug**
All REST requests to mini-jinad currently are done using `requests` module., which blocks the event-loop on main-jinad. We should use `aiohttp` to send async requests.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200510406202377) by [Unito](https://www.unito.io)

deepankarm:
Fixes #2746


nan-wang:
The `dump_and_reload` test failed. https://github.com/jina-ai/jina/pull/2948/checks?check_run_id=3069763856#step:5:1115

deepankarm:
> The `dump_and_reload` test failed. https://github.com/jina-ai/jina/pull/2948/checks?check_run_id=3069763856#step:5:1115

Fixed now

daemon/__init__.py
```diff
@@ -160,17 +151,43 @@ def _start_consumer():
...
+    """Setup steps for JinaD"""
     _update_default_args()
     pathlib.Path(__root_workspace__).mkdir(parents=True, exist_ok=True)
     if not jinad_args.no_fluentd:
         Thread(target=_start_fluentd, daemon=True).start()
     _start_consumer()
     _start_uvicorn(app=_get_app(mode=jinad_args.mode))
+
+
+def teardown():
```
jacobowitz, line 176: Shouldn't we stop the consumer thread and uvicorn here?

daemon/stores/flows.py
```diff
@@ -1,92 +1,80 @@
...
+        :param kwargs: keyword args
+        :return: response from mini-jinad
+        """
 
-            if r.status_code != requests.codes.ok:
+        self._logger.debug(f'sending PUT request to mini-jinad on {uri}/{self._kind}')
+        async with aiohttp.request(
+            method='PUT', url=f'{uri}/{self._kind}', params=params
+        ) as response:
+            response_json = await response.json()
```
jacobowitz, line 54: What happens if there is some issue with the json deserialization? Same goes for other cases. Does this work then here or throw an exception?

tests/conftest.py
```diff
@@ -188,3 +189,14 @@ def test_timeout_ctrl_time(monkeypatch):
 def tmpfile(tmpdir):
     tmpfile = f'jina_test_{next(tempfile._get_candidate_names())}.db'
     return tmpdir / tmpfile
+
+
+@pytest.yield_fixture(scope="session")
+def event_loop(request):
```
jacobowitz, line 195: Cant we use the default loop fixture? If so (I think the scope is different), the comment below seems to be wrong as the scope here session and not 'for each test'

jacobowitz (changes requested):
only some minor comments/questions.

Otherwise looks good :)

daemon/__init__.py
deepankarm: Consumer thread is a daemon thread, so no need to close it manually.
Uvicorn handles its own `shutdown` logic in `serve` itself for different signal handlers.

daemon/stores/flows.py
deepankarm: That gets caught in `containers.py` exception handler and raised to the invoker. But our exception sending logic to the user is poor now. It needs quite some work and cannot be done in this PR

tests/conftest.py
deepankarm: You're right. The comment is wrong. It needs to be a session, else we need to handle event loop (start and closure) logic in aiohttp client session, which is not required in the actual code (as uvicorn takes care of it), and needed only in tests. 

I had to spent >2hrs just to fix this :)

jacobowitz (dismissed):
lgtm :+1:

jacobowitz (approved):

PR #2949, opened 2021-07-14
chore remove event banner

alexcg1:
Commenting out event banner until next time we need it

- chore: polish wording
- chore: comment out event banner til next time
- Revert "chore: comment out event banner til next time"
- chore: comment out event banner til next time


nan-wang (approved):
LGTM👍

PR #2953, opened 2021-07-15
fix(importer): add custom sys meta path finder

numb3r3:
This PR aims at fixing the (internal) extra dependency issues in our hub . 

For instance, base on the executor from  https://github.com/jina-ai/executor-audio-VGGishEncoder/tree/fix-module-not-found , i found some weird observations. 
```
from jina import Executor, Flow
from jina.executors import BaseExecutor
import threading

# # => case1: works
# from vggish_audio_encoder import VggishAudioEncoder
# encoder = VggishAudioEncoder()

# # => case2: works
# def run():
#     encoder = BaseExecutor.load_config('config.yml')
# threading.Thread(target=run).start()

# # => case3: works
encoder = BaseExecutor.load_config('config.yml')

# # => case4: works
# f = Flow().add(uses='config.yml')
# with f:
#     f.block()
```
All of the use cases above work perfectly. However, if we want to use `jina executor/pod/pea`  cli to load executor, we will get errors:
```
jina executor --uses config.yml   # => Failed, throw exception: `vggish` module cannot be found
```

### DIFF
--- a/jina/importer.py
+++ b/jina/importer.py
@@ -3,6 +3,7 @@ import sys
 import warnings
 from types import SimpleNamespace, ModuleType
 from typing import Optional
+from importlib.abc import MetaPathFinder, Loader
 
 from . import __resources_path__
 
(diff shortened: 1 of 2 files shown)


jina/importer.py
```diff
@@ -95,6 +96,57 @@ def __exit__(self, exc_type, exc_val, traceback):
...
+        load_path = module.__spec__.origin
+        init_file = '__init__.py'
+        if load_path.endswith(init_file):
+            module.__path__ = [load_path[: -len(init_file) - 1]]
+
+
+_loader = ModuleLoader()
+
+
+class PathFinder:
```
JoanFM, line 117: may be better to rename this to JinaExecPathFinder, according to https://pyoxidizer.readthedocs.io/en/stable/oxidized_importer_meta_path_finders.html there is a PathFinder class involved and could become confusing?

numb3r3: thanks for your suggestions. I have renamed it as `ExtensionPathFinder`

jina/importer.py
```diff
@@ -95,6 +96,57 @@ def __exit__(self, exc_type, exc_val, traceback):
...
+    @classmethod
+    def exec_module(cls, module):
+        """Initialize an extension module
+
+        :param module: a namespace module
+        """
+        load_path = module.__spec__.origin
+        init_file = '__init__.py'
+        if load_path.endswith(init_file):
+            module.__path__ = [load_path[: -len(init_file) - 1]]
```
winstonww, line 111: For my understanding, what's the use case for setting the `module.__path__` variable?

winstonww (approved):
LGTM! Just a quick question for my understanding.

Issue #2944, opened 2021-07-14
cannot get embedding from reloaded DocumentArrayMemmap 


numb3r3:
**Describe the bug**

You can reproduce the error with the following sample scripts
```python
from jina import Document
from jina.types.arrays.memmap import DocumentArrayMemmap
import numpy as np

def random_docs(num_docs):
    for j in range(num_docs):
        doc_id = j

        d = Document(id=doc_id)
        d.tags['id'] = doc_id
        
        d.embedding = np.random.random(
            [10 + np.random.randint(0, 1)]
        )
        yield d

dam = DocumentArrayMemmap('tmp_memmap_toy')
docs = list(random_docs(100))
dam.extend(docs)

dam2 = DocumentArrayMemmap('tmp_memmap_toy')
x = dam2.get_attributes('embedding')
```
The error log is:
```
Traceback (most recent call last):
  File "toys/toy_memmap.py", line 22, in <module>
    x = dam2.get_attributes('embedding')
  File "/Users/fengwang/Jina/workspace/jina/jina/types/arrays/document.py", line 62, in get_attributes
    return self.get_attributes_with_docs(*fields)[0]
  File "/Users/fengwang/Jina/workspace/jina/jina/types/arrays/document.py", line 79, in get_attributes_with_docs
    for doc in self:
  File "/Users/fengwang/Jina/workspace/jina/jina/types/arrays/memmap.py", line 208, in __iter__
    yield self[k]
  File "/Users/fengwang/Jina/workspace/jina/jina/types/arrays/memmap.py", line 160, in __getitem__
    with mmap.mmap(self._body_fileno, offset=p, length=l) as m:
ValueError: mmap length is greater than file size
```

**Describe how you solve it**

---

**Environment**

```
- jina 2.0.8                                                                                                                            
- jina-proto 0.0.84                                                                                                                     
- jina-vcs-tag (unset)                                                                                                                  
- libzmq 4.3.4                                                                                                                          
- pyzmq 1.18.1                                                                                                                          
- protobuf 3.17.3                                                                                                                       
- proto-backend cpp                                                                                                                     
- grpcio 1.38.1                                                                                                                         
- pyyaml 5.4.1                                                                                                                          
- python 3.8.3                                                                                                                          
- platform Darwin                                                                                                                       
- platform-release 20.3.0                                                                                                               
[...]

PR #2956, opened 2021-07-15
fix: cannot get embedding from reloaded DocumentArrayMemmap

winstonww:
This PR aims to close issue #2944 . Thank you.

### DIFF
--- a/jina/types/arrays/memmap.py
+++ b/jina/types/arrays/memmap.py
@@ -82,7 +82,7 @@ class DocumentArrayMemmap(TraversableSequence, DocumentArrayGetAttrMixin, Itr):
         open(self._body_path, mode).close()
 
         self._header = open(self._header_path, 'r+b')
-        self._body = open(self._body_path, 'r+b')
+        self._body = open(self._body_path, 'a+b')
 
         tmp = np.frombuffer(
             self._header.read(),
@@ -147,7 +147,7 @@ class DocumentArrayMemmap(TraversableSequence, DocumentArrayGetAttrMixin, Itr):
             ).tobytes()
         )
         self._header_map[doc.id] = (len(self._header_map), p, r, r + l)
-        self._start = p + r + l
+        self._start = p + r + l  # <==> self._start = self._start + l
         self._body.write(value)
         if flush:
             self._header.flush()
--- a/tests/unit/types/arrays/test_memmap.py
+++ b/tests/unit/types/arrays/test_memmap.py
@@ -173,3 +173,19 @@ def test_convert_dm_to_dam(tmpdir, mocker):
     mock.assert_called()
     assert len(da) == 0
     assert len(dam) == 100
+
+
+@pytest.mark.parametrize('embed_dim', [10, 10000])
+def test_extend_and_get_attribute(tmpdir, embed_dim):
+    dam = DocumentArrayMemmap(tmpdir)
+    docs = list(random_docs(100, start_id=0, embed_dim=embed_dim))
+    dam.extend(docs)
+
+    dam2 = DocumentArrayMemmap(tmpdir)
+    x = dam2.get_attributes('embedding')
+    assert len(dam2) == 100
+
+    docs = list(random_docs(100, start_id=100, embed_dim=embed_dim))
+    dam2.extend(docs)
+    x = dam2.get_attributes('embedding')
+    assert len(dam2) == 200
--- a/tests/unit/types/arrays/test_memmap.py
+++ b/tests/unit/types/arrays/test_memmap.py
@@ -184,8 +184,11 @@ def test_extend_and_get_attribute(tmpdir, embed_dim):
     dam2 = DocumentArrayMemmap(tmpdir)
     x = dam2.get_attributes('embedding')
     assert len(dam2) == 100
-
+    assert len(x) == 100
+    assert x[0].shape == (embed_dim,)
     docs = list(random_docs(100, start_id=100, embed_dim=embed_dim))
     dam2.extend(docs)
     x = dam2.get_attributes('embedding')
     assert len(dam2) == 200
+    assert len(x) == 200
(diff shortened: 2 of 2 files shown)


JoanFM (commented):
Let's add a test that shows the motivation for the change

winstonww:
@JoanFM Sure, thanks for the suggestion.

JoanFM (changes requested):
remember to add test

numb3r3 (changes requested):
This PR needs improvement. The following test cannot pass:
```
from jina import Document
import numpy as np

def random_docs(num_docs):
    for j in range(num_docs):
        doc_id = j

        d = Document(id=doc_id)
        d.tags['id'] = doc_id
        
        d.embedding = np.random.random(
            [10 + np.random.randint(0, 1)]
        )
        yield d

dam = DocumentArrayMemmap('tmp_memmap_toy')
docs = list(random_docs(100))
dam.extend(docs)

dam2 = DocumentArrayMemmap('tmp_memmap_toy')
x = dam2.get_attributes('embedding')
assert len(dam2) == 100


docs = list(random_docs(100))
dam2.extend(docs)
x = dam2.get_attributes('embedding')
assert len(dam2) == 200
```
BTW., please convert the above script to a pytest

winstonww:
Thank you for the suggestion. I think the error happens because in the following:
>```
> docs = list(random_docs(100))
> dam2.extend(docs)
> x = dam2.get_attributes('embedding')
> assert len(dam2) == 200
> ```

docs generated by this `random_docs` has the same id as the docs generated by the previous `random_docs`. As a result, docs are getting replaced instead of extended. 

If we replace `random_docs`  implementation with the following:
```
def random_docs(num_docs, start=0):
    for j in range(start, num_docs+start):
       ...
```
Then the proposed case passes.  I will continue to see if there's any other issues with the PR in the meantime.

numb3r3:
> Thank you for the suggestion. I think the error happens because in the following:
> 
> > ```
> > docs = list(random_docs(100))
> > dam2.extend(docs)
> > x = dam2.get_attributes('embedding')
> > assert len(dam2) == 200
> > ```
> 
> docs generated by this `random_docs` has the same id as the docs generated by the previous `random_docs`. As a result, docs are getting replaced instead of extended.
> 
> If we replace `random_docs` implementation with the following:
> 
> ```
> def random_docs(num_docs, start=0):
>     for j in range(start, num_docs+start):
>        ...
> ```
> 
> Then the proposed case passes. I will add the above to pytest and continue to see if there's any other issues with the PR in the meantime.

I see, the name `DocumentArrayMemmap` is very confusing. This is named as an array, but this is actually a map-style data.

tests/unit/types/arrays/test_memmap.py
```diff
@@ -173,3 +173,19 @@ def test_convert_dm_to_dam(tmpdir, mocker):
...
+    dam.extend(docs)
+
+    dam2 = DocumentArrayMemmap(tmpdir)
+    x = dam2.get_attributes('embedding')
+    assert len(dam2) == 100
+
+    docs = list(random_docs(100, start_id=100, embed_dim=embed_dim))
+    dam2.extend(docs)
+    x = dam2.get_attributes('embedding')
+    assert len(dam2) == 200
```
numb3r3, line 191:
```suggestion
    dam2 = DocumentArrayMemmap(tmpdir)
    x = dam2.get_attributes('embedding')
    assert len(dam2) == 100
    assert len(x) == 100
    assert x[0].shape == (embed_dim,)

    docs = list(random_docs(100, start_id=100, embed_dim=embed_dim))
    dam2.extend(docs)
    x = dam2.get_attributes('embedding')
    assert len(dam2) == 200
    assert len(x) == 200
    assert x[0].shape == (embed_dim,)
```

numb3r3 (changes requested):

tests/unit/types/arrays/test_memmap.py
winstonww: Addressed, please see [e5dea0f](https://github.com/jina-ai/jina/pull/2956/commits/e5dea0f45fbd88593f8ae29405a8c43085a99996) for the change, thanks.

numb3r3 (approved):
LGTM

florian-hoenicke:
I would recommend:
Whenever there is a corrupted DocumentArray on the disk, Jina should log a warning and initialize an empty DocumentArray.

tests/unit/types/arrays/test_memmap.py
```diff
@@ -173,3 +173,22 @@ def test_convert_dm_to_dam(tmpdir, mocker):
     mock.assert_called()
     assert len(da) == 0
     assert len(dam) == 100
+
```
florian-hoenicke, line 176: Can you test if manual interruption of the writing process causes corruptions and therefore this error message?

florian-hoenicke (changes requested):

florian-hoenicke:
Issue @alexcg1 discovered:
```
Clone https://github.com/alexcg1/jina-shortest-program
Run python sensible.py
Ctrl-C
Don't delete workspace
Re-run the script
Try to query via curl
```

jina/types/arrays/memmap.py
```diff
@@ -147,7 +147,7 @@ def append(self, doc: 'Document', flush: bool = True) -> None:
             ).tobytes()
         )
         self._header_map[doc.id] = (len(self._header_map), p, r, r + l)
-        self._start = p + r + l
+        self._start = p + r + l  # <==> self._start = self._start + l
```
hanxiao, line 150: can't understand the comment here, suggestion: either delete it or write it clearly

```suggestion
        self._start = p + r + l
```

hanxiao (changes requested):
see comment above

hanxiao:
#2944 looks like a fake issue. hold the merge

PR #2959, opened 2021-07-15
fix: open path as file for DictReader

georgeamccarthy:
🐛 **Bug Fix** `jina.types.document.generators.from_csv` was not behaving as expected or as documented.

### Purpose
- Makes `from_csv` behave as [expected and documented: ](https://github.com/jina-ai/jina/blob/master/.github/2.0/cookbooks/Document.md#construct-from-json-csv-ndarray-and-files)
> from_csv - Yield Document from a CSV file. Each line is a Document object
- `from_csv` takes `fp: Iterable[str]` and is documented as `:param fp: file paths` however it is passed directly to csv.DictReader(fp).
- [However csv.DictReader() takes a file not a string.](https://docs.python.org/3/library/csv.html#csv.DictReader) 
- Hence if `from_csv` is given a string path, it does not yield Documents correctly, instead it tries to yield the path string as Documents.

### Mentions
- Discussed with @florian-hoenicke and Pei-Shan at weekly Jina-MLH maintainer meeting.
- Came up when working on [protein_search](https://github.com/georgeamccarthy/protein_search) with @fissoreg

### Recreate

Create `app.py` with

```
from jina import DocumentArray
from jina.types.document.generators import from_csv

da = DocumentArray(
    from_csv("example_data.csv", field_resolver={"Name": "text"})
)

for d in da:
    print(d)
```

Create `example_data.csv` with

```
Name
George
Gian
Mable
```

Before change `python app.py` prints

```
{'id': '2f7b0cc6-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'x'}, 'content_hash': '1cdf8d85c151209c'}
{'id': '2f7b2f4e-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'a'}, 'content_hash': 'e07c484b04c14a3e'}
{'id': '2f7b3548-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'm'}, 'content_hash': 'cabd9a82bf030bbd'}
{'id': '2f7b3a20-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'p'}, 'content_hash': 'a343518e4bc62320'}
{'id': '2f7b3ea8-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'l'}, 'content_hash': 'a790cf78e61bd2ab'}
{'id': '2f7b4312-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'e'}, 'content_hash': '9ecb4fbaf4f87b84'}
{'id': '2f7b477c-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': '_'}, 'content_hash': '2092d99a84ea60b9'}
{'id': '2f7b4c04-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'd'}, 'content_hash': '0745985bbc0c789c'}
{'id': '2f7b503c-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'a'}, 'content_hash': 'e07c484b04c14a3e'}
{'id': '2f7b5460-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 't'}, 'content_hash': '9de8fdeb8a8f6503'}
{'id': '2f7b5898-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'a'}, 'content_hash': 'e07c484b04c14a3e'}
{'id': '2f7b5ce4-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': '.'}, 'content_hash': '996076630c89b68a'}
{'id': '2f7b60fe-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'c'}, 'content_hash': '5a446e2e6a821b74'}
{'id': '2f7b6518-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 's'}, 'content_hash': 'f5928cb47f90f3df'}
{'id': '2f7b6928-e5af-11eb-9174-1e00d23adf51', 'tags': {'e': 'v'}, 'content_hash': '2506fbb5bbf76bbd'}
```

Notice how the path has been gobbled up by DictReader.

After this change `python app.py` prints

```
{'id': '0e01a874-e5b0-11eb-82ce-1e00d23adf51', 'text': 'George', 'content_hash': 'd4d3f2f18e2bf4c0'}
{'id': '0e01d7f4-e5b0-11eb-82ce-1e00d23adf51', 'text': 'Gian', 'content_hash': '1b20aee0af11cf46'}
{'id': '0e01db96-e5b0-11eb-82ce-1e00d23adf51', 'text': 'Mable', 'content_hash': '8f416cbfebd3cf3c'}
```

(fixed)

### DIFF
--- a/jina/types/document/generators.py
+++ b/jina/types/document/generators.py
@@ -106,7 +106,7 @@ def from_csv(
     """
     from ..document import Document
 
-    lines = csv.DictReader(fp)
+    lines = csv.DictReader(open(fp))
     for value in _subsample(lines, size, sampling_rate):
         if 'groundtruth' in value and 'document' in value:
             yield Document(value['document'], field_resolver), Document(


nan-wang:
@georgeamccarthy I'm afraid this is not the expected usage of `from_csv`. We expect two ways of usages,

1. from file objects
  ```python
    with open('docs.csv') as fp:
        result = list(from_csv(fp))
    assert isinstance(result[0], Document)
  ```
1. from string
  ```python
    with open('docs.csv') as fp:
        lines = fp.readlines()
    result = list(from_csv(lines))
    assert isinstance(result[0], Document)
  ```

But you are right, we should make it clearer in the cookbook that we are using `csv.reader` under the hood. https://docs.python.org/3/library/csv.html#csv.reader

nan-wang:
Instead of changing the codes, I suggest you make a PR to improve the docstrings at https://github.com/jina-ai/jina/blob/master/jina/types/document/generators.py#L98

georgeamccarthy:
Hi 🙂 As a new user, passing a string path to a csv was what I expected this generator to do. Is there a reason why that is not the preferred way? The code snippets above have some redundancy in 

> with open('docs.csv') as fp:
> with open('docs.csv') as fp: 

Alternatively if instead we keep fp as file object then because the documentation and code are in disagreement then we can improve the docs to make it clearer that from_csv requires an open file. In that case because `fp` meaning file path implies a file path _string_ I think the alternative changes could be:

1. Change `fp` to `f` denoting _file_
2. Change docustring to: 'f: open csv file'
2. Change type from `Iterable[str]` to someone which describes an open file.

Rubix982:
@georgeamccarthy would this change bring about any other breaking changes on modules that are using this module?

fissoreg:
Hi everyone!

Based on @nan-wang examples, I agree that the documentation needs improvement.

Also, according to `csv.reader` documentation:

>  If csvfile is a file object, it should be opened with newline=''. 

which means that the first example snippet should be 
```python
 with open('docs.csv', newline='') as fp:
      result = list(from_csv(fp))
  assert isinstance(result[0], Document)
```
Just using `open('docs.csv')` probably doesn't pose any problems most of the time (it didn't in our case), but it might be less robust?

Passing a `str` to `from_csv` could be handy and it would avoid the above problem, but it transitions the control of opening/closing the file to the function. This is not a bad thing in general, but in combination with the `yield` it would cause problems given that if you don't immediately consume the returned generator you would keep the file open (possibly) indefinitely.

Clarifying the documentation is thus necessary, while passing a `str` to the function is probably not something you want to allow unless there are some very good use cases.

My 2 cents, hope this helps :)

georgeamccarthy:
Thanks Gian and Saif for your input! I will make some changes to the PR that improve the documentation while leaving the functionality as it is.

Issue #2870, opened 2021-07-07
add match and find functions to `DocumentArray`


JoanFM:
**Describe the feature**
To give more power to `DocumentArray`, add the following methods to `DocumentArray`.

```
def match(self, another: `DocumentArray`, metric: str = 'cosine', limit: Optional[int]=None) -> None:
  """Compute embedding based nearest neighbour in `another` for each Document in `self`,
    and store results in `matches`.
    :param another: the other DocumentArray to match against
    :param metric: the distance metric
    :param limit: the maximum number of matches, when not given all Documents in `another` are considered as matches
    """
```

```
def find(self, tags_regex: Dict[str, str]) -> DocumentArray:
    """Return a DocumentArray based on the regex matching result on `tags`.
    :param tags_regex: a mapping of the tag name, and regular expression.
    """
```

nan-wang:
Do we have some codes to show the expected behaviors? @JoanFM

davidbp:
I will put a draft here https://github.com/jina-ai/jina/pull/2942

JoanFM:
Once this `PR` is finished, needs to be put and used in `NumpySearcher` in the hub

PR #2963, opened 2021-07-16
feat: filter in documentarray

bwanglzu:
#2870


bwanglzu:
close this, I need to think of a better way

PR #2972, opened 2021-07-19
refactor: change the conditions to use router dealer

JoanFM:
Would require rebump of all hub executors

### DIFF
--- a/jina/peapods/peas/__init__.py
+++ b/jina/peapods/peas/__init__.py
@@ -240,12 +240,12 @@ class BasePea:
 
     def activate_runtime(self):
         """ Send activate control message. """
-        if self._is_dealer:
+        if self._is_worker:
             self._retry_control_message('ACTIVATE')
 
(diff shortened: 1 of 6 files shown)


jina/peapods/peas/__init__.py
```diff
@@ -404,11 +407,3 @@ def role(self) -> 'PeaRoleType':
 
         .. #noqa: DAR201"""
         return self.args.pea_role
-
-    @property
```
winstonww, line 408: Just for my understanding, what's the rationale for removing the `_is_inner_pea` property?

winstonww (approved):
LGTM

jina/peapods/peas/__init__.py
JoanFM: that nobody is using and is private

winstonww: understood, thank you

Issue #2715, opened 2021-06-21
document granularity off when building chunks bottom up


Kelton8Z:
e.g. With 
```python
from jina import Document

text = 'blah'
doc = Document(
    id='root1',
    text=text,
    chunks=[
        Document(
            id='chunk11',
            text=text,
            chunks=[
                Document(id='chunk111', text=text),
                Document(id='chunk112', text=text),
            ]
        ),
        Document(id='chunk12', text=text),
        Document(id='chunk13', text=text),
    ]
)
```

`doc.plot()` gives the attached image which has granularity at 1 for all chunks

**Screenshots**
![image](https://user-images.githubusercontent.com/22567795/122757527-4eab2d80-d2ca-11eb-8bdd-78de2d369504.png)

and printing flow response shows the same problem

![image](https://user-images.githubusercontent.com/22567795/122758975-0987fb00-d2cc-11eb-9698-2d77ee20f537.png)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200479679796904/1200498270493800) by [Unito](https://www.unito.io)

JoanFM:
It is not the plot the problem. It is the creation.

Kelton8Z:
> It is not the plot the problem. It is the creation.

I just realized after printing flow response

JoanFM:
I believe this is due to the fact that when the `latest` chunks are built, it detects its parent as granularity 0. 

So building from the largest depth to the shallowest, it will have this problem.

JoanFM:
@bwanglzu can u share ur thoughts on why is not doable

bwanglzu:
it looks not correct while we can hardly change this. Remember the `granularity` of chunks are set to `ref_doc + 1`, which means at init time, the `Document` will try to find it's parent document, then add 1 to it's parent document's `granularity`. 

In this case, all the documents are created in one line, and at chunk creation time, it didn't know anything about it's parent (root) document. So these chunks will be initialised as a root.

I would say it is not a recommended way to create Jina `Document`, it might mess up recursive document structure. Let's create `Documents` and it's `Chunks` with a more deterministic manner:

```python
d = Document(text='blah')
d.chunks.append(Document(text='chunk1')  # append one chunk
d.chunks.extend([Document(text='chunk2'), Document(text='chunk3')]). # append more chunks
```

JoanFM:
[...]

PR #2983, opened 2021-07-21
docs(cookbook): add document creation clean code

bwanglzu:
related to #2715

### DIFF
--- a/.github/2.0/cookbooks/CleanCode.md
+++ b/.github/2.0/cookbooks/CleanCode.md
@@ -336,7 +336,7 @@ tips to help you write beautiful and efficient code.
    </tr>
    </table>
 
-1. Add `Chunks` to root `Document`, do not create them in one line to keep recursive document structure correct. This is because `chunks` use `ref_doc` to control it's `granularity`, at `chunk` creation time, it didn't know anything about it's parent, and will get a wrong `granularity` value.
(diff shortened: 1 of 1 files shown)


tadejsv (commented):
To me (as a user) it is not clear why the Do version should be used instead of Don't. Could you add a sentence or two describing what goes wrong with the Don't version?

.github/2.0/cookbooks/CleanCode.md
```diff
@@ -336,7 +336,7 @@ tips to help you write beautiful and efficient code.
    </tr>
    </table>
 
-1. Add `Chunks` to root `Document`, do not create them in one line to keep recursive document structure correct.
+1. Add `Chunks` to root `Document`, do not create them in one line to keep recursive document structure correct. This is because `chunks` use `ref_doc` to control it's `granularity`, at `chunk` creation time, it didn't know anything about it's parent, and will get a wrong `granularity` value.
```
JoanFM, line 339:
```suggestion
1. Add `Chunks` to root `Document`, do not create them in one line to keep recursive document structure correct. This is because `chunks` use `ref_doc` to control its `granularity`, at `chunk` creation time, it didn't know anything about its parent, and will get a wrong `granularity` value.
```

JoanFM (changes requested):

PR #2985, opened 2021-07-21
feat: change default entrypoint and allow runtimes in main process

JoanFM:
Allow runtimes to work in main Processes and respond to terminate signals from both Kubernetes and our own Peas/Pods in the same way.
It does:

- Adds a new zed_runtime api client, and has executor be an alias for it. It starts a ZEDRuntime in the main process which is what would be expected when working inside Docker.

- Adds SIGNAL handeling for Gateway runtimes so that they can be scaled by Kubernetes.

- Hides implementation details from runtimes for the Pea without the need of serializing objects by using static methods.

- By changing jina executor to run jina runtime minimizes more the chances of having a version problem with jinahub since all the Pea/Pod components are not instantiated in the docker container

- It completes 100% the decoupling between Pea and Runtime having a clearer layer differentiation


cli/api.py
```diff
@@ -32,8 +32,31 @@ def pea(args: 'Namespace'):
...
+    Start a ZEDRuntime
+
+    :param args: arguments coming from the CLI.
+    """
+    from jina.peapods.runtimes.zmq.zed import ZEDRuntime
+    from jina.peapods.zmq import Zmqlet
+    import multiprocessing
+
+    # This event is not needed but ZEDRuntime was born to communicate to the Pea and needs an event
+    dummy_event = multiprocessing.Event()
```
JoanFM, line 46: maybe we need to revisit the dependency of this `Event` in `ZEDRuntime`

jina/peapods/runtimes/zmq/asyncio.py
```diff
@@ -67,6 +68,19 @@ def __init__(self, *args, **kwargs):
...
+
+        try:
+            for signame in {'SIGINT', 'SIGTERM'}:
+                self._loop.add_signal_handler(
+                    getattr(signal, signame),
+                    lambda *args, **kwargs: self.is_cancel.set(),
+                )
+        except ValueError as exc:
+            self.logger.warning(
+                f' The runtime {self.__class__.__name__} will not be able to handle signal handlers. '
```
JoanFM, line 80:
```suggestion
                f' The runtime {self.__class__.__name__} will not be able to handle termination signals. '
```

winstonww (approved):
LGTM!

> Adds a new zed_runtime api client, and has executor be an alias for it. It starts a ZEDRuntime in the main process which is what would be expected when working inside Docker.

Trying to understand, what happens in the current case when the runtime is started in a non-main process in docker?  What's the drawback?

JoanFM:
> LGTM!
> 
> > Adds a new zed_runtime api client, and has executor be an alias for it. It starts a ZEDRuntime in the main process which is what would be expected when working inside Docker.
> 
> Trying to understand, what happens in the current case when the runtime is started in a non-main process in docker? What's the drawback? I guess starting the runtime in the main process is a more direct hence better implementation?

Yes, it is more about having a clearer implementation and clarifying responsabilities and layers

winstonww:
> Yes, it is more about having a clearer implementation and clarifying responsabilities and layers

Got it, thank you.

jina/peapods/runtimes/container/__init__.py
```diff
@@ -263,3 +280,135 @@ def _is_container_alive(self) -> bool:
...
+        """
+        Check if the runtime has successfully started
+
+        :param timeout: The time to wait before readiness or failure is determined
+        :param ready_or_shutdown_event: the multiprocessing event to detect if the process failed or succeeded
+        :param kwargs: extra keyword arguments
+
+        :return: True if is ready or it needs to be shutdown
+        """
+        return ready_or_shutdown_event.wait(timeout)
```
maximilianwerk, line 300: Seems like this is the default method implementation, which is only changed by `Zedlet`. Can we put this implementation in the base class and only let the `Zedlet` overwrite it? Much less verbose code. Same might hold true for the `_wait_for_cancel` method, even thought I am not utterly sure.

jina/peapods/peas/__init__.py
```diff
@@ -158,41 +152,18 @@ def __init__(self, args: 'argparse.Namespace'):
...
-                ctrl_host = self.args.host
-
-            self._zed_runtime_ctrl_address = Zmqlet.get_ctrl_address(
-                ctrl_host, self.args.port_ctrl, self.args.ctrl_with_ipc
-            )[0]
-        else:
-            self._zed_runtime_ctrl_address = Zmqlet.get_ctrl_address(
-                self.args.host, self.args.port_ctrl, self.args.ctrl_with_ipc
-            )[0]
+        self._zed_runtime_ctrl_address = self.runtime_cls.get_control_address(
```
maximilianwerk, line 162: should this parameter be renamed to just `_runtime_ctrl_address`, since it is for different runtimes?

jina/peapods/runtimes/container/__init__.py
```diff
@@ -12,13 +13,30 @@
...
 
-    def __init__(self, args: 'argparse.Namespace', ctrl_addr: str, **kwargs):
-        super().__init__(args, ctrl_addr, **kwargs)
+    def __init__(self, args: 'argparse.Namespace', **kwargs):
+        super().__init__(args, **kwargs)
+        self.ctrl_addr = self.get_control_address(
+            args.host,
+            args.port_ctrl,
+            docker_kwargs=self.args.docker_kwargs,
+        )
```
maximilianwerk, line 33: Why is this different from the `_zed_runtime_ctrl_address` above?

maximilianwerk (commented):
I think there is some variable renaming in the `BasePea` missing. Otherwise, I like it a lot - especially the initializing of only a runtime inside a docker container. Is the following understanding correct:

A `BasePea` is now only the context manager of a runtime, when you want to start a runtime either:

- from a Flow
- from a Pod
- inside a docker container from a Pea

Or in other words: We give the user the capability of starting:

- only a runtime on its own with all the needed configurations
  - The user needs to take care of properly shutting it down. This is highly non-trivial!
  - This actually should only be done in a docker container
- a Pea, which might then start a runtime inside a docker container or just plainly. The Pea will take care of properly shutting it down, whenever the Pea gets a cancel signal
  - This is useful in k8s, when you want to take care of load balancing by k8s and not an own HeaPea. Anyhow, you need to take care of sharding/replicas yourself.
- a Pod, when you want to start a fully grown Pod inside a Flow
  - This is useful, e.g. when you want to use external Pods in multiple Flows
- a Flow => captain obvious sends greetings.

Is this a proper summary? If so, let's document this in the `Serving.md` Cookbook (not necessarily in this PR.)

JoanFM:
> I think there is some variable renaming in the `BasePea` missing. Otherwise, I like it a lot - especially the initializing of only a runtime inside a docker container. Is the following understanding correct:
> 
> A `BasePea` is now only the context manager of a runtime, when you want to start a runtime either:
> 
> * from a Flow
> * from a Pod
> * inside a docker container from a Pea
> 
> Or in other words: We give the user the capability of starting:
> 
> * only a runtime on its own with all the needed configurations
>   
>   * The user needs to take care of properly shutting it down. This is highly non-trivial!
>   * This actually should only be done in a docker container
> * a Pea, which might then start a runtime inside a docker container or just plainly. The Pea will take care of properly shutting it down, whenever the Pea gets a cancel signal
>   
>   * This is useful in k8s, when you want to take care of load balancing by k8s and not an own HeaPea. Anyhow, you need to take care of sharding/replicas yourself.
> * a Pod, when you want to start a fully grown Pod inside a Flow
>   
>   * This is useful, e.g. when you want to use external Pods in multiple Flows
> * a Flow => captain obvious sends greetings.
> 
> Is this a proper summary? If so, let's document this in the `Serving.md` Cookbook (not necessarily in this PR.)

Some clarifications.

BasePea is not a runtime context manager, it starts and closes the process where the runtime lives.

When using in Kubernetes u should have as entrypoint a jina executor or jina gateway, which will start simply a runtime. So when using Kubernetes, Peas or Pods from Jina play no role. Then starting and closing are handled by Kubernetes through the aliveness check and the termination signals

jina/peapods/runtimes/container/__init__.py
JoanFM: an alternative would be to add another argument to the parser, but I do not want to overkill with args

jina/peapods/peas/__init__.py
JoanFM: This is on purpose .I want to make here clear that the control address is only valid for the zedruntime living anywhere (remote via jinad or via docker). The other Runtimes do not have control address that one cares about.Actually I think that ContainrrRuntime and JinaDRuntime should not exist at the same level as ZEDRuntime, I think that Pea should eventually just start extra processes to stream logs from docker and from jinad. But this is out of scope for now.

jina/peapods/runtimes/container/__init__.py
JoanFM: I am not so sure how inherited methods behave with static method. I agree there is place for improvement, but can be done in a next PR

maximilianwerk:
OK can you merge master once again and then we are good to merge, I think.

jina/peapods/peas/__init__.py
```diff
@@ -240,23 +211,43 @@ def _retry_control_message(self, command: str, num_retry: int = 3):
...
-        else:
-            self.cancel_event.set()
+    def _wait_for_ready_or_shutdown(self, timeout: Optional[float]):
+        """
+        Waits for the process to be ready or to know it has failed.
+
+        :param timeout: The time to wait before readiness or failure is determined
+            .. # noqa: DAR201
+        """
+        return self.runtime_cls.wait_ready_or_shutdown(
```
nan-wang, line 244: Keep consistent
```suggestion
        return self.runtime_cls.wait_for_ready_or_shutdown(
```

jina/peapods/peas/__init__.py
```diff
@@ -240,23 +211,43 @@ def _retry_control_message(self, command: str, num_retry: int = 3):
...
-            self._retry_control_message('DEACTIVATE')
+    def _cancel_runtime(self, skip_deactivate: bool = False):
+        """
+        Send terminate control message.
 
-    def _cancel_runtime(self):
-        """Send terminate control message."""
-        from ..runtimes.zmq.zed import ZEDRuntime
-        from ..runtimes.container import ContainerRuntime
+        :param skip_deactivate: flag to tell if deactivate signal may be missed.
```
nan-wang, line 225:
```suggestion
        :param skip_deactivate: Mark that the DEACTIVATE signal may be missed if set to True
```

jina/peapods/runtimes/container/__init__.py
nan-wang, line 300: This method is duplicated in `GatewayRuntime`, `ContainerRuntime`, and `JinaDRuntime`.

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -498,3 +502,170 @@ def envelope(self) -> 'jina_pb2.EnvelopeProto':
...
+                num_retry=3,
+                logger=logger,
+            )
+        ZEDRuntime._retry_control_message(
+            ctrl_address=control_address,
+            timeout_ctrl=timeout_ctrl,
+            command='TERMINATE',
+            num_retry=3,
+            logger=logger,
+        )
```
nan-wang, line 631:
```suggestion
        _cmd = 'DEACTIVATE' if not skip_deactivate and socket_in_type == SocketType.DEALER_CONNECT else 'TERMINATE'
        ZEDRuntime._retry_control_message(
            ctrl_address=control_address,
            timeout_ctrl=timeout_ctrl,
            command=_cmd,
            num_retry=3,
            logger=logger,
        )
```

jina/peapods/runtimes/zmq/zed.py
```diff
@@ -498,3 +502,170 @@ def envelope(self) -> 'jina_pb2.EnvelopeProto':
...
+        Get the control address for a runtime with a given host and port
+
+        :param host: the host where the runtime works
+        :param port: the control port where the runtime listens
+        :param kwargs: extra keyword arguments
+        :return: The corresponding control address
+        """
+        from ...zmq import Zmqlet
+
+        return Zmqlet.get_ctrl_address(host, port, False)[0]
```
nan-wang, line 671: duplicated codes at `ZMQRuntime`, `JinaDRuntime`, and `ContainerRuntime`

jina/peapods/runtimes/container/__init__.py
```diff
@@ -263,3 +280,135 @@ def _is_container_alive(self) -> bool:
...
+                num_retry=3,
+                logger=logger,
+            )
+        ContainerRuntime._retry_control_message(
+            ctrl_address=control_address,
+            timeout_ctrl=timeout_ctrl,
+            command='TERMINATE',
+            num_retry=3,
+            logger=logger,
+        )
```
nan-wang, line 361: check out the comments for `ZEDRuntime`
```suggestion
        if not skip_deactivate and socket_in_type == SocketType.DEALER_CONNECT:
            ContainerRuntime._retry_control_message(
                ctrl_address=control_address,
                timeout_ctrl=timeout_ctrl,
                command='DEACTIVATE',
                num_retry=3,
                logger=logger,
            )
        ContainerRuntime._retry_control_message(
            ctrl_address=control_address,
            timeout_ctrl=timeout_ctrl,
            command='TERMINATE',
            num_retry=3,
            logger=logger,
        )
```

nan-wang (changes requested):
Plus, I find the docstrings of the `BaseRuntime` are out-of-dated and need to be adapted to these changes.

jina/peapods/peas/__init__.py
nan-wang: >  ContainrrRuntime and JinaDRuntime should not exist at the same level as ZEDRuntime
This might sound a bit confusing for the developers especially when they are all named as `Runtime`

jina/peapods/runtimes/container/__init__.py
JoanFM: Yes, I am not sure how inheritance works for staticmethods may try to make it better

jina/peapods/runtimes/zmq/zed.py
JoanFM: Not valid, because if DEALER_CONNECT, u want to have both `DEACTIVATE` and `TERMINATE

jina/peapods/runtimes/container/__init__.py
JoanFM: What's the change here?

jina/peapods/runtimes/zmq/zed.py
JoanFM: Yes, but they do not have a common parent, and I do not want to add new Mixin just for this

nan-wang (approved):
LGTM👍

PR #2993, opened 2021-07-22
fix: apply islice on iterator object of data in batch iterator

alaeddine-13:
fixes the following example:
```
from jina.helper import batch_iterator
from jina import Document

dam = DocumentArrayMemmap("test_dam")

dam.append(Document())

bi = batch_iterator(dam, 3)

# prints infinitely
for batch in bi:
    for doc in batch:
        print(doc)
```

### DIFF
--- a/tests/integration/issues/github_2992/test_flow.py
+++ b/tests/integration/issues/github_2992/test_flow.py
@@ -1,0 +1,17 @@
+from jina import Document, Flow
+from jina.types.arrays.memmap import DocumentArrayMemmap
+
+
+def test_dam_flow(tmpdir):
+    def _assert_called_once(response):
+        assert len(response.docs) == 1
+        assert not _assert_called_once.called
+        _assert_called_once.called = True
+
+    _assert_called_once.called = False
+
+    f = Flow().add()
+    dam = DocumentArrayMemmap(tmpdir)
+    dam.append(Document())
+    with f:
+        f.post('/', dam, on_always=_assert_called_once)
--- a/tests/integration/issues/github_2992/test_flow.py
+++ b/tests/integration/issues/github_2992/test_flow.py
@@ -3,17 +3,10 @@ from jina.types.arrays.memmap import DocumentArrayMemmap
 
 
 def test_dam_flow(tmpdir):
-    def _assert_called_once(response):
-        assert len(response.docs) == 1
-        assert not _assert_called_once.called
-        _assert_called_once.called = True
-
-    _assert_called_once.called = False
-
     f = Flow().add()
     dam = DocumentArrayMemmap(tmpdir)
     dam.append(Document())
     with f:
-        response = f.post('/', dam, on_always=_assert_called_once, return_results=True)
+        response = f.post('/', dam, return_results=True)
     assert len(response) == 1
     assert len(response[0].docs) == 1
--- a/tests/integration/issues/github_2992/test_flow.py
+++ b/tests/integration/issues/github_2992/test_flow.py
@@ -14,4 +14,5 @@ def test_dam_flow(tmpdir):
     dam = DocumentArrayMemmap(tmpdir)
     dam.append(Document())
     with f:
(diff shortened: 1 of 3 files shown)


JoanFM (changes requested):
Please add unit test

JoanFM (approved):

JoanFM (approved):
Can u also add an integration tests under `issues` with the number of the issue?

JoanFM (changes requested):
Sorry, want to request change.

Really nice work

alaeddine-13:
> Can u also add an integration tests under `issues` with the number of the issue?

I'll create an issue then

alaeddine-13:
> Can u also add an integration tests under `issues` with the number of the issue?

added

tests/integration/issues/github_2992/test_flow.py
```diff
@@ -0,0 +1,17 @@
+from jina import Document, Flow
+from jina.types.arrays.memmap import DocumentArrayMemmap
+
+
+def test_dam_flow(tmpdir):
```
JoanFM, line 5: use return_results or mock as in other tests

JoanFM (changes requested):

tests/integration/issues/github_2992/test_flow.py
alaeddine-13: is it okay now ? Or you were rejecting the usage of on_always ?

tests/integration/issues/github_2992/test_flow.py
```diff
@@ -0,0 +1,18 @@
...
+        assert not _assert_called_once.called
+        _assert_called_once.called = True
+
+    _assert_called_once.called = False
+
+    f = Flow().add()
+    dam = DocumentArrayMemmap(tmpdir)
+    dam.append(Document())
+    with f:
+        response = f.post('/', dam, on_always=_assert_called_once, return_results=True)
```
JoanFM, line 17: remove the `on_always` and the callback it is not needed anymore

JoanFM (changes requested):

tests/integration/issues/github_2992/test_flow.py
```diff
@@ -0,0 +1,18 @@
...
+    assert len(response.docs) == 1
```
JoanFM, line 18:
```suggestion
   assert len(responses) == 1
   assert len(responses[0].docs) == 1
```

JoanFM (changes requested):

tests/integration/issues/github_2992/test_flow.py
alaeddine-13: but the issue is that in a failing case, the interpreter keeps processing f.post forever. This means the response will never be used. I think on_always is the best way to detect the bug efficiently because if on_always is called more than once it means there's a bug.

JoanFM: if u check that only one response is processed then it is enough. Having callbacks as tests is hard since very little guarantee that the assert is called. We have seen this problem before

alaeddine-13: Honestly I don't know a way to catch the bug (infinitely posting documents, without ever existing the f.post line) without using on_always. Can you help with that ? Maybe there's a timeout parameter that I can use ?

tests/integration/issues/github_2992/test_flow.py
alaeddine-13: applied

tests/integration/issues/github_2992/test_flow.py
JoanFM: responses are not returned until all the requests are handled, so the bug is already caught like this

JoanFM: So passing this tedt without the callback already guarantees the fixing of the bug. U can try without the fix and u will see

JoanFM (approved):

PR #2994, opened 2021-07-22
feat(daemon): python client

deepankarm:
- Introduced `JinaDClient`  & `AsyncJinaDClient` for all operations with JinaD (workspaces, peas, pods, flows, logs etc)
- Updated cookbook for JinaD
- Fixed the error propagation for `client -> daemon -> partial-daemon -> daemon -> client` to make debugging easier.


daemon/clients/base.py
```diff
@@ -0,0 +1,216 @@
...
+                        self._logger.debug(
+                            f'successfully cancelled log streaming task for {id}'
+                        )
+                        break
+        finally:
+            for logger in remote_loggers.values():
+                logger.close()
+
+
+class BaseClient(AsyncBaseClient):
```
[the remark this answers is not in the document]
JoanFM: I think this just should have a different name and be a `mixin` declaring `alive` and `status` as functions that any class should implement

JoanFM (commented):
This PR will conflict with PR #2985

daemon/clients/base.py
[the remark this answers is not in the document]
deepankarm: > quite weird to have `BaseClient` inherit from `AsyncBaseClient`?

`aiohttp` is async in nature, so `BaseClient` inherits from the async one.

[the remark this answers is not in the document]
deepankarm: > I think this just should have a different name and be a `mixin` declaring `alive` and `status` as functions that any class should implement

yah. let me refactor that.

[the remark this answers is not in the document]
deepankarm: done

jina/peapods/runtimes/jinad/__init__.py
```diff
@@ -53,92 +75,70 @@ async def _wait_for_cancel(self):
...
-                break
+    async def _sleep_forever(self):
+        """Sleep forever, no prince will come."""
+        await asyncio.sleep(1e10)
+
+    def _mask_args(self, args: 'argparse.Namespace'):
+        _args = copy.deepcopy(args)
+
+        # reset the runtime to ZEDRuntime or ContainerRuntime
+        if _args.runtime_cls == 'JinadRuntime':
```
jacobowitz, line 101: Can we somehow reuse the existing logic here for finding the correct runtime class ? This looks error prone to future changes in runtime class config?

daemon/clients/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+        logger = logger or JinaLogger(self.__class__.__name__)
+        self.kwargs = {'uri': uri, 'logger': logger, 'timeout': timeout}
+
+    @property
+    def peas(self):
+        """Pea Client
+
+        :return: Pea Client
+        """
+        return self._pea_cls(**self.kwargs)
```
jacobowitz, line 51: should we somehow cache all of these individual clients instead of always returning a new instance? Right now it seems cheap enough to create new instances all the time, but maybe it gets more expensive in the future?

daemon/clients/pods.py
```diff
@@ -0,0 +1,13 @@
+from .peas import AsyncPeaClient
+from .mixin import AsyncToSyncMixin
+
+
+class AsyncPodClient(AsyncPeaClient):
+    """Async Client to create/update/delete Peods on remote JinaD"""
+
+    _kind = 'pod'
+    _endpoint = '/pods'
```
jacobowitz, line 9: Would it maybe be a good idea to reuse the definition from the Api here? So its always in sync?

jacobowitz (changes requested):
I think this will make using JinaD much more conivnient, great work!

I have only very small comments/questions.

jina/peapods/runtimes/jinad/__init__.py
deepankarm: This is same as the previous logic, just moved from a different file.

daemon/clients/__init__.py
deepankarm: Hmm, we possibly can. Maybe in the next iteration?

deepankarm:
[Docs are updated in the cookbook. ](https://github.com/jina-ai/jina/blob/feat-daemon-client/.github/2.0/cookbooks/Daemon.md)

.github/2.0/cookbooks/Daemon.md
```diff
@@ -2,60 +2,159 @@
...
+
+### API Docs
+
+- [Static docs with redoc](https://api.jina.ai/daemon/)
+
+- [Interactive swagger docs](http://localhost:8000/docs) (works once JinaD is started)
+
+### JinaDClient
+
+With v2.0.11, we introduce a Python Client to work with JinaD servers. It is async by default as it uses `aiohttp` internally. You can use `JinaDClient` or (for your async code) feel free to use `AsyncJinaDClient` which makes all following code awaitables.
```
jacobowitz, line 69: That paragraph is slightly confusing? You are saying that by default the client is async, but then there are async and sync versions to choose from? Maybe lets rather say that there exist sync and async versions and name them here as you do?

daemon/clients/base.py
```diff
@@ -0,0 +1,183 @@
...
+from .mixin import AsyncToSyncMixin
+from ..models.id import DaemonID, daemonize
+from ..helper import error_msg_from, if_alive
+
+
+class AsyncBaseClient:
+    """
+    JinaD baseclient (Async)
+
+    :param host: the host address of ``jinad`` instance
```
jacobowitz, line 22: The docstrings are wrong? there is no host/port, only uri?

daemon/clients/pods.py
deepankarm: We can move this to enums in a later PR.

.github/2.0/cookbooks/Daemon.md
```diff
@@ -2,60 +2,159 @@
...
 
 ------
 
+## Using JinaD
+
+`JinaD` docker image is published on [Docker Hub](https://hub.docker.com/r/jinaai/jina/tags?page=1&ordering=last_updated&name=-daemon) along with core & follows the [standard image versioning](https://github.com/jina-ai/jina/blob/master/RELEASE.md#docker-image-versioning) used in Jina.
+
+### Run
+
+To deploy JinaD, SSH into a server (e.g.- ec2 instance) & run the below command.
```
jacobowitz, line 48: Not sure, but maybe we should point also here to the options to change the port and root workspace directory? People might have reasons to use or not use 8000 and /tmp/jinad and will wonder this when reading this part. (maybe just link somewhere with the args or so)

.github/2.0/cookbooks/Daemon.md
```diff
@@ -2,60 +2,159 @@
...
+        - [Build](#build)
+        - [Run](#run-1)
+        - [Why?](#why)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-# JinaD 2.0
+# JinaD (Daemon)
 
 JinaD is a daemon for deploying & managing Jina on remote via a RESTful interface. It achieves isolation of deployments by defining a `workspace` for each Jina object.
```
jacobowitz, line 38: I would put slightly more content here. For us its obvious what `managing Jina` means, but I would give examples? Like create  flows etc.

Small language thing: I would rather say `managing Jina remotely` or `managing Jina on remote hosts` (not just remote)

jacobowitz (changes requested):
some small suggestions for the cookbook part

jacobowitz (approved):
looks good now :)

hanxiao (approved):

PR #2997, opened 2021-07-23
feat: find method for DocumentArray

davidbp:
Allow users to find Documents in a  DocumentArray that match several regular expressions per tag

### DIFF
--- a/.github/2.0/cookbooks/Document.md
+++ b/.github/2.0/cookbooks/Document.md
@@ -1001,6 +1001,62 @@ match emb = [1.  0.1 0.  0.  0. ] score = 1.6763054132461548
 
 
 
+### Selecting a subset of Documents from a `DocumentArray` using `.find`
+
+`DocumentArray` provides function `.find` that finds the documents in the `DocumentArray`  whose tag values match a dictionary of user provided regular expressions. Since a `Document` can have many tags, the function expects one regular expression for each tag that a user wants to consider.
+
+The simplest way to use this function is to provide only  the `regexes` dict. In this case, documents will be selected if all regular expressions passed are matched. Sometimes, a user might want to be less restrictive and might want to select documents only if a subset of the regular expressions is verified. In this case, `threshold` can be used to set the number of regular expressions that need to be matched in order to select a document.
+
+Let us consider the following example, where we want to select documents form a `DocumentArray` if the `city` tag contains a city that starts with `'B'`.
+
+```python
+d1 = Document(tags={'city': 'Barcelona', 'phone':'None'})
+d2 = Document(tags={'city': 'Berlin','phone':'648907348'})
+d3 = Document(tags={'city': 'Paris', 'phone': 'None'})
+d4 = Document(tags={'city': 'Brussels', 'phone': 'None'})
+
+docarray = DocumentArray([d1, d2, d3, d4])
+
+regexes = {'city':r'B.*'}
+docarray_filtered = docarray.find(regexes=regexes)
+print(f'len(docarray_filtered)={len(docarray_filtered)}')
+for d in docarray_filtered:
+    print(f'dict(d.tags)={dict(d.tags)}')
+```
+
+Will print
+
+```python
+len(docarray_filtered)=3
+dict(d.tags)={'phone': 'None', 'city': 'Barcelona'}
+dict(d.tags)={'phone': '648907348', 'city': 'Berlin'}
+dict(d.tags)={'phone': 'None', 'city': 'Brussels'}
+```
+
+We can consider more conditions, for example a subset of the previous documents that have `None` in the `'phone'` tag. We could do this as follows:
+
+```python
+regexes = {'city':r'B.*', 'phone':'None' }
+docarray_filtered = docarray.find(regexes=regexes)
+print(f'len(docarray_filtered)={len(docarray_filtered)}')
+for d in docarray_filtered:
+    print(f'dict(d.tags)={dict(d.tags)}')
+```
+
+Will print
+
+```
+len(docarray_filtered)=2
+dict(d.tags)={'city': 'Barcelona', 'phone': 'None'}
+dict(d.tags)={'phone': 'None', 'city': 'Brussels'}
+```
+
+
+
+
+
+
+
 ## `DocumentArrayMemmap` API
 
 When your `DocumentArray` object contains a large number of `Document`, holding it in memory can be very demanding. You may want to use `DocumentArrayMemmap` to alleviate this issue. A `DocumentArrayMemmap` stores all Documents directly on the disk, while only keeps a small lookup table in memory. This lookup table contains the offset and length of each `Document`, hence it is much smaller than the full `DocumentArray`. Elements are loaded on-demand to memory during the access.
--- a/jina/types/arrays/search_ops.py
+++ b/jina/types/arrays/search_ops.py
@@ -1,6 +1,6 @@
 import re
 import operator
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
 
 if False:
@@ -21,7 +21,7 @@ class DocumentArraySearchOpsMixin:
 
     def find(
         self,
-        regexes: Dict[str, str],
+        regexes: Dict[str, Union[str, re.Pattern]],
         traversal_paths: list = ['r'],
         operator: str = '>=',
         value: Optional[int] = None,
@@ -59,7 +59,8 @@ class DocumentArraySearchOpsMixin:
         value = value or len(regexes)
 
         for tag_name, regex in regexes.items():
-            regexes[tag_name] = re.compile(regex)
+            if type(regex) == str:
+                regexes[tag_name] = re.compile(regex)
 
         for pos, doc in enumerate(iterdocs):
             counter = 0
--- a/jina/types/arrays/search_ops.py
+++ b/jina/types/arrays/search_ops.py
@@ -3,6 +3,10 @@ import operator
 from typing import Dict
 
 
+if False:
+    from .document import DocumentArray
+
+
 class DocumentArraySearchOpsMixin:
     """ A mixin that provides search functionality to DocumentArrays"""
 
--- a/jina/types/arrays/search_ops.py
+++ b/jina/types/arrays/search_ops.py
@@ -1,10 +1,10 @@
 import re
-from typing import Union, Dict
+from typing import Dict
 
 import numpy as np
 
-# if False:
-from .document import DocumentArray
+if False:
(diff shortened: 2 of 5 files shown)


jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,78 @@
...
+            regexes[tag_name] = re.compile(regex)
+
+        for pos, doc in enumerate(iterdocs):
+            for tag_name, pattern in regexes.items():
+                tag_value = doc.tags.get(tag_name, None)
+                if tag_value:
+                    if pattern.match(tag_value):
+                        matched_counts[pos] += 1
+
+        if operator == '<':
```
JoanFM, line 58: keep this in a dictionary mapping from `operator_name` to operation function or partialfunc

JoanFM (changes requested):

jina/types/arrays/search_ops.py
davidbp: good catch, done

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,69 @@
...
+        '>=': operator.ge,
+    }
+
+    def find(
+        self,
+        regexes: Dict,
+        traversal_paths: list = ['r'],
+        operator: str = '>=',
+        value: int = 1,
+    ) -> None:
```
tadejsv, line 26: You are not returning `None`

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        Find Documents that match the regular expressions in `regexes`.
```
tadejsv, line 28: You might want to reword this to `Find Documents whose tags match` - on first reading it was not clear to me what these regexes are compared to.

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,69 @@
...
+        for pos, doc in enumerate(iterdocs):
+            for tag_name, pattern in regexes.items():
+                tag_value = doc.tags.get(tag_name, None)
+                if tag_value:
+                    if pattern.match(tag_value):
+                        matched_counts[pos] += 1
+
+        operator_func = self.operators[operator]
+        coordinate_flags = operator_func(matched_counts, value)
+        indices = np.where(coordinate_flags)[0].tolist()
```
tadejsv, line 64: Maybe that's just a matter of taste, but it seems to me that it would be simpler to add documents to filtered list in the "main" loop - seems like gains from vectorization here would be quite small.

jina/types/arrays/search_ops.py
davidbp: I totally agree

jina/types/arrays/search_ops.py
davidbp: Agree, the regular expression match takes longer than anything else.

tests/unit/types/arrays/test_search_ops.py
```diff
@@ -0,0 +1,45 @@
...
+    d2 = Document(tags={'city': 'Berlin', 'phone': '648907348'})
+    d3 = Document(tags={'city': 'Paris', 'phone': 'None'})
+    d4 = Document(tags={'city': 'Brussels', 'phone': 'None'})
+    docs = [d1, d2, d3, d4]
+    return docs
+
+
+def test_single_regex(list_doc_examples, tmpdir):
+    regexes = {'city': r'B.*'}
+    D = DocumentArray(list_doc_examples)
```
JoanFM, line 20: if possible, do not use capital letters as variables

JoanFM (changes requested):

tests/unit/types/arrays/test_search_ops.py
davidbp: I will do it but I don't feel it's a great idea, capitals for Matrices/Tensors/DocumentArrays make a lot of sense, non capital versions for vectors/documents also make sense. Makes everything easier to read.

JoanFM: this is standard programming standards, capitals are reserved for class names and not for vars or instances

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,68 @@
...
+        '>': operator.gt,
+        '==': operator.eq,
+        '!=': operator.ne,
+        '<=': operator.le,
+        '>=': operator.ge,
+    }
+
+    def find(
+        self,
+        regexes: Dict,
```
JoanFM, line 24: are we sure this complies with the definition of a `regex`?

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,68 @@
...
+        traversal_paths: list = ['r'],
+        operator: str = '>=',
```
JoanFM, line 26: I do not understand what this operator is for

tests/unit/types/arrays/test_search_ops.py
```diff
@@ -0,0 +1,47 @@
...
+
+    filtered_doc_array = doc_array.find(regexes=regexes)
+    filtered_doc_array_memmap = doc_array_memmap.find(regexes=regexes)
+
+    # Examples with Barcelona, Berlin, Brussels should match
+    assert len(filtered_doc_array) == 3
+    assert len(filtered_doc_array_memmap) == 3
+
+
+def test_multiple_regex(list_doc_examples, tmpdir):
```
JoanFM, line 32: instead of doing the test twice, please parametrize with DocArray and DocArrayMemmap

JoanFM (changes requested):

jina/types/arrays/search_ops.py
davidbp: I guess you mean `Dict[str,str]` ?

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,68 @@
...
+        value: int = 1,
```
JoanFM, line 27:
```suggestion
        value: Optional[int] = None,
```

And then in the code say.

```python
value = value or len(regexes)
```

JoanFM (changes requested):

tests/unit/types/arrays/test_search_ops.py
davidbp: Done

jina/types/arrays/search_ops.py
davidbp: yes, done

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,74 @@
...
+            operator in self.operators
+        ), f'operator={operator} is not a valid operator from {self.operators.keys()}'
+
+        operator_func = self.operators[operator]
+        iterdocs = self.traverse_flat(traversal_paths)
+        filtered = DocumentArray()
+
+        value = value or len(regexes)
+
+        for tag_name, regex in regexes.items():
```
JoanFM, line 61: I wonder if the interface should be this or provide already `compiled` regexes? Can we check if the items are already `regexes` and we do not need to `compile` them? This way we could actually use in an executor

JoanFM (changes requested):

jina/types/arrays/search_ops.py
davidbp: I think it is more convenient for the user to simply pass a string rather than having to compile them

tests/unit/types/arrays/test_search_ops.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.mark.parametrize('doc_array_creator', [docarray_type, docarray_memmap_type])
+def test_single_regex(doc_array_creator, list_doc_examples, tmpdir):
+    regexes = {'city': r'B.*'}
+    doc_array = doc_array_creator(list_doc_examples, tmpdir)
+    filtered_doc_array = doc_array.find(regexes=regexes)
+
+    # Examples with Barcelona, Berlin, Brussels should match
+    assert len(filtered_doc_array) == 3
```
bwanglzu, line 35: can you assert the result is correct?

tests/unit/types/arrays/test_search_ops.py
```diff
@@ -0,0 +1,48 @@
...
+def test_multiple_regex(doc_array_creator, list_doc_examples, tmpdir):
+    regexes = {'city': r'B.*', 'phone': 'Non'}
+    doc_array = doc_array_creator(list_doc_examples, tmpdir)
+
+    filtered_doc_array = doc_array.find(
+        regexes=regexes, traversal_paths=['r'], operator='==', value=2
+    )
+
+    # Examples with Barcelona, Brussels should match
+    assert len(filtered_doc_array) == 2
```
bwanglzu, line 48: same as above

bwanglzu (changes requested):

jina/types/arrays/search_ops.py
```diff
@@ -0,0 +1,75 @@
...
+            if type(regex) == str:
```
JoanFM, line 62:
```suggestion
            isinstance(regex, str):
```

JoanFM (changes requested):

.github/2.0/cookbooks/Document.md
```diff
@@ -1001,6 +1001,62 @@ match emb = [1.  0.1 0.  0.  0. ] score = 1.6763054132461548
...
+d1 = Document(tags={'city': 'Barcelona', 'phone':'None'})
+d2 = Document(tags={'city': 'Berlin','phone':'648907348'})
+d3 = Document(tags={'city': 'Paris', 'phone': 'None'})
+d4 = Document(tags={'city': 'Brussels', 'phone': 'None'})
+
+docarray = DocumentArray([d1, d2, d3, d4])
+
+regexes = {'city':r'B.*'}
+docarray_filtered = docarray.find(regexes=regexes)
+print(f'len(docarray_filtered)={len(docarray_filtered)}')
```
JoanFM, line 1022: see how in the rest of the cookbook we show output without actually writing `print`

JoanFM (changes requested):

.github/2.0/cookbooks/Document.md
davidbp: I see it full of prints! Besides, as  a reader isn't it better to get pretty printed results?

tests/unit/types/arrays/test_search_ops.py
davidbp: yes, done

bwanglzu (approved):
LGTM!

JoanFM (approved):
test with regex as values in the `regex` dict

PR #3001, opened 2021-07-23
feat: enabled benchmarking into ci

maateen:
# Changes:

- feat: enabled benchmarking into ci


deepankarm:
Is this tested on jina-terraform directly? If yes, can you share the link for the run?

maateen:
Yes, this has been tested. Please check [here](https://github.com/jina-ai/jina-terraform/actions/runs/1059710340).

winstonww:
LGTM. Just for the sake of my understanding, do I understand correctly that the mechanism for triggering benchmarking in another target repo is as follows:
1. Add a new job to the workflow and use `benc-uk/workflow-dispatch@v1` in the newly added job
2. Define `workflow_dispatch` in target repo to trigger benchmarking (in this case, `jina-ai/terraform`)
3. Provide other fields such as `ref`, `token` and `inputs` to specify the branch, grant access and provide input parameters respectively

Thank you.

hanxiao (approved):

Issue #2975, opened 2021-07-19
Make DocumentArray and DocumentArrayMemmap similar


alaeddine-13:
• Switching between DA <> DAM requires minimum effort.
• DA & DAM should implement the same common methods & interfaces & magic functions.
• Any Ops that work on DA should work on DAM as well.

slack discussion: https://jinaai.slack.com/archives/C018F60RBL5/p1626675464298200

- [x] add slice support to __getitem__ and __delitem__
- [x] fix updating existing document
- [x] integrate buffer pool
- [x] add tests
- [x] add DocumentArrayNeuralOperationsMixin to DocumentArrayMemmap
- [x] pass checks

PR #3002, opened 2021-07-23
feat: unify da dam interface

alaeddine-13:
feat: unify da dam interface

closes: https://github.com/jina-ai/jina/issues/2975


jina/types/arrays/memmap.py
```diff
@@ -154,32 +154,52 @@ def append(self, doc: 'Document', flush: bool = True) -> None:
             self._header.flush()
             self._body.flush()
 
-    def __getitem__(self, key: Union[int, str]) -> 'Document':
+    def _iteridx_by_slice(self, s: slice):
+        start, stop, step = s.start or 0, s.stop or self.__len__(), s.step or 1
+        if (
+            start <= -self.__len__()
+            or start >= self.__len__()
+            or stop <= -self.__len__()
```
winstonww, line 162: LGTM overall, just a nit/question:

In python list implementation, slicing with out of bounds indices is allowed, for instance:
```
>>> a = [1,2,3,4,5,6,7]
>>> a[0:-6]
[1]
>>> a[0:-7]
[]
>>> a[0:-8]
[]
>>> a[0:-9]
[]
>>> a[5:]
[6, 7]
>>> a[8:]
[]
>>> a[8:-10]
[]
```
Should we remove this constraint to align with the python implementation?

alaeddine-13: I've added support for out of bound slices + some tests. lmk if it's okay now

winstonww: Sure, LGTM!

jina/types/arrays/buffer.py
```diff
@@ -0,0 +1,97 @@
+from collections import OrderedDict
+from typing import Union
+
+from ... import Document
+
+if False:
+    from .memmap import DocumentArrayMemmap
+
+
+class BufferPoolManager:
```
hanxiao, line 10: given that `buffer` means binary string in Python context, i would rename this module as `buffer_pool.py` or `bpm.py`

hanxiao:
🔴 note the latency-bot reports that DAM extending efficiency drops significantly 
image

hanxiao (changes requested):
🔴 note the latency-bot reports that DAM extending efficiency drops significantly

alaeddine-13:
>  note the latency-bot reports that DAM extending efficiency drops significantly

I see why, will take care of it

alaeddine-13:
>  note the latency-bot reports that DAM extending efficiency drops significantly

@hanxiao to keep documents persisted after they change, we need to check whether they changed or not. Content hash/hashing means we need to serialize the document => not efficient
maybe we can maintain a version number of the document, each time a setter is called

jina/types/arrays/buffer.py
alaeddine-13: renamed

jina/types/document/helper.py
```diff
@@ -45,3 +46,39 @@ def chunks(self) -> Iterable['DocGroundtruthPair']:
         assert len(self.doc.chunks) == len(self.groundtruth.chunks)
         for doc, groundtruth in zip(self.doc.chunks, self.groundtruth.chunks):
             yield DocGroundtruthPair(doc, groundtruth)
+
+
+class VersionedMixin(object):
```
hanxiao, line 51:
```suggestion
class VersionedMixin:
```

jina/types/document/helper.py
```diff
@@ -45,3 +46,39 @@ def chunks(self) -> Iterable['DocGroundtruthPair']:
...
+    """
+    Helper class to add versioning to an object. The version number is incremented each time an attribute is set.
+    """
+
+    version = 0
+    ON_GETATTR: List = []
+
+    def _increase_version(self):
+        super(VersionedMixin, self).__setattr__('version', self.version + 1)
```
hanxiao, line 60:
```suggestion
        super().__setattr__('version', self.version + 1)
```

jina/types/document/helper.py
```diff
@@ -45,3 +46,39 @@ def chunks(self) -> Iterable['DocGroundtruthPair']:
...
+
+    def __setattr__(self, attr, value):
+        super(VersionedMixin, self).__setattr__(attr, value)
```
hanxiao, line 63:
```suggestion
        super().__setattr__(attr, value)
```

jina/types/arrays/bpm.py
```diff
@@ -0,0 +1,101 @@
+from collections import OrderedDict
+from typing import Union
+
+from ... import Document
+
+if False:
+    from .memmap import DocumentArrayMemmap
+
+
+class BufferPoolManager:
```
JoanFM, line 10: this class requires specific isolated `unittests`

jina/types/arrays/memmap.py
```diff
@@ -150,39 +163,122 @@ def append(self, doc: 'Document', flush: bool = True) -> None:
...
+        """
+        self._update_or_append(doc, idx=idx, flush=flush, update_buffer=update_buffer)
+
+    def _iteridx_by_slice(self, s: slice):
+        start, stop, step = (
+            s.start or 0,
+            s.stop if s.stop is not None else self.__len__(),
+            s.step or 1,
+        )
+        if 0 > stop >= -self.__len__():
```
JoanFM, line 217: compute `length` once at the beginning of the method and use it

JoanFM (changes requested):
I would not mix things in this PR. I would remove versioning and maybe add it later in another PR

jina/types/arrays/bpm.py
alaeddine-13: I added unit tests in test_buffer.py but the problem is that the class is coupled with DAM to access disk. In fact the code would be cleaner if I moved all disk operations to another class (DiskManager), and there would be less coupling: 
DAM <== BufferPoolManager <== DiskManager
what do you suggest here?

JoanFM: but anyway u would be coupled with another, I think it would be overkilling. It can be unittested usning `DocumentArrayMemmap` as input, that's clear

jina/types/arrays/bpm.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        Clears the memory buffer
+        """
+        self.doc_map.clear()
+        self.buffer = []
+
+    def __getitem__(self, key: Union[str, int]):
+        if isinstance(key, int):
+            key = self.dam._int2str_id(key)
+            return self[key]
```
maximilianwerk, line 88: I think this is not (and should not be) needed. The DAM actually never uses this method. The whole index -> key abstraction should (and does) happen in the DAM.

Apart from that: The `self.dam._int2str_id(key)` is a smell, since it accesses a private variable of another object.

jina/types/arrays/memmap.py
```diff
@@ -98,11 +108,11 @@ def _load_header_body(self, mode: str = 'a'):
         )
         self._header_entry_size = 24 + 4 * self._key_length
 
-        self._header_map = {
-            r[0]: (idx, r[1], r[2], r[3])
-            for idx, r in enumerate(tmp)
-            if not np.array_equal((r[1], r[2], r[3]), HEADER_NONE_ENTRY)
-        }
+        self._header_map = OrderedDict()
```
maximilianwerk, line 111: Is this still needed or can it be reverted to a generic dict, since we use the `buffer_pool` now? All the `move_to_end` logic inside the DAM could be removed as well.

jina/types/arrays/memmap.py
```diff
@@ -150,39 +163,121 @@ def append(self, doc: 'Document', flush: bool = True) -> None:
...
+            # the idx of the appended document is the idx of the last doc + 1
+            idx = (
+                next(reversed(self._header_map.items()))[1][0] + 1
+                if len(self._header_map)
+                else 0
+            )
+            self._header_map[doc.id] = (idx, p, r, r + l)
+            self._header_map.move_to_end(doc.id)
+        else:
+            self._header_map.update({doc.id: (idx, p, r, r + l)})
```
maximilianwerk, line 176: Not sure about it, but does the `update` performs equally well as the pure assignment? In the end, the update here needs to first create a new dictionary in memory with one entry and then assign. This seems to be a performance loss.

jina/types/arrays/memmap.py
```diff
@@ -214,15 +326,27 @@ def __iter__(self) -> Iterator['Document']:
...
+
+                # allows overwriting an existing document
+                if str_key != value.id:
+                    entry = self._header_map.pop(value.id)
+                    self._header_map = OrderedDict(
+                        [
+                            (value.id, entry) if k == str_key else (k, v)
+                            for k, v in self._header_map.items()
+                        ]
+                    )
```
maximilianwerk, line 341: This is super expensive. is there any reason why this is needed?

maximilianwerk (changes requested):
I think there are two many intermediate dictionaries created due to this PR. Please see my comments below.

jina/types/arrays/memmap.py
alaeddine-13: yes, the problem is when we want to call setitem with another document (different ID), we need to change the key in the header map but keep the same order (because __iter__ relies on the order).
I don't know a datastructure that supports inserting at some position efficiently

maximilianwerk: Can you point me to a test that fails, when we don't use the `OrderedDict`?

alaeddine-13: I don't think there are any. Now I'm aware that using a normal dict is fine.
I can point you to a case where updating at this point without reconstructing the dict corrupts the order of the entries

jina/types/arrays/bpm.py
alaeddine-13: removed

jina/types/arrays/memmap.py
alaeddine-13: reverted to generic dict

jina/types/arrays/memmap.py
alaeddine-13: I've changed it

jina/types/arrays/memmap.py
alaeddine-13: Here's a valid reason why _header_map should be an ordered dict:
due to this issue: https://github.com/jina-ai/jina/issues/3076, when adding a new entry to the header map, the idx should not be the len but rather the largest idx + 1
I think the most efficient way is to use next(reversed(self._header_map.items()))
in python3.7, this is not supported for generic dicts: 
https://stackoverflow.com/questions/58413076/why-are-python-dictionaries-not-reversible-for-python3-7

jina/types/arrays/memmap.py
```diff
@@ -214,15 +326,27 @@ def __iter__(self) -> Iterator['Document']:
...
+                    if str_key in self.buffer_pool.doc_map:
+                        self.buffer_pool.doc_map.pop(str_key)
             else:
                 raise IndexError(f'`key`={key} is out of range')
         elif isinstance(key, str):
-            value.id = key
-            self.append(value)
+            if key != value.id:
+                raise ValueError('key must be equal to document id')
+            self.update(value, key)
```
maximilianwerk, line 348: Update expects an `idx` and not a `key`.

alaeddine-13: fixed

jina/types/arrays/bpm.py
alaeddine-13: I added separate unit tests to `BufferPoolManager`

jina/types/arrays/bpm.py
```diff
@@ -0,0 +1,103 @@
...
+    def __getitem__(self, key: Union[str, int]):
+        if isinstance(key, str):
+            doc = self.buffer[self.doc_map[key][0]]
+            self.doc_map.move_to_end(key)
+            return doc
+        else:
+            raise TypeError(f'`key` must be str, but receiving {key!r}')
+
+    def __delitem__(self, key):
+        buffer_idx = self.doc_map.pop(key)[0]
```
JoanFM, line 99:
```suggestion
        buffer_idx,  _ = self.doc_map.pop(key)
```

jina/types/arrays/bpm.py
```diff
@@ -0,0 +1,103 @@
...
+        return result
+
+    def clear(self):
+        """
+        Clears the memory buffer
+        """
+        self.doc_map.clear()
+        self.buffer = []
+
+    def __getitem__(self, key: Union[str, int]):
```
JoanFM, line 90: the type hint is wrong according to the implementation

jina/types/document/__init__.py
```diff
@@ -142,6 +143,8 @@ class Document(ProtoTypeMixin):
 
     """
 
+    ON_GETATTR = ['matches']
```
JoanFM, line 146: it seems weird that matches is in a list where chunks is not. Why is that?

JoanFM (changes requested):

jina/types/document/__init__.py
alaeddine-13: oh forgot to add it back,  fixing it

alaeddine-13: fixed

jina/types/arrays/bpm.py
alaeddine-13: fixed

jina/types/arrays/bpm.py
alaeddine-13: done

bwanglzu:
run this piece of code:
```python
from jina import Document
import numpy as np
import time


dam = DocumentArrayMemmap('tmp')
for idx in range(10000):
    dam.append(Document(embedding=np.random.rand(512)))
start = time.perf_counter()
dam.get_attributes('embedding')
print(time.perf_counter() - start)
start = time.perf_counter()
dam.get_attributes('embedding')
print(time.perf_counter() - start)
```

this feature branch seems slower than current master branch

jina/types/arrays/memmap.py
```diff
@@ -150,39 +164,121 @@ def append(self, doc: 'Document', flush: bool = True) -> None:
...
-            return self[self._int2str_id(key)]
-        else:
-            raise TypeError(f'`key` must be int or str, but receiving {key!r}')
+        :param doc: The doc needs to be appended.
+        :param update_buffer: If set, update the buffer.
+        :param flush: If set, then flush to disk on done.
+        """
+        self._update_or_append(doc, flush=flush, update_buffer=update_buffer)
+
+    def update(
```
JoanFM, line 196: Are we also sure we want this to be part of the interface?

JoanFM (commented):
If we want to make this clear, what is shared between DA and DAM, we should define a clear `trait` (mixin) defining what public interface is shared and force DA and DAM to implement those, while the rest of the methods being private

hanxiao:
@bwanglzu please note that for DAM `.extend()` is marginally faster than looping & `.append`

> run this piece of code:
> 
> ```python
> import time
> import numpy as np
> 
> from jina import Document
> from jina.types.arrays.memmap import DocumentArrayMemmap
> 
> 
> dam = DocumentArrayMemmap('tmp')
> for idx in range(10000):
>     dam.append(Document(embedding=np.random.rand(512)))
> start = time.perf_counter()
> dam.get_attributes('embedding')
> print(time.perf_counter() - start)
> start = time.perf_counter()
> dam.get_attributes('embedding')
> print(time.perf_counter() - start)
> ```
> 
> ~this feature branch seems much slower than current master branch~
> ignore the previous message, the time spend is roughly the same

jina/types/arrays/memmap.py
alaeddine-13: made it protected

jina/types/arrays/abstract.py
```diff
@@ -0,0 +1,228 @@
...
+            When defining the :param:``traversal_paths`` with multiple paths, the returned
+            :class:``Documents`` are determined at once and not on the fly. This is a different
+            behavior then in :method:``traverse`` and :method:``traverse_flattened_per_path``!
+
+        :param traversal_paths: a list of string that represents the traversal path
+        """
+        pass
+
+    @abstractmethod
+    def match(
```
JoanFM, line 69: I am not sure `match` is offered by `DocumentArrayMemmap`, if so, make sure is tested

alaeddine-13: yes it's offered and `tests/unit/types/arrays/test_neural_ops.py` now tests various combinations of DA and DAM

JoanFM: `NeuralOpsMixin ` is not implemented by `DocumentArrayMemmap`

JoanFM (changes requested):

jina/types/arrays/abstract.py
alaeddine-13: image

I'm confused, why do you think so ?

jina/types/arrays/abstract.py
```diff
@@ -21,7 +21,7 @@ def get_attributes(self, *fields: str) -> Union[List, List[List]]:
 
         :param fields: Variable length argument with the name of the fields to extract
         """
-        pass
```
JoanFM: declaring as `Abstractmethod` is enough u just need to do

```python
   ...
```

jina/types/arrays/abstract.py
```diff
@@ -21,7 +21,7 @@ def get_attributes(self, *fields: str) -> Union[List, List[List]]:
...
+        raise NotImplementedError
```
JoanFM, line 24:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -32,7 +32,7 @@ def get_attributes_with_docs(
...
```
JoanFM, line 35:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -44,7 +44,7 @@ def traverse(
 
         :param traversal_paths: a list of string that represents the traversal path
         """
-        pass
+        raise NotImplementedError
```
JoanFM, line 47:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -56,7 +56,7 @@ def traverse_flat_per_path(
...
```
JoanFM, line 59:
```suggestion
        ...
```

jina/types/arrays/abstract.py
JoanFM: Not in master currently

jina/types/arrays/abstract.py
```diff
@@ -112,7 +112,7 @@ def match(
         :param use_scipy: use Scipy as the computation backend
         :param metric_name: if provided, then match result will be marked with this string.
         """
-        pass
+        raise NotImplementedError
```
JoanFM, line 115:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -139,7 +139,7 @@ def visualize(
         :param show_axis: If set, axis and bounding box of the plot will be printed.
 
         """
-        pass
+        raise NotImplementedError
```
JoanFM, line 142:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -168,7 +168,7 @@ def find(
         :param threshold: Number of regex that should match the operator to accept a Document.
                           If no value is provided `threshold=len(regexes)`.
         """
-        pass
+        raise NotImplementedError
```
JoanFM, line 171:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -178,7 +178,7 @@ def sample(self, k: int, seed: Optional[int] = None) -> 'DocumentArray':
         :param seed: initialize the random number generator, by default is None. If set will
             save the state of the random function to produce certain outputs.
         """
-        pass
+        raise NotImplementedError
```
JoanFM, line 181:
```suggestion
        ...
```

jina/types/arrays/abstract.py
```diff
@@ -187,43 +187,43 @@ def shuffle(self, seed: Optional[int] = None) -> 'DocumentArray':
...
```
JoanFM, line 190:
```suggestion
        ...
```

JoanFM (changes requested):

jina/types/arrays/abstract.py
alaeddine-13: okay, I'll see if it's excluded from codecov

alaeddine-13: Done

jina/types/arrays/abstract.py
alaeddine-13: yeah but the purpose of this PR is to make it possible to add neural ops. So `DocumentArrayNeuralOpsMixin` are added to `DocumentArrayMemmap` in this PR

hanxiao: @alaeddine-13 is right.

tests/unit/types/arrays/test_neural_ops.py
```diff
@@ -52,19 +131,22 @@ def docarrays_for_embedding_distance_computation_sparse():
...
-    docarrays_for_embedding_distance_computation, limit, batch_size
-):
-    D1, D2 = docarrays_for_embedding_distance_computation
-    D1.match(D2, metric='sqeuclidean', limit=limit, batch_size=batch_size)
-    for m in D1.get_attributes('matches'):
-        if limit is None:
-            assert len(m) == len(D2)
-        else:
-            assert len(m) == limit
+def test_matching_retrieves_correct_number(docarray_combinations, limit, batch_size):
```
JoanFM, line 142: this `docarray_combinations` loop should not be done by us, but this should be part of the `pytest.mark.parametrize`

JoanFM: Otherwise we cannot get a report on which combination failed the test.

JoanFM (changes requested):

tests/unit/types/arrays/test_neural_ops.py
alaeddine-13: fixed

tests/unit/types/arrays/test_neural_ops.py
```diff
@@ -52,13 +84,17 @@ def docarrays_for_embedding_distance_computation_sparse():
...
+@pytest.mark.parametrize('gen_doc_arrays', gen_docarray_combinations())
 @pytest.mark.parametrize(
     'limit, batch_size', [(1, None), (2, None), (None, None), (1, 1), (1, 2), (2, 1)]
 )
-def test_matching_retrieves_correct_number(
-    docarrays_for_embedding_distance_computation, limit, batch_size
-):
-    D1, D2 = docarrays_for_embedding_distance_computation
+def test_matching_retrieves_correct_number(gen_doc_arrays, limit, batch_size):
+    D1, D2 = gen_doc_arrays()
```
JoanFM, line 97: I do not think this should be a callable, this should be the `docarrays` themselves

JoanFM (approved):
LGTM! We can improve testing in another PR

.github/2.0/cookbooks/Document.md
```diff
@@ -1247,10 +1247,11 @@ This table summarizes the interfaces of `DocumentArrayMemmap` and `DocumentArray
 | `__add__`, `__iadd__` | ❌|✅|
 | `__bool__` |✅|✅|
 | `__eq__` |✅|✅|
-| `save`, `load` |❌ unnecessary |✅|
+| saving and loading |`save()` and `reload()` | `save(filename)` and `load(filename)` |
 | `sample` |✅ |✅|
 | `shuffle` |✅ |✅|
-| `match` |❌|✅|
+| `match` (L/Rvalue) |✅|✅|
+| `visualize` (L/Rvalue) |✅|✅|
```
hanxiao, line 1254:
```suggestion
| `visualize` |✅|✅|
```

.github/2.0/cookbooks/Document.md
```diff
@@ -1247,10 +1247,11 @@ This table summarizes the interfaces of `DocumentArrayMemmap` and `DocumentArray
 | `__add__`, `__iadd__` | ❌|✅|
 | `__bool__` |✅|✅|
 | `__eq__` |✅|✅|
-| `save`, `load` |❌ unnecessary |✅|
+| saving and loading |`save()` and `reload()` | `save(filename)` and `load(filename)` |
```
hanxiao, line 1250:
```suggestion
```

hanxiao (approved):

PR #3009, opened 2021-07-26
fix(hubble): more informed logging for fetch_meta

numb3r3:


winstonww (approved):
LGTM

hanxiao (changes requested):
i tend to reject this, as we do not want to disclose Hubble URL. unnecessary exposure of this URL attracts unnecessary (possible adversary) visits

I would accept if there is a separate `self.logger.debug(f'hubble URL: {...}')`

numb3r3:
> i tend to reject this, as we do not want to disclose Hubble URL. unnecessary exposure of this URL attracts unnecessary (possible adversary) visits
> 
> I would accept if there is a separate `self.logger.debug(f'hubble URL: {...}')`

@hanxiao I agree to trim the protocol prefix `jinahub`. But i still suggest to show more detailed info to user. If the **flow includes different jinahub executors**, the current `rich` status bar would display repeated messages, which would be a bit confusing at first glance.

hanxiao (approved):

PR #3016, opened 2021-07-26
refactor(types): move common math function to jina.math module

hanxiao:


davidbp (dismissed):
Updated some dosctrings with inconsistent X and x variables referring to the input matrix.  
I know PEP8 encourages variables non capital, but capital single letters for matrices are common in Math (even sklearn uses them) so I think using `x_mat` to tell the user that the input is a matrix is very convenient ( and maybe `x_vec` for vectors).
LGTM!

PR #3017, opened 2021-07-26
Refactor distances to math module

davidbp:
refactor distances to  math module


davidbp:
done in https://github.com/jina-ai/jina/pull/3016

PR #3019, opened 2021-07-27
feat(types): allow scipy metric in match and add minmax normalization

hanxiao:
improve `.match` by
- allowing scipy metric
- allowing 2-arity function
- allowing min-max normalization

### DIFF
--- a/.github/2.0/cookbooks/Document.md
+++ b/.github/2.0/cookbooks/Document.md
@@ -961,10 +961,10 @@ The following image shows how `DocumentArrayA` finds `limit=5` matches from the 
 
 ![match_illustration_5](https://github.com/jina-ai/jina/blob/master/.github/images/match_illustration_5.svg)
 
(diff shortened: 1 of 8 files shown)


JoanFM (commented):
why did we remove the option to have similarity?

hanxiao:
it is changed min-max normalization for better generalbility and professionality

davidbp (approved):
LGTM. Note that this changes the default behaviour. Previously all returned results were "higher the better" by default. Now it depends on the metric the user chooses (which is fine for me).
- One question arises: How is the ordering taking into account if larger values are better or lower values?

JoanFM:
> it is changed to min-max normalization for better generalbility and professionality

normalization does not change the meaning of the score to similarity.

davidbp:
My only concern is the updated `top_k` that returns negative values
```
    if descending:
        values = -values
```

hanxiao:
> My only concern is the updated `top_k` that returns negative values
> 
> ```
>     if descending:
>         values = -values
> ```

plz read the full code

hanxiao:
> > it is changed to min-max normalization for better generalbility and professionality
> 
> normalization does not change the meaning of the score to similarity.

https://github.com/jina-ai/jina/pull/3019/files#diff-0f9b5cdc7560c5f53ae28daa201efd41371ff58e80340df23991ad2d27bb9f84R42

PR #3021, opened 2021-07-27
feat: add shuffle to documentarray

bwanglzu:
add shuffle to document array, it's implemented as sample the entire document array.


davidbp (approved):
LGTM

PR #3032, opened 2021-07-28
fix: fashion hello world fix 

davidbp:
Fixes the image representation that contained a bug created when the image was represented as a 3d array


hanxiao (approved):

bwanglzu (approved):
LGTM!

PR #3033, opened 2021-07-28
test(daemon): py module

deepankarm:


daemon/Dockerfiles/default.Dockerfile
```diff
@@ -1,6 +1,6 @@
 # This is the default dockerfile used for all containers created by JinaD
 
-ARG JINA_VERSION=latest
+ARG JINA_VERSION=master
```
deepankarm, line 3: Since we are making many changes to jinad, with this change, the default build will pull from `master` branch (This is used for starting partial-jinad).  In a later PR, we can probably use the jinad's base image as the base image for partial-jinad too.

hanxiao (approved):

Issue #3034, opened 2021-07-29
DocumentArray.match on itself produces recursive results


hanxiao:
**Describe the bug**

```python
from jina import DocumentArray, Document
```

```python
import numpy as np
```

```python
da = DocumentArray([Document(embedding=np.array([1,2,3])), 
                    Document(embedding=np.array([1,0,1])),
                    Document(embedding=np.array([1,1,2]))])
```

Now match against itself.

```python
da.match(da)
```

Check the structure of the first document, this looks correct.

```python
da[0]
```

[...]

PR #3040, opened 2021-07-29
fix: match with same documents

bwanglzu:
for #3034 , I would not say it's a bug, it's a side effect. Calling match on same document does not really make sense. But here is a workaround. not a perfect solution, but works. Feel free to close this if there is a better solution

before: 
1D590F4A-B5CB-47C4-81F5-A30E53B4354B

after:

3FE4B9E8-90EF-4729-9F1B-687952086F2E

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -75,7 +75,7 @@ class DocumentArrayNeuralOpsMixin:
             if isinstance(normalization, (tuple, list)):
                 dist = minmax_normalize(dist, normalization)
 
-        source_docarray_ids = [d.id for d in self]
+        source_docarray_ids = {d.id for d in self}
         m_name = metric_name or (metric.__name__ if callable(metric) else metric)
         for _q, _ids, _dists in zip(self, idx, dist):
             _q.matches.clear()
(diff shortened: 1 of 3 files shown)


tadejsv (dismissed):
Looks good to me

jina/types/arrays/neural_ops.py
```diff
@@ -75,10 +75,16 @@ def match(
             if isinstance(normalization, (tuple, list)):
                 dist = minmax_normalize(dist, normalization)
 
+        source_docarray_ids = [d.id for d in self]
```
hanxiao, line 78: use set

jina/types/arrays/neural_ops.py
```diff
@@ -75,10 +75,16 @@ def match(
...
         m_name = metric_name or (metric.__name__ if callable(metric) else metric)
         for _q, _ids, _dists in zip(self, idx, dist):
             _q.matches.clear()
             for _id, _dist in zip(_ids, _dists):
                 d = Document(darray[int(_id)], copy=True)
+                # Note, when match self with other, or both of them share the same Document
+                # we might have recursive matches .
+                # checkout https://github.com/jina-ai/jina/issues/3034
+                if d.id in source_docarray_ids:
```
hanxiao, line 87: list is low-efficient

hanxiao (changes requested):
do not merge

Issue #3036, opened 2021-07-29
error for embedding


yangsp5:
I follow the cookbook and use `csr_matrix`, set  `use_scipy=False` and got wrong :
```
Traceback (most recent call last):
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/peapods/runtimes/zmq/zed.py", line 313, in _msg_callback
    processed_msg = self._callback(msg)
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/peapods/runtimes/zmq/zed.py", line 299, in _callback
    self._pre_hook(msg)._handle()._post_hook(msg)
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/peapods/runtimes/zmq/zed.py", line 251, in _handle
    groundtruths_matrix=self.groundtruths_matrix,
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/executors/__init__.py", line 186, in __call__
    self, **kwargs
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/executors/decorators.py", line 103, in arg_wrapper
    return fn(*args, **kwargs)
  File "index.py", line 66, in bar
    docs.match(self._docs, metric='cosine', limit=20, use_scipy=True)
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/types/arrays/neural_ops.py", line 65, in match
    dists = cdist(X, Y, metric)
  File "/opt/conda/lib/python3.7/site-packages/scipy/spatial/distance.py", line 2931, in cdist
    raise ValueError('XA must be a 2-dimensional array.')
ValueError: XA must be a 2-dimensional array.
```

set `use_scipy=False`,  I got this wrong
```
Traceback (most recent call last):
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/peapods/runtimes/zmq/zed.py", line 313, in _msg_callback
    processed_msg = self._callback(msg)
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/peapods/runtimes/zmq/zed.py", line 299, in _callback
    self._pre_hook(msg)._handle()._post_hook(msg)
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/peapods/runtimes/zmq/zed.py", line 251, in _handle
    groundtruths_matrix=self.groundtruths_matrix,
  File "/home/oppoer/.local/lib/python3.7/site-packages/jina/executors/__init__.py", line 186, in __call__
[...]

PR #3041, opened 2021-07-29
feat: add sparse versions for pairwise distances

davidbp:
A user was rightly pointing out that `.match` was not working with sparse data https://github.com/jina-ai/jina/issues/3036. This PR aims to add functions and update `.match` to allow working with sparse data.

This PR also removes from  `tests/unit/types/arrays/test_neural_ops.py` the tests that affect only `jina.math.distance` and puts them in `tests/unit/math/test_distance.py` (with more tests for sparse data)


jina/types/arrays/neural_ops.py
```diff
@@ -50,26 +53,46 @@ def match(
                                 the min distance will be rescaled to `a`, the max distance will be rescaled to `b`
                                 all values will be rescaled into range `[a, b]`.
         :param use_scipy: use Scipy as the computation backend
+        :param is_sparse: boolean stating if inpyt is a scipy.sparse object
```
hanxiao, line 56:
```suggestion
        :param is_sparse: boolean stating if input is a scipy.sparse object
```

PR #3045, opened 2021-07-29
refactor: try to speed up update content hash

bwanglzu:
the idea:
1. when calling `update_content_hash`, we actually never change parameters, so `include_fields` never being used except test. We also get rid of the mutually exclusive check.
2. maybe this is a better solution for hashing, we don’t specify `exclude_fields` ,but specify `fields`  necessary for hashing, safer
2. `CopyFrom` and `FieldMask` is expensive, we manage to only use `FieldMask` without creating the `empty_proto`
3. `SerializeToString ` = `SerializePartialToString ` + check if message is initialised, we initialise the pb message in the first line of the method, so checking is unnecessary please refer to [here](https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.SerializePartialToString).

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -448,7 +448,7 @@ class Document(ProtoTypeMixin):
         FieldMask(paths=fields_to_hash).MergeMessage(self._pb_body, masked_d)
 
         self._pb_body.content_hash = blake2b(
-            masked_d.SerializeToString(), digest_size=DIGEST_SIZE
+            masked_d.SerializePartialToString(), digest_size=DIGEST_SIZE
         ).hexdigest()
 
     @property
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -428,25 +428,27 @@ class Document(ProtoTypeMixin):
 
     def update_content_hash(
         self,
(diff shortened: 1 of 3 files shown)


hanxiao:
> SerializeToString = SerializePartialToString + check if message is initialised, we initialise the pb message in the first line of the method, so checking is unnecessary.

if this is true and brings more efficiency, then we should replace it globally

bwanglzu:
> if this is true and brings more efficiency, then we should replace it globally
it is true [link](https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.SerializePartialToString), checking if it brings more efficiency

bwanglzu:
on master: update content hash account for 27.71% of execution time
0288B4E6-2286-4C92-BA39-AE78EFA83E59

on feature branch, update content hash accounts for 16.06% of entire execution time:

30D497FE-3281-4053-8D23-CA7642AAAB20

bwanglzu:
match on 200 random docs avg time
on master branch :  --- 13.43830680847168 seconds ---
on feature branch: --- 11.4239821434021 seconds --- 

not significant performance increase, speed up 15% percent

PR #3051, opened 2021-07-30
feat: connect to pred config

maximilianwerk:
This allows to overwrite our remote-local detection magic explicitly for special use-cases (e.g. running on multiple instances in a VPN and additionally having one remote Pod).


JoanFM (commented):
When we were trying to have remote external Pods it was very hard to understand what are the needed arguments related to `dynamic` routing. I am not sure if this adds yet more noise to this.

In which context have we encountered this need?

cli/autocomplete.py
```diff
@@ -114,6 +113,9 @@
             '--external',
             '--peas-hosts',
             '--pod-role',
+            '--dynamic-routing',
+            '--no-dynamic-routing',
```
JoanFM, line 117: it makes not much sense to have these 2 options at the same time and redundant

JoanFM (changes requested):

jina/parsers/peapods/pod.py
```diff
@@ -88,3 +88,27 @@ def mixin_base_pod_parser(parser):
...
+    parser.add_argument(
+        '--dynamic-routing',
+        action='store_true',
+        default=True,
+        help='The Pod will setup the socket types of the HeadPea and TailPea depending on this argument.'
+        if _SHOW_ALL_ARGS
+        else argparse.SUPPRESS,
+    )
+
+    parser.add_argument(
```
JoanFM, line 101: How can this and the other coexits and be true by default?

maximilianwerk:
> When we were trying to have remote external Pods it was very hard to understand what are the needed arguments related to `dynamic` routing. I am not sure if this adds yet more noise to this.
> 
> In which context have we encountered this need?

We are preparing a Cookbook section for explaining how to setup hosts correctly. This PR is a pre-step for it in order to remove make the setup easier with this Cookbook.

maximilianwerk:
@JoanFM the cookbook will be changed in #3066

JoanFM (approved):

PR #3055, opened 2021-07-30
Perf match fast

davidbp:
This PR plays with an alternative match to avoid creating as many Documents as in the original match.
This is related to https://github.com/jina-ai/jina/pull/3045 , making Document creation faster would make a huge impact on `.match`. 

#### Problem with current  `.match` 

The following benchmark shows that most of the time in `.match`  is not spend computing distances or ranking by distance. It is creating python objects (in particular, `Document` objects):
Note that lines 82 and line 89 take 49+33 = 82% of the overall running time of `.match`

```
    73         1    1068161.0 1068161.0      3.1          dist, idx = top_k(dists, limit, descending=False)
    74         1          1.0      1.0      0.0          if normalization is not None:
    75                                                       if isinstance(normalization, (tuple, list)):
    76                                                           dist = minmax_normalize(dist, normalization)
    77                                           
    78         1          2.0      2.0      0.0          m_name = metric_name or (metric.__name__ if callable(metric) else metric)
    79     10001     409283.0     40.9      1.2          for _q, _ids, _dists in zip(self, idx, dist):
    80     10000     177530.0     17.8      0.5              _q.matches.clear()
    81    110000     288019.0      2.6      0.8              for _id, _dist in zip(_ids, _dists):
    82    100000   17141824.0    171.4     49.0                  d = Document(darray[int(_id)], copy=True)
    83                                                           # Note, when match self with other, or both of them share the same Document
    84                                                           # we might have recursive matches .
    85                                                           # checkout https://github.com/jina-ai/jina/issues/3034
    86    100000     572428.0      5.7      1.6                  if d.id in self:
    87                                                               d.pop('matches')
    88    100000     858797.0      8.6      2.5                  d.scores[m_name] = _dist
    89    100000   11562266.0    115.6     33.0                  _q.matches.append(d)

Total time: 38.9032 s
File: test_simple_docs.py
Function: main_func at line 5

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     5                                           @profile
     6                                           def main_func():
     7         1         10.0     10.0      0.0      da1 = DocumentArray()
     8         1          2.0      2.0      0.0      da2 = DocumentArray()
     9     10001       9610.0      1.0      0.0      for i in range(10000):
    10     10000    1710010.0    171.0      4.4          da1.append(Document(embedding=np.random.rand(512)))
    11     10000    1712220.0    171.2      4.4          da2.append(Document(embedding=np.random.rand(512)))
    12         1          2.0      2.0      0.0      start_time = time.time()
    13         1   35471312.0 35471312.0     91.2      da1.match(da2, limit=10)
    14         1         49.0     49.0      0.0      print("--- %s seconds ---" % (time.time() - start_time))

```

#### Benchmark match_fast

Store this script in **`test_simple_docs_fast.py `**

```
from jina import DocumentArray, Document
import time
import numpy as np

def main_func():
    da1 = DocumentArray()
    da2 = DocumentArray()
    for i in range(10_000):
        da1.append(Document(embedding=np.random.rand(512)))
        da2.append(Document(embedding=np.random.rand(512)))
    start_time = time.time()
    da1.match_fast(da2, limit=10)
    print("--- %s seconds ---" % (time.time() - start_time))

main_func()
```

```
python test_simple_docs_fast.py 
--- 10.381131172180176 seconds ---
```

#### Benchmark match

Store this script in **`test_simple_docs.py`**
```
from jina import DocumentArray, Document
import time
import numpy as np

def main_func():
    da1 = DocumentArray()
    da2 = DocumentArray()
    for i in range(10000):
        da1.append(Document(embedding=np.random.rand(512)))
        da2.append(Document(embedding=np.random.rand(512)))
    start_time = time.time()
    da1.match(da2, limit=10)
    print("--- %s seconds ---" % (time.time() - start_time))

main_func()
```

python test_simple_docs.py
--- 18.579392910003662 seconds ---

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -87,3 +87,63 @@ class DocumentArrayNeuralOpsMixin:
                     d.pop('matches')
                 d.scores[m_name] = _dist
                 _q.matches.append(d)
+
+    def match_fast(
+        self,
+        darray: Union['DocumentArray', 'DocumentArrayMemmap'],
+        metric: Union[
+            str, Callable[['np.ndarray', 'np.ndarray'], 'np.ndarray']
+        ] = 'cosine',
+        limit: Optional[int] = inf,
+        normalization: Optional[Tuple[int, int]] = None,
+        use_scipy: bool = False,
+        metric_name: Optional[str] = None,
+    ) -> Union[DocumentArray, DocumentArrayMemmap]:
+        """
+        Alternative match that returns does not append matches
+        :param darray: the other DocumentArray or DocumentArrayMemmap to match against
+        :param metric: the distance metric
+        :param limit: the maximum number of matches, when not given
+                      all Documents in `another` are considered as matches
+        :param normalization: a tuple [a, b] to be used with min-max normalization,
+                                the min distance will be rescaled to `a`, the max distance will be rescaled to `b`
+                                all values will be rescaled into range `[a, b]`.
+        :param use_scipy: use Scipy as the computation backend
+        :param metric_name: if provided, then match result will be marked with this string.
+        :return : DocumentArray with matches
+        """
+
+        from .document import DocumentArray
+
+        X = np.stack(self.get_attributes('embedding'))
+        Y = np.stack(darray.get_attributes('embedding'))
+        limit = min(limit, len(darray))
+
+        if isinstance(metric, str):
+            if use_scipy:
+                from scipy.spatial.distance import cdist
+            else:
+                from ...math.distance import cdist
+            dists = cdist(X, Y, metric)
+        elif callable(metric):
+            dists = metric(X, Y)
+        else:
+            raise TypeError(
+                f'metric must be either string or a 2-arity function, received: {metric!r}'
+            )
+
+        dist, idx = top_k(dists, limit, descending=False)
+        if normalization is not None:
+            if isinstance(normalization, (tuple, list)):
+                dist = minmax_normalize(dist, normalization)
+
+        m_name = metric_name or (metric.__name__ if callable(metric) else metric)
+
+        docarray = DocumentArray()
+        for ids, dist_vec in zip(idx, dist):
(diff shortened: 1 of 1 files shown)


jina/types/arrays/neural_ops.py
```diff
@@ -87,3 +87,63 @@ def match(
...
+        metric: Union[
+            str, Callable[['np.ndarray', 'np.ndarray'], 'np.ndarray']
+        ] = 'cosine',
+        limit: Optional[int] = inf,
+        normalization: Optional[Tuple[int, int]] = None,
+        use_scipy: bool = False,
+        metric_name: Optional[str] = None,
+    ) -> Union['DocumentArray', 'DocumentArrayMemmap']:
+        """
+        Alternative match that returns does not append matches
```
tadejsv, line 103: ?

davidbp: This is not meant to be an update for match, just some code to show the bottlenecks

jina/types/arrays/neural_ops.py
```diff
@@ -87,3 +87,63 @@ def match(
...
+                dist = minmax_normalize(dist, normalization)
+
+        m_name = metric_name or (metric.__name__ if callable(metric) else metric)
+
+        docarray = DocumentArray()
+        for ids, dist_vec in zip(idx, dist):
+            for i, distance in zip(ids, dist_vec):
+                d = self[int(i)]
+                d.scores[m_name] = distance
+                docarray.append(d)
```
hanxiao, line 147: without `.matches.append`, returned results like this (i.e. `docarray`) seems useless.

hanxiao:
note that the abs timing number in this PR's body is not reliable anymore after https://github.com/jina-ai/jina/pull/3056 and  https://github.com/jina-ai/jina/pull/3058 , master is now much faster. But the spirit stays, **the most time-consuming part of `match()` is still about `.matches.append()`, which takes >50% time**. So this is still unsolved.

PR #3058, opened 2021-07-31
refactor(proto): remove content_hash from builtin proto

hanxiao:


bwanglzu:
how about add `@cached_property` the content hash?

hanxiao:
> how about add @cached_property the content hash?

no, that will introduce inconsistency issue, leads to more bug. In general, `@cached_property` always leads to inconsistency to some extent, but in this case `Document` fields change much more frequently than other cases. so no.

PR #3069, opened 2021-08-02
refactor: change bind to all inbound

slettner:
This PR introduces the following change:

For all runtimes, the gateway will bind to all inbound connections on the localhost, i.e. `0.0.0.0` instead of the parameter `args.host`. 

This change is motivated by a use-case from kubernetes where the gateway host deployment is hidden behind a kubernetes service with corresponding dns (this is a typical kubernetes pattern). 
Clients can reach the gateway via this dns from the outside but the host cannot bind to this address as it is the IP of the service not the deployment where the gateway runs.

### DIFF
--- a/jina/peapods/runtimes/gateway/grpc/__init__.py
+++ b/jina/peapods/runtimes/gateway/grpc/__init__.py
@@ -2,10 +2,12 @@ import os
 
 import grpc
 
-from ..prefetch import PrefetchCaller
-from ...zmq.asyncio import AsyncNewLoopRuntime
-from ....zmq import AsyncZmqlet
+from jina import __default_host__
+
 from .....proto import jina_pb2_grpc
+from ....zmq import AsyncZmqlet
+from ...zmq.asyncio import AsyncNewLoopRuntime
+from ..prefetch import PrefetchCaller
 
 __all__ = ['GRPCRuntime']
 
@@ -42,7 +44,7 @@ class GRPCRuntime(AsyncNewLoopRuntime):
         self.zmqlet = AsyncZmqlet(self.args, logger=self.logger)
         self._prefetcher = GRPCPrefetchCall(self.args, self.zmqlet)
         jina_pb2_grpc.add_JinaRPCServicer_to_server(self._prefetcher, self.server)
-        bind_addr = f'0.0.0.0:{self.args.port_expose}'
+        bind_addr = f'{__default_host__}:{self.args.port_expose}'
         self.server.add_insecure_port(bind_addr)
         await self.server.start()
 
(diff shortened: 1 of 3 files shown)


maximilianwerk:
@JoanFM @deepankarm any objections?

JoanFM:
it seems good. It surprises me how it used to work before? Binding is always local to one self right?

JoanFM:
What I do not understand is then why `uvicorn` exposes the `host` argument.

In the meanwhile, u can simply set properly the `host` argument right @slettner .

I would wait for @deepankarm input on this

JoanFM (changes requested):
Wait for @deepankarm input

hanxiao:
>  It surprises me how it used to work before? Binding is always local to one self right?

Because `arg.host` by default is `0.0.0.0`

Note that this PR is not complete, if in those contexts `host` is no longer a parameter, then argparse should not expose them to the users. The parser side needs to be refactored a bit to remove `host` for those context.

maximilianwerk:
> > It surprises me how it used to work before? Binding is always local to one self right?
> 
> Because `arg.host` by default is `__default_host__`
> 
> Note that this PR is not complete, if in those contexts `host` is no longer a parameter but a constant, then `argparse` should not expose them to the users. The parser side needs to be refactored a bit to remove `host` for those context.

That's wrong. We still need the `host` to tell the last Pod how to reach the Gateway again. For this, we need to configure the host of the Gateway. With dynamic routing, you never define `host_in` and `host_out` by yourself, but only tell each Pod on which host it runs. The Flow then computes the relevant information for either the `routing_table` or sets `host_in` if we have a remote-local situation.

maximilianwerk:
To give some more motivation, imagine the following Flow:

```
Flow.add(name='p1').add(name='p2').add(name='p3', needs=['p1', 'p2'])
```

You can now easily put them on different machines via setting `host` of each Pod. Anyhow, if you want to configure the `host_in` of `p3`, what would it be - `p1.host` or `p2.host`? Same holds true for `host_out` of `p1` - `p2.host` or `p3.host`?

hanxiao:
> The parser side needs to be refactored a bit to remove host for those context.

`context` means every context mentioned in this PR. Not beyond this PR.

maximilianwerk:
> > The parser side needs to be refactored a bit to remove host for those context.
> 
> `context` means every context mentioned in this PR. Not beyond this PR.
> 
> Every parser behind these `contexts` should remove `host` as argument. Any parser that not related to this PR should stay the same.

The `host` argument is currently used in four places:

- `Daemon` => still needed
- `Pea` => still needed
- `Gateway` => still needed as described above
- `ClientCLI` => still needed or at least this PR does not change anything with the `ClientCLI`

The Gateway also uses the `mixin_http_gateway_parser`, which actually does not has the `host` argument and thus, we don't need to change anything. Perhaps this should have had the `host` argument beforehand, but it does not. `GRPC` and `websocket` do not have their own mixins at all.

Am I getting you right, if you would prefer to split the configuration into further mixins with these arguments specified more detailed and as low level as possible?

JoanFM (approved):

PR #3097, opened 2021-08-04
fix: add sparse support

davidbp:
Add sparse support to distances and neural_ops

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -6,6 +6,7 @@ import numpy as np
 from ... import Document
 from ...importer import ImportExtensions
 from ...math.helper import top_k, minmax_normalize
+from ...logging.predefined import default_logger
 
 if False:
     from .document import DocumentArray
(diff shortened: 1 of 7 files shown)


jina/types/arrays/neural_ops.py
```diff
@@ -113,10 +135,10 @@ def visualize(
 
         """
 
-        x_mat = np.stack(self.get_attributes('embedding'))
+        x_mat = np.stack(self.get_attributes("embedding"))
         assert isinstance(
             x_mat, np.ndarray
-        ), f'Type {type(x_mat)} not currently supported, use np.ndarray embeddings'
+        ), f"Type {type(x_mat)} not currently supported, use np.ndarray embeddings"
```
JoanFM, line 141: single quote

JoanFM (changes requested):

jina/types/arrays/neural_ops.py
davidbp: updated

PR #3111, opened 2021-08-05
feat(daemon): rolling update via client

deepankarm:


daemon/clients/pods.py
```diff
@@ -8,6 +16,35 @@ class AsyncPodClient(AsyncPeaClient):
...
+        async with aiohttp.request(
+            method='PUT',
+            url=f'{self.store_api}/{daemonize(id, self._kind)}',
+            params={
+                'kind': UpdateOperation.ROLLING_UPDATE.value,
+                'dump_path': dump_path,
+            },
+        ) as response:
+            response_json = await response.json()
+            if response.status != HTTPStatus.CREATED:
```
jacobowitz, line 40: Shouldnt this rather be HTTPStatus.OK?

jacobowitz (changes requested):
only 1 comment regarding the HTTP code for the update, looks good otherwise

daemon/clients/pods.py
deepankarm: You're right. Fixed

jacobowitz (approved):
:+1:

PR #3116, opened 2021-08-06
feat: supporting jina hub push --tag and --docker-file

mapleeit:


.github/2.0/cookbooks/Hub.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+You can specify `-t` or `--tags` parameters to tag one executor. In additional, you can specify `-f` or `--docker-file` parameters to use a specific docker file to build your executor.
+
+## 3. Update Executor in JinaHub
+
+Everything is iterating in Internet world. We also provide a way to update your existing executor in JinaHub. To update one executor, you must have both `UUID` and `SECRET`.
+
+```bash
+$ jina hub push [--public/--private] --force <UUID> --secret <SECRET> <path_to_executor_folder>
```
tadejsv, line 66: For better presentation, I think this should be explained before tags

.github/2.0/cookbooks/Hub.md
```diff
@@ -0,0 +1,155 @@
...
+## 1. Create Executor
+
+To create your Executor, you just need to run this command in your terminal:
+
+```terminal
+jina hub new
+```
+
+A wizard will ask you some questions about the Executor. For the basic configuration, you will be asked two things: The Executor’s name and where it should be saved. For this tutorial, we will call ours RequestLogger. And you can save it wherever you want to have your project. The wizard will ask if you want to have a more advanced configuration, but it is unnecessary for this tutorial.
+
```
alaeddine-13, line 35: I think it would be great if we inform users about what the command will generate and maybe provide some details about the the created files
```suggestion
After running the command, a project with the following structure will be generated:
```text
MyExecutor/
├── Dockerfile	        # Optional
├── manifest.yml
├── config.yml
├── README.md
├── requirements.txt
└── executor.py
```

.github/2.0/cookbooks/Hub.md
mapleeit: addressed

.github/2.0/cookbooks/Hub.md
mapleeit: addressed

mapleeit:
@tadej-redstone @alaeddine-13 Thanks for your suggestions. I already made some changes. It could be nice if you can review again.

numb3r3 (approved):
LGTM

alaeddine-13:
LGTM too

PR #3119, opened 2021-08-06
perf(dam): cache memmap handler

hanxiao:
inlining `mmap.mmap` instead of opening it everytime saves about 30% time

on this branch:
```
get embedding ...	get embedding takes 4 seconds (4.64s)
get embedding ...	get embedding takes 4 seconds (4.84s)
get embedding ...	get embedding takes 4 seconds (4.68s)
```

on master (2aee836):
```
get embedding ...	get embedding takes 7 seconds (7.41s)
get embedding ...	get embedding takes 6 seconds (6.66s)
get embedding ...	get embedding takes 6 seconds (6.27s)

```

```python
from tests import random_docs

dam1 = DocumentArrayMemmap('./dam2')
dam1.clear()
dam1.extend(random_docs(100000))

with TimeContext('get embedding'):
    a = dam1.get_attributes('embedding')
    assert len(a) == 100000
```

### DIFF
--- a/jina/types/arrays/memmap.py
+++ b/jina/types/arrays/memmap.py
@@ -17,6 +17,7 @@ from .document import DocumentArrayGetAttrMixin
 from .search_ops import DocumentArraySearchOpsMixin
 from .traversable import TraversableSequence
 from ..document import Document
+from ...helper import cached_property
 
 HEADER_NONE_ENTRY = (-1, -1, -1)
 PAGE_SIZE = mmap.ALLOCATIONGRANULARITY
@@ -65,6 +66,10 @@ class DocumentArrayMemmap(
         self._key_length = key_length
         self._load_header_body()
 
+    @cached_property
+    def _mmap(self):
+        return mmap.mmap(self._body_fileno, length=0)
+
     def reload(self):
         """Reload header of this object from the disk.
 
@@ -104,6 +109,7 @@ class DocumentArrayMemmap(
             if not np.array_equal((r[1], r[2], r[3]), HEADER_NONE_ENTRY)
         }
         self._body_fileno = self._body.fileno()
+
         self._start = 0
         if self._header_map:
             self._start = tmp[-1][1] + tmp[-1][3]
@@ -160,9 +166,8 @@ class DocumentArrayMemmap(
     def __getitem__(self, key: Union[int, str]) -> 'Document':
         if isinstance(key, str):
             pos_info = self._header_map[key]
-            _, p, r, l = pos_info
-            with mmap.mmap(self._body_fileno, offset=p, length=l) as m:
-                return Document(m[r:])
+            _, p, r, r_plus_l = pos_info
+            return Document(self._mmap[(p + r) : (p + r_plus_l)])
         elif isinstance(key, int):
             return self[self._int2str_id(key)]
         else:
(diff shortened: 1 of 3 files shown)


jina/types/arrays/memmap.py
```diff
@@ -65,6 +66,10 @@ def __init__(self, path: str, key_length: int = 36):
         self._key_length = key_length
         self._load_header_body()
 
+    @cached_property
+    def _mmap(self):
+        return mmap.mmap(self._body_fileno, length=0, prot=mmap.PROT_READ)
```
JoanFM, line 71: Are we sure that opening the file only once makes it work when `append` and `update` operations are done?

JoanFM: I do not think this will work when the `file` size is increased.


```
class mmap.mmap(fileno, length, flags=MAP_SHARED, prot=PROT_WRITE|PROT_READ, access=ACCESS_DEFAULT[, offset])
(Unix version) Maps length bytes from the file specified by the file descriptor fileno, and returns a mmap object. If length is 0, the maximum length of the map will be the current size of the file when mmap is called.
```

JoanFM (changes requested):

jina/types/arrays/memmap.py
bwanglzu: we want to invalidate the mmap?

hanxiao: That's true, converting to draft.

jina/types/arrays/memmap.py
```diff
@@ -64,12 +64,12 @@ def __init__(self, path: str, key_length: int = 36):
         self._header_path = os.path.join(path, 'header.bin')
         self._body_path = os.path.join(path, 'body.bin')
         self._key_length = key_length
-        self._invalidated_mmap = False
-        self._last_mmap = None
+        self._invalidated_mmap = False  # a boolean flag to say mmap is invalidated, set on the change of mmap
```
JoanFM, line 67: having only one of these should be enough right? having `self._last_mmap = None` means it was invalidated the last time

jina/types/arrays/memmap.py
```diff
@@ -17,6 +17,7 @@
 from .search_ops import DocumentArraySearchOpsMixin
 from .traversable import TraversableSequence
 from ..document import Document
+from ...helper import cached_property
```
JoanFM, line 20: remove this import

jina/types/arrays/memmap.py
```diff
@@ -218,11 +233,13 @@ def __setitem__(self, key: Union[int, str], value: 'Document') -> None:
                 self.append(value)
                 self._header_map[self._int2str_id(key)] = self._header_map[value.id]
                 del self[value.id]
+                self._last_mmap = None
```
alaeddine-13, line 236: maybe invalidations of `self._last_mmap` should happen inside each disk operation to the DAM, like append and delitem, not in methods that use them

jina/types/arrays/memmap.py
```diff
@@ -156,13 +170,13 @@ def append(self, doc: 'Document', flush: bool = True) -> None:
         if flush:
             self._header.flush()
             self._body.flush()
+        self._last_mmap = None
```
alaeddine-13, line 173: since the invalidation happens in extend, we can invalidate just when flushing I think

JoanFM: every operation that invalidates the cache should invalidate right?

PR #3129, opened 2021-08-09
feat: grpc for requests

jacobowitz:
This PR adds the possibility to use gRPC instead of ZMQ for communicating between Pods. This works only for an entire Flow, its not possible to mix gRPC and ZMQ. Also some of the advanced features of ZMQ like 'send_as_bind' are not availble with grpc. This means all Pods need to be able to connect to each other.

There are a few open questions, I think mainly around what kind of runtime GrpcDataRuntime actually should be. Otherwise this PR should be ready for review

Open design questions:
* ~~The dynamic routing was implemented before in the ZMQlet. Now I also need the same logic in the gRPClet. We should move it to some shared place.~~
* I've duplicated the merging of incomplete messages for head peas (in the pre hook function). Thats kind of ugly, but it seems not easy to unify atm
* ~~The new Runtime actually uses a flag in the ZEDRutime parser. This flag should be put somewhere else, in the Flow parser I guess?~~
* We have now two Grpc based runtimes which is slightly confusing. We may want to change some naming? One runtime is used only in the Gateway. The new runtime will be used for regular pod/peas.
* In case of fan out scenarios (1 to many communication, e.g. with shards and polling ALL) we will serialize one message one time per target, which is quite wasteful as it is the same message


JoanFM:
Some thoughts

> * The dynamic routing was implemented before in the ZMQlet. Now I also need the same logic in the gRPClet. We should move it to some shared place. => We can have a helper class or module, definitely
> * The new Runtime actually uses a flag in the ZEDRuntime parser. This flag should be put somewhere else, in the Flow parser I guess? I think for now it is safe to keep it here, as is a granularity not much exposed to the user.
> * We have now two Grpc based runtimes which is slightly confusing. We may want to change some naming? One runtime is used only in the Gateway. The new runtime will be used for regular pod/peas. I agree that `runtimes` naming should be changed. We can have any name we want and then we can decide exactly which interfaces are needed for `regular` Runtimes and for `gateway` ones.

jina/peapods/peas/__init__.py
```diff
@@ -159,12 +159,15 @@ def __init__(self, args: 'argparse.Namespace'):
...
     def _set_ctrl_adrr(self):
         """Sets control address for different runtimes"""
-        self._zed_runtime_ctrl_address = self.runtime_cls.get_control_address(
+        self.runtime_ctrl_address = self.runtime_cls.get_control_address(
             host=self.args.host,
             port=self.args.port_ctrl,
             docker_kwargs=getattr(self.args, 'docker_kwargs', None),
         )
 
+        if not self.runtime_ctrl_address:
```
JoanFM, line 168: Does it ever happen?

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,266 @@
...
+            args=self.args,
+            message_callback=self._callback,
+            logger=self.logger,
+        )
+
+    def run_forever(self):
+        """Start the `Grpclet`."""
+        self.logger.debug('Run GRPCDataRuntime')
+        self._grpclet_task = self._loop.create_task(self._grpclet.start())
+        try:
```
JoanFM, line 51: I do not think this is needed, in theory in the `__exit__` of Runtime this should be logged

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,266 @@
...
+
+        :return: True if status is ready else False.
+        """
+
+        try:
+            response = Grpclet.send_ctrl_msg(ctrl_address, 'STATUS')
+        except RpcError:
+            return False
+
+        return True
```
JoanFM, line 97: it should check the value of the response? Not sure about this

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,266 @@
...
+        self.logger.debug('Teardown GRPCDataRuntime')
+
+        self._data_request_handler.close()
+        self._loop.stop()
+        self._loop.close()
+
+        super().teardown()
+
+    def _close_grpclet(self):
+        self._loop.create_task(self._grpclet.close())
```
jacobowitz, line 68: I think this is not quite right, but maybe its not dangerous. I should somehow make sure to wait for this task to finish, but I dont know how to do this really. I cant just run it on self._loop as this loop is already running.

jacobowitz:
I actually checked the dynamic routing overlap and noticed its almost none. Most logic is already in the RoutingTable so I believe it should be okay now as is.

jina/peapods/peas/__init__.py
jacobowitz: yes. the new runtime has no ctrl_address and we should use `port_in` instead.

jina/peapods/runtimes/grpc/__init__.py
jacobowitz: With this mechanism there is no real response. All gRPC calls return some empty protobuf struct, which is always the same. So the only thing 'STATUS' actually test here is that we can do the gRPC call. So if some RpcError occurs (no connection etc) we know that status is not ready. Otherwise we assume its okay.

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+                    ('grpc.max_send_message_length', -1),
+                    ('grpc.max_receive_message_length', -1),
+                ],
+            )
+
+        stub = jina_pb2_grpc.JinaDataRequestRPCStub(channel)
+
+        return stub
+
+    async def _add_envelope(self, msg, routing_table):
```
deepankarm, line 97: This doesn't look like an async funtion.

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+            pod_address = target.active_target_pod.full_address
+
+            new_message = await self._add_envelope(msg, target)
+
+            if pod_address not in self._stubs:
+                self._stubs[pod_address] = Grpclet._create_grpc_stub(pod_address)
+
+            try:
+                self.msg_sent += 1
+                self._stubs[pod_address].Call(new_message)
```
deepankarm, line 57: `_create_grpc_stub` by default seems to be async, hence this should be awaited?

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+    async def Call(self, msg, *args):
+        """Processes messages received by the GRPC server
+        :param msg: The received message
+        :param args: Extra positional arguments
+        :return: Empty protobuf struct, necessary to return for protobuf Empty
+        """
+        if self.callback:
+            if inspect.iscoroutinefunction(self.callback):
+                self._loop.create_task(self.callback(msg))
```
deepankarm, line 139: why can we not do `asyncio.create_task(...)` or `await callback` directly if we need to wait for completion. Do we need the loop object?

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+from jina.peapods.grpc import Grpclet
+from jina.peapods.runtimes.base import BaseRuntime
+from jina.peapods.runtimes.request_handlers.data_request_handler import (
+    DataRequestHandler,
+)
+from jina.proto import jina_pb2
+from jina.types.routing.table import RoutingTable
+
+
+class GRPCDataRuntime(BaseRuntime, ABC):
```
deepankarm, line 21: This suits better to inherit from `AsyncNewLoopRuntime`. @JoanFM ?

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+            message_callback=self._callback,
+            logger=self.logger,
+        )
+
+    def run_forever(self):
+        """Start the `Grpclet`."""
+        self.logger.debug('Run GRPCDataRuntime')
+        self._grpclet_task = self._loop.create_task(self._grpclet.start())
+        try:
+            self._loop.run_until_complete(self._grpclet_task)
```
deepankarm, line 50: Don't need to create a task, & then do `run_until_complete`. `self._loop.run_until_complete(self._grpclet.start())` should work.

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        timeout_ns = 1000000000 * timeout if timeout else None
+        now = time.time_ns()
+        while timeout_ns is None or time.time_ns() - now < timeout_ns:
+            if shutdown_event.is_set() or GRPCDataRuntime.is_ready(ctrl_address):
+                return True
+            time.sleep(0.1)
+
+        return False
+
+    def _callback(self, msg: 'Message') -> None:
```
deepankarm, line 145: This seems to be used only in `Grpclet.Call`, why can we not make it `async _callback(...)` & use await syntax inside (instead of using loop to create task)?  And inside `Grpclet.Call` we `await` or `create_task`?

jina/proto/jina.proto
```diff
@@ -360,3 +361,13 @@ service JinaRPC {
     rpc Call (stream RequestProto) returns (stream RequestProto) {
     }
 }
+
+/**
+ * jina gRPC service for DataRequests.
+ */
+service JinaDataRequestRPC {
+    // Pass in a Message, wrapping a DataRequest
+    rpc Call (MessageProto) returns (google.protobuf.Empty) {
```
deepankarm, line 370: Would unary call be enough here and we don't need a stream? And what does `returns ..Empty` mean?

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -16,34 +16,42 @@
 class PrefetchCaller:
     """An async zmq request sender to be used in the Gateway"""
 
-    def __init__(self, args: argparse.Namespace, zmqlet: 'AsyncZmqlet'):
+    def __init__(
+        self,
+        args: argparse.Namespace,
+        zmqlet: 'AsyncZmqlet' = None,
+        grpclet: 'Grpclet' = None,
```
deepankarm, line 23: We can probably accept 1 mandatory arg `iolet` here, which is `Union[AsyncZmqlet, Grpclet]`?

deepankarm (changes requested):

jina/peapods/runtimes/grpc/__init__.py
JoanFM: Maybe, but to be honest the inheritance in Runtime may need to change. If it is useful we can use that.

JoanFM:
Just to make sure I am aligned, the logic to send to ANY of the shards will be handled by this?

jina/peapods/runtimes/grpc/__init__.py
jacobowitz: But I actually want to cancel this task later, I tried the other way around and had some issues that it was never completing

jina/proto/jina.proto
jacobowitz: Every gRPC call needs to be return something, there is nothing like None or void. Empty is kind of the equivalent, it means that there will be an empty protobuf struct sent back as a result (and needs to be returned on the server side as well).
We dont want that really, but if we use RPC semantics I think there is no way around this hack

jacobowitz:
@JoanFM 
Regarding the shards: Not really? The head still needs to know to which shard to send and what its address might be. 
The grpc runtime should then support sending the request, but the POLLING logic itself is not part of this code here

jina/peapods/grpc/__init__.py
jacobowitz: I will use `asyncio.create_task()`.

I am not awaiting the result because I actually don't care for it and want to respond to the gRPC call right away. For backpressure/bottleneck detection blocking here might be useful but I am not sure we should do that here now. Maybe later?

jina/peapods/runtimes/grpc/__init__.py
jacobowitz: I agree that the runtime naming/structure could use some refactoring, but I cant use `AsyncNewLoopRuntime` here. Its inheriting from `AsyncZMQRuntime` which is not correct here. I dont want ZMQ

jina/peapods/runtimes/grpc/__init__.py
jacobowitz: good idea, this gets rid of the coroutine check in the grpclet.
Will do everything as suggested, except the awaiting as the same argument as above goes here: We are not really interested in the result and want to finish the gRPC call asap

jina/peapods/grpc/__init__.py
jacobowitz: Not awaited (if we dont want to wait actually), but I had to wrap it actually in a task

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+        await self._grpc_server.wait_for_termination()
+
+    async def Call(self, msg, *args):
+        """Processes messages received by the GRPC server
+        :param msg: The received message
+        :param args: Extra positional arguments
+        :return: Empty protobuf struct, necessary to return for protobuf Empty
+        """
+        if self.callback:
+            asyncio.create_task(self.callback(msg))
```
winstonww, line 144: LGTM overall, just wondering in this case if we need to check if `self.callback` is a coroutine or a normal function before invoking it using `asyncio.create_task`?

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+
+        return new_message
+
+    async def close(self, *args, **kwargs):
+        """Stop the Grpc server
+        :param args: Extra positional arguments
+        :param kwargs: Extra key-value arguments
+        """
+        self._logger.debug('Close grpc server')
+        await self._grpc_server.stop(grace=True)
```
deepankarm, line 117: [`grace` needs to be an optional float value. ](https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.Server.stop)

jina/peapods/runtimes/grpc/__init__.py
deepankarm: Except for the name. there's no zmq in `AsyncZMQRuntime` anymore :D We can do this later.

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+        self._data_request_handler = DataRequestHandler(args, self.logger)
+        self._grpclet = Grpclet(
+            args=self.args,
+            message_callback=self._callback,
+            logger=self.logger,
+        )
+
+    def run_forever(self):
+        """Start the `Grpclet`."""
+        self.logger.debug('Run GRPCDataRuntime')
```
deepankarm, line 47: leftover debug message?

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+
+        return False
+
+    async def _callback(self, msg: 'Message') -> None:
+        try:
+            msg = self._post_hook(self._handle(self._pre_hook(msg)))
+            asyncio.create_task(self._grpclet.send_message(msg))
+        except RuntimeTerminated:
+            # this is the proper way to end when a terminate signal is sent
+            self._close_grpclet()
```
deepankarm, line 147: await all `_close_grpclet` calls?

```suggestion
            await self._close_grpclet()
```

jina/peapods/grpc/__init__.py
jacobowitz: I just removed this check :D
In theory you are right, but we know all places where this is used now and its always async. If we every want to pass a blocking callback, I think we can add the check then

winstonww: Ah missed the previous commits. Gotcha, will resolve this convo, thanks

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+
+            try:
+                self.msg_sent += 1
+
+                # this wraps the awaitable object from grpc as a coroutine so it can be used as a task
+                # the grpc call function is not a coroutine but some _AioCall
+                async def task_wrapper(new_message, pod_address):
+                    await self._stubs[pod_address].Call(new_message)
+
+                asyncio.create_task(task_wrapper(new_message, pod_address))
```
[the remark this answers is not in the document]
jacobowitz: the grpc call does not create a coroutine for some reason. The returned object is awaitable, but you cant create task with it...

jina/peapods/grpc/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @staticmethod
+    def send_ctrl_msg(pod_address: str, command: str):
+        """
+        Sends a control message via gRPC to pod_address
+        :param pod_address: the pod to send the command to
+        :param command: the command to send (TERMINATE/ACTIVATE/...)
+        :returns: Empty protobuf struct
+        """
+        stub = Grpclet._create_grpc_stub(pod_address, is_async=False)
```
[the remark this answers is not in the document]
jacobowitz: but this is a static method so I dont really want to keep state here? Also the control message are very rare.
Also in the case of the ready check it will make not much of a difference. While its running in the loop all grpc calls will fail as there is no connection. Once everything works, the loop stops

jina/peapods/peas/__init__.py
JoanFM: ah ok, I did not see is port_in

tests/unit/peapods/runtimes/grpc/test_grpc_data_runtime.py
```diff
@@ -32,7 +32,9 @@ def test_grpc_data_runtime(mocker):
     runtime._data_request_handler.handle.assert_called()
 
     runtime.cancel(f'{args.host}:{args.port_in}')
+    runtime_thread.join()
```
JoanFM, line 35: inside the thread, start runtime and context manage it

JoanFM: and a new thread is dangerous

florian-hoenicke (approved):

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -16,34 +17,37 @@
...
     """An async zmq request sender to be used in the Gateway"""
 
-    def __init__(self, args: argparse.Namespace, zmqlet: 'AsyncZmqlet'):
+    def __init__(
+        self, args: argparse.Namespace, iolet: Union['AsyncZmqlet', 'Grpclet']
+    ):
         """
         :param args: args from CLI
-        :param zmqlet: zeromq object
+        :param iolet: One of AsyncZmqlet or Grpclet. Used for sending/receiving data to/from the Flow
```
JoanFM, line 25: to/from the client I would say, not the Flow

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -0,0 +1,290 @@
...
+        self._grpclet_task = self._loop.create_task(self._grpclet.start())
+        self._loop.run_until_complete(self._grpclet_task)
+
+    def teardown(self):
+        """Close the `Grpclet` and `DataRequestHandler`."""
+        self.logger.debug('Teardown GRPCDataRuntime')
+
+        self._data_request_handler.close()
+        start = time.time()
+        while self._pending_tasks and time.time() - start < 1.0:
```
JoanFM, line 63: what is the logic about?

jina/peapods/runtimes/grpc/__init__.py
JoanFM: But for `status` we may want to provide different results right? Maybe not now, but it can be important in the future?

JoanFM (approved):
Couple of minor things, but LGTM

jina/peapods/runtimes/grpc/__init__.py
jacobowitz: I am collecting pending async task during the lifetime (they get removed once completed). In the end I want to wait for a second to let them complete. Thats like a simpel implementation of graceful shutdown

PR #3133, opened 2021-08-09
Add black logo

Roshanjossey:
For using it in zeromq.org


Yongxuanzhang:
is it the same as https://github.com/jina-ai/jina/blob/add-black-logo/docs/_static/logo-dark.svg? 
I think the reason is that although it is said to be 'dark' or 'black', it is actually transparent. And ZMQ uses white color as the background, so that's why we cannot see the dark logo?

PR #3136, opened 2021-08-10
fix(hubble): docker pull progressbar

numb3r3:


jina/hubble/hubio.py
```diff
@@ -584,13 +584,54 @@ def _fetch_meta(
...
+                            task_id,
+                            completed=pg_detail['current'],
+                            total=pg_detail['total'],
+                            description=status,
+                        )
+                    elif not pg_detail:
+                        progress.update(task_id, advance=0, description=status)
+
+            except Exception as ex:
+                raise ex
```
hanxiao, line 627: if it just simply catches and raises then we do not need this `try` block

numb3r3: @hanxiao `try` block has been removed

PR #3147, opened 2021-08-11
test: increase hub test coverage

alaeddine-13:
closes https://github.com/jina-ai/hubble/issues/115


tests/unit/hubble/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM jinaai/jina:latest
```
JoanFM, line 1: this needs to use `test-pip` that is the one we use to test in CI

JoanFM (changes requested):

tests/unit/hubble/Dockerfile
alaeddine-13: changed

hanxiao (approved):

PR #3148, opened 2021-08-11
refactor: no copy on append

JoanFM:
Regading how `append` to Matches behave, I do not see any point on offering `copy` option. I think is better to add it to the list, and then edit afterwards


davidbp (approved):
LGTM!

PR #3157, opened 2021-08-12
feat: add efficient access to embeddings

davidbp:
This PR adds a `.embeddings` property for faster acces to the embeddings of a `DocumentArray` or a `DocumentArrayMemmap` . This is done iterating over the bytes of the embeddings at proto level avoiding the creation of intermediate Python obejcts. 

This PR also provides
- A speed up of 3x on the current `.match` implementation
- A speed up of 2x on the current `.match` implementation with minibatches

### Example:

```
from jina import Document, DocumentArray
import numpy as np
d = Document(embedding=np.random.random(128))
docarray = DocumentArray([d for _ in range(100_000)])
```

Using `get_attributes`
```
%timeit np.vstack(docarray.get_attributes('embedding'))
```
```
1.65 s ± 95.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Using  `.embeddings`
```
%timeit docarray.embeddings
```
```
649 ms ± 16.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Impact on `.match`

```
d1 = Document(embedding=np.array([0,0,0,0,1]))
da_1 = DocumentArray([d1])
da_2 = DocumentArray([Document(embedding=x) for x in np.random.random((1_000_000, 5))])
```

This PR
```
%time da_1.match(da_2, metric=metric, limit=3)
5.54 s ± 209 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Current master branch
```
%time da_1.match(da_2, metric=metric, limit=3)
15.2 s ± 400 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Impact on `.match` with batching

This PR also provides `_get_embeddings` which can be used to get a slice of the embeddings which is relevant for computing in minibatches.

With the current master branch
```
dam = DocumentArrayMemmap('./my-memmap')
dam.extend([Document(embedding=x) for x in np.random.random((100_000, 5)) ])
%timeit da_1.match(dam, metric=metric, limit=3, batch_size=10_000)
```
```
5.03 s ± 278 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
In this PR
```
%timeit da_1.match(dam, metric=metric, limit=3, batch_size=10_000)
```
```
3.26 s ± 159 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Warning 

We have to give user responsability to construct Documents in the DocumentArray  with  embeddings of the same dtype and shape. 

We need to warn users to construct DocumentArrays with embeddings of the same type.
```
d1 = Document(embedding=np.array([1,2,3]))
d2 = Document(embedding=np.array([2.2,2.,5.])
aux = DocumentArray([d1,d2] )
aux.embeddings
array([[                  1,                   2,                   3],
       [4612136378390124954, 4611686018427387904, 4617315517961601024]])
```

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -252,7 +252,9 @@ class DocumentArrayNeuralOpsMixin:
     def embeddings(self) -> np.ndarray:
         """Return a `np.ndarray` stacking all the `embedding` attributes as rows.
 
-        Warning: This operation assumes all embeddings have the same shape.
+        Warning: This operation assumes all embeddings have the same shape and dtype.
+                 All dtype and shape values are assumed to be equal to the values of the
+                 first element in the DocumentArray / DocumentArrayMemmap
 
         Warning: This operation currently does not support sparse arrays.
 
@@ -273,7 +275,9 @@ class DocumentArrayNeuralOpsMixin:
         Example: `self._get_embeddings(10:20)` will return 10 embeddings from positions 10 to 20
                   in the `DocumentArray` or `DocumentArrayMemmap`
 
-        Warning: This operation assumes all embeddings have the same shape.
+        Warning: This operation assumes all embeddings have the same shape and dtype.
+                 All dtype and shape values are assumed to be equal to the values of the
+                 first element in the DocumentArray / DocumentArrayMemmap
 
         Warning: This operation currently does not support sparse arrays.
 
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -102,8 +102,9 @@ class DocumentArrayNeuralOpsMixin:
         is_sparse = False
 
         if isinstance(darray[0].embedding, np.ndarray):
-            x_mat = np.stack(self.embeddings)
-            y_mat = np.stack(darray.embeddings)
+            x_mat = self.embeddings
+            y_mat = darray.embeddings
+
         else:
             import scipy.sparse as sp
 
@@ -150,9 +151,7 @@ class DocumentArrayNeuralOpsMixin:
         def batch_generator(y_darray: 'DocumentArrayMemmap', n_batch: int):
             n_max = len(y_darray)
             for i in range(0, len(y_darray), n_batch):
-                y_mat = np.vstack(
-                    [y_darray[j].embedding for j in range(i, min(i + n_batch, n_max))]
-                )
+                y_mat = y_darray._get_embeddings(slice(i, i + n_batch))
                 yield y_mat, i
 
         y_batch_generator = batch_generator(darray, batch_size)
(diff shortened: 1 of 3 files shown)


jina/types/arrays/neural_ops.py
```diff
@@ -248,3 +248,27 @@ def visualize(
             plt.savefig(output, bbox_inches='tight', pad_inches=0.1)
         else:
             plt.show()
+
+    @property
+    def embeddings(self, indices: Optional[slice] = None) -> np.ndarray:
```
JoanFM, line 253: It assumes all of the documents in slice do have `slides`.

And there is no possibility to add arguments to `property`

tests/unit/types/arrays/test_neural_ops.py
```diff
@@ -392,3 +393,29 @@ def test_match_inclusive_dam(tmpdir):
...
+    da = DocumentArrayMemmap(tmpdir)
+    da.extend(Document(embedding=np.array([1, 2, 3, 4])) for _ in range(100))
+    np.testing.assert_almost_equal(da.get_attributes('embedding'), da.embeddings)
+
+
+def test_da_embeddings_slice(tmpdir):
+    da = DocumentArrayMemmap(tmpdir)
+    da.extend(Document(embedding=np.array([1, 2, 3, 4])) for _ in range(100))
+    np.testing.assert_almost_equal(
+        da.get_attributes('embedding')[10:20], da.embeddings[10:20]
```
JoanFM, line 408: this is not doing what u expect

jina/types/arrays/neural_ops.py
davidbp: True, updated with a private method to handle slice

tests/unit/types/arrays/test_neural_ops.py
davidbp: True, updated tests

jina/types/arrays/neural_ops.py
```diff
@@ -250,11 +249,28 @@ def visualize(
...
 
-        Example: `self.embeddings[10:20]` will return 10 embeddings from positions 10 to 20
+        Warning: This operation assumes all embeddings have the same shape.
+
+        Warning: This operation currently does not support sparse arrays.
+
+        :return: embeddings stacked per row as `np.ndarray`.
+        """
+
+        x_mat = b''.join(d.proto.embedding.dense.buffer for d in self)
```
JoanFM, line 262: I would recommend calling `self._get_embeddings(None)`?

jina/types/arrays/neural_ops.py
```diff
@@ -248,3 +247,48 @@ def visualize(
...
+        :param indices: slice of data from where to retrieve embeddings.
+        :return: embeddings stacked per row as `np.ndarray`.
+        """
+        if indices is None:
+            indices = slice(0, len(self))
+
+        x_mat = bytearray()
+        len_slice = 0
+        for d in self[indices]:
+            x_mat += d.proto.embedding.dense.buffer
```
numb3r3, line 289: I would prefer to use `assert` to guarantee all of embedding is dense. 

```suggestion
            assert d.proto.embedding.dense, 'does not support sparse arrays'
            x_mat += d.proto.embedding.dense.buffer
```

numb3r3 (changes requested):

jina/types/arrays/neural_ops.py
davidbp: That would kill performance. Operations at this level are not usually checked.

jina/types/arrays/neural_ops.py
davidbp: It is a sensible recomendation, but this non-for loop version is a bit more efficient. I just don´t do it in the other case because I can´t (since there are slices involved)

jina/types/arrays/neural_ops.py
```diff
@@ -248,3 +247,48 @@ def visualize(
             plt.savefig(output, bbox_inches='tight', pad_inches=0.1)
         else:
             plt.show()
+
+    @property
+    def embeddings(self) -> np.ndarray:
```
JoanFM, line 252: This does not belong in this mixin, I think it should belong in the GetAttrMixin

JoanFM (changes requested):

jina/types/arrays/neural_ops.py
davidbp: gut GetAttrMixin is only for DocumentArray right? we want this for both.

JoanFM: ok then keep it here for now

jina/types/arrays/neural_ops.py
numb3r3: I agree that the checking procedure should happen at the DocumentArray level.

numb3r3 (approved):
LGTM

JoanFM (approved):

PR #3162, opened 2021-08-13
ci: inject Document overload during CI

alaeddine-13:
closes https://github.com/jina-ai/internal-tasks/issues/144

### DIFF
--- a/scripts/inject-document-props-as-overload.py
+++ b/scripts/inject-document-props-as-overload.py
@@ -1,0 +1,87 @@
+import inspect
+import re
+import warnings
+from operator import itemgetter
+from typing import Optional, Tuple, List
+
+from jina import Document
+
+
+def get_properties(cls) -> List[Tuple[str, Optional[str], Optional[str]]]:
+    src = inspect.getsource(cls)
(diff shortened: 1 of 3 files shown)


scripts/inject-document-props-as-overload.py
```diff
@@ -0,0 +1,87 @@
...
+        for property_name, type_hint, _ in properties
+    ]
+    args_str = ', \n'.join(kwargs + [f'{indent}{indent}**kwargs'])
+    doc_str = '\n'.join(
+        f'{indent}{indent}:param {property_name}: {description}'
+        for property_name, _, description in properties
+    )
+
+    noqa_str = '\n'.join(
+        f'{indent}{indent}.. # noqa: DAR{j}' for j in ['202', '101', '003']
```
JoanFM, line 52: what is this `noqa` doing? It is weird to rely on this being present right?

alaeddine-13: you're right
I just wanted to be extra safe but I don't think we'll ever be in a case of tolerating these darglint errors

JoanFM (approved):

PR #3167, opened 2021-08-15
feat: add split prepare for trainer

bwanglzu:


jina/types/arrays/search_ops.py
```diff
@@ -112,3 +113,21 @@ def shuffle(self, seed: Optional[int] = None) -> 'DocumentArray':
...
+        :return: a dict where Documents with the same value on `attribute` are grouped together, their orders are
+        preserved from the original :class:`DocumentArray`.
+        """
+        from .document import DocumentArray
+
+        rv = defaultdict(DocumentArray)
+        for doc in self:
+            value = doc.tags.get(attribute)
+            if not value:
+                continue
```
nan-wang, line 131: We need to add a warning message when the `attribute` does not exist in the `Document`. At least this skipping behavior should be claimed in the docstring.
```suggestion
                continue
```

nan-wang (changes requested):

jina/types/arrays/search_ops.py
hanxiao: please only label this behavior in docstring

numb3r3 (approved):
LGTM

JoanFM (commented):
Shouldn't this function be generalized to a `groupby` with a predicate as input?

bwanglzu:
@JoanFM can you name an example?

JoanFM:
> @JoanFM can you name an example?

Would it be natural:

```python

da = DocumentArray(...)

groups = da.groupby(lambda x: x.tags['category'])
```

bwanglzu:
this also looks good to me, how do you think cc @nan-wang @hanxiao

hanxiao:
groupby has diff semantics,  https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html

> A groupby operation involves some combination of splitting the object, applying a function, and combining the results.

I would just call this function `split`

jina/types/arrays/search_ops.py
```diff
@@ -112,3 +113,24 @@ def shuffle(self, seed: Optional[int] = None) -> 'DocumentArray':
         from .document import DocumentArray
 
         return DocumentArray(self.sample(len(self), seed=seed))
+
+    def split(self, attribute: str) -> Dict[Any, 'DocumentArray']:
```
hanxiao, line 117: If it is only implemented with splitting by tags, then this arg should be called as `tag`

```suggestion
    def split(self, tag: str) -> Dict[Any, 'DocumentArray']:
```

also in docstring

hanxiao (approved):

PR #3172, opened 2021-08-16
feat: unknown args loggin

maximilianwerk:
This will log a warning, when an unknown argument is used via the JAML syntax inside the Pod arguments. See the tests for examples with warnings.

I am not sure if `warn_unknown_args` and `DEPRECATED_ARGS_MAPPING` belongs into `jina/parsers/helper.py`. Happy about other suggestions

### DIFF
--- a/cli/__init__.py
+++ b/cli/__init__.py
@@ -14,12 +14,9 @@ def _get_run_args(print_args: bool = True):
         args, unknown = parser.parse_known_args()
 
         if unknown:
-            import warnings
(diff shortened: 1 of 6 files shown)


jacobowitz (approved):
lgtm, I was just wondering if there is a way to tell the user where the illegal args was used? It could be in a lot of places? But I guess that would be a fair amount of work to add this type of tracking to the args processing, so probably not for this PR

PR #3174, opened 2021-08-16
feat: add conda recipe script

tadejsv:
This PR adds a script that produces a conda recipe, to be used for creating a conda-forge package.

Some changes from the pypi package:
- only conda-forge packages can be used as dependencies. This creates some version conflicts, which unfortunately makes it impossible to install the `daemon`, `demo`, and `devel` versions.
- conda does not support extras - you can not do `conda install jina[perf]`. Instead what I have done here is to create multiple packages, so that you do `conda install jina-perf` - this is inspired by DVC, which also faces the same problem - see [their recipe](https://github.com/conda-forge/dvc-feedstock/blob/master/recipe/meta.yaml). So in the end, we would offer `jina-core`, `jina-perf` and `jina` (standard)

The process for making our package available on conda-forge would go like this:
1. after we agree on the recipe, I will open a PR on their recipes repo. This will run some CI (which may show some more problems, so that we need to make adjustments to our recipe here), and after that is approved, our package will be available.
2. When that PR is merged, we can also merge this one, since then we know that the format of the script is final [I will also add a CD script to re-generate the recipe on release before that].
3. conda-forge has an auto update bot, which works well for just updating the version, if there are no requirement changes. If there are requirement changes, one of us can go the the feedstock repo and create a PR with the changes (just copy/pasting the recipe from our repo), and merge it. We could consider automating this step, but I think it would create more work than it saves.

Left to do:
- [x] figure out the easiest way to make jinja and pyyaml work together
- [x] add version and package name (jinja) to template - parse from pypi or take from repo?
- [x] parse the SHA string from pypi
- [ ] add workflow action so that bot automatically generates script (on release?)
- [x] test recipe locally 
- [x] open PR on conda-forge feedstock repo


scripts/create-conda-recipe.py
```diff
@@ -135,12 +174,24 @@ def ignore_aliases(self, data):
 }
 
 
+#####################################
+# Write the recipe to conda/meta.yaml
```
JoanFM, line 178:
```suggestion
# Write the recipe to conda/recipe.yaml
```

tadejsv: The recipe file is called `meta.yaml` in conda
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html

JoanFM: But the name of the file where u are writing is wrong then?

JoanFM: Line 196

tadejsv: Oh the name of the file was wrong. Fixed now, thanks

nan-wang (approved):
LGTM👍

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
+import re
+from collections import defaultdict
+from copy import deepcopy
+
+import requests
+import yaml
+from bs4 import BeautifulSoup
+
+
+def get_extra_requires(path):
```
winstonww, line 10: Can you add a test to test `get_extra_requires` and show what this function is doing?

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+    try:
+        with open(path) as fp:
+            extra_deps = defaultdict(set)
+            for k in fp:
+                if k.strip() and not k.startswith('#'):
+                    tags = set()
+                    if ':' in k:
+                        k, v = k.split(':')
```
winstonww, line 18: It might be better to use more descriptive variable names rather than letter variables such as `k` and `v` to improve readability?

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+                    # Can not use extra requirements in conda - i.e. uvicorn[standard]
+                    k[0] = re.sub(r'\[\w+\]', '', k[0])
+
+                    # The docker package is docker-py on conda-forge
+                    if k[0] == 'docker':
+                        k[0] = 'docker-py'
+
+                    # Pytorch package is pytorch-cpu (gpu not needed for demo) on forge
+                    if k[0] == 'torch':
+                        k[0] = 'pytorch-cpu'
```
winstonww, line 38: How about something along the lines of:
```
    k[0] = {'docker': 'docker-py', 'torch': 'pytorch-cpu'}.get(k[0], k[0])
```
This way you only have to extend the dictionary instead of having to modify the  if checks every time?

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+
+                    for t in tags:
+                        extra_deps[t].add(k)
+
+        return extra_deps
+    except FileNotFoundError:
+        return {}
+
+
+class RecipeDumper(yaml.SafeDumper):
```
winstonww, line 51: Can you add a test for `RecipeDumper`?

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def increase_indent(self, flow=False, *args, **kwargs):
+        return super().increase_indent(flow=flow, indentless=False)
+
+
+#######################################################
+# Get requirements from the extra-requirements.txt file
+#######################################################
+
+extra_deps = get_extra_requires('extra-requirements.txt')
```
winstonww, line 70: Is it possible to wrap each block into a function so that each function can be tested?

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+reqs['core'] = extra_deps['core']
+reqs['perf'] = reqs['core'].union(extra_deps['perf'])
+reqs['standard'] = reqs['perf'].union(extra_deps['standard'])
+
+
+######################################
+# Get latest version and SHA from pypi
+######################################
+
+
```
winstonww, line 84: > Is it possible to wrap each block into a function so that each function can be tested? 
Same with the above

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+    if table_row.select_one('th').contents[0] == "SHA256":
+        jina_sha = table_row.select_one('button')['data-clipboard-text']
+
+
+###################
+# Create the recipe
+###################
+
+# Create yaml file as a dictionary
+test_object = {
```
winstonww, line 101: > Is it possible to wrap each block into a function so that each function can be tested? 
Same with the above

scripts/create-conda-recipe.py
```diff
@@ -0,0 +1,188 @@
...
+    },
+    'extra': {'recipe-maintainers': ['tadejsv', 'maateen', 'nan-wang', 'hanxiao']},
+}
+
+
+#####################################
+# Write the recipe to conda/meta.yaml
+#####################################
+
+recipe = yaml.dump(
```
winstonww, line 172: > Is it possible to wrap each block into a function so that each function can be tested? 
Same with the above

winstonww (changes requested):
Great Work! Just a few questions from my side. In general, do we need tests for this PR?

tadejsv:
@winstonww This is not something we need tests for - this script is not part of the Jina package, and it is not meant to be general - it has only one specific purpose, to produce the conda script for Jina.

Writing tests would not really server any purpose here, as this script has only one possible output - the recipe, which needs to be manually inspected anyway. The only reason I wrote the script (instead of the recipe itself directly) is for reproducibility.

maateen (approved):
LGTM

hanxiao (approved):

PR #3184, opened 2021-08-17
perf: use embeddings property in cache

bwanglzu:
add `embeddings` property similar as DA to DAM, use  cache if exist, otherwise build and cache it.

### DIFF
--- a/jina/types/arrays/memmap.py
+++ b/jina/types/arrays/memmap.py
@@ -522,13 +522,13 @@ class DocumentArrayMemmap(
     def embeddings(self) -> np.ndarray:
         """Return a `np.ndarray` stacking all the `embedding` attributes as rows.
 
-        Warning: This operation assumes all embeddings have the same shape and dtype.
-                 All dtype and shape values are assumed to be equal to the values of the
-                 first element in the DocumentArray / DocumentArrayMemmap
-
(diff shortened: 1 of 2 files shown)


jina/types/arrays/memmap.py
```diff
@@ -447,10 +447,10 @@ def get_attributes(self, *fields: str) -> Union[List, List[List]]:
         fields = list(fields)
         if 'embedding' in fields:
             if self._embeddings_memmap is None:
-                embeddings = [doc.get_attributes('embedding') for doc in self]
-                self._embeddings_memmap = np.asarray(embeddings)
+                embeddings = self.embeddings  # type: np.ndarray
```
JoanFM, line 450: Also the `embeddings` property should profit from the `cache` maybe?

jina/types/arrays/neural_ops.py
```diff
@@ -267,12 +267,21 @@ def embeddings(self) -> np.ndarray:
 
         :return: embeddings stacked per row as `np.ndarray`.
         """
+        from .memmap import DocumentArrayMemmap
```
JoanFM, line 270: Let's just reimplement this method in `DocumentArrayMemmap`.

JoanFM (changes requested):

JoanFM (approved):

PR #3188, opened 2021-08-17
refactor(cli): 💥 remove pod alias, pea & pod subcommand

hanxiao:
todo: also track 💥 as breaking changes with release script


JoanFM:
I do not think this PR is needed or good. I think we may only want to hide `zed_runtime`.

So the point here is what we want to offer.

I think when a Flow does.

```python
f = Flow().add(uses='jinahub+docker://foo')
```
The `Process` running in `Docker` should instantiate a `runtime` and not a `Pod` or a `Pea`. This way the different layers separation is clear. And then running inside a `Container` works the same inside `Kubernetes` and with `Peas/Pods`.

But at the same time we want to have the possibility to start a remote `executor` by doing :+1: 

```bash
jina executor --uses....
```

while having this Runtime context managed by a `Pea` or `Pod`. 

The other alternative we have is to change the entrypoint in `ContainerRuntime` to force `jina runtime`, but this affects the freedom of executors in the hub and outside the hub

cli/api.py
```diff
@@ -8,13 +8,21 @@ def executor(args: 'Namespace'):
...
+        uses.startswith('jinahub+docker://')
+        or uses.startswith('docker://')
+        or uses.startswith('jinahub://')
+    ):
+        from jina.peapods.pods.factory import PodFactory
+
+        try:
+            with PodFactory.build_pod(args) as p:
+                p.join()
+        except KeyboardInterrupt:
```
JoanFM, line 22: I think this is already handled inside the Pod context manager

JoanFM (changes requested):

cli/api.py
hanxiao: what is already handled?

hanxiao: keyboard interrupt?

hanxiao: yes, handled, confirm

tests/integration/issues/github_3124/test_cli_executor.py
```diff
@@ -27,7 +27,9 @@ def test_executor_cli_docker(docker_image):
 
 
 def test_zed_runtime_cli_docker(docker_image):
-    process = subprocess.Popen(['jina', 'runtime', '--uses', 'docker://clitest:latest'])
+    process = subprocess.Popen(
+        ['jina', 'executor', '--uses', 'docker://clitest:latest']
```
JoanFM, line 31: but we do not want to fail. Or we expect this not to work? We will need to do changes in the tests in the hub

hanxiao:
i'm closing this PR and cherry pick some content into a new one.

PR #3190, opened 2021-08-17
feat: add setter to embeddings

davidbp:
Allow setting  in `.embeddings` at `DocumentArray` level

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -267,12 +267,22 @@ class DocumentArrayNeuralOpsMixin:
 
         :return: embeddings stacked per row as `np.ndarray`.
         """
-
         x_mat = b''.join(d.proto.embedding.dense.buffer for d in self)
 
         return np.frombuffer(x_mat, dtype=self[0].proto.embedding.dense.dtype).reshape(
             (len(self), self[0].proto.embedding.dense.shape[0])
         )
+
+    @embeddings.setter
+    def embeddings(self, emb: np.ndarray):
+
+        assert len(emb) == len(self), (
+            'the number of rows in the input ({len(emb)}),'
+            'should match the number of Documents ({len(self)})'
+        )
+
+        for d, x in zip(self, emb):
+            d.embedding = x
 
     def _get_embeddings(self, indices: Optional[slice] = None) -> np.ndarray:
         """Return a `np.ndarray` stacking  the `embedding` attributes as rows.
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -255,35 +255,6 @@ class DocumentArrayNeuralOpsMixin:
         else:
             plt.show()
 
-    @property
-    def embeddings(self) -> np.ndarray:
-        """Return a `np.ndarray` stacking all the `embedding` attributes as rows.
-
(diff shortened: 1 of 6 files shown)


jina/types/arrays/neural_ops.py
```diff
@@ -267,13 +267,23 @@ def embeddings(self) -> np.ndarray:
...
         """
-
         x_mat = b''.join(d.proto.embedding.dense.buffer for d in self)
 
         return np.frombuffer(x_mat, dtype=self[0].proto.embedding.dense.dtype).reshape(
             (len(self), self[0].proto.embedding.dense.shape[0])
         )
 
+    @embeddings.setter
+    def embeddings(self, emb: np.ndarray):
```
JoanFM, line 277: not sure it belongs here, but good for now

tests/unit/types/arrays/test_neural_ops.py
```diff
@@ -425,3 +425,16 @@ def test_dam_get_embeddings(tmpdir):
     np.testing.assert_almost_equal(
         da.get_attributes('embedding')[10:20], da._get_embeddings(slice(10, 20))
     )
+
+
+def test_embeddings_setter():
+    emb = np.random.random((100, 128))
+    da = DocumentArray([Document() for _ in range(100)])
+    da.embeddings = emb
+    np.testing.assert_almost_equal(da.embeddings, emb)
```
JoanFM, line 434: can u also assert that one by one each document has the right embedding vector?

JoanFM (changes requested):

tests/unit/types/arrays/test_neural_ops.py
davidbp: I would say both statements are basically the same, right? since to get .embeddings  we iterate over each document and get the embeddig vector

jina/types/arrays/neural_ops.py
davidbp: it it is the only mixin that we have that covers DA and DAM that is why it's here..

tests/unit/types/arrays/test_neural_ops.py
JoanFM: yes, but I want to specifically check that accessing each of them gives same result

tests/unit/types/arrays/test_neural_ops.py
```diff
@@ -433,8 +433,14 @@ def test_embeddings_setter():
     da.embeddings = emb
     np.testing.assert_almost_equal(da.embeddings, emb)
 
+    for x, y in zip(emb, da.embeddings):
```
JoanFM, line 436: THis is not what I mean.
I mean for `for emb, doc in zip(embeddings, da)` assert `doc.embedding == emb` or something like this

JoanFM (changes requested):

tests/unit/types/arrays/test_neural_ops.py
davidbp: Ok, updated with the doc.embedding test

tests/unit/types/arrays/test_neural_ops.py
davidbp: updated

JoanFM (approved):

PR #3193, opened 2021-08-17
perf: save a single snapshot of warning filters and use it

alaeddine-13:
Document's constructor uses `warnings.filterwarnings` which compiles a regex expression and performs remove/insert operations.
In the code snippet 2, `warnings.filterwarnings` consumes 9% of the total time.

image
The goal of this PR is to compile and construct the warnings filter once, save it as a snapshot and then use it each time it's needed.

Benchmark:
1) create a DAM with random docs:
```python
from tests import random_docs

dam = DocumentArrayMemmap('mydam')
dam.extend(random_docs(100000))
```

2) test iteration:
```python

dam = DocumentArrayMemmap('mydam')
with TimeContext('iter'):
    for doc in dam:
        pass
```
on master: `iter ...	iter takes 5 seconds (5.28s)`
on feature branch: `iter ...	iter takes 4 seconds (4.22s)`

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -6,6 +6,7 @@ import os
 import urllib.parse
 import urllib.request
 import warnings
+from copy import copy
 from hashlib import blake2b
 from typing import (
     Iterable,
@@ -25,7 +26,7 @@ from google.protobuf import json_format
 from google.protobuf.field_mask_pb2 import FieldMask
 
 from .converters import png_to_buffer, to_datauri, to_image_blob
-from .helper import versioned, VersionedMixin
+from .helper import versioned, VersionedMixin, get_warning_filters
 from ..mixin import ProtoTypeMixin
 from ..ndarray.generic import NdArray, BaseSparseNdArray
 from ..score import NamedScore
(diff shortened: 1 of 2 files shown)


jina/types/document/__init__.py
```diff
@@ -295,16 +300,15 @@ def _update_doc(d: Dict):
...
-                    warnings.filterwarnings(
-                        'error', 'Unexpected end-group tag', category=RuntimeWarning
-                    )
-                    try:
-                        self._pb_body.ParseFromString(document)
-                    except RuntimeWarning as ex:
-                        raise BadDocType(
-                            f'fail to construct a document from {document}'
-                        ) from ex
+                filters = warnings.filters
```
numb3r3, line 303:
```suggestion
                filters = copy(warnings.filters)
```
I'm not sure whether the `copy` can be ignored here? Maybe you should **add a unit test** for this feature.

numb3r3 (changes requested):

jina/types/document/__init__.py
alaeddine-13: @numb3r3 do you know cases where not using `copy` can break something ?
here, we're just switching the reference of warnings.filters which preserves the integrity of the filters object and performs faster than copying the underlying objects inside the list

JoanFM: I am also not sure it makes so much sense this copy, u will be changing the same underlying object right?

JoanFM: Can u add a tests that forces `ParseFromString` to emit some warnings and capture it properly

JoanFM (changes requested):

hanxiao (changes requested):
hmm, but i would suggest completely remove `filterwarnings` context manager, i.e. completely removing this warning-catch-turn-into-exception logic.

This protection is not exhaustive anyway, as one can always find a case that converts an arbitrary binary string into a random `Document`.

If I remember this correctly, I added this warning-catch heuristics to support implicity `Document(buffer=)` setter, basically allowing `Document(b'blah')` directly in the constructor. So the logic was to first try construct as a full doc, if not success, put it into `.buffer`

Issue #3185, opened 2021-08-17
The core and perf install options do not work with pypi - standard is installed


tadejsv:
Try this: create a new python environment, and execute

```
JINA_PIP_INSTALL_CORE=1 pip install jina
```

You will see that this will install jina with all the **standard** dependencies (for example, fastapi). Note tha this is for 2.0.18, which already has https://github.com/jina-ai/jina/pull/3145 merged in.

I think the reason is that when installing from pypi, the `setup.py` script does not get executed, or simply does not have access to env vars.

Also, here https://github.com/jina-ai/jina/blob/6403c4b19c66ba4ca16d306f09b5de53248fd782/setup.py#L121-L139
the `perf_deps` actually contain only 2 dependencies, and do not contain any core dependencies at all. They should be a union with `core_deps`, just like with `standard_deps`.

hanxiao:
```Dockerfile
FROM python:3.7-slim

RUN JINA_PIP_INSTALL_CORE=1 pip install jina

ENTRYPOINT ["jina"]
```

`docker build --platform amd64 -t jinaai/test-pip .`

works for me

PR #3203, opened 2021-08-18
fix(setup): add core deps to perf deps

hanxiao:
#3185 partially


hanxiao:

PR #3208, opened 2021-08-18
refactor(types): move up embeddings getter setter to mixin

hanxiao:


JoanFM (commented):
We saw this does not work on tests. Somehow for Memmap it does not like to have the property getter and setter implemented at different levels

hanxiao:
aha, seems like a python feature, good to know

PR #3212, opened 2021-08-18
fix: fix listview insert, append

JoanFM:


tests/unit/types/test_list_view.py
```diff
@@ -153,3 +153,26 @@ def test_list_view_access_out_of_bounds(list_proto):
...
+    view = ListView(lv)
+    assert len(view) == 0
+    view.insert(0, 10)
+    assert len(view) == 1
+    assert view[0] == 10
+    view.insert(20, 10)
+    assert len(view) == 2
+    assert view[1] == 10
+    view.insert(0, 20)
+    assert view[0] == 20
```
Yongxuanzhang, line 169:
```suggestion
    view.insert(20, 20)
    assert len(view) == 2
    assert view[1] == 20
    view.insert(0, 30)
    assert len(view) == 3
    assert view[0] == 30
```

Yongxuanzhang: would it be better to insert a different value other than `10`?

tests/unit/types/test_list_view.py
```diff
@@ -153,3 +153,27 @@ def test_list_view_access_out_of_bounds(list_proto):
...
+    view = ListView(lv)
+    assert len(view) == 0
+    view.insert(0, 10)
+    assert len(view) == 1
+    assert view[0] == 10
+    view.insert(20, 20)
+    assert len(view) == 2
+    assert view[1] == 20
+    view.insert(0, 30)
+    assert len(view) == 3
```
JoanFM, line 169:
```suggestion
    assert len(view) == 2
```

hanxiao (approved):

PR #3219, opened 2021-08-19
fix: no warning for metas in executor

maximilianwerk:
If you use `metas` inside an Executor definition, there was a warning, that it is an unknown argument beforehands. Even thought it is not really used inside the argparse, this is the correct way in my opinion.

### DIFF
--- a/cli/autocomplete.py
+++ b/cli/autocomplete.py
@@ -104,7 +104,6 @@ ac_table = {
             '--pea-id',
             '--pea-role',
             '--noblock-on-start',
-            '--metas',
         ],
         'flow': [
             '--help',
@@ -181,7 +180,6 @@ ac_table = {
             '--pea-id',
             '--pea-role',
             '--noblock-on-start',
-            '--metas',
             '--routing-table',
             '--dynamic-routing',
             '--connect-to-predecessor',
@@ -257,7 +255,6 @@ ac_table = {
             '--pea-id',
             '--pea-role',
             '--noblock-on-start',
-            '--metas',
         ],
         'pod': [
             '--help',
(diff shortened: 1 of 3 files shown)


maximilianwerk:
Can I get some opinions here, if that makes sense. So the point is:

`metas` is used as an argument for Peas. If we check for known/unknown arguments the parser is the source of truth. Thus, if we are able to specify `metas` it should consequently also be a parameter, non?

@JoanFM @deepankarm @hanxiao

JoanFM:
I am not sure what's the strategy. I think having the metas defined in the `executor` config is more handy that having it in the `Flow` yaml.

hanxiao (changes requested):
i need time to figure out why this is necessary, will come back later

maximilianwerk:
> i need time to figure out why this is necessary, will come back later

Following line of thought:

The general target of `warn_unknown_args` is to give the user a warning they use a not-existing or deprecated argument. This is usually done via checking the `unknown_args` returning from `argparse`. If `metas` is only used somewhere hidden without defined as a proper argument, it will always be treated as an unknown argument. Since we anyhow rely on the power of argparse, all user defined arguments should be represented by one parser or another.

I am not sure, if I 100% agree with this line of thought above, but users complained, that they did not got any warning, when specifying deprecated arguments. So this was my idea of solving it: Get more rigid with arg parsing.

hanxiao:
the test is incorrect, that YAML in the test is not a valid Flow anyway.

You may want to use the below instead, which gives a valid Flow with one executor and executor has metas 

```python
yaml = '''jtype: Flow
version: 1
executors:
    - uses: 
        jtype: BaseExecutor
        metas:
            name: hello
    '''
with Flow().load_config(yaml):
    pass
```

which gives no warning.

jacobowitz (approved):

PR #3222, opened 2021-08-19
feat(routing): add static routing table feature

jacobowitz:
This adds the new parameter `static_routing_table` to the `Flow`. By default this parameter is disabled. 

If this parameter is enabled, the Flow will precompute the routing table for all Pods and set it via the args. If the routing table is statically set, all pods in this Flow will not send the routing table in the data requests and just use their own static routing table.

This does not work with `external` Pods as we cant set the routing table via args there.

If we want to make this feature work with external Pods as well, we would need to implement some mechanism to distribute the routing table outside of data requests. That might be desirable soon, but I think at this point it is better and safer to do this rather minimal change, which will only do exactly what we need now.


JoanFM (commented):
Did not think it through much, but doesn't it seem that we are using 2 arguments that are not compatible to activate only one feature? it seems we are using 2 bits to provide 1 bit of info?

jina/peapods/runtimes/grpc/__init__.py
```diff
@@ -217,6 +216,16 @@ def _handle(self, msg: Message) -> Message:
 
         return msg
 
+    def _get_expected_parts(self, msg):
+        if msg.is_data_request:
+            if not self._static_routing_table:
+                graph = RoutingTable(msg.envelope.routing_table)
+                return graph.active_target_pod.expected_parts
+            else:
```
JoanFM, line 224: Are we sure if static routing table, that `num_part` will always be properly populated?

jacobowitz:
I used two arguments, because on argument is a Flow global argument which defined 'this flow uses static routing tables' and the other args is defining per pod if it should send the routing table or not. At the moment the second argument is always automatically populated by the Flow depending on its configuration.
In the current implementation this might be unnecessary, but I think we might soon have a situation where pods within the same Flow do this differently (external/not external). It makes it easier to implement then if you dont have to change args. But of course one could also argue that we should only introduce this as needed later then.

jina/peapods/runtimes/grpc/__init__.py
jacobowitz: I think so. At least for all non shared Pods (non external/regular) we know the information beforehand. For head peas it should be derived from `needs` and for tail peas it should be derived from `parallel`. As far as I can see it should be correct. But it might be worth adding more tests checking this, that is true

jacobowitz: Actually it did not work for the `parallell` case because the pea args were not updated after the Flow had set the Pod args. I fixed this and added tests for this case.

jacobowitz:
I've added another optimization for grpc and static routing table: The target pods cant change so they are pre computed in __init__

JoanFM (approved):

PR #3227, opened 2021-08-20
feat: make docs behavior consistent :boom: 

maximilianwerk:
The expected behavior from Executor side should be:

```
if Documents are returned:
    forward them to next Pod
else:
    forward whatever the docs object contains after Executor execution to next Pod
```

Subtasks

- [x] repair all tests (this will test the new behavior sufficiently!)
- [x] adopt cookbooks

### DIFF
--- a/jina/peapods/pods/compound.py
+++ b/jina/peapods/pods/compound.py
@@ -105,7 +105,6 @@ class CompoundPod(BasePod, ExitStack):
             self._enter_pea(self.head_pea)
             for _args in self.replicas_args:
                 _args.noblock_on_start = True
-                _args.polling = PollingType.ALL
                 self._enter_replica(Pod(_args))
             tail_args = self.tail_args
             tail_args.noblock_on_start = True
@@ -119,7 +118,6 @@ class CompoundPod(BasePod, ExitStack):
                 self.head_pea = Pea(head_args)
                 self._enter_pea(self.head_pea)
                 for _args in self.replicas_args:
-                    _args.polling = PollingType.ALL
                     self._enter_replica(Pod(_args))
                 tail_args = self.tail_args
                 self.tail_pea = Pea(tail_args)
@@ -223,6 +221,7 @@ class CompoundPod(BasePod, ExitStack):
             _args.port_ctrl = helper.random_port()
             _args.socket_out = SocketType.PUSH_CONNECT
             _args.socket_in = SocketType.DEALER_CONNECT
+            _args.polling = PollingType.ALL
             _args.dynamic_routing = False
             # ugly trick to avoid Head of Replica to have wrong host in
             tmp_args = copy.deepcopy(_args)
--- a/jina/peapods/runtimes/request_handlers/data_request_handler.py
+++ b/jina/peapods/runtimes/request_handlers/data_request_handler.py
@@ -189,6 +189,11 @@ class DataRequestHandler:
 
     @staticmethod
     def replace_docs(msg, docs):
+        """Replaces the docs in a message with new Documents.
+
+        :param msg: The message object
+        :param docs: the new docs to be used
+        """
         msg.request.docs.clear()
         msg.request.docs.extend(docs)
 
--- a/jina/peapods/runtimes/request_handlers/data_request_handler.py
+++ b/jina/peapods/runtimes/request_handlers/data_request_handler.py
@@ -132,6 +132,15 @@ class DataRequestHandler:
             self.logger.debug(
                 f'skip executor: mismatch request, exec_endpoint: {msg.envelope.header.exec_endpoint}, requests: {self._executor.requests}'
             )
+            if partial_requests:
+                DataRequestHandler.replace_docs(
+                    msg,
+                    docs=_get_docs_from_msg(
+                        msg,
+                        partial_request=partial_requests,
+                        field='docs',
+                    ),
+                )
             return
 
         params = self._parse_params(msg.request.parameters, self._executor.metas.name)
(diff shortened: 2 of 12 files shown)


JoanFM (changes requested):
May u add a test proving ur point? What is the problem trying to be solved here?

maximilianwerk:
> May u add a test proving ur point? What is the problem trying to be solved here?

@jacobowitz can you add the test that shows your problem? Thanks!

jina/peapods/runtimes/request_handlers/data_request_handler.py
```diff
@@ -176,6 +176,9 @@ def handle(
                 # this means the returned DocArray is a completely new one
                 msg.request.docs.clear()
                 msg.request.docs.extend(r_docs)
+        elif partial_requests:
+            msg.request.docs.clear()
```
JoanFM, line 180: Why do u clear and extend? Shouldn't it be equivalent to keep them?

maximilianwerk: not in the `partial_requests` case.

JoanFM: and why do u need to send all the docs. If the meesages that are not the last partial are not forwarded to next Pod?

maximilianwerk: Because it is consistent behavior regarding the Executor. Giving the Executor all Documents inside `docs`, but not forwarding them is confusing.

maximilianwerk: Imagine

```
class MyMergerExecutor):
    @requests
    def merge(self, docs, docs_matrix, **kwargs):
        for partial_docs in docs_matrix:
            for doc in partial_docs:
                assert doc in docs
```

will never fail. Then you expect: Well all Documents will be forwarded, since that is the usual behavior of an Executor.

JoanFM: you have to return them as said in the cookbook for now. I think a more agnostic way to do this is to compare r_docs is not Nona and docs != msg.request.docs?

maximilianwerk: Even if it is in the cookbook, I believe it is not a good design. It is rather confusing.

JoanFM: ok, if partial_requests is only reached for the last partial message because that is skipped in the `pre_hook`

maximilianwerk:
The failing test is actually quite interesting. Imagine:

```
Flow()
.add(name='foo').add(
    **external_args,
    name='external_fake',
    external=True,
    needs=['gateway', 'foo'],
)
```

Without a merging strategy, the second Pod will forward all Documents duplicated. Once from `gateway` and once from `foo`. I would argue, this is to-be-expected. Before, it would forward one RANDOM set of Documents. You don't even have deterministic behavior which one will be forwarded.

This makes the PR even more valid in my eyes.

JoanFM (commented):
Make sure `cookbook` is updated and tests added

maximilianwerk:
Following example

```
from jina import Document, Executor, Flow, requests
import random
import time


class MyExecutor(Executor):
    def __init__(self, value, *args, **kwargs):
        self.value = value

    @requests
    def any(self, docs, *args, **kwargs):
        for doc in docs:
            doc.text = self.value
        sleep_interval = random.random()
        print(sleep_interval)
        time.sleep(sleep_interval)


f = (
    Flow()
    .add(uses='MyExecutor', uses_with={'value': 'a'})
    .add(uses='MyExecutor', uses_with={'value': 'b'}, needs=['gateway'])
    .needs_all()
)

with f:
    result = f.post(on='/search', inputs=[Document(text='0')], return_results=True)
    print({doc.text for doc in result[0].data.docs})
```

produces randomly `{'a'}` or `{'b'}`. I would prefer `{'a', 'b'}`.

jina/peapods/pods/compound.py
```diff
@@ -119,7 +118,6 @@ def start(self) -> 'CompoundPod':
                 self.head_pea = Pea(head_args)
                 self._enter_pea(self.head_pea)
                 for _args in self.replicas_args:
```
JoanFM, line 120: this is important to be kept. For the inner replica pod right?

jina/peapods/runtimes/request_handlers/data_request_handler.py
```diff
@@ -132,18 +132,27 @@ def handle(
             self.logger.debug(
                 f'skip executor: mismatch request, exec_endpoint: {msg.envelope.header.exec_endpoint}, requests: {self._executor.requests}'
             )
+            if partial_requests:
+                DataRequestHandler.replace_docs(
```
JoanFM, line 136: I think here u should do nothing right?

jina/peapods/pods/compound.py
maximilianwerk: It is kept. I just moved it to the place, where all args are set. Inside `_set_replica_args`. I prefer to have only one place in the code, where args are set.

jina/peapods/runtimes/request_handlers/data_request_handler.py
maximilianwerk: Clarified offline. For the record:

The reason I added this: When you do a `Flow.add(name='1').add(name='2', needs['gateway', '1'])` by default it uses the `BaseExecutor` in the HeadPea of `2`. But the `BaseExecutor` has no endpoint defined. Anyhow, it should merge the the docs from both pathways to one big `DocumentArray`.

maximilianwerk:
> Make sure cookbook is updated and tests added

Actually the Cookbook was already stating the new behavior. See https://github.com/jina-ai/jina/blob/master/.github/2.0/cookbooks/Flow.md#joiningmerging.

jacobowitz (approved):
looks good, thank you very much for this improvement :+1:

PR #3232, opened 2021-08-20
perf: define embedding as instance of ndarray

davidbp:
This PR improves setting embedding time

- Before this PR:
```
x =  np.array([1,2,3])
%timeit d.embedding = x
6.28 µs ± 82.5 ns per loop
```
After this PR:
```
x =  np.array([1,2,3])
%timeit d.embedding = x
5.31 µs ± 74 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```


jina/types/document/__init__.py
```diff
@@ -227,6 +227,8 @@ def __init__(
                 assert d.tags['good'] == 'bye'  # true
         """
         self._pb_body = jina_pb2.DocumentProto()
+        self._embedding = NdArray(self._pb_body.embedding)
```
JoanFM, line 230: this will not let the user dynamically change the embedding and its type. Plus it will make every Document object take twice the memory space in terms of embedding

alaeddine-13: I don't think this consumes twice memory because the allocated space is for the protobuf body which is the same

JoanFM: ah okey, we are not getting the value yet. I see.

JoanFM (dismissed):
Ok I understant the PR now. I would rename _embedding to _embedding_wrapper or something

alaeddine-13:
keeping the embedding proto in an embedding wrapper breaks a lot of unit tests. This is because many parts rely on `self._pb_body.ClearField`.
This means that the embedding wrapper needs to be rebuilt each time ClearField is performed, therefore there is no speed up at that point.
Furthermore, the constructor becomes slower.
I suggest just keeping the commit that makes os.getenv static

alaeddine-13:
closing the PR because it's not possible to speed up without breaking tests

PR #3235, opened 2021-08-20
perf: remove environment variable reading from init

davidbp:
This PR removes runtime environment variable read for DenseNDarray creation

- Before this PR
            ```
            %timeit aux = jina.types.ndarray.dense.numpy.DenseNdArray()
            2.81 µs ± 177 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
            ```
            
- After this PR
            ```
            %timeit aux = jina.types.ndarray.dense.numpy.DenseNdArray()
            1.48 µs ± 11.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
            ```

### DIFF
--- a/jina/types/ndarray/dense/numpy.py
+++ b/jina/types/ndarray/dense/numpy.py
@@ -8,6 +8,8 @@ from ....proto import jina_pb2
 
 __all__ = ['BaseDenseNdArray']
 
+QUANTIZE = os.environ.get('JINA_ARRAY_QUANT')
+
 
 class DenseNdArray(BaseDenseNdArray):
     """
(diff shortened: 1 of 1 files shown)


jina/types/ndarray/dense/numpy.py
```diff
@@ -8,6 +8,8 @@
 
 __all__ = ['BaseDenseNdArray']
 
+QUANTIZE = os.environ.get('JINA_ARRAY_QUANT')
```
JoanFM, line 11: I think is better to have this as class attribute so that remote jinas can reset this and just be computed at the first import

JoanFM (approved):
forget what I said

PR #3236, opened 2021-08-21
feat(daemon): add timeout to client

deepankarm:
This PR adds 3 things:
- Adds timeout for JinaDClient
- Sets `/workspace` to `$HOME` in all partial-jinad based containers. 
- Adds `jinahub+docker://Executor` and `jinahub://Executor` schemes to jinadruntime.

### DIFF
--- a/daemon/clients/base.py
+++ b/daemon/clients/base.py
@@ -1,6 +1,7 @@
 import os
 import json
 import asyncio
+from numbers import Number
 from http import HTTPStatus
 from typing import Dict, Optional, Union
 
@@ -34,7 +35,9 @@ class AsyncBaseClient:
         timeout: int = None,
     ):
         self._logger = logger
-        self.timeout = timeout
+        self.timeout = aiohttp.ClientTimeout(
+            timeout if isinstance(timeout, Number) else 10 * 60
+        )
         self.http_uri = f'http://{uri}'
         self.store_api = f'{self.http_uri}{self._endpoint}'
         self.logstream_api = f'ws://{uri}/logstream'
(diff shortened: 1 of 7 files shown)


daemon/clients/base.py
```diff
@@ -34,7 +35,9 @@ def __init__(
         timeout: int = None,
     ):
         self._logger = logger
-        self.timeout = timeout
+        self.timeout = aiohttp.ClientTimeout(
+            timeout if isinstance(timeout, Number) else 10 * 60
```
JoanFM, line 39:
```suggestion
            timeout or 10 * 60
```

JoanFM (changes requested):

daemon/clients/base.py
deepankarm: We need to confirm that it is a number, else default to 10secs

JoanFM: but it can be None or a Number as per API, if the called do not respect that is not its fault right?

deepankarm: If we allow any value it will throw an error, it's better to avoid that.
```python
async with aiohttp.request(
        method='GET',
        url='https://gorest.co.in/public/v1/users/123/posts',
        timeout=aiohttp.ClientTimeout('abc')
    ) as r:
        print(await r.json())

# TypeError: '>' not supported between instances of 'str' and 'int'
```

JoanFM: ok I accept, but I think that if typehimt does not allow str u should assume it does not come, otherwise it is untractable

JoanFM (approved):

Issue #3243, opened 2021-08-23
Enable Grpc runtime for Flows with HTTP and Websocket gateways


jacobowitz:
Enable Grpc runtime for Flows with HTTP and Websocket gateways

PR #3244, opened 2021-08-23
fix(grpc): enable grpc runtime for http and websocket gateway

jacobowitz:
This PR enables the GRPCDataRuntime to be run with HTTP and Websocket Flows. Before those were ignoring `grpc_data_requests` and always starting an AsyncZmqlet.

This fixes https://github.com/jina-ai/jina/issues/3243


florian-hoenicke (approved):
Thanks for this implementation. That's very useful for us.

PR #3249, opened 2021-08-23
cicd: fix test on mac

Yongxuanzhang:


tests/unit/peapods/runtimes/container/test_container_runtime.py
```diff
@@ -210,7 +210,7 @@ def test_tail_host_docker2local():
         .add(name='d13')
     )
     with f:
-        assert getattr(f._pod_nodes['d12'].tail_args, 'host_out') == localhost
```
alaeddine-13, line 213: `localhost` is no more used, maybe the variable should be removed

Yongxuanzhang: just realized it is used in the other 2 tests. Maybe just keep it

PR #3251, opened 2021-08-23
feat(daemon): support envvars

deepankarm:
## Description
A lot of times, we have Flow/Executor yamls that have env vars that need to be set while starting the Flow. This PR adds the support to do the same. The user journey remains the same.

1. Create a workspace with all your files. These files might use env var syntax like `${{VAR_NAME}}` or `$VAR_NAME`.
2. Create a Flow in the above workspace. The change in this PR comes in this step. 
   - We add a new query param `envs`. It supports passing multiple strings, in `key=value` format.
   - These env vars are parsed & set while starting the containers in JinaD.
   - JAML already supports env var substitution, which gets used inside the container, while starting Flow/Executor.

#### JinaDClient
```python
from daemon.clients import JinaDClient

client = JinaDClient(host=HOST, port=PORT)
client.flows.create(workspace_id=workspace_id, filename='flow.yml', envs={'PORT_EXPOSE': 12345, 'VAR_NAME': 'VAR_VALUE'})
```

#### cURL
```bash
curl -X 'POST' \
  'http://localhost:8000/flows?workspace_id=<workspace-id>&filename=flow.yml&envs=PORT_EXPOSE%3D12345&envs=FUNC%3Dfoo'
```

#### Python requests
```python
import requests
requests.post(
    'http://localhost:8000/flows', 
    params=[('workspace_id',  workspace_id), ('filename', filename), ('envs', 'PORT_EXPOSE=23456'), ('envs', 'FUNC=foo')]
)
```

### DIFF
--- a/daemon/api/dependencies.py
+++ b/daemon/api/dependencies.py
@@ -1,29 +1,63 @@
 from pathlib import Path
 from http import HTTPStatus
+from contextlib import ExitStack
 from typing import Dict, List, Optional
 
-from fastapi import HTTPException, UploadFile, File
 from pydantic import FilePath
 from pydantic.errors import PathNotAFileError
+from fastapi import HTTPException, UploadFile, File, Query, Depends
 
 from jina import __docker_host__, Flow
 from jina.enums import PeaRoleType, SocketType, RemoteWorkspaceState
 from jina.helper import cached_property, random_port
-from ..helper import get_workspace_path
+from .. import daemon_logger
 from ..models import DaemonID, FlowModel, PodModel, PeaModel
+from ..helper import get_workspace_path, change_cwd, change_env
 from ..stores import workspace_store as store
+
+
+class Environment:
+    """Parses environment variables to be set inside the containers"""
+
+    _split_char = '='
+    _strip_chars = ''.join(['\r', '\n', '\t'])
+
+    def __init__(self, envs: List[str] = Query([])) -> None:
+        self.vars = {}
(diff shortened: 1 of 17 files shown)


daemon/api/dependencies.py
```diff
@@ -46,13 +80,20 @@ def localpath(self) -> Path:
...
+        Before loading the flow yaml, change CWD to workspace dir.
 
         :return: port_expose
         """
-        f = Flow.load_config(str(self.localpath()))
-        return f.port_expose or random_port()
+        with ExitStack() as stack:
+            if self.envs:
+                for key, val in self.envs.items():
+                    stack.enter_context(change_env(key, val))
```
jacobowitz, line 93: This is changing env variables back and forth on the jinad host and not in the child docker container right? Isnt that dangerous as it could happen concurrently and messes things up?

jacobowitz (commented):
One question regarding the env variable setting, otherwise looks good :)

daemon/api/dependencies.py
deepankarm: Good point. This is changing the env on the main jinad process. We can possibly add an `asyncio.Lock` to avoid other events conflicting. I would do it in a different PR though, as adding this lock via `async..await` needs to be propagated to the endpoint level.

jacobowitz (approved):
lgtm but lets raise another issue for the locking thing for the env setting?

PR #3254, opened 2021-08-24
Feat port renaming

maximilianwerk:
Backwards compatible renaming of the `port_expose` argument. That got more complicated than expected. 

### The following naming changed:

Client args: `port_expose` => `port_gateway`
Gateway: `port_expose` is kept for all three ways of communication: http, websocket and GRPC
remote runtime: `port_expose` => `port_jinad`
jinad: `port_expose` => `port`

### Backwards compatibility

For `port_gateway`, `port` and `port_jinad` I kept `port_expose` as a potential synonym to make it still backwards compatible.

I am not super happy about the overall result and would like to get some opinion, whether you see this as an usability improvement @deepankarm @JoanFM @florian-hoenicke @hanxiao


florian-hoenicke:
I like that we don't have port expose for executors anymore.
This was just confusing in the Kubernetes context.

jina/clients/__init__.py
```diff
@@ -17,7 +17,7 @@
 def Client(
     asyncio: Optional[bool] = False,
     host: Optional[str] = '0.0.0.0',
-    port_expose: Optional[int] = None,
+    port_gateway: Optional[int] = None,
```
deepankarm, line 20: If `Client` always connects to gateway, should we just call it `port`?

jina/parsers/peapods/runtimes/remote.py
```diff
@@ -5,27 +5,80 @@
...
+    _add_host(gp)
 
     gp.add_argument(
-        '--host',
-        type=str,
-        default=__default_host__,
-        help=f'The host address of the runtime, by default it is {__default_host__}.',
+        '--port-jinad',
+        type=int,
+        default=12345,  # helper.random_port(),
```
deepankarm, line 18: Default port for JinaD for now is `8000`, let's set that as the default. But we should have a better default :)

jina/parsers/peapods/runtimes/remote.py
```diff
@@ -5,27 +5,80 @@
...
+        help='The port of the remote machine for usage with JinaD.',
     )
 
+
+def mixin_remote_jinad_parser(parser):
```
deepankarm, line 23: If this is only used in `daemon` codebase, let's move this to `daemon/parser.py` and set the default `port` to `8000`

jina/parsers/peapods/runtimes/remote.py
```diff
@@ -5,27 +5,80 @@
...
+    gp.add_argument(
+        '--port-gateway',
         type=int,
         default=helper.random_port(),
-        help='The port of the host exposed to the public',
+        help='The port of the Gateway, which the client should connect to.',
     )
 
+
+def mixin_grpc_gateway_parser(parser):
```
deepankarm, line 54: Except `proxy` arg. this can be called `mixin_gateway_parser`?

deepankarm (commented):
Should we call `port_expose` as `port_gateway` for the gateway port?

.github/2.0/cookbooks/Flow.md
```diff
@@ -90,7 +90,7 @@ Client:
 ```python
 from jina import Client, Document
 
-c = Client(port_expose=12345)
+c = Client(port_jinad=12345)
```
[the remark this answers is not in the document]
maximilianwerk: Oh yes, that was my mistake. `port_gateway` is the correct one.

jina/parsers/peapods/runtimes/remote.py
maximilianwerk: What do you mean with "Except `proxy` arg"?

jina/clients/__init__.py
maximilianwerk: Hmm fair enough.

jina/parsers/peapods/runtimes/remote.py
maximilianwerk: Got. it. renamed the function.

tests/distributed/test_workspaces/envvars_ws/flow.yml
```diff
@@ -1,7 +1,7 @@
 jtype: Flow
 with:
   protocol: http
-  port_expose: ${{PORT_EXPOSE}}
+  port_expose: $PORT_EXPOSE
```
deepankarm, line 4: This test used `${{PORT_EXPOSE}}`  because we support this syntax and we want to validate if it works.

maximilianwerk: OK, I'll revert.

PR #3266, opened 2021-08-25
fix(daemon): manage ports for remote flows

deepankarm:
## Remote Flow <-> ContainerRuntime (via JinaD)

https://github.com/jina-ai/jina/blob/92eb63fdc113157f5eebcbb6c3e5e53058931829/daemon/api/dependencies.py#L97-L115


daemon/api/dependencies.py
```diff
@@ -77,27 +116,77 @@ def localpath(self) -> Path:
...
+        :return: return filepath to save flow config in
         """
-        Sets `port_expose` for the Flow started in `partial-daemon`.
-        This port needs to be exposed before starting `partial-daemon`, hence set here.
-        If env vars are passed, set them in current env, to make sure all values are replaced.
-        Before loading the flow yaml, change CWD to workspace dir.
+        return get_workspace_path(self.workspace_id, f'{self.id}_{self.filename}')
 
-        :return: port_expose
+    def load_and_dump(self) -> None:
```
deepankarm, line 128: @maximilianwerk This is the discussion we had about `load` & `save` config for Flows before starting containers.

JoanFM (dismissed):

daemon/api/dependencies.py
```diff
@@ -77,27 +85,95 @@ def localpath(self) -> Path:
...
+        """
+        every Flow created inside JinaD lives inside a container. It is important to know the
+        list of ports to be published with localhost before actually starting the container.
+
+        1. `load` the flow yaml here.
+            - yaml is stored in `workspace` directory, so we'll `cd` there
+            - yaml might include env vars. so we'll set them (passed via query params)
+        2. `build` the Flow so that `gateway` gets added.
+            - get the list of ports to be published (port_expose, port_in, port_out, port_ctrl)
```
jacobowitz, line 106: Just FYI: For GprcDataRuntime we dont need port_out and port_ctrl - But it also does not hurt to expose them as well for now. Eventually we might want to be a bit tighter here though

daemon/stores/partial.py
```diff
@@ -75,32 +77,41 @@ def add(self, args: Namespace, **kwargs) -> PartialStoreItem:
...
+
+            self.object: Flow = Flow.load_config(yaml_source).build()
+            self.object.workspace_id = jinad_args.workspace_id
+
+            # In dependencies, we set `runs_in_docker` for the `gateway` and for `CompoundPod` we need
+            # `runs_in_docker` to be False. Since `Flow` args are sent to all Pods, `runs_in_docker` gets set
+            # for the `CompoundPod`, which blocks the requests. Below we unset that (hacky & ugly)
+            for pod in self.object._pod_nodes.values():
+                runtime_cls = update_runtime_cls(pod.args, copy=True).runtime_cls
+                if runtime_cls in ['ZEDRuntime'] + list(GATEWAY_RUNTIME_DICT.values()):
```
jacobowitz, line 110: Lets add a comment here explaining that particular list of runtime classes? E.g. GrpcDataRuntime is not here, which I think is correct as it should not be a CompoundPod.

tests/daemon/unit/api/flow3.yml
```diff
@@ -0,0 +1,20 @@
+jtype: Flow
+with:
```
jacobowitz, line 2: Lets add a test `--grpc-data-requests` and no compound pod?

jacobowitz (changes requested):
Only some small suggestions, but I think @florian-hoenicke should read this PR and check if it works with the way he uses a Flow in K8s? Also this load/dump mechanism might be something useful for his PR as well?

daemon/api/dependencies.py
deepankarm: right. We'll be strict here with port mapping, as we improve on it.

tests/daemon/unit/api/flow3.yml
deepankarm: Done

jacobowitz (approved):
lgtm now :) :+1:

Microsoft-account-png:
Hi gamers

PR #3275, opened 2021-08-26
feat(k8s): add K8s connection pooling for grpc

jacobowitz:
This PR adds the concept of a `ConnectionPool` and its concrete implementations `GrpcConnectionPool`  and`K8sGrpcConnectionPool`. 

The Grpclet will either use `GrpcConnectionPool`  or`K8sGrpcConnectionPool`. The pool itself can be disabled though, which means that the class is still used but it basically behaves as before and just creates/send messages as told.

TODO: 
- [x] Add necessary documentation for `--k8s-connection-pool`
- [x] Check hanging tests


jacobowitz:
@maximilianwerk 
This PR will not allow any kind of sophisticated traffic management schemes. The mechanism here is basically a 'fat client' which will make decision on its own (here: based on what k8s replicas are available).
To support your usecase we would need to introduce some centrally managed service, which can make those decisions and distribute it to the clients. I think this is out of scope for this PR, but we can evolve the mechanism over time in becoming something like this.
For now we would need to use proxy based load balancing to be able to do what you described.

maximilianwerk:
> @maximilianwerk
> This PR will not allow any kind of sophisticated traffic management schemes. The mechanism here is basically a 'fat client' which will make decision on its own (here: based on what k8s replicas are available).
> To support your usecase we would need to introduce some centrally managed service, which can make those decisions and distribute it to the clients. I think this is out of scope for this PR, but we can evolve the mechanism over time in becoming something like this.
> For now we would need to use proxy based load balancing to be able to do what you described.

Totally got this. My comment was rather meant to highlight the possibility to develop the system in this direction.

jina/peapods/pods/k8slib/kubernetes_deployment.py
```diff
@@ -96,6 +96,23 @@ def deploy_service(
         },
         custom_resource_dir=custom_resource_dir,
     )
+
+    logger.info(f'🔋\tCreate necessary permissions"')
```
florian-hoenicke, line 100:
```suggestion
    logger.info(f'🔑\tCreate necessary permissions"')
```

florian-hoenicke: :)

florian-hoenicke:
Great PR! This is a really nice feature, which makes us independent of proxy-based load balancing.
It will introduce a small breaking change we have to be aware of.
People who use custom templates now have to provide these additional templates as well.
We need to make it possible to only use a subset of custom templates in a different pr.

florian-hoenicke (approved):

jacobowitz:
* Changed the log icon as florian suggested
* fixed some issue I found in manual testing with benchmarking
* rebased

jacobowitz:
* Enabled connection pooling by default (if in k8s). I think this should be done because by default there will be no service mesh deployed, so all load balancing mechanisms would be broken without the pool. If the user wants to use a service mesh, they need to disable the pool then

jacobowitz:
> Great PR! This is a really nice feature, which makes us independent of proxy-based load balancing.
> It will introduce a small breaking change we have to be aware of.
> People who use custom templates now have to provide these additional templates as well.
> We need to make it possible to only use a subset of custom templates in a different pr.

Actually its not a breaking change, because I am intentionally not looking in the custom resource directory for those permission templates. This should not be exposed to the users as we only do this for allowing our internal mechanisms like connection pooling to work. 
If we want the user to be able to do custom permissions, we should add this mechanism separately

jina/peapods/networking.py
```diff
@@ -1,8 +1,322 @@
...
+        """
+        pass
+
+    def close(self):
+        """
+        Closes the connection pool
+        """
+        self._connections.clear()
+
+    def _send_message(self, msg: Message, connection):
```
JoanFM, line 136: u can use `abstractmethod` better rather than just NotImplementedError.

JoanFM (commented):
Just a comment,

It seems that the `ConnectionPool` class seems to handle 2 things that are perpendicular (Getting list of connections (replicas or instances) and how to send it).

This would separate K8s, Local, Swarm logic with `GRPC` vs `ZMQ` logic on how messages are sent.

Not a big thing, but as it evolves, we may want to break into 2 traits related to ConnectionListGetters and the actual message sending

jacobowitz:
@JoanFM 
Addressed your comment regarding the abstract method.
I also had to change kubernetes from devel to standard with this [commit](https://github.com/jina-ai/jina/pull/3275/commits/771971ababe999666d7f18a684c3e6b6c2c7c884) . I am not sure about that, but with this change we need kubernetes dependency in every container running in K8s (before this was only necessary where the Flow is created). So I think devel is no longer correct, standard is also not the best, but without adding a new tag I dont see any other option?

I also rebased against master

jacobowitz:
Updated the K8s doc, mainly removing the limitation regarding load balancing as this gets removed by this PR

extra-requirements.txt
```diff
@@ -72,7 +72,7 @@ jsonschema:                 cicd
 dgl:                        cicd
 black==20.8b1:              test
 matplotlib:                 devel
-kubernetes>=18.20.0:        devel
+kubernetes>=18.20.0:        standard, devel
```
JoanFM, line 75: if it is standard, u can remove from `devel` I'd say

jina/peapods/grpc/__init__.py
```diff
@@ -1,11 +1,12 @@
 import argparse
 import asyncio
-from typing import Optional, Callable
+from typing import Optional, Callable, Type
 
 import grpc
 from google.protobuf import struct_pb2
 
 from jina.logging.logger import JinaLogger
+
```
JoanFM, line 9: It is actually weird, in the rest of the project we use relative paths

jina/peapods/grpc/__init__.py
```diff
@@ -26,10 +27,14 @@ def __init__(
         message_callback: Callable[['Message'], None],
         logger: Optional['JinaLogger'] = None,
     ):
+        from jina.peapods.networking import create_connection_pool
```
JoanFM, line 30: why is this import hidden?

JoanFM: Is there any cyclic dependency?

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+    :param port: port to use for the connections
+    """
+
+    def __init__(self, port: int):
+        self.port = port
+        self._connections = []
+        self._address_to_connection_idx = {}
+        self._rr_counter = 0
+
+    def add_connection(self, address: str, connection):
```
JoanFM, line 31: can u add type hint for `connection`?

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+        Remove connection with ip from the connection list
+        :param address: Remove connection for this address
+        :returns: The removed connection or None if there was not any for the given ip
+        """
+        if address in self._address_to_connection_idx:
+            return self._connections.pop(self._address_to_connection_idx.pop(address))
+
+        return None
+
+    def get_connection(self):
```
JoanFM, line 52: should it be named `get_next_connection` maybe?

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+
+class ConnectionPool:
+    """
+    Manages a list of connections.
+
+    :param logger: the logger to use
+    :param on_demand_connection: Flag to indicate if connections should be created on demand
+    """
+
+    def __init__(self, logger: JinaLogger = None, on_demand_connection=True):
```
JoanFM, line 94:
```suggestion
    def __init__(self, logger: Optional[JinaLogger] = None, on_demand_connection=True):
```

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+    :param namespace: K8s namespace to operate in
+    :param client: K8s client
+    :param logger: the logger to use
+    """
+
+    def __init__(
+        self,
+        namespace: str,
+        client,
+        logger: JinaLogger = None,
```
JoanFM, line 186:
```suggestion
        logger: Optional[JinaLogger] = None,
```

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+
+    :param namespace: K8s namespace to operate in
+    :param client: K8s client
+    :param logger: the logger to use
+    """
+
+    def __init__(
+        self,
+        namespace: str,
+        client,
```
JoanFM, line 185: type hint for `client`

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+        self._k8s_event_queue = asyncio.Queue()
+        self.enabled = False
+
+        self._fetch_initial_state()
+
+        from kubernetes import watch
+
+        self._api_watch = watch.Watch()
+
+        self.update_thread = Thread(target=self.run, daemon=True)
```
JoanFM, line 202: I panic when I see `threads` but I guess there is no better alternative

jina/peapods/networking.py
```diff
@@ -1,8 +1,341 @@
...
+        while self.enabled:
+            event = await self._k8s_event_queue.get()
+            self._process_item(event)
+
+    def run(self):
+        """
+        Subscribes on MODIFIED events from list_namespaced_pod AK8s PI
+        """
+
+        self.enabled = True
```
JoanFM, line 227: Why is this important? I think that the correct way to do things would be to instantiate the `_k8s_client` locally in the `stack` of being run better than being a shared resource between the process and its thread

JoanFM: Same for `loop` Is not clear to me which are the `shared` resources with the `thread`. And are they safe to be accessed without `locking`?

JoanFM (changes requested):

florian-hoenicke (dismissed):

jina/peapods/grpc/__init__.py
jacobowitz: indeed, I checked and it seems we are using both? But absolute should be more correct, no?

jina/peapods/networking.py
jacobowitz: actually I can't, because connection can be anything from this point of view (grpc stubs, zmq sockets, raw sockets, ..). Also this class does not interact with the object at all, it just stores and returns the object.

jina/peapods/networking.py
jacobowitz: yeah its annoying, but we need the `Thread` for the blocking call in the kubernetes API due to https://github.com/kubernetes-client/python/issues/323

jina/peapods/networking.py
jacobowitz: Lets clear up the shared client in a follow up issue.
The loop I need to do 
`asyncio.run_coroutine_threadsafe(self._k8s_event_queue.put(event['object']), self._loop )`

I use an asyncio.Queue for communication between the Threads and this Queue is not threadsafe so I use the loop here to access the queue in a safe (but ugly) manner

deepankarm (commented):
Minor comments

extra-requirements.txt
jacobowitz: actually it fails if I remove it? like here https://github.com/jina-ai/jina/pull/3275/checks?check_run_id=3719205391

jacobowitz: also it seems that everything in standard is also in devel atm? (and daemon but I think thats not necessary here)

jina/parsers/peapods/runtimes/zed.py
```diff
@@ -202,3 +202,22 @@ def mixin_zed_runtime_parser(parser):
         if _SHOW_ALL_ARGS
         else argparse.SUPPRESS,
     )
+
+    gp.add_argument(
+        '--k8s-connection-pool',
+        action='store_true',
+        default=True,
```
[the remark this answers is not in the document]
jacobowitz: it should be unless --k8s-no-connection-pool is specified? Maybe I made a mistake there, but its the same pattern we use for [--dynamic-routing](https://github.com/jina-ai/jina/blob/master/jina/parsers/__init__.py#L105)

extra-requirements.txt
jacobowitz: not sure that works? all other requirements are basically standard, daemon, devel?

JoanFM (approved):

JoanFM (changes requested):
Wait for windows release to be made

JoanFM (approved):

extra-requirements.txt
hanxiao: is our k8s support mature enough to be in `standard`? its docs is still in the experimental section now.

suggest remove `standard`, keep `devel`

JoanFM: This relates to a problem raised in hubble slack channel. Kubernetes is now needed in the hub executors.

hanxiao: > Kubernetes is now needed in the hub executors.
what does this mean? do you mean Hub executor should do `pip install kubernetes`  in each Dockerfile?

hanxiao (approved):
there are some styling issues on relative imports, but overall good to go

PR #3288, opened 2021-08-30
Update.py

Aabid02:
The code which are committed suggest these changes.
[ LINE 47,51 ] - Changed from pathlib import path - to - fs from open_fs 
Reason - Compatibility - Considering pathlib interface can point to object by passing an object. However If one needs to pass both fs and path inside. In this condition problem arises because here data consist of two different files having a different suffix. For such instance, using fs.path manipulates paths. This Library makes sures that underlying storage recieves correctly formatted path.

[ LINE 59 ] - In PyFileSystem the abstraction is not a path but a directory. Like we have replaced pathlib to FS. So open_fs ( ' . ' ) returns FS object for the current working directory. This object contains methods for making directories and moving files.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,11 +44,11 @@ except FileNotFoundError:
 
 
 def register_ac():
-    from pathlib import Path
+    from fs import open_fs
     import os
     import re
 
-    home = str(Path.home())
+    home = str(open_fs.home())
     resource_path = 'jina/resources/completions/jina.%s'
     regex = r'#\sJINA_CLI_BEGIN(.*)#\sJINA_CLI_END'
     _check = {'zsh': '.zshrc', 'bash': '.bashrc', 'fish': '.fish'}
@@ -56,7 +56,7 @@ def register_ac():
     def add_ac(k, v):
         v_fp = os.path.join(home, v)
         if os.path.exists(v_fp):
-            with open(v_fp) as fp, open(resource_path % k) as fr:
+            with open('.') as fp, open(resource_path % k) as fr:
                 sh_content = fp.read()
                 if re.findall(regex, sh_content, flags=re.S):
                     _sh_content = re.sub(regex, fr.read(), sh_content, flags=re.S)


Mazonh2020:
Git nit

Aabid02:
> Git nit
> 

Please specify the reason for the above comment

setup.py
```diff
@@ -44,19 +44,19 @@
...
+    home = str(open_fs.home())
     resource_path = 'jina/resources/completions/jina.%s'
     regex = r'#\sJINA_CLI_BEGIN(.*)#\sJINA_CLI_END'
     _check = {'zsh': '.zshrc', 'bash': '.bashrc', 'fish': '.fish'}
 
     def add_ac(k, v):
         v_fp = os.path.join(home, v)
         if os.path.exists(v_fp):
-            with open(v_fp) as fp, open(resource_path % k) as fr:
+            with open('.') as fp, open(resource_path % k) as fr:
```
hanxiao, line 59: it is designed to write under `~/`, so this change is wrong.

hanxiao:
> However If one needs to pass both fs and path inside.

For `setup.py`, it is a fixed procedure, user can not pass anything here.

hanxiao:
closing, as this change is unnecessary.

Aabid02:
> > However If one needs to pass both fs and path inside.
> 
> For `setup.py`, it is a fixed procedure, user can not pass anything here.

One can use enviroment variables, these however needs to be set on windows while prefixing them works on Linux/OS X (FOO=bar python setup.py) And we can extend disutils own cmd_class which can implement new features, so one can use it to change variables in given script. (python setup.py foo install) will execute the foo command. Mreover using argparse is useful too ,it helps passing the remaining arguments into other script or program.

setup.py
Aabid02: However a content type can be specified by rst extension or markdown corresponding to no formatting incurs in code.

PR #3296, opened 2021-08-30
Refactor don't resolve host

florian-hoenicke:
As discussed with @jacobowitz, we can not see any benefit of resolving the host. We rather lose information by converting the URL to an IP. 
In addition, the flow context can not know the actual IP when the pod is orchestrated by another system like k8s.


jacobowitz (approved):
I think that's a very good idea, I dont see a reason to resolve the names. Both gRPC and zmq should be able to do it on their own. E.g. zmq (http://api.zeromq.org/3-2:zmq-tcp): 

A peer address may be specified by either of the following:

```
The DNS name of the peer.
The IPv4 or IPv6 address of the peer, in its numeric representation.
```

ggg112234 (approved):

PR #3300, opened 2021-08-30
feat(math): add LSH matching

konstantinkutzkov:
A prototype implementation of locality sensitive hashing for fast similarity search.

### DIFF
--- a/jina/math/lsh.py
+++ b/jina/math/lsh.py
@@ -1,0 +1,53 @@
+import numpy as np
+
+class LSH:
+    """
+    param: k: the number of hyperplanes
+    param: d: the dimensionality of the random hyperplanes 
+    """
+
+    def __init__(self, k, d):
+        self.k = k
+        self.d = d
+        self.H = np.random.normal(0, 1, (k, d))
+
+    def hash_to_binary_code(self, U, nr_perms, nr_coordinates):
+        """
+        param: U: data to be distributed to bins
+        param: nr_perms: how many permutations to use
+        param: nr_coordinates: the prefix length in each permutation
+        """
+        assert self.H.shape[1] == U.shape[1]
+        hashes = np.sign(self.H.dot(U.T))
+        hashes = hashes.T
+        print("hashes shape", hashes.shape)
+        bins = [{} for _ in range(nr_perms)]
+        random_seed = []
+        for perm in range(nr_perms):
+            s = np.random.choice(np.arange(hashes.shape[1]), size=nr_coordinates, replace=False)
+            random_seed.append(s)
+            for i in range(hashes.shape[0]):
+                t=tuple(hashes[i][s])
+                bins[perm].setdefault(t, [])
+                bins[perm][t].append(i)
+        return bins, random_seed
+
+
+    def get_candidates_for_queries(self, Q, bins, random_seed):
+        """
+        param: Q: data matrix with the query vectors
+        param: bins: the bins with vectors in the different permutations
+        param: random_seed: the permutations, i.e. the bit positions, defined when hashing the data
+        """
+        assert self.H.shape[1] == Q.shape[1]
+        hashes = np.sign(self.H.dot(Q.T))
+        hashes = hashes.T
+        candidates = {}
+        for bin, s in enumerate(random_seed):
+            for i in range(hashes.shape[0]):
+                t=tuple(hashes[i][s])
+                if t in bins[bin]:
+                    candidates.setdefault(i, set())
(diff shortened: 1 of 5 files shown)


bwanglzu:
hi @kingkonk81 thanks so much for contributing! However we need more discussion on that, should LSH part of core or an executor, and the implementation might have performance issue, need more benchmarking. If possible can you create an issue, raise some discussion first then go to the real implementation?

bwanglzu:
@kingkonk81 we decided to not add LSH into jina core, but jina executors. Are you interested in raise a PR in [executor repo](https://github.com/jina-ai/executors)?

kingkonk81:
Hello Wang,

This was a "fake pull request" as part of my application for the machine
learning scientist position.

I don't think that creating an executor would be right way for doing LSH.
At a high level we need to create an inverted index for documents that
allows fast retrieval of candidate documents. With my (rather limited)
understanding of Jina, I would say this should go into Document and
DocumentArray. Each document should have a bucket id, similarly to its id
but the number of buckets will be much smaller than the number of
documents. The DocumentArray class should support efficient queries of the
form: "Give me all documents that belong to bucket <bucket id>." The latter
would involve a method that creates a dictionary with buckets as keys and
document ids in each bucket as values. This is similar to
_update_id_to_index_map(self) which, as written in the method
documentation, is costly. Also, other methods will need to be updated like
the ones for deletion and insertion, and new methods created.

As you see this would require discussions and planning with your engineers
as it goes into the core of Jina. And I don't know if you would be at all
interested in having an indexing scheme like this one. Moreover, there are
other options for fast similarity retrieval. I can create an issue if you
would like to have a more formal discussion on the topic.

Best,
Konstantin

jina/math/lsh.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        param: U: data to be distributed to bins
+        param: nr_perms: how many permutations to use
+        param: nr_coordinates: the prefix length in each permutation
+        """
+        assert self.H.shape[1] == U.shape[1]
+        hashes = np.sign(self.H.dot(U.T))
+        hashes = hashes.T
+        print("hashes shape", hashes.shape)
+        bins = [{} for _ in range(nr_perms)]
```
bwanglzu, line 24: `defaultdict` is a better choice: https://docs.python.org/3/library/collections.html

jina/parsers/helloworld.py
```diff
@@ -139,10 +139,17 @@ def set_hw_parser(parser=None):
...
     gp.add_argument(
-        '--top-k', type=int, default=50, help='Top-k results to retrieve and visualize'
+        '--top-k', type=int, default=30, help='Top-k results to retrieve and visualize'
+    )
+    gp.add_argument(
+        '--lsh',
+        type=str,
+        default=False,
+        choices=["False", "True"],
+        help='Use locality sensitive hashing for nearest neighbor search'
```
bwanglzu, line 152: same, do not touch until proven it works.

jina/types/arrays/neural_ops.py
```diff
@@ -85,6 +103,71 @@ def match(
...
+                                all values will be rescaled into range `[a, b]`.
+        :param nr_hashes: the number od separating hyperplanes in LSH
+        """
+
+        if isinstance(darray[0].embedding, np.ndarray):
+            x_mat = self.embeddings
+            y_mat = darray.embeddings
+
+        else:
+            raise Exception("Not implemented")
```
bwanglzu, line 127: never use `Exception`, use `NotImplementedError`

bwanglzu (changes requested):
hi @kingkonk81 I left some comments in the PR. And as discussed, this won't be able to merge into master. Sorry I miss some context before, now I knows.

In general, my feeling is that the code quality needs to be improved. No unitest, lack of docstrings, sometimes code is a bit difficult to read. From functional point of view, we might needs some explanation text in PR that indicates locally sensitive hashing works better than current implementation, with some figures etc. 

Before proving it works, we should not touch helloworld examples, since this is the first thing a user will do when they download Jina.

For the inverted index part. Maybe we need a inverted index at some point, I'm not sure about your idea on implementing it inside `Document` and `DocumentArray`. As far as I can imaging, it should be a separate thing.

JoanFM:
I think this LSH should go into a specific executor, where an Indexer based on `DocumentArray` is built using LSH

JoanFM:
Hello @konstantinkutzkov ,

This sort of behavior should be moved into some specific `Executor`. You are more than welcome to push such an `Executor` to our hub with this implementation.

Thanks for the contribution

PR #3304, opened 2021-08-31
fix(daemon): port mapping for compound pods

deepankarm:
This addresses issues with port-mapping, when inside a remote Flow, we start a CompoundPod which uses ContainerRuntime. So the `head` & `tail` Peas live inside the Flow container & they need to be set/exposed correctly inside main-daemon & partial-daemon. With this PR, we have introduced `PortMappings` which take a json of `pod_name`, `pea_name` & corresponding Ports.


jacobowitz (approved):
I cant follow in every detail, but generally looks like a good solution for compound pods to me :+1:

PR #3310, opened 2021-08-31
fix:  throw exception for positional arguments

numb3r3:
**In current PR, i removed the CI workflow for macOS, as the macOS runner does not support Docker natively. **

- [x] catch illegal positional argument, e.g., `f = Flow().add('docker://dummy_executor')` will throw a `ValueError` exception
- [x] ci tests on MacOS runner (**status**: `tests/unit/peapods/runtimes` failed on CI workflow)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,8 +233,10 @@ jobs:
         run: |
           sudo apt-get install jq
           echo "::set-output name=matrix::$(bash scripts/get-all-test-paths.sh)"
+          echo "::set-output name=macos::$(bash scripts/get-all-test-paths.sh macos)"
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
+      macos: ${{ steps.set-matrix.outputs.macos }}
 
   core-test:
     needs: prep-testbed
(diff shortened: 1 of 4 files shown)


tadejsv:
Python already has a feature for that
https://stackoverflow.com/questions/2965271/forced-naming-of-parameters-in-python/14298976#14298976

hanxiao:
Tadej is right, simply use `*`

numb3r3:
> Python already has a feature for that
> https://stackoverflow.com/questions/2965271/forced-naming-of-parameters-in-python/14298976#14298976

@tadej-redstone @hanxiao Here, I use `*args` to identify the illegal parameters, which will enable us to **throw an explicit exception**. If we just want to ignore them without any warning/exceptions, I'm ok with simple `*`

.github/workflows/ci.yml
```diff
@@ -315,6 +315,46 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Test
+        id: test
+        run: |
+          pytest --suppress-no-test-exit-code --force-flaky --min-passes 1 --max-runs 5 --timeout=360 -v -s tests/system/*.py
+        timeout-minutes: 20
+        env:
+          JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
+          JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
tadejsv, line 356: Probably not needed?

.github/workflows/ci.yml
```diff
@@ -315,6 +315,46 @@ jobs:
...
+          docker-machine create default
+          docker-machine env default
+          eval $(docker-machine env default)
+          brew services start docker-machine
+          docker build -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
+          python -m pip install --upgrade pip
+          python -m pip install wheel
+          pip install ".[standard,test,cicd]" --no-cache-dir
+          jina
+          export JINA_LOG_LEVEL="ERROR"
```
tadejsv, line 345: Because each step in github actions is run in its own process, this will not persist. Set it when you do the test, or globally for the whole job

.github/workflows/ci.yml
```diff
@@ -315,6 +315,46 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Prepare enviroment
+        run: |
+          brew install docker-machine docker
+          docker-machine create default
+          docker-machine env default
+          eval $(docker-machine env default)
+          brew services start docker-machine
+          docker build -f Dockerfiles/pip.Dockerfile -t jinaai/jina:test-pip .
```
tadejsv, line 340: So this just tests that the image is built, right? It is not used for testing?

numb3r3: Sorry, this PR is not ready yet.

tadejsv: Oh ok

hanxiao (approved):

PR #3312, opened 2021-08-31
fix: rolling update with container runtime

deepankarm:


jina/flow/base.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         daemon: Optional[bool] = False,
         default_swagger_ui: Optional[bool] = False,
         description: Optional[str] = None,
+        dump_path: Optional[str] = '',
```
cristianmtr, line 113: So there's no way to have dump_path be preserved in the args without adding it here? It might be seen as polluting the namespace, with something that is very executor-specific

cristianmtr (dismissed):

jina/flow/base.py
deepankarm: We can hide it using `SHOW_ALL_ARGS`. Let me change that.

deepankarm: Removed using `SHOW_ALL_ARGS`

tests/distributed/test_rolling_update_container_runtime/test_rolling_update.py
```diff
@@ -0,0 +1,111 @@
...
+    query_flow_id = client.flows.create(
+        workspace_id=workspace_id,
+        filename='flow_query.yml',
+        envs={'JINAD_WORKSPACE': f'/tmp/jinad/{workspace_id}'},
+    )
+    return dbms_flow_id, query_flow_id, workspace_id
+
+
+@pytest.mark.parametrize('docker_compose', [compose_yml], indirect=['docker_compose'])
+def test_dump_dbms_remote(executor_images, docker_compose):
```
cristianmtr, line 57: Can u parametrize this to show what works and what doesn't? 

`replicas = 1,2`
`parallel = 1,2`

cristianmtr (changes requested):

cristianmtr (approved):

PR #3314, opened 2021-08-31
perf: load image to array in cache sizes

davidbp:
This PR fills in images in 'mini-batches' (actually mini-chunks of each image) to avoid going over  the standard L2-cache. This provides almost 2x performance and less memory allocations.

Benchmark
```
from jina.types.document.converters import to_image_blob
%timeit im_numpy = to_image_blob(`path_to_photo.jpg`)
```

Master branch
```
203 ms ± 12 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This PR:
```
114 ms ± 1.79 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### DIFF
--- a/jina/types/document/converters.py
+++ b/jina/types/document/converters.py
@@ -78,29 +78,6 @@ def png_to_buffer(arr: 'np.ndarray', width: int, height: int, resize_method: str
     return png_bytes
 
 
-def _raw_img_to_numpy(raw_img):
-    """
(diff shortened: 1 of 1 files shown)


anukaal (approved):

JoanFM:
As discussed, this optimizations should be moved maybe in some specific `Executor` and not being taken care by `core`

PR #3324, opened 2021-09-02
feat(flow): add install_requirments args when using jinahub

numb3r3:
This PR aims to pass `install_requirments` arg to enable `f = Flow().add(uses="jinahub://dummy", install_requirements=True)`


tests/integration/hub_usage/dummyhub/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jina
```
hanxiao, line 1: do not use `jina` in any of our own CICD, this pollutes our download statistics for Dev Rel team.

hanxiao (changes requested):
having `jina` package installed in CICD is forbidden, as it pollutes our statistics

tests/integration/hub_usage/dummyhub/requirements.txt
numb3r3: addressed in latest commit

PR #3327, opened 2021-09-02
perf: Document embedding getter

davidbp:
This PR doubles the performance for retrieving embeddings using `doc.embedding` without any change on the API.

This is achieved removing the instantiation of  a `NdArray`  created only to get it's `.value` (which is a numpy array), essentially doubling performance.

Benchmark

```
d = Document(embedding = np.random.rand((128)))
d.embedding
```

- This PR
      ```
      %timeit aux = d.embedding
      2.4 µs ± 50.6 ns per loop
      ```

- Master
      ```
       %timeit aux = d.embedding
       6.2 µs ± 173 ns per loop 
      ```

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -531,7 +531,13 @@ class Document(ProtoTypeMixin, VersionedMixin):
 
         :return: the embedding of this Document
         """
-        return NdArray(self._pb_body.embedding).value
+        if self._pb_body.embedding.HasField('dense'):
+            dense_proto = self._pb_body.embedding.dense
+            return np.frombuffer(dense_proto.buffer, dtype=dense_proto.dtype).reshape(
+                dense_proto.shape
+            )
+        else:
+            return NdArray(self._pb_body.embedding).value
 
     def get_sparse_embedding(
         self, sparse_ndarray_cls_type: Type[BaseSparseNdArray], **kwargs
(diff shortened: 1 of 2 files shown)


numb3r3 (approved):
LGTM 👍

jina/types/document/__init__.py
```diff
@@ -531,7 +531,13 @@ def embedding(self) -> 'SparseArrayType':
 
         :return: the embedding of this Document
         """
-        return NdArray(self._pb_body.embedding).value
+        if self._pb_body.embedding.HasField('dense'):
+            dense_proto = self._pb_body.embedding.dense
+            return np.frombuffer(dense_proto.buffer, dtype=dense_proto.dtype).reshape(
+                dense_proto.shape
+            )
```
hanxiao, line 538: this is simply `DenseNdArray(self._pb_body.embedding).value`

hanxiao (changes requested):
tend to reject, as this basically bypass the NdArray class we implemented, sacrificing all quantization and corner case coverage.

jina/types/document/__init__.py
davidbp: > tend to reject, as this basically bypasses the `NdArray` class we implemented, sacrificing all quantization and corner case coverage, i.e. use simpler implementation to trade for faster speed.
> 
> I believe the key is to make `NdArray` class faster as we would still need it in advance usecase.

I see, Maybe we can do the same using`NdArray` but instead of creating an instance every time we can make it so that a `staticmethod` returns the numpy array. Will investigate further.

PR #3330, opened 2021-09-02
docs: add recommend way to perform heavy lift jobs

bwanglzu:


docs/fundamentals/clean-code.md
```diff
@@ -263,4 +263,64 @@ with f:
 ```
 ````
 
+## Power heavy-lifting job by the Flow, not Client
+   
+Wrap heavy lifting jobs inside an `Executor` and part of the `Flow`.
+For instance, loading high resolution images from Client into `Document` and `DocumentArray`
+could be time consuming.
+Put it inside an Executor could leverage Jina's capability to scale it up.
```
alaeddine-13, line 271: I view that this best practice mainly helps to reduce data sent over the network, if access to the uri is the same for the server and the client. Also this helps perform a lazy conversion, meaning it's converted on the fly.
I don't understand how Executor and Flow help with scaling up, unless the server's machine is faster than the client

docs/fundamentals/clean-code.md
```diff
@@ -263,4 +263,64 @@ with f:
...
+
+````{tab} ✅ Do
+```{code-block} python
+---
+emphasize-lines: 12
+---
```
alaeddine-13, line 277: note that this will highlight the line number 12, starting from line 278 (considered number 1)
So unless you have another important line to highlight, I suggest you replace with:
```suggestion
```python
```

docs/fundamentals/clean-code.md
```diff
@@ -263,4 +263,64 @@ with f:
...
+with f:
+    f.post('/foo', inputs=_load_data())
+```
+````
+
+````{tab} 😔 Don't
+```{code-block} python
+---
+emphasize-lines: 12
+---
```
alaeddine-13, line 308: note that this will highlight the line number 12, starting from line 278 (considered number 1)
So unless you have another important line to highlight, I suggest you replace with:
```suggestion
```python
```

docs/fundamentals/clean-code.md
bwanglzu: your comment make sense, I see it helps because scaling up a flow, e.g. using `parallel=n` or scale flow with k8s is easier. Not because server is faster than client.

docs/fundamentals/clean-code.md
```diff
@@ -263,4 +263,59 @@ with f:
...
+
+f = Flow().add(uses=MyExecutor)
+
+def _load_data():
+    da = DocumentArray()
+    image_uris = glob.glob('/.workspace/*.png')  # load high resolution images.
+    for image_uri in image_uris:
+        doc = Document(uri=image_uri)
+        da.append(doc)
+    return da
```
hanxiao, line 295: in fact, you just violated the first best practice: https://docs.jina.ai/fundamentals/clean-code/#generator-as-flow-input

hanxiao (changes requested):
reject, best practice violates in the best practice section/

PR #3336, opened 2021-09-03
fix: small changes to conda script

tadejsv:
Changes that were requested by conda-forge team

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,13 +1,13 @@
 {% set name = "jina" %}
-{% set version = "2.0.19" %}
+{% set version = "2.0.23" %}
 
 package:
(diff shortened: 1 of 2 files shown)


conda/meta.yaml
```diff
@@ -1,20 +1,21 @@
 {% set name = "jina" %}
-{% set version = "2.0.19" %}
+{% set version = "2.0.23" %}
```
hanxiao, line 2: this needs to be changed on every release right? could you add it in `scripts/release.sh` L45-49

tadejsv: Will do

Microsoft-account-png (commented):
```_

Issue #2582, opened 2021-06-08
_mermaid_id computed every time


JoanFM:
**Describe the bug**
The _mermaid_id cannot be used to link nodes in a graph, because it is changes when put in a ChunkArray. We should use `doc.id` as `_mermaid_id`

```python
>>> from jina import Document
>>> c1 = Document(text='c1')
>>> c2 = Document(text='c2')
>>> d = Document()
>>> d.chunks.append(c1)
<jina.types.document.Document id=dcbd74ac-c852-11eb-8d87-e86a64801cb1 mime_type=text/plain text=c1 granularity=1 parent_id=e353e45e-c852-11eb-8d87-e86a64801cb1 content_hash=62fa94ba00bf953a at 140635546411472>
>>> d.chunks.append(c2)
<jina.types.document.Document id=e07656ae-c852-11eb-8d87-e86a64801cb1 mime_type=text/plain text=c2 granularity=1 parent_id=e353e45e-c852-11eb-8d87-e86a64801cb1 content_hash=7c9d9e27f13a2c95 at 140635521309328>
>>> c1._mermaid_id
'0abe8a80-bd2c-4cb5-808a-21be0e69c295'
>>> d.chunks[0]._mermaid_id
'41cae67b-be0e-4888-8f3a-e99fa5659abc'
>>> d.chunks[0]._mermaid_id
'd965a7a3-7180-406f-b3de-a1a87698554f'
>>> d.chunks[0]._mermaid_id
'96cfd552-47fc-46cf-994c-99cfe5ff4b7b'

```

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINACOREV1-1034) by [Unito](https://www.unito.io)

PR #3339, opened 2021-09-03
perf: document fast init

alaeddine-13:
This speeds up DocumentArray.__iter__ heavily since Document constructor becomes much faster
Consequently, all methods that use __iter__ become faster.
When developing this, we had to set mermaid_id efficiently. A similar issue was raised before: https://github.com/jina-ai/jina/issues/2582

To solve it, we used a class attribute counter (to make mermaid_id unique based on document id which is already unique)

results:
|    |      master      |  perf-document-fast-init ( ff87e72 ) | perf-document-fast-init (ec5961d) | perf-document-fast-init (latest) |
|----------|:-------------:|------:|------:|------:|
| document array iter |  1.16 s | 0.32 s | 0.50 s | 0.36 s |
| document array get_attributes('text') |    1.32 s   |   0.42 s | 0.61 s | 0.43 s |
| document array get_attributes('text', 'embedding') | 3.32 s | 2.06 s | 2.21 s | 2.17 s |
| document construct by pb_body | 1.3 s | 0.68 s | 0.6 s | 0.45 s |

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -229,14 +229,15 @@ class Document(ProtoTypeMixin, VersionedMixin):
         """
         Document.CNT += 1
 
-        self._pb_body = jina_pb2.DocumentProto()
         try:
             if isinstance(document, jina_pb2.DocumentProto):
                 if copy:
+                    self._pb_body = jina_pb2.DocumentProto()
                     self._pb_body.CopyFrom(document)
                 else:
                     self._pb_body = document
             elif isinstance(document, bytes):
+                self._pb_body = jina_pb2.DocumentProto()
                 self._pb_body.ParseFromString(document)
             elif isinstance(document, (dict, str)):
                 if isinstance(document, str):
@@ -270,6 +271,7 @@ class Document(ProtoTypeMixin, VersionedMixin):
                     )
                 )
 
+                self._pb_body = jina_pb2.DocumentProto()
                 if support_fields.issuperset(user_fields):
                     json_format.ParseDict(document, self._pb_body)
                 else:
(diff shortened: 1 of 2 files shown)


davidbp:
> To solve it, we used a class attribute counter (to make mermaid_id unique based on document id which is already unique)

I want to leave a comment here because the solution might seem unjustified/weird.
While we were doing this with Alaeddine we found out that creating the unique ids took a lot of time (to me this was not exected at least). You pay a  big price  for iterating  and calling the previous unique id mechanism. 

Just as a reference:
```
In [7]: %timeit [uuid.uuid4() for i in range(100_000)]
318 ms ± 1.78 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit [i for i in range(100_000)]
2.87 ms ± 27.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Amanda31byrd (changes requested):
Can these be fixed to normal

hanxiao (changes requested):
hold, as this is 🕊️ column-access we were discussing in https://jinaai.slack.com/archives/C018F60RBL5/p1630674830115600

jina/types/document/__init__.py
```diff
@@ -226,96 +229,110 @@ def __init__(
...
-
-                user_fields = set(document)
-                support_fields = set(
-                    self.attributes(
-                        include_proto_fields_camelcase=True, include_properties=False
-                    )
+        Document.CNT += 1
+        if pb_only:
+            if not isinstance(document, jina_pb2.DocumentProto) or document is None:
+                raise ValueError(
```
hanxiao, line 235: can you restructure you `if` here so that the rest of the code remains untouched. Right now I see a lot of lines of changes but it seems that they only differs in ident

AMSAJ0: s3_s3a

alaeddine-13: I refactored the constructor

hanxiao:
i moved the Proto constructor inside, this should give a small performance boost and should result in similar performance as your first version. Could you give it a check?

alaeddine-13:
> i moved the Proto constructor inside, this should give a small performance boost and should result in similar performance as your first version. Could you give it a check?

Yes LGTM
I'll check how it performs and update the performance table

alaeddine-13:
thanks for your help @hanxiao 
I updated the table, now back to the first version performance
Also, after the refactor commit, also DAM got a boost:

- dam iter:
     * perf-document-fast-init : 2.32 s
     * master : 3.17 s

- dam get_attributes('text'):
     * perf-document-fast-init : 2.03 s
     * master : 3.36 s

Issue #3351, opened 2021-09-06
unexpected args 


cristianmtr:
I can reproduce it with this test https://github.com/jina-ai/executors/blob/fbd11ff89b4242d254205f832884a8d3f9917b80/tests/integration/psql_import/test_import_psql.py#L173

The problem is that the arguments are passed to the MatchMerger as well https://github.com/jina-ai/jina/blob/c613fa7799e25c8385f826c3f26bce1ea55e3e7d/jina/jaml/parsers/executor/legacy.py#L78

PR #3353, opened 2021-09-06
fix: #3351

cristianmtr:


jina/jaml/parsers/executor/legacy.py
```diff
@@ -98,6 +99,14 @@ def parse(self, cls: Type['BaseExecutor'], data: Dict) -> 'BaseExecutor':
         obj.is_updated = False
         return obj
 
+    @staticmethod
+    def is_tail_or_head(data):
+        try:
+            name = getattr(getattr(data, 'runtime_args', {}), 'name', '')
+            return '/' in name
+        except Exception as e:
+            print(f'========= {e}')
```
hanxiao, line 108:
```suggestion
        except Exception:
            pass  # intentionally pass as <write why there is no reason to handle it>
```

jina/jaml/parsers/executor/legacy.py
```diff
@@ -98,6 +99,16 @@ def parse(self, cls: Type['BaseExecutor'], data: Dict) -> 'BaseExecutor':
         obj.is_updated = False
         return obj
 
+    @staticmethod
+    def is_tail_or_head(data: Dict) -> bool:
+        """Based on name, compute if this is a tail/head Pea or a main Pea
+
+        :param data: the data for the parser
+        :return: boolean of whether it is tail/head or main
```
jacobowitz, line 107: that description is slightly misleading? I assume this is True if it is head/tail . 

So I'd write :return: boolean True if it is tail/head., False otherwise

jina/jaml/parsers/executor/legacy.py
```diff
@@ -78,7 +78,8 @@ def parse(self, cls: Type['BaseExecutor'], data: Dict) -> 'BaseExecutor':
         arguments_from_cls = LegacyParser._get_all_arguments(cls)
         arguments_from_yaml = set(data.get('with', {}))
         difference_set = arguments_from_yaml - arguments_from_cls
-        if any(difference_set):
+        # only log warnings about unknown args for main Pea
+        if any(difference_set) and LegacyParser.is_tail_or_head(data):
```
jacobowitz, line 82: I somehow dont get this one? Your comment says you only want to do this for a main pea but then you enter the if here if it is tail/head? I guess I am missing something here?

jacobowitz (changes requested):
One small comment on a comment :) And one question as I am confused by the if

jina/jaml/parsers/executor/legacy.py
cristianmtr: adjusted, thanks

jacobowitz (approved):
looking good :+1:

jina/jaml/parsers/executor/legacy.py
cristianmtr: I fixed that. Initially I wanted the method to return if it's a main Pea. But switched it around and forgot to change here. The pains of not having a proper test for logging

PR #3362, opened 2021-09-07
fix: name can be None in args

cristianmtr:
`runtime_args.name` can be `None` in some tests, and we don't want to make tests require more scaffolding

Ex https://github.com/jina-ai/executors/runs/3534309301#step:4:781

### DIFF
--- a/jina/jaml/parsers/executor/legacy.py
+++ b/jina/jaml/parsers/executor/legacy.py
@@ -106,8 +106,11 @@ class LegacyParser(VersionedYAMLParser):
         :param data: the data for the parser
         :return: True if it is tail/head, False otherwise
         """
-        name = data.get('runtime_args', {}).get('name', '')
-        return 'head' in name or 'tail' in name
+        try:
+            name = data.get('runtime_args', {}).get('name', '')
+            return 'head' in name or 'tail' in name
+        except Exception as _:
+            pass  # name can be None in tests since it's not passed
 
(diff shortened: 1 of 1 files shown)


jina/jaml/parsers/executor/legacy.py
```diff
@@ -106,8 +106,11 @@ def is_tail_or_head(data: Dict) -> bool:
         :param data: the data for the parser
         :return: True if it is tail/head, False otherwise
         """
-        name = data.get('runtime_args', {}).get('name', '')
-        return 'head' in name or 'tail' in name
+        try:
+            name = data.get('runtime_args', {}).get('name', '')
+            return 'head' in name or 'tail' in name
+        except Exception as _:
```
jacobowitz, line 112: KeyError? Exception seems to broad?

jacobowitz (changes requested):
I'd be more specific with the exception?

jina/jaml/parsers/executor/legacy.py
cristianmtr: It can also be `IterableError`, because they key can be there, but it can be `None`

jacobowitz: then? KeyError, IterableError ? :P

jacobowitz (approved):
If you know the errors specifically, it could be more specific, but doesnt matter really I think.

Issue #3197, opened 2021-08-17
Raise error when using grpc data runtime with shards


jacobowitz:
The new GrpcDataRuntime (https://github.com/jina-ai/jina/blob/master/jina/peapods/runtimes/grpc/__init__.py) does not support sharding at the moment. Thus a Flow should not specify this runtime and a sharded pod at the same time.

Right now this will not raise any error, but silently produce unexpected behavior. This is undesirable and the Flow should rather check this condition and raise an appropriate error so that the user can fix its Flow configuration.

unnati914:
Can I take up this issue?
Thank you!

maximilianwerk:
@unnati914 sure. If you have any questions, please join our [slack](https://slack.jina.ai/) and ask in the #contributor channel. :)

unnati914:
tysm :)

jacobowitz:
@unnati914 thx for picking it up!

As an example, this Flow definition is not working and should raise an error:

`Flow(grpc_data_requests=True).add(shards=2)`

jacobowitz:
@unnati914 
Are there any updates on this one? Can I assist you with anything?

bhattalok49:
hi, can  i work on this issue ?

CatStark:
Hi @bhattalok49, i think @unnati914 is unavailable at the moment so yes you can work on this :) I'll re assign it to you

RaghavPrabhakar66:
@CatStark Can I work on this issue?

jacobowitz:
@bhattalok49 did not react, so I guess its fine.

Thank you @RaghavPrabhakar66 :) If you have any question, I am happy to help. Here or via our community slack

PR #2939, opened 2021-07-13
feat: kubernetes mvp

florian-hoenicke:
MVP implementation of simple

PR #3364, opened 2021-09-07
fix(flow): raise error when using grpc data runtime with shards

RaghavPrabhakar66:
Fixes Issue: #3197

Expected Behavior: `Flow(grpc_data_requests=True).add(shards=2)` should raise an error.

Implemented Behavior:
```python
>>> Flow(grpc_data_requests=True).add(shards=2)
  File "<stdin>", line 1, in <module>
  File "/workspaces/jina/jina/flow/builder.py", line 33, in arg_wrapper
    return func(self, *args, **kwargs)
  File "/workspaces/jina/jina/flow/base.py", line 716, in add
    raise NotImplementedError("GRPC data runtime does not support sharding")
NotImplementedError: GRPC data runtime does not support sharding
>>> Flow(grpc_data_requests=True).add(shards=None)
<jina.flow.base.Flow object at 0x7fb19e5c8b80>
>>> Flow(grpc_data_requests=True).add()
<jina.flow.base.Flow object at 0x7fb19e5c8b80>
```

### DIFF
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -709,7 +709,7 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
         )
 
         # grpc data runtime does not support sharding at the moment
-        if args.grpc_data_requests and kwargs.get('shards') is not None:
+        if args.grpc_data_requests and kwargs.get('shards') > 1:
             raise NotImplementedError("GRPC data runtime does not support sharding")
 
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -713,6 +713,7 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
             args.grpc_data_requests
             and kwargs.get('shards') is not None
             and kwargs.get('shards', 1) > 1
+            and self.args.infrastructure != InfrastructureType.K8S
         ):
             raise NotImplementedError("GRPC data runtime does not support sharding")
 
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -711,6 +711,10 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
             args.runtime_cls = 'GRPCDataRuntime'
 
+            # grpc data runtime does not support sharding at the moment
+            if kwargs.get('shards') is not None:
+                raise NotImplementedError("GRPC data runtime does not support sharding")
+
         # pod workspace if not set then derive from flow workspace
         args.workspace = os.path.abspath(args.workspace or self.workspace)
 
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -708,12 +708,12 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
             kwargs, parser, True, fallback_parsers=FALLBACK_PARSERS
         )
 
+        # grpc data runtime does not support sharding at the moment
+        if kwargs.get('shards') is not None:
+            raise NotImplementedError("GRPC data runtime does not support sharding")
+
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
             args.runtime_cls = 'GRPCDataRuntime'
-
-            # grpc data runtime does not support sharding at the moment
-            if kwargs.get('shards') is not None:
-                raise NotImplementedError("GRPC data runtime does not support sharding")
 
         # pod workspace if not set then derive from flow workspace
         args.workspace = os.path.abspath(args.workspace or self.workspace)
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -709,7 +709,7 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
         )
 
         # grpc data runtime does not support sharding at the moment
-        if kwargs.get('shards') is not None:
+        if args.grpc_data_requests and kwargs.get('shards') is not None:
             raise NotImplementedError("GRPC data runtime does not support sharding")
 
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -709,7 +709,11 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
         )
 
         # grpc data runtime does not support sharding at the moment
-        if args.grpc_data_requests and kwargs.get('shards') > 1:
+        if (
+            args.grpc_data_requests
+            and kwargs.get('shards') is not None
+            and kwargs.get('shards', 1) > 1
+        ):
             raise NotImplementedError("GRPC data runtime does not support sharding")
 
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
--- a/tests/unit/flow-construct/test_flow.py
+++ b/tests/unit/flow-construct/test_flow.py
@@ -836,4 +836,5 @@ def test_flow_grpc_with_shard():
         Flow(grpc_data_requests=True).add(shards=2)
         Flow(grpc_data_requests=True).add(shards=None)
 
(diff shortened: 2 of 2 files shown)


jina/flow/base.py
```diff
@@ -711,6 +711,10 @@ def add(
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
             args.runtime_cls = 'GRPCDataRuntime'
 
+            # grpc data runtime does not support sharding at the moment
+            if kwargs.get('shards') is not None:
```
jacobowitz, line 715: I would check that condition outside the if in 711, because it could happen that the runtime_cls is actually not 'ZEDRuntime' but something else

jacobowitz (changes requested):
Thank you very much for your contribution!

Can you maybe also add a unit test, which covers this case? It should go somewhere in tests/unit/flow-construct/ I believe

AMSAJ0:
Ok

jina/flow/base.py
AMSAJ0: good

RaghavPrabhakar66:
> Can you maybe also add a unit test, which covers this case? It should go somewhere in tests/unit/flow-construct/ I believe

@jacobowitz  Should I write test in a new file or add into an existing file? For now I have added it to `test_flow.py`.

jina/flow/base.py
RaghavPrabhakar66: Makes sense, fixed it.

bdr2486 (approved):

bdr2486 (approved):

bdr2486 (approved):

jina/flow/base.py
```diff
@@ -708,6 +708,10 @@ def add(
             kwargs, parser, True, fallback_parsers=FALLBACK_PARSERS
         )
 
+        # grpc data runtime does not support sharding at the moment
+        if kwargs.get('shards') is not None:
+            raise NotImplementedError("GRPC data runtime does not support sharding")
+
         if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':
             args.runtime_cls = 'GRPCDataRuntime'
```
agarwalrounak, line 717: Just pitching in here because I had been following this issue for a while. 

Since, we are concerned about not using GrpcDataRuntime with shards, I believe that the if that has been added in line 712 should come in line 718 after the completion of if where we are assigning runtime_cls to GRPCDataRuntime. Also, there should be a check to see if runtime_cls is GRPCDataRuntime before raising the NotImplementedError. The current implementation will not allow sharding with any runtime. For example, when args.grpc_data_requests is false and args.runtime_cls == 'ZEDRuntime', we move ahead with ZEDRuntime but are still not allowing sharding. This should not be the case.

jacobowitz (changes requested):
Thanks for adding the test :)
Mayebe also add anothe one testing the good case? `shards=1` should work (even though shards is specified).

jina/flow/base.py
jacobowitz: I think it should be independent of the actual `runtime_cls` so I would not do the check inside `if args.grpc_data_requests and args.runtime_cls == 'ZEDRuntime':`

But you are right that the current check will prevent any sharding, so the check should be something like 

if args.grpc_data_requests and kwargs.get('shards') is not None:
    raise NotImplementedError("GRPC data runtime does not support sharding")

RaghavPrabhakar66: If we check for grpc_data_requests and shards together, then I think current placement is fine.

tests/unit/flow-construct/test_flow.py
```diff
@@ -827,3 +827,13 @@ def test_connect_to_predecessor():
...
+
+
+def test_flow_grpc_with_shard():
+    with pytest.raises(
+        NotImplementedError, match='GRPC data runtime does not support sharding'
+    ):
+        Flow(grpc_data_requests=True).add(shards=2)
+        Flow(grpc_data_requests=True).add(shards=None)
+
+    Flow(grpc_data_requests=False).add(shards=-1)
```
jacobowitz, line 839: Let's use the good case here which is 1. Also I would also add a good case without any shard like `Flow(grpc_data_requests=False).add()`
I believe the no shard case will fail with your change, because kwargs does not contain 'shards' then and you might get a KeyError.

```suggestion
    Flow(grpc_data_requests=False).add(shards=1)
```

jina/flow/base.py
```diff
@@ -708,6 +708,10 @@ def add(
             kwargs, parser, True, fallback_parsers=FALLBACK_PARSERS
         )
 
+        # grpc data runtime does not support sharding at the moment
+        if args.grpc_data_requests and kwargs.get('shards') is not None:
```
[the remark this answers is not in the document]
RaghavPrabhakar66: If None is fine, then shouldnt check should be:
`if args.grpc_data_requests and kwargs.get('shards') is not None and kwargs.get('shards') > 1:`

tests/unit/flow-construct/test_flow.py
RaghavPrabhakar66: With condition:
```python
if args.grpc_data_requests and kwargs.get('shards') is not None and kwargs.get('shards') > 1:
            raise NotImplementedError("GRPC data runtime does not support sharding")
```
I ran `Flow(grpc_data_requests=False).add()` case and it didnt gave any error.
```python
>>> Flow(grpc_data_requests=False).add()
<jina.flow.base.Flow object at 0x7f1c765636a0>
```

jina/flow/base.py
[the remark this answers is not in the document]
jacobowitz: yes indeed, to make the if a bit shorter you can provide a default value to the .get() function in case its not existing:
`kwargs.get('shards', 1) `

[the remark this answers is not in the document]
RaghavPrabhakar66: This could work. But what about cases `Flow(grpc_data_requests=True).add(shards=None) ` because it overrides defailt value 1 and gives error during if-condition.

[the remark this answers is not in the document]
jacobowitz: ok, true so I guess we always need to check for None then as well.

jacobowitz:
@RaghavPrabhakar66 
Looks good now :) 
One last thing that we should add is the K8s check. Recently we added support for [k8s deployment](https://github.com/jina-ai/jina/pull/2939) and is this case grpc/shards is actually a working combination.

You can check it with something like this
`args.infrastructure == InfrastructureType.K8S`

RaghavPrabhakar66:
@jacobowitz  We can add something like `and self.args.infrastructure != InfrastructureType.K8S` to if-condition. I checked, there are only two Infrastructure Types K8s and Jina, so I think this could work.

tests/unit/flow-construct/test_flow.py
```diff
@@ -827,3 +827,15 @@ def test_connect_to_predecessor():
...
+def test_flow_grpc_with_shard():
+    with pytest.raises(
+        NotImplementedError, match='GRPC data runtime does not support sharding'
+    ):
+        Flow(grpc_data_requests=True).add(shards=2)
+        Flow(grpc_data_requests=True).add(shards=None)
+
+    Flow(grpc_data_requests=False).add(shards=1)
+    Flow(grpc_data_requests=False).add()
+    Flow(grpc_data_requests=False, infrastructure=InfrastructureType.K8S).add(shards=2)
```
jacobowitz, line 841:
```suggestion
    Flow(grpc_data_requests=True, infrastructure=InfrastructureType.K8S).add(shards=2)
```

The point is to validate check grpc and K8s is a legal combination

jacobowitz (changes requested):
One minor comment left, lets address it and then we can check if CI runs through.

tests/unit/flow-construct/test_flow.py
RaghavPrabhakar66: Done.

jacobowitz:
@RaghavPrabhakar66 
thx for the update :) lets get CI to pass now. Atm its complaining about the commit message being too long:

` input: fix(flow): update grpc with shard test and grpc with shard check condition
✖   header must not be longer than 72 characters, current length is 74 [header-max-length]`

https://github.com/jina-ai/jina/pull/3364/checks?check_run_id=3609792590

The rules are defined here: https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md

We probably will need multiple iterations until CI passes, I will run it whenever you update the PR

jacobowitz (approved):
all green and change looks good to me! thx a lot for this PR :)

PR #3347, opened 2021-09-06
refactor(protobuf): use .uri as an independent attribute

numb3r3:

PR #3366, opened 2021-09-08
fix(type): fix check on unique content field

hanxiao:


jacobowitz (approved):
lgtm

just for reference, the underlying change of uri semantics is here: https://github.com/jina-ai/jina/pull/3347

PR #3370, opened 2021-09-08
feat: k8s benchmark

jacobowitz:
DONT MERGE
This is also opened for reference, we dont want this on master.
This code was used to compare proxy based load balancing in k8s vs pooling


jacobowitz:
Benchmark is done, not needed to be merged

PR #3374, opened 2021-09-09
fix: Remove unused class MyConverter

gaocegege:
We already implement the conversion logic in [MyEncoder](https://github.com/jina-ai/jina/blob/4c4f69b150bc1f7103f31aea8ceade1f36081a7e/jina/helloworld/fashion/my_executors.py#L69). Thus there is no need to import it.

BTW, should we remove the definition of `MyConverter` from my_executors.py?

### DIFF
--- a/jina/helloworld/fashion/app.py
+++ b/jina/helloworld/fashion/app.py
@@ -12,7 +12,7 @@ if __name__ == '__main__':
         index_generator,
         query_generator,
     )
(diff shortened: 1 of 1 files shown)


niuzs-alan (dismissed):
LGTM! 😁

gaocegege:
commit message is not valid, I am re-pushing to fix it.

Issue #3323, opened 2021-09-02
yaml `ConstructorError` while using multiprocessing spawn start method


deepankarm:
```python
# executor.py
from jina import Executor, requests

class Custom(Executor):
    @requests
    def foo(*args, **kwargs):
        print('foo111')


# app.py
import os
os.environ['JINA_LOG_LEVEL'] = 'DEBUG'


def main1():
    from jina import Flow
    from executor import Custom

    with Flow().add(uses=Custom) as f:
        pass


def main2():
    from jina import Flow
    from executor import Custom

    with Flow().add(uses=Custom,
                    py_modules=os.path.join(os.path.dirname(__file__), 'executor.py')) as f:
        pass

if __name__ == '__main__':
    main2() # works
    main1() # fails with the following error
    """
    yaml.constructor.ConstructorError: could not determine a constructor for the tag '!Custom'
      in "<unicode string>", line 1, column 1:
        !Custom
    """
```

Run  `JINA_MP_START_METHOD=spawn python app.py`

tadejsv:
I tried, but unfortunately I was unable to fix. The problem is the following:

The global state of the yaml module, which has the representer for `MyExecutor`, does not get preserved when using the spawn method.

What I thought would work:
- pass global state of `yaml` module to `Process.run`. Does not work, module not pickeable. This approach also feels not very elegant

What could work:
- pass `MyExecutor` to `Process.run` - trigers re-import, and thus representer properly registered.

Problem with this idea:
- Currently `Flow.add(uses=MyExecutor)` works by transforming all arguments to CLI options (`--uses=MyExecutor`), so only strings get passed to `Process.run`. If we are willing to modify that, then this problem is solvable. Otherwise I think not

PR #3380, opened 2021-09-10
fix: uses=Executor should work with spawn start method

tadejsv:
This provides a "workaround" fix for  #3323.

However, there is still an issue if `uses="Executor"` is used instead of `uses=Executor`. Deeper refactoring would be needed to fix that - namely, the `jaml` module should not rely on global state of `pyyaml`, but used solely instantiated objects.


jina/flow/base.py
```diff
@@ -712,6 +712,12 @@ def add(
             kwargs, parser, True, fallback_parsers=FALLBACK_PARSERS
         )
 
+        # Temporary workaround to re-import executor module when using mp spawn start
+        # method, so thta the executor is re-registered with pyyaml. A better solution
```
winstonww, line 716: minor: typo `thta` --> 'that`

jina/flow/base.py
```diff
@@ -712,6 +712,12 @@ def add(
...
+        # probably involves deeper refactoring
+        if isinstance(kwargs.get('uses'), type(JAMLCompatible)):
+            args._exec_cls = kwargs['uses']
```
winstonww, line 719: Just trying to understand, I tried grepping for ` _exec_cls` but it seems like it's not used anywhere except here where it's being set and in  `jina/jaml/__init__.py` where it's deleted from `kwargs`. What does this `_exec_cls` do?

tadejsv: Exactly, it's not used anywhere. 

It is only passed, because when you pass the class to `multiprocessing.Process.start` method (which is what happens under the hood when jina starts the pod/pea) when using `'spawn'` start method, it will re-import that class from its module. And this re-importing is the action that is needed to trigger it's registration with `pyyaml`.

winstonww: gotcha

maximilianwerk (approved):
LGTM, even thought all the YAML magic actually looks like black magic.

PR #3384, opened 2021-09-10
Update importer.py

angela-zhao:
## Code Change
Check if user module is in sys.builtin_module_names instead of sys.modules.

## Why
Previously, it was throwing a user warning whenever I was importing an executor from one of my own modules because all imported modules are listed in sys.modules.

## Minimum Working Example of Error
app.py
```
from jina import Flow
from my_executor import Foo

flow = Flow().load_config('flow.yml')

with flow:
    flow.post(on="/index")
```

flow.yml
```
jtype: Flow
version: '1'
pods:
  - name: foo
    uses: Foo
    py_modules: 'my_executor.py'
    needs: gateway
```

my_executor.py
```
from jina import Executor, requests

class Foo(Executor):
    @requests
    def foo(self, **kwargs):
        print("foo")
```

When running `python app.py`, I get the following user warning:

> UserWarning: 
>             my_executor shadows one of built-in Python module name.
>             It is imported as `user_module.my_executor`          
>             Affects:
>             - Either, change your code from using `from my_executor import ...` 
>               to `from user_module.my_executor import ...`
>             - Or, rename my_executor to another name

### DIFF
--- a/jina/importer.py
+++ b/jina/importer.py
@@ -105,7 +105,7 @@ def _path_import(absolute_path: str):
         if user_module_name == '__init__':
             # __init__ can not be used as a module name
             spec_name = default_spec_name
-        elif user_module_name not in sys.modules:
+        elif user_module_name not in sys.builtin_module_names:
             spec_name = user_module_name
         else:
             warnings.warn(


tadejsv:
Hi @angela-zhao , can you make a simple example reproducing your issue?

angela-zhao:
> Hi @angela-zhao , can you make a simple example reproducing your issue?

I just added an example

tadejsv:
Thanks @angela-zhao . However, I would say that here the warning is working as intended.

What is happening is that you first import the executor's module in the main process, and then, because you are specifying `py_modules`, you import it again in the executor's process. Because executor's process inherits the state of the main process, the new import will override the import from the main process - this is what the warning is about.

In your situation you can simply ignore this warning, or, even better, remove the import from `app.py`, as it is not really needed there.

In general, this warning is there because jina's import system is not as flexible as python's (file name is always used, you can not create an alias), and so you could unknowingly override some existing package, simply because you named the python file with the same name.

Let me know if this makes sense to you.

angela-zhao:
Thanks for your explanation. In the minimum working example, the error is gone after I remove the Foo import.

In my project, the user warning is still there even though I don't import my executors in app.py. I'm not sure where it's being imported twice.

tadejsv:
Until recently there has been a bug related to that, but it was fixed. Can you try updating to the latest version?

```
pip install jina -U
```

angela-zhao:
> Until recently there has been a bug related to that, but it was fixed. Can you try updating to the latest version?
> 
> ```
> pip install jina -U
> ```

This fixed it. Thank you!

PR #3386, opened 2021-09-10
feat(util): add get_docs_batch_generator for documentarray

### LINKED ISSUE #3333: Add  `get_docs_batch_generator` for DocumentArray
This is probably the most used function from jina commons. I think it would make sense to add it to jina - maybe not as a DocumentArray method, but as a util somewhere. Would enable us to eliminate jina commons requirement from a lot of executors

agarwalrounak:
Closes #3333 

This PR aims to add `get_docs_batch_generator` for DocumentArray. This change would enable us to eliminate jina commons requirement from a lot of executors. 

The existing import statement `from jina_commons.batching import get_docs_batch_generator` in executors can be updated to `from jina.util import get_docs_batch_generator`. Another way could be to add `from jina.util import get_docs_batch_generator` to `jina/__init__.py` file and update the existing import statement in executors to `from jina import get_docs_batch_generator`.

### DIFF
--- a/jina/types/arrays/traversable.py
+++ b/jina/types/arrays/traversable.py
@@ -2,6 +2,8 @@ import itertools
 from abc import abstractmethod
 from typing import Iterable
 
+from ...util import get_docs_batch_generator
+
 
 def _check_traversal_path_type(traversal_paths):
     if isinstance(traversal_paths, str):
@@ -95,3 +97,22 @@ class TraversableSequence:
     @abstractmethod
     def _flatten(sequence):
         ...
+
+    @staticmethod
+    def batch(docs, traversal_path, batch_size, needs_attr):
+        """
+        Calls `get_docs_batch_generator` and returns the `Generator` created.
+
+        Example Usage:
+
+        >>> for batch in batch(docs, ['r'], 10, needs_attr='blob'):
+        >>>     ...
+
+        :param docs: A document array.
+        :param traversal_path: Specifies along which "axis" the document shall be traversed. (defaults to ['r'])
+        :param batch_size: Size of each generated batch (except the last one, which might be smaller, default: 32)
+        :param needs_attr: Optionally, you can filter out docs which don't have this attribute
+
+        :return: Generator
+        """
+        return get_docs_batch_generator(docs, traversal_path, batch_size, needs_attr)
--- a/jina/util/__init__.py
+++ b/jina/util/__init__.py
@@ -1,0 +1,55 @@
+"""Module for Jina helper functions for batching documents in executors."""
+
+from typing import List, Generator, Optional
+
+from jina import DocumentArray
+
+
+def _batch_generator(
+    data: DocumentArray, batch_size: int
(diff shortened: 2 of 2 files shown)


hanxiao (changes requested):
`.batch` should be a method of `DocumentArray` & `DocumentArrayMemap`

agarwalrounak:
Hi @hanxiao, thank you for your review! 
I had a query. By _**`.batch` should be a method of `DocumentArray` & `DocumentArrayMemap`**_, do you mean that add a method `batch` to the classes `DocumentArray` & `DocumentArrayMemap`, which then calls the `get_docs_batch_generator` method added in `util/__init__.py`? Please let me know if I understood it correctly.

hanxiao:
> By doing so, we would then be able to change the occurrences of get_docs_batch_generator(...) in executors to DocumentArray.batch(...) or DocumentArrayMemap.batch(...)

Yes, that would be the user experience.

I would suggest to implement `.batch` in `jina/types/arrays/traversable.py:TraversableSequence` mixin, as this mixin class is shared by both `DocumentArray` and `DocumentArrayMemap`.

agarwalrounak:
Thank you for your suggestion. I have implemented `batch` method in `TraversableSequence` mixin class. `batch` calls `get_docs_batch_generator` method added in `util/__init__.py` and returns the `Generator` created.

jina/types/arrays/traversable.py
```diff
@@ -95,3 +97,22 @@ def traverse_flat(self, traversal_paths: Iterable[str]) -> 'TraversableSequence'
     @abstractmethod
     def _flatten(sequence):
         ...
+
+    @staticmethod
```
hanxiao, line 101: can you simply move `get_docs_batch_generator` function into `.batch` and make `.batch` **not** staticmethod?

The reason is that `DocumentArray` to jina is like `ndarray` to numpy, it is our basic data type, and we'd like to have any method that relate to `DocumentArray` or `DocumentArrayMemmap` as its "native" class method. We do not prefer "misc"/"util" module to store orphan methods, it is not part of our design. In Jina, every method should have a clear parent (at least to the best reasonable effort).

jina/util/__init__.py
```diff
@@ -0,0 +1,55 @@
+"""Module for Jina helper functions for batching documents in executors."""
```
hanxiao, line 1: Delete this file.

In general, we do not prefer some "misc"/"util" module to store orphan methods, it is not part of our design. In Jina, every method should have a clear parent (at least to the best reasonable effort).

hanxiao (changes requested):
- move `get_docs_batch_generator` inside `jina/types/arrays/traversable.py` and **un-**staticmethod the `.batch` func
- need unit tests

agarwalrounak:
@hanxiao I was away from my laptop for a while, so I couldn't work on the changes you suggested. Sorry for that! I see that you opened a new PR with the changes. 
I will pick up a different issue to work on.

PR #3390, opened 2021-09-11
refactor: support path inputs

Bharat123rox:
https://www.notion.so/Bharat123rox-ac61ddc162d2428795fa893e5f7ee805

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -14,6 +14,7 @@ import warnings
 from argparse import ArgumentParser, Namespace
 from datetime import datetime
 from itertools import islice
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


tadejsv:
Hi @Bharat123rox 

Unfortunately I do not know of any such script - you'll have to try different things here yourself. I would it is not so important that we find 100% of them - we can always continue with another PR.

For now, find as many as you can, and do not forget to add unit tests for all the changes you make. We can start with the review process once this is done.

JoanFM:
Hello @Bharat123rox,

There has been no progress on this thread for quite some time. Are you having problems to address the issue?

PR #3407, opened 2021-09-14
fix(spawn): reregister all jaml classes

deepankarm:
This reverts the hacks added in #3380 and #3404 and passes all `JAMLCompatible` classes registered in main process to the `run` method in Peas.


hanxiao:
reregister = gister

tadejsv (approved):
Seems good!

PR #3417, opened 2021-09-15
feat: implement blobs getter and setter

tadejsv:
This properly implements `blobs` getter and setter for DocumentArray, and adds it to DocumentArrayMemmap (by simply relying on `get_attributes`)


alaeddine-13 (commented):
Can we also keep a test for blobs of Document array ?

tadejsv:
@alaeddine-13 it is still there, I did not do anything with it.

jina/types/arrays/document.py
```diff
@@ -534,9 +527,49 @@ def embeddings(self, emb: np.ndarray):
...
+
+        :param emb: The blob array to set. The first axis is the "row" axis.
+        """
+
+        if len(blobs) != len(self):
+            raise ValueError(
+                f'the number of rows in the input ({len(blobs)}), should match the'
+                f'number of Documents ({len(self)})'
+            )
+
```
alaeddine-13, line 573: this is replicated and the same code as in DA
maybe we should keep the setter in the parent class

jina/types/arrays/memmap.py
```diff
@@ -554,15 +554,55 @@ def embeddings(self) -> np.ndarray:
...
-            'the number of rows in the input ({len(emb)}),'
-            'should match the number of Documents ({len(self)})'
-        )
+        :param emb: The embedding matrix to set
+        """
+        if len(emb) != len(self):
+            raise ValueError(
+                f'the number of rows in the input ({len(emb)}), should match the'
+                f'number of Documents ({len(self)})'
+            )
```
alaeddine-13, line 565: let's add a test for this case

jina/types/arrays/memmap.py
```diff
@@ -554,15 +554,55 @@ def embeddings(self) -> np.ndarray:
...
+        """Set the blobs of the Documents
+
+        :param emb: The blob array to set. The first axis is the "row" axis.
+        """
+
+        if len(blobs) != len(self):
+            raise ValueError(
+                f'the number of rows in the input ({len(blobs)}), should match the'
+                f'number of Documents ({len(self)})'
+            )
```
alaeddine-13, line 600: let's also add a test for this

jina/types/arrays/document.py
```diff
@@ -534,9 +527,49 @@ def embeddings(self, emb: np.ndarray):
...
         """
 
-        assert len(emb) == len(
-            self
-        ), f'the number of rows in the input ({len(emb)}), should match the number of Documents ({len(self)})'
+        if len(emb) != len(self):
+            raise ValueError(
+                f'the number of rows in the input ({len(emb)}), should match the'
+                f'number of Documents ({len(self)})'
+            )
```
alaeddine-13, line 535: let's add a test for this case

alaeddine-13:
> @alaeddine-13 it is still there, I did not do anything with it.

alright all good then

jina/types/arrays/document.py
tadejsv: I guess so. Since the mixin does not seem to be an abstract class (does not raise `NotImplementedError`), this would be ok I think

tadejsv: It turns out if subclass overrides the getter, it must also override the setter. For that reason this is not possible

tadejsv: Found a way around it

hanxiao (approved):

PR #3444, opened 2021-09-20
feat(gateway): https support to uvicorn based servers

deepankarm:


jacobowitz (approved):
Nice feature!

I think we should provide an example in the docs though how to setup Jina with HTTPS and some certificate.

Issue #3393, opened 2021-09-13
Pass arguments to `jina hub new`


tadejsv:
Right now when using `jina hub new` the only option is to do everything interactively. It would be great if some commands could be passed as options, to CLI, for example `--name=MyExecutor`

sauravgarg540:
Hey @tadejsv, I would like to work on this issue.

CatStark:
Hi @sauravgarg540 feel free to open a PR you would like to help

PR #3448, opened 2021-09-20
feat: add parser to jina hub new

sauravgarg540:
This PR aims at resolving #3393.

Resolution:
- A new.py file is added to add a parser to Jina hub new.
- added arguments in test_new of test_hubio.py
- Updated hubio.py to make use of command-line arguments.

### DIFF
--- a/jina/parsers/hubble/new.py
+++ b/jina/parsers/hubble/new.py
@@ -1,0 +1,70 @@
+"""Argparser module for hub new"""
+
+import argparse
+
+from ..helper import add_arg_group
+
+
+def mixin_hub_new_parser(parser):
+    """Add the arguments for hub new to the parser
+    :param parser: the parser configure
+    """
+    gp = add_arg_group(parser, title='New')
+    gp.add_argument(
+        '-n',
+        '--name',
+        help='Setup name of the executor',
+        action='store',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-p',
+        '--path',
+        help='Path to store an executor',
+        action='store',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-ac',
+        '--advance_configuration',
+        help='Set up advance configuration',
+        choices=('True', 'False'),
+        default=None,
+    )
+
+    gp.add_argument(
+        '-d',
+        '--description',
+        help='Give a short description of your executor',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-k',
+        '--keywords',
+        help='Give some keywords to help people search your executor(separated by space)',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-u',
+        '--url',
+        help='URL of your GitHub repo',
+        action='store',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-ad',
+        '--add_dockerfile',
+        help='Set True to create your own docker file',
+        choices=('True', 'False'),
+        default=None,
+    )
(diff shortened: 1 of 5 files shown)


jina/parsers/hubble/new.py
```diff
@@ -0,0 +1,70 @@
...
+from ..helper import add_arg_group
+
+
+def mixin_hub_new_parser(parser):
+    """Add the arguments for hub new to the parser
+    :param parser: the parser configure
+    """
+    gp = add_arg_group(parser, title='New')
+    gp.add_argument(
+        '-n',
```
hanxiao, line 14: can you remove the single letter alias, we don't have such convention for `argsparse`. Reason is tho it may be convenient, but because we have so many args in our CLI, it can easily introduce conflict & hard to manage

jina/parsers/hubble/new.py
```diff
@@ -0,0 +1,70 @@
...
+        '--name',
+        help='Setup name of the executor',
+        action='store',
```
hanxiao, line 17:
```suggestion

```
no need, same as all below

jina/parsers/hubble/new.py
```diff
@@ -0,0 +1,70 @@
...
+        '--path',
+        help='Path to store an executor',
+        action='store',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-ac',
+        '--advance_configuration',
```
hanxiao, line 33: dash over underscore
```suggestion
        '--advance-configuration',

```

jina/parsers/hubble/new.py
```diff
@@ -0,0 +1,70 @@
...
+        help='Set up advance configuration',
+        choices=('True', 'False'),
```
hanxiao, line 35:
```suggestion
        action='store_true'

```

jina/parsers/hubble/new.py
```diff
@@ -0,0 +1,70 @@
...
+        '--url',
+        help='URL of your GitHub repo',
+        action='store',
+        type=str,
+        default=None,
+    )
+
+    gp.add_argument(
+        '-ad',
+        '--add_dockerfile',
```
hanxiao, line 66: dash over underscore

jina/parsers/hubble/new.py
```diff
@@ -0,0 +1,70 @@
...
+        help='Set True to create your own docker file',
+        choices=('True', 'False'),
```
hanxiao, line 68: store_true

hanxiao (changes requested):
1. need style fix
2. need unit test

jina/parsers/hubble/new.py
sauravgarg540: Hey @hanxiao, currently `--advance-configuration` is totally optional and keeps the program fully interactive. There are three scenarios here:

1. With `--advance-configuration True` the program tries to set advance configuration
2. With `--advance-configuration False` the program will skip setting advance configuration
3.  if `--advance-configuration` is not passed a prompt will ask the user to decide

Could you please share the required use case If I set `action='store_true'`
1. Make `--advance-configuration` compulsory to set up advance configuration and skip setting advance configuration if `--advance-configuration`  is not passed.
2. If the user did not pass `--advance-configuration` should there be a prompt to ask to confirm if the user wants to set up advance configuration. And what if user did not pass `--advance-configuration` as they want to skip setting advance configuration in that case the prompts will become annoying.

jina/parsers/hubble/new.py
sauravgarg540: I have a similar doubt here as for `--advance-configuration`

sauravgarg540:
Hey @hanxiao, I have made the changes as suggested by you. Waiting for your feedback.

bwanglzu:
interesting, latency tracking seems added 2 times to ci:

30973AFC-8DF2-41FB-B14C-6237D79A70A1

JoanFM (changes requested):
Tests are skipped, we have to check why

JoanFM (approved):
Thank you very much for this great contribution!

PR #3449, opened 2021-09-20
feat: k8s custom port expose

florian-hoenicke:
Currently, the only possible port to reach the k8s gateway is 8080.
This pr makes it possible to specify the port number in the flow


cli/autocomplete.py
```diff
@@ -189,8 +189,24 @@
             '--pea-id',
             '--pea-role',
             '--noblock-on-start',
-            '--dynamic-routing',
+            '--uses-before',
```
JoanFM, line 192: This  is wrong, no uses before for gateway. Or parallel , shards and more

JoanFM (changes requested):

jina/parsers/__init__.py
```diff
@@ -89,6 +89,8 @@ def set_gateway_parser(parser=None):
     mixin_comm_protocol_parser(parser)
     mixin_gateway_parser(parser)
     mixin_pea_parser(parser)
+    mixin_base_pod_parser(parser)
```
JoanFM, line 92: this mixin base should not be added I think

JoanFM (changes requested):

jina/parsers/__init__.py
florian-hoenicke: What is the reason it should not be added? It is a gateway pod and would be the first step towards scaling the gateway.

cli/autocomplete.py
florian-hoenicke: If we say the gateway is a pod, it should have all features of a pod

JoanFM: Right now this is not true, when this is true it may happen. But not now.

florian-hoenicke: true, I changed it. Thanks

jina/parsers/__init__.py
florian-hoenicke: changed it since we don't scale the gateway for now

JoanFM (approved):

PR #3459, opened 2021-09-22
ci(docs): deploy to netlify on pr

deepankarm:


.github/workflows/label-pr.yml
```diff
@@ -1,7 +1,7 @@
 name: PR
 
 on:
-  pull_request_target:
+  pull_request:
```
deepankarm, line 4: Only done to validate. Will be reverted

PR #3463, opened 2021-09-23
fix(hubble): disk cache compatible across python versions

numb3r3:


maximilianwerk (changes requested):
Can we add some test? Perhaps prepare a pickle file in each python version and check if they are readable in all python versions?

numb3r3:
> Can we add some test? Perhaps prepare a pickle file in each python version and check if they are readable in all python versions?

In this PR, I just clean the cache file when it cannot be loaded correctly.

hanxiao:
@alaeddine-13

maximilianwerk (approved):
LGTM

PR #3468, opened 2021-09-23
feat: sort top_k and return top_k Document from DocumentArray

### LINKED ISSUE #3467: DocumentArray `sort` should also get `top_k` to allow for more efficient sorting
**Describe the feature**
Right now `sort` is done on all the `Document` collection, I would like to sort only up to a `top_k`.

sauravgarg540:
This PR resolves #3467

## Resolution:
- Sort top_k and return top_k `Document` from `DocumentArray`
- added unit test

### DIFF
--- a/jina/types/arrays/document.py
+++ b/jina/types/arrays/document.py
@@ -375,14 +375,19 @@ class DocumentArray(
             hi_idx -= 1
         self._update_id_to_index_map()
 
-    def sort(self, key=None, *args, **kwargs):
+    def sort(self, key=None, top_k: int = None, *args, **kwargs):
         """
         Sort the items of the :class:`DocumentArray` in place.
 
         :param key: key callable to sort based upon
+        :param top_k: sort only upto top_k Document from DocumentArray
         :param args: variable set of arguments to pass to the sorting underlying function
         :param kwargs: keyword arguments to pass to the sorting underlying function
         """
+        # Only for DocumentArray
+        if isinstance(self._pb_body, list):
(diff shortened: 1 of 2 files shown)


jina/types/arrays/document.py
```diff
@@ -375,14 +375,19 @@ def reverse(self):
...
         """
         Sort the items of the :class:`DocumentArray` in place.
 
         :param key: key callable to sort based upon
+        :param top_k: sort only upto top_k Document from DocumentArray
         :param args: variable set of arguments to pass to the sorting underlying function
         :param kwargs: keyword arguments to pass to the sorting underlying function
         """
+        # Only for DocumentArray
+        if isinstance(self._pb_body, list):
```
JoanFM, line 388: Maybe it was not clear from the documentation. The point is to only sort the `first top_k` but of the sorted list.

There are sorting algorithms that do not sort a complete collection but that only guarantee that they first `top_k` elements are sorted while the rest may not be

JoanFM (changes requested):

hanxiao:
i dont think this implementation is needed , how is it different from `.sort()[:k]`

closing

JoanFM:
> i dont think this implementation is needed , how is it different from `.sort()[:k]`
> 
> closing

agree, this implementation is not the proper one. This is why there are changes requested

sauravgarg540:
Hey @JoanFM, thank you for the clarification on the issue description. I was already in doubt that why only sort first K values in a collection. I will create another PR with the correct implementation.

PR #3469, opened 2021-09-24
fix(daemon): remote flow with parallel>1

deepankarm:


daemon/stores/partial.py
```diff
@@ -144,6 +145,12 @@ def add(
                             pod.replicas_args = CompoundPod._set_replica_args(
                                 pod.args, pod.head_args, pod.tail_args
                             )
+                else:
+                    # avoid setting runs_in_docker for Pods with parallel > 1 and using `ZEDRuntime`
+                    # else, replica-peas would try connecting to head/tail-pea via __docker_host__
+                    if runtime_cls == 'ZEDRuntime' and pod.args.parallel > 1:
```
JoanFM, line 151: can u add this to `GRPCDataRuntime` if possible?

tests/distributed/test_topologies/test_topologies.py
```diff
@@ -144,3 +152,56 @@ def test_complex_needs(parallels, mocker):
...
+            host=__default_host__,
+            port=args['port_expose'],
+            protocol=args['protocol'],
+        ).post(
+            on='/',
+            inputs=(Document(blob=np.random.random([1, 100])) for _ in range(NUM_DOCS)),
+            on_done=response_mock,
+            show_progress=True,
+        )
+        response_mock.assert_called()
```
cristianmtr, line 184: here you should create a new Mock object, right? The first time it can be called but the second time it might not but the assert will still pass, no?

cristianmtr (approved):
LGTM, but maybe one small thing

cristianmtr (approved):

PR #3472, opened 2021-09-24
fix(hubble): enforce download specific tag

numb3r3:


cristianmtr (changes requested):
Can you add a quick test?

jina/hubble/hubio.py
```diff
@@ -533,7 +533,7 @@ def _get_prettyprint_usage(self, console, executor_name, usage_kind=None):
     @disk_cache_offline(cache_file=str(_cache_file))
     def fetch_meta(
         name: str,
-        tag: Optional[str] = None,
```
JoanFM, line 536: According to the code, it seems to be an Optional argument indeed

JoanFM (changes requested):

jina/hubble/hubio.py
numb3r3: According to the implementation of `disk_cache_offline`,  the optional argument is not included in the cache key. That's the reason I made this revision to cache the meta info regarding specific tag.

nan-wang (changes requested):

jina/hubble/hubio.py
```diff
@@ -533,7 +533,7 @@ def _get_prettyprint_usage(self, console, executor_name, usage_kind=None):
...
+        tag: str,
         secret: Optional[str] = None,
         force: bool = False,
```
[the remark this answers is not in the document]
numb3r3: it is used by `disk_cache_offline`

tests/unit/hubble/test_hubio.py
```diff
@@ -167,7 +167,7 @@ def _mock_get(url, headers=None):
     monkeypatch.setattr(requests, 'get', _mock_get)
     args = set_hub_pull_parser().parse_args(['jinahub://dummy_mwu_encoder'])
 
-    executor = HubIO(args).fetch_meta('dummy_mwu_encoder')
+    executor = HubIO(args).fetch_meta('dummy_mwu_encoder', None)
```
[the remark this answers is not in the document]
numb3r3: in some cases, the `tag=None`

nan-wang (approved):
LGTM👍

jina/hubble/hubio.py
numb3r3: The `tag` is not optional in the function's signature. But `self.fetch_meta('dummy_exe', None)` will be executed when pulling with `uses='jinahub://dummy_exec'` without specific tag.

JoanFM: Please, add note as discussed

numb3r3:
> Can you add a quick test?

Addressed by latest commit

cristianmtr (dismissed):

jina/hubble/hubio.py
```diff
@@ -533,13 +533,13 @@ def _get_prettyprint_usage(self, console, executor_name, usage_kind=None):
...
     ) -> HubExecutor:
         """Fetch the executor meta info from Jina Hub.
         :param name: the UUID/Name of the executor
-        :param tag: the version tag of the executor
+        :param tag: the tag of the executor if available, otherwise, use `None` as the value
```
JoanFM, line 542: The comment should be that `tag` should be passed via `args` and `force` and `secret` as `kwargs`, otherwise `cache` does not work

JoanFM (changes requested):

PR #3473, opened 2021-09-24
fix: fix pod default naming in flow

hanxiao:


maximilianwerk:
Let me take it from here.

deepankarm (approved):

PR #3445, opened 2021-09-20
fix: fix plot, naming and kubernetes

JoanFM:

PR #3474, opened 2021-09-24
refactor: RAII for pods

JoanFM:
**Changes introduced**
- Whenever a `Pod` is initialized, all its internal components must be initialized, and then entering context is delegated at `start` time.

This is needed to refactor in a nice manner the plotting of Plots #3445


JoanFM:
Will most likely not work for `deepcopy` in Flow

jacobowitz (dismissed):

deepankarm (approved):

PR #3476, opened 2021-09-24
docs readme webinar promo

alexcg1:
Bing mentioned we should put a banner on the repo to promote the webinar. Et voila, here it is!

- docs(readme): add webinar banner
- docs(readme): upload image


hanxiao:
image
please use this workflow instead

PR #3477, opened 2021-09-24
perf(daemon): reload files across flows/executors

deepankarm:
If a remote `Executor` downloads a pre-trained model, we want to make sure, during the next run the Executor can reload from the previously downloaded model (from disk). This can be done by reusing the previous workspace. This PR adds tests to show how it can be used & fixes few bugs related to the same.

### DIFF
--- a/jina/flow/base.py
+++ b/jina/flow/base.py
@@ -1586,6 +1586,10 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
 
     @workspace.setter
     def workspace(self, value: str):
+        """set workspace dir for flow & all pods
+
+        :param value: workspace to be set
+        """
         self.args.workspace = value
         for k, p in self:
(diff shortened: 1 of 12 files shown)


jina/flow/base.py
```diff
@@ -1627,6 +1631,26 @@ def workspace_id(self, value: str):
...
+        return self.args.env
+
+    @env.setter
+    def env(self, value: Dict[str, str]):
+        """set env vars for flow & all pods.
+        This can be used by jinad to set envs for Flow and all child objects
+
+        :param value: value to be set
+        """
+        if isinstance(value, dict):
```
JoanFM, line 1649: no need to set this `if`. If ur typehint suggests this there is no point to check.

JoanFM (changes requested):

nan-wang (approved):
LGTM👍

PR #3481, opened 2021-09-26
feat: optimize memory in match when traversal path is root

winstonww:
This PR aims to address [this](https://github.com/jina-ai/executor-simpleindexer/issues/13) issue. Thanks.

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -63,7 +63,7 @@ class DocumentArrayNeuralOpsMixin:
         """
 
         lhv = self
-        if traversal_ldarray:
+        if traversal_ldarray and traversal_ldarray != ['r']:
             lhv = self.traverse_flat(traversal_ldarray)
 
             from .document import DocumentArray
@@ -72,7 +72,7 @@ class DocumentArrayNeuralOpsMixin:
                 lhv = DocumentArray(lhv)
 
         rhv = darray
-        if traversal_rdarray:
+        if traversal_rdarray and traversal_rdarray != ['r']:
             rhv = darray.traverse_flat(traversal_rdarray)
 
             from .document import DocumentArray


hanxiao (changes requested):
- this patch logic should be implemented **inside** `.traverse_` of `TraversableSequence` instead of here
- the name of this PR should be changed to `perf(types): ...`

winstonww:
@hanxiao Title is now updated as suggested. So I looked into the `_traverse` function and found that if `path` is `'r'`, docs get yielded as-is. And `traverse_flat` will wrap this with `itertools.chain.from_iterable`. I imagine if we move the patch logic to `traversable.py`, it would look something like:

```python
class TraversableSequence:
    def traverse(self, traversal_paths):
        if traversal_paths == ['r']: return self
        ...

   def traverse_flat(self, traversal_paths):
        if traversal_paths == ['r']: return self
        ...
```

Is this the desired approach? It also seems like both `_traverse` and `traverse_flat` won't take up as much memory as the operation to create a new `DocumentArray` in the subsequent step on L72 of the `match` function, since `_traverse` is just a generator and `_traverse_flat` returns a chain object. Do you have any thoughts on this? thanks.

maximilianwerk (approved):
LGTM

JoanFM:
> @hanxiao Title is now updated as suggested. So I looked into the `_traverse` function and found that if `path` is `'r'`, docs get yielded as-is. And `traverse_flat` will wrap this with `itertools.chain.from_iterable`. I imagine if we move the patch logic to `traversable.py`, it would look something like:
> 
> ```python
> class TraversableSequence:
>     def traverse(self, traversal_paths):
>         if traversal_paths == ['r']: return self
>         ...
> 
>    def traverse_flat(self, traversal_paths):
>         if traversal_paths == ['r']: return self
>         ...
> ```
> 
> Is this the desired approach? It also seems like both `_traverse` and `traverse_flat` won't take up as much memory as the operation to create a new `DocumentArray` in the subsequent step on L72 of the `match` function, since `_traverse` is just a generator and `_traverse_flat` returns a chain object. Do you have any thoughts on this? thanks.

This seems logic, check what implication it has to flat itself to see if the check can only be applied to `traverse`.

Also I thinnk the safest check could be:

```python
if len(traversal_paths) == 1 and 'r' in traversal_paths:
```
This way it is sensible to multiple types of sequences

hanxiao:
Joan is right @winstonww

jina/types/arrays/neural_ops.py
```diff
@@ -72,7 +72,7 @@ def match(
                 lhv = DocumentArray(lhv)
 
         rhv = darray
-        if traversal_rdarray:
```
JoanFM, line 75: The solution should be in traverse to do the check and in flatten avoid flattening if only one sequence is passed

JoanFM (changes requested):

winstonww:
@JoanFM @hanxiao @maximilianwerk Thank you for reviewing and the comments. As there are some urgent EAP issues that need to be fixed, I will put this PR on hold today and tomorrow. Will address suggestions after the EAPs are fixed. Thank you.

PR #3482, opened 2021-09-26
feat(daemon): custom dockerfile

deepankarm:
A hacky implementation to support custom dockerfiles in JinaD workspaces. Since JinaD relies on `partial-daemon` inside the child containers, in the custom dockerfile, the user must have a `daemon` base image mentioned, e.g.
1. `FROM jinaai/jina:<version/master/test>-daemon` must be included inside the dockerfile.  
2. Add an entry `dockerfile = custom.dockerfile` in `.jinad` file.

As a test, I have added a `DummyRedisIndexer` that relies on the `redis-server` cli to start the server during Executor init and that gets used during index & query. Installing of redis-server from `apt` is done inside the custom dockerfile. The same can be extended to enable using GPUs from the host machine for both Flows & Executors. (Note: Enabling `--gpus all` is still not added)

https://github.com/jina-ai/jina/blob/e838fe3cd9546998260c09393f7045a11b65af74/tests/distributed/test_workspaces/redis_executor.py#L8-L17

### DIFF
--- a/daemon/files.py
+++ b/daemon/files.py
@@ -86,7 +86,7 @@ class DaemonFile:
     def dockerfile(self, value: Union[DaemonDockerfile, str]):
         """Property setter for dockerfile
 
-        :param dockerfile: allowed values in DaemonDockerfile
+        :param value: allowed values in DaemonDockerfile
         """
         try:
             self._dockerfile = os.path.join(
--- a/daemon/files.py
+++ b/daemon/files.py
@@ -1,21 +1,21 @@
-import glob
 import os
+import re
 from itertools import chain
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict, List, Union
 
 from fastapi import UploadFile
 
-from jina.helper import cached_property
 from jina.logging.logger import JinaLogger
+from jina.excepts import DaemonInvalidDockerfile
 from . import __rootdir__, __dockerfiles__, jinad_args
 from .helper import get_workspace_path
 from .models import DaemonID
-from .models.enums import DaemonBuild, PythonVersion
+from .models.enums import DaemonDockerfile, PythonVersion
 
 
-def workspace_files(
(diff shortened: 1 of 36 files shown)


daemon/files.py
```diff
@@ -43,33 +40,13 @@ def workspace_files(
         logger.debug(f'saved uploads to {dest}')
 
 
-def _merge_requirement_file(dest: str, f: UploadFile) -> None:
-    """Merge requirement files
+def is_requirements_txt(filename) -> bool:
```
deepankarm, line 43: Since this adds support to any requirements.txt file in the format `'.*requirements.*\.txt$'. merging requirements.txt files don't make sense anymore. Hence I have removed that part of the code @jacobowitz

daemon/files.py
```diff
@@ -188,38 +178,33 @@ def ports(self, ports: str):
...
         :return: space separated values
         """
-        # TODO: merge this with _read_requirements_file()
-        _req = f'{self._workdir}/requirements.txt'
-        if not Path(_req).is_file():
+        requirements = ''
+        for filename in os.listdir(self._workdir):
+            if is_requirements_txt(filename):
+                with open(os.path.join(self._workdir, filename)) as f:
+                    requirements += ' '.join(f.read().splitlines())
```
jacobowitz, line 191: do we have something similiar (req file parsing) maybe for hubble? I was just wondering, because this will miss special cases like `-f https://some-custom-repo.com`

jacobowitz (approved):
neat feature! only one small idea for the requirements file parsing

daemon/files.py
deepankarm: Hmm. We only have tested with requirements that have 1 word in them (e.g. - `jina`, `git+...`). If need arises, let us change the logic in the dockerfile. Rather than passing the requirements as an argument during `docker build`, let's pass names of the requirements files as an argument and let `pip install -r` handle the test.

jacobowitz (approved):

PR #3483, opened 2021-09-27
fix(kubernetes): Fix the service and deployment log

gaocegege:
This PR is to:

- change the log output about service since v1/service resource on Kubernetes does not need the image
- change the log output about deployment since we confuse the name and image name


JoanFM (approved):
Thank you very much for your contribution!

PR #3484, opened 2021-09-27
feat: start portforwarding when flow.post in K8s

florian-hoenicke:
This pr enables to run k8s flows via context manager.

# Example:
## Before:

```
    Flow(
        name='test-flow', port_expose=8080, infrastructure='K8S'
    ).start()
    
    # ... setup port forward ...
    with k8s_cluster_namespaced.port_forward(
        resp = requests.post(
            f'http://localhost:{port_expose}/{endpoint}',
            json={'data': [{} for _ in range(10)]},
        )

```

## After:
```
    with Flow(
        name='test-flow', port_expose=8080, infrastructure='K8S'
    ):
        flow.post(endpoint, [Document() for _ in range(10)])
```

In order to archive this, it uses port-forwarding during `flow.post`. 
For now, each call of `flow.post` creates and terminates port-forwarding. This means, calling `flow.post` multiple times creates an overhead, which can be tackled in a future pr.

If you are fine with the draft, I will add the docs as well in the same pr.


jina/peapods/pods/k8s.py
```diff
@@ -65,8 +65,12 @@ def port_expose(self) -> int:
...
     @property
     def host(self) -> str:
-        """Not implemented"""
-        raise NotImplementedError
+        """Currently, when deploying on Kubernetes, Jina does not expose a public host.
+        Instead Jina sends requests via port-forward and runs the requests against localhost.
+
+        :return: localhost
+        """
+        return 'localhost'
```
jacobowitz, line 73: Thats not correct I think. Its only localhost from the perspective of the port forwarding, but this property might be used in other contexts as well.

jina/clients/mixin.py
```diff
@@ -57,18 +60,31 @@ async def _get_results(*args, **kwargs):
...
-            on_done=on_done,
-            on_error=on_error,
-            on_always=on_always,
-            exec_endpoint=on,
-            target_peapod=target_peapod,
-            parameters=parameters,
-            request_size=request_size,
-            **kwargs,
-        )
+        with ExitStack() as stack:
```
jacobowitz, line 63: Isnt this also done in the regular Client and not only in the Flow context? Can we guarantee that this never is executed if a user just uses the regular Client?

jacobowitz: Maybe you can use self.client to set up the context somehow? So you dont need to do it every time?

jacobowitz (changes requested):
Interesting that there even exists a python package for this task :D

I have some comments on the specific solution, but I think generally its okay do it like this

jina/peapods/pods/k8slib/kubernetes_tools.py
```diff
@@ -167,7 +165,9 @@ def get_port_forward_contextmanager(
 
     :return: context manager which sets up and terminates the port-forward
     """
-    _wait_for_gateway(namespace, timeout)
+    import portforward
+
+    _wait_for_flow_ready(namespace, timeout)
```
JoanFM, line 170: this should not be implemented here. This should be part of every `K8sPod` `wait_for_ready` logic.

JoanFM: Let's have that before this

JoanFM (changes requested):

extra-requirements.txt
```diff
@@ -72,8 +72,8 @@ jsonschema:                 cicd
 dgl:                        cicd
 black==20.8b1:              test
 matplotlib:                 devel
-kubernetes>=18.20.0:        devel
-portforward>=0.2.4:         devel
```
JoanFM: i am not sure about this. I would still keep in devel since this is for debugging and not needed in any image

JoanFM (changes requested):

jina/peapods/pods/k8s.py
florian-hoenicke: What would you suggest?
In theory, there could be multiple ingress setup for the same gateway, so it could be even a list of hosts.

jina/flow/base.py
```diff
@@ -1075,6 +1075,14 @@ def start(self):
 
         self._build_level = FlowBuildLevel.RUNNING
 
+        if (
+            self.args.infrastructure == InfrastructureType.K8S
+            and self.args.k8s_startup_time > 0
```
JoanFM, line 1082: what is this argument?

jina/parsers/flow.py
```diff
@@ -52,6 +52,12 @@ def mixin_k8s_parser(parser):
         if _SHOW_ALL_ARGS
         else argparse.SUPPRESS,
     )
+    gp.add_argument(
+        '--k8s-startup-time',
+        type=int,
+        default=60,
+        help='Time to wait until the kubernetes flow started. In the future, we will solve this using probes.',
```
JoanFM, line 59: Do not add this, this will be soon solved

JoanFM (changes requested):

jina/clients/mixin.py
```diff
@@ -61,14 +61,17 @@ async def _get_results(*args, **kwargs):
                 return result
 
         with ExitStack() as stack:
-            if self.args.infrastructure == InfrastructureType.K8S:
+            if (
+                not hasattr(self.args, 'infrastructure')
```
JoanFM, line 65: why it should be the case??? It always has them no?

jina/clients/mixin.py
JoanFM: there is no point on using a stack when u have only one context manager to handle

jina/clients/mixin.py
florian-hoenicke: not when running a client directly

jina/parsers/flow.py
florian-hoenicke: I have to add this otherwise the code will not work.
Let's have the independent feature development for probes and port-forward.
This attribute can be removed when introducing probes.

jina/flow/base.py
florian-hoenicke: The time to wait between the startup of the flow and the first request sent

extra-requirements.txt
florian-hoenicke: thanks done

jina/peapods/pods/k8slib/kubernetes_tools.py
florian-hoenicke: thanks removed

jina/clients/mixin.py
florian-hoenicke: > isn't this also done in the regular Client and not only in the Flow context? Can we guarantee that this never is executed if a user just uses the regular Client?


Thanks. When just running the client, there is no infrastructure attribute and therefore the normal context is used.

florian-hoenicke: > there is no point on using a stack when u have only one context manager to handle

The point is to have more compact code.

jina/parsers/flow.py
JoanFM: No, this will wait until the other, they are part of the same, do not want to add an unnecessary attribute in the wrong direction. The other PR will be ready soon

jina/peapods/pods/k8slib/kubernetes_tools.py
```diff
@@ -11,6 +13,9 @@
     cur_dir, '..', '..', '..', 'resources', 'k8s', 'template'
 )
 
+if False:
+    from contextlib import GeneratorContextManager
```
deepankarm, line 17: For which python version `GeneratorContextManager` exists? I can find a `_GeneratorContextManager`, but not this.

jina/clients/mixin.py
```diff
@@ -57,18 +60,34 @@ async def _get_results(*args, **kwargs):
...
+            elif self.args.infrastructure == InfrastructureType.K8S:
+                stack.enter_context(
+                    kubernetes_tools.get_port_forward_contextmanager(
+                        self.args.name, self.port_expose
+                    )
+                )
+            else:
+                raise NotImplemented(
+                    f'Infrastructure type {self.args.infrastructure} unknown.'
+                )
```
deepankarm, line 78: I would change this to
```suggestion
        context_mgr = nullcontext()
        if self.args.infrastructure == InfrastructureType.K8S:
            context_mgr = kubernetes_tools.get_port_forward_contextmanager(
                self.args.name, self.port_expose
            )

        with context_mgr:
            ...
```

deepankarm (changes requested):

extra-requirements.txt
```diff
@@ -72,6 +72,7 @@ jsonschema:                 cicd
 dgl:                        cicd
 black==20.8b1:              test
 matplotlib:                 devel
+portforward>=0.2.4:         devel
```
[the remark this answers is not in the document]
JoanFM: This would be ideal @deepankarm but we have a problem about the hub images, there is currently no way to force them to have a different set of dependencies

JoanFM (changes requested):
PR to wait for #3509. startup time should not be added to remove later.

jina/clients/mixin.py
florian-hoenicke: thanks, changed. Much better

jina/peapods/pods/k8slib/kubernetes_tools.py
florian-hoenicke: I use python3.7

jina/peapods/pods/k8s.py
```diff
@@ -311,8 +311,12 @@ def port_expose(self) -> int:
 
     @property
     def host(self) -> str:
```
JoanFM, line 313: I think is better to remove this property, not needed apparently

JoanFM (changes requested):

extra-requirements.txt
```diff
@@ -72,7 +72,7 @@ jsonschema:                 cicd
 dgl:                        cicd
 black==20.8b1:              test
 matplotlib:                 devel
+portforward>=0.2.4:         perf, standard, devel
```
JoanFM, line 75: this is not needed in `perf`

JoanFM (changes requested):

extra-requirements.txt
florian-hoenicke: good point

jina/peapods/pods/k8s.py
florian-hoenicke: It is needed  when sending data.

florian-hoenicke: Gateway host is needed.

florian-hoenicke:
```
@property
    def client(self) -> 'BaseClient':
        """Return a :class:`BaseClient` object attach to this Flow.

        .. # noqa: DAR201"""

        kwargs = dict(
            host=self.host,
            port=self.port_expose,
            protocol=self.protocol,
        )
        kwargs.update(self._common_kwargs)
        return Client(**kwargs)
```

in base flow

jina/peapods/pods/k8slib/kubernetes_tools.py
florian-hoenicke: changed to Generator thanks

jina/resources/k8s/template/deployment.yml
```diff
@@ -17,6 +17,14 @@ spec:
       containers:
       - name: executor
         image: {image}
+#        resources:
+#          requests:
```
JoanFM, line 21: remember to remove these

JoanFM (changes requested):

JoanFM (approved):

PR #3500, opened 2021-09-28
fix: progress bar overwrite other logs

maximilianwerk:
The current progress bar overwrites the last log line printed before the bar starts. Old behavior:

```
PASSED                                                                                              
PASSED                                                                                            
PASSED                                                                                              
PASSED                                                                                              
⠴   CANCELED ━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0:00:00 8.7 step/s FAILED            
```

New behavior:

```
tests/unit/test_bla.py:my_test1
PASSED
tests/unit/test_bla.py:my_test2
PASSED
tests/unit/test_bla.py:my_test3
PASSED
tests/unit/test_bla.py:my_test4
PASSED
tests/unit/test_bla.py:my_test5
⠴   CANCELED ━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0:00:00 8.7 step/s FAILED            
```
Canceled from me doing ctrl+c

### DIFF
--- a/jina/logging/profile.py
+++ b/jina/logging/profile.py
@@ -227,6 +227,7 @@ class ProgressBar(TimeContext):
         self._num_update_called = 0
         self._on_done = message_on_done
         self._stop_event = threading.Event()
+        print()
 
     def update(
         self,


jina/logging/profile.py
```diff
@@ -227,6 +227,7 @@ def __init__(
         self._num_update_called = 0
         self._on_done = message_on_done
         self._stop_event = threading.Event()
+        print()
```
JoanFM, line 230: can u add a comment on top explaining the reason, to avoid people from removing it?

JoanFM (changes requested):

jina/logging/profile.py
maximilianwerk: What we actually should do on each update of the log line: Check if the content on the current log line is what we wrote there the last time. Only then update, otherwise create a new line.

JoanFM (approved):

PR #3502, opened 2021-09-28
feat: add configmap template

bwanglzu:


jina/peapods/pods/k8slib/kubernetes_deployment.py
```diff
@@ -101,6 +103,19 @@ def deploy_service(
         custom_resource_dir=custom_resource_dir,
     )
 
+    kubernetes_tools.create(
+        'configmap',
+        {
+            'name': name,
+            'target': name,
+            'namespace': namespace,
+            'data': envs,
```
jacobowitz, line 112: this means that the defaults in the template are always overriden no? So if the user does not provide any env variables things like JINA_LOG_LEVEL will be missing?

jacobowitz (commented):
Looks convinient to use! We should check with @deepankarm though what he things and how we do the same in jinad.

I also had two comments/questions on things work with this config map. Never used it myself before.

jina/resources/k8s/template/configmap.yml
```diff
@@ -0,0 +1,9 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  namespace: {namespace}
+  name: config-map
+data:
+  jina.log.level: "DEBUG"
```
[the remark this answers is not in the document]
bwanglzu: lower case will be automatically convert to upper case

jina/peapods/pods/k8slib/kubernetes_deployment.py
bwanglzu: PR updated

JoanFM:
@bwanglzu are the 2 tasks listed done?

bwanglzu:
@JoanFM needs some alignment with 2 ppls, maybe next Monday

jina/resources/k8s/template/configmap.yml
```diff
@@ -0,0 +1,8 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: config-map
+data:
+  jina_log_level: "DEBUG"
```
JoanFM, line 6: this comes in capital letters no?

jina/resources/k8s/template/configmap.yml
```diff
@@ -0,0 +1,8 @@
...
+  pythonunbuffered: "1"
```
JoanFM, line 7: all of these actually?

JoanFM (changes requested):

jina/resources/k8s/template/configmap.yml
bwanglzu: i remembered is a common practice to use lowercase in configmap. It will be converted in upper case automatically.

JoanFM: okk

JoanFM:
@bwanglzu make it a draft then

jina/peapods/pods/k8slib/kubernetes_tools.py
```diff
@@ -89,13 +89,15 @@ def networking_v1_beta1_api(self):
 
 def create(
     template: str,
```
florian-hoenicke, line 91: I think the parameters `kind` and template are redundant. Let's only keep one of them.

florian-hoenicke (changes requested):

jina/peapods/pods/k8slib/kubernetes_tools.py
```diff
@@ -138,3 +143,22 @@ def _get_yaml(template: str, params: Dict, custom_resource_dir: Optional[str] =
         for k, v in params.items():
             content = content.replace(f'{{{k}}}', str(v))
     return content
+
+
+def _patch_yaml(template: str, params: Dict, custom_resource_dir: Optional[str] = None):
+    import yaml
+
+    if custom_resource_dir:
```
florian-hoenicke, line 151: the custom resource dir logic should not be handled in this function. Instead, the function should just know about the `resource_dir`

jina/resources/k8s/template/deployment-init.yml
```diff
@@ -37,13 +37,9 @@ spec:
...
-        env:
-        - name: JINA_LOG_LEVEL
-          value: debug
-        - name: PYTHONUNBUFFERED
-          value: "1"
-        - name: WORKER_CLASS
-          value: "uvicorn.workers.UvicornH11Worker"
+        envFrom:
+          - configMapRef:
+              name: config-map
```
florian-hoenicke, line 42: If you have multiple executors within the same namespace, they will overwrite the config-maps. Therefore, they need a different name, e.g., {executor_dns_name}-env.

jina/peapods/pods/k8slib/kubernetes_deployment.py
```diff
@@ -101,18 +105,32 @@ def deploy_service(
         custom_resource_dir=custom_resource_dir,
     )
 
+    kubernetes_tools.create(
```
florian-hoenicke, line 108: I think it is important to create the config map before you create the deployment.

florian-hoenicke (changes requested):

jina/parsers/peapods/pod.py
```diff
@@ -145,3 +145,13 @@ def mixin_k8s_pod_parser(parser):
         if _SHOW_ALL_ARGS
         else argparse.SUPPRESS,
     )
+
+    parser.add_argument(
+        '--k8s-envs',
```
jacobowitz, line 150: Why do we need a new argument? Cant we just use the [existing one](https://github.com/jina-ai/jina/blob/master/jina/parsers/peapods/pea.py#L48)?

jacobowitz (changes requested):
I checked the existing env variable args and I think we can just reuse that?

jina/parsers/peapods/pod.py
deepankarm: yes. @bwanglzu This variable `env` exists for `Pea`, `Pod` & `Flow`. Let's reuse them here.

jina/peapods/pods/k8slib/kubernetes_tools.py
bwanglzu: you're correct

jina/parsers/peapods/pod.py
bwanglzu: good, I'll make use `env`, thanks!

PR #3504, opened 2021-09-28
refactor: small connection pool cleanup

JoanFM:


jina/peapods/networking.py
```diff
@@ -225,28 +219,34 @@ async def _process_events(self):
...
-        Subscribes on MODIFIED events from list_namespaced_pod AK8s PI
+        Subscribes on MODIFIED events from list_namespaced_pod K8s API
         """
+        from kubernetes import watch
+        from ..peapods.pods.k8slib.kubernetes_tools import K8SClients
 
         self.enabled = True
+        k8s_client = K8SClients().v1
+        api_watch = watch.Watch()
+        loop = get_or_reuse_loop()
```
jacobowitz, line 230: this will create a new loop in this thread instead of using the existing one from the other thread. I am not sure this is desirable:

- somebody needs to manage the loop (run/stop)
- I am using the loop to [safely call ](https://github.com/jina-ai/jina/blob/master/jina/peapods/networking.py#L237)the non threadsafe asyncio queue from the other thread, I think this only works if we use the same loop?

jacobowitz (commented):
I am not sure about the loop, the rest looks good to me

PR #3510, opened 2021-09-28
feat(daemon): allow gpus arg for remote Executors

deepankarm:
This PR
1. Enables any remote Executors to set `gpus` argument with `Flow` syntax. This argument was only valid for local + dockerized Executors. Since all remote Executors are dockerized, this enables setting up `docker devices` on any remote containers.
2. All remote Flows are by default started with `gpus: all` devices set, as we don't have `gpus` arg for Flows. If needed and used, we can enable it later.

Note: If on remote, docker cannot access the device passed, we reset the device & start the container again. 

Usage with remote Executor
----

```python
f = Flow().add(
    uses='CustomExecutor',
    host=CLOUDHOST,
    port_jinad=8000,
    py_modules=['/path/to/pymodule.py'],
    upload_files=['/path/to/requirements.txt'],
    gpus='all'
)
```

### DIFF
--- a/daemon/api/dependencies.py
+++ b/daemon/api/dependencies.py
@@ -12,6 +12,7 @@ from jina import Flow, __docker_host__
 from jina.helper import cached_property
 from jina.peapods import CompoundPod
 from jina.peapods.peas.helper import update_runtime_cls
+from jina.peapods.runtimes.container.helper import get_gpu_device_requests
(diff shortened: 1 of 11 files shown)


JoanFM (dismissed):
LGTM!

JoanFM (approved):

PR #3514, opened 2021-09-29
feat: expose apps api from k8s

JoanFM:


jina/peapods/pods/k8slib/kubernetes_tools.py
```diff
@@ -49,42 +51,52 @@ def k8s_client(self):
...
-        if self.__networking_v1_beta1_api is None:
-            self.__instantiate()
-        return self.__networking_v1_beta1_api
+        if self._networking_v1_beta1_api is None:
+            self._instantiate()
+        return self._networking_v1_beta1_api
 
 
-__k8s_clients = K8SClients()
+_k8s_clients = K8sClients()
```
bwanglzu, line 99: shall we consider to make it public? I made a very first draft without creating template [here](https://github.com/jina-ai/jina/pull/3502/files#diff-763dd0d69c0c069bce44805ebe35b455d4c9bbea3819a122ca64333e514f2040R104) which takes a client as the first argument.

bwanglzu (approved):
LGTM, minor comment

jina/peapods/pods/k8slib/kubernetes_tools.py
JoanFM: We can consider later, I am not sure why it is so important to keep this as "singleton" style, should not create such big objects. I'd rather have every user instantiate one

PR #3526, opened 2021-09-30
ci: try to fix cd yml

JoanFM:


.github/workflows/cd.yml
```diff
@@ -112,25 +115,8 @@ jobs:
           release_token: ${{ secrets.JINA_CORE_RELEASE_TOKEN }}
 
   prep-testbed:
-    needs: [update-doc, update-schema]
```
hanxiao, line 115: keep the update-schema, otherwise chore job will trigger it

PR #3535, opened 2021-09-30
feat(sharding): align shard and replica concept with k8s

### LINKED ISSUE #3489: Align Jina native Replicas/shards with K8s style

jacobowitz:
WIP: This is meant to unify shard/replica concepts

* With this PR a `CompoundPod` manages shards and a `Pod` manages replicas (before it was the other way around)
* Its consistent with the `K8sPod` implementation
* I want to eventualy remove parallel args, now it is just an alias for shards and I use shards everywhere. Parallel can still be used to avoid a breaking change.

This PR is not a breaking change, but complex Flows using shards and replicas might see slightly different behaviour as shards and replicas are swapt around.

Fixes #3489


jina/helloworld/fashion/app.py
```diff
@@ -73,7 +73,7 @@ def hello_world(args):
     # load index flow from a YAML file
     f = (
         Flow()
-        .add(uses=MyEncoder, parallel=2)
+        .add(uses=MyEncoder, shards=2)
```
JoanFM, line 76: here we should use `replicas`, because it is about hitting only one instance

jina/helloworld/chatbot/app.py
```diff
@@ -20,7 +20,7 @@ def _get_flow(args):
     """Ensure the same flow is used in hello world example and system test."""
     return (
         Flow(cors=True)
-        .add(uses=MyTransformer, parallel=args.parallel)
+        .add(uses=MyTransformer, shards=args.shards)
```
JoanFM, line 23: same here

jina/parsers/helloworld.py
```diff
@@ -170,10 +170,10 @@ def set_hw_chatbot_parser(parser=None):
         help='The port of the host exposed to the public',
     )
     parser.add_argument(
-        '--parallel',
+        '--shards',
```
JoanFM, line 173: I would say `parallel` now should point to `replicas`

JoanFM (commented):
Seems to be in the right direction.

JoanFM:
Just a minor comment, I think now `parallel` should become an `alias` of `replicas` since it is the most common behavior to have `ANY`

jacobowitz:
> Just a minor comment, I think now `parallel` should become an `alias` of `replicas` since it is the most common behavior to have `ANY`

@JoanFM 
I had this in a first version as I agree that it makes more sense. But then I dont want to introduce a breaking change here and changing `parallel` to an alias of `replicas` would be a breaking change. If anyone uses `parallel` with polling `ALL` this would break, because replicas with dont support polling.
With the current version of the PR all existing uses of `parallel` should still work as nothing changes there.

JoanFM:
@jacobowitz may u add a screenshot of the Flow.pot() before and after this code when using replicas and shards, combined and alone?

PR #3537, opened 2021-09-30
fix(gateway): set default prefetch size to 1

hanxiao:
prev. `50` is quite arbitrary and it gives quite misleading feeling that Jina is slow whereas the fact is the generator is slow, and collecting 50  req x 100 docs (default `request_size`) is slow


AMSAJ0 (dismissed):

JoanFM (approved):

hanxiao:
i think this one beyond just one-line change.
- tests are broken on `prefetch=1`
- seems no way to completely disable prefetching

JoanFM:
> i think this one is beyond just a one-line change.
> 
> * tests are broken on `prefetch=1`
> * seems no way to completely disable prefetching
> 
> will leave it to @jacobowitz

Isn't the override in the specific test enough?

JoanFM:
Another thing @hanxiao and @jacobowitz is why the test fails with prefetch one? does it mean that one does not prefetch next until the other is done?

hanxiao:
> ? does it mean that one does not prefetch next until the other is done?

yes, this is exactly why I say this PR needs further investigation. I think part of the code in gateway is written in a way that `prefetch` is always nonzero number

PR #3540, opened 2021-09-30
refactor: wait start success async not thread

JoanFM:
**Changes introduced**
Removes the `threads` started by Jina Flow process to check for `successfull` started of Pods and change them to be asynchronous.

**Motivations**
- Python does not work well with threads
- Suspicious that our testing CI does not behave so well with threads and this may be causing more `hanging` tests since the change was introduced
- Avoid the potential presence of daemon threads or concurrency issues, since Flows and Pods are not thread safe objects


Anthony00107 (dismissed):

jina/flow/base.py
```diff
@@ -1071,26 +1071,28 @@ def start(self):
             if not getattr(v.args, 'external', False):
                 self.enter_context(v)
 
-        self._wait_until_all_ready()
+        import asyncio
+
+        if not asyncio.run(self._wait_until_all_ready()):
```
deepankarm, line 1076: This would fail in jupyter like enviornments, as the event-loop is already running and this code executes in the main process. 

We have a `jina.helper.run_async` method, which can be used to fix it. But executes `asyncio.run(func())` in a separate thread, which is (mostly) not different from keeping this as-is & running them directly on different threads.

jina/helper.py
```diff
@@ -1190,14 +1195,20 @@ def run(self):
...
-            try:
+            result = getattr(thread, 'result', None)
+            if result is not None:
                 return thread.result
-            except AttributeError:
-                from .excepts import BadClient
+            else:
+                thread_exc = getattr(thread, 'exception', None)
+                if thread_exc is not None:
+                    print(f' Joan here thread exception {thread_exc}')
```
JoanFM, line 1204:
```suggestion
```

PR #3542, opened 2021-10-01
fix(kubernetes): add kubernetes dep to perf

### LINKED ISSUE #3536: Kubernetes dependency not installed in Executor images
Since 2.1.5 jina is broken in K8s, because the kubernetes dependency is missing from all executor images.
This is because kubernetes is only part of `standard` and not part of `core`. 

We either need to install `standard` in hubble or add kubernetes to `core`.

jacobowitz:
fixes #3536

Perf is needed so that the requirement is actually part of all hubble built docker images.

### DIFF
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -72,7 +72,7 @@ jsonschema:                 cicd
 dgl:                        cicd
 black==20.8b1:              test
 matplotlib:                 devel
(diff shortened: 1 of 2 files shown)


JoanFM (dismissed):
Not a big fan of this solution but it is needed

davidofell5 (approved):

JoanFM (approved):

PR #3549, opened 2021-10-01
fix: conda requirements

tadejsv:
This updates conda requirements, due to addition of `kubernetes` requirement

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -105,12 +105,12 @@ outputs:
         - fastapi
         - filelock
         - grpcio >=1.33.1
-        - kubernetes >=18.20.0
         - lz4 <3.1.2
         - numpy
         - pathspec
         - protobuf >=3.13.0
         - pydantic
+        - python-kubernetes >=18.20.0
         - python-multipart
         - pyyaml >=5.3.1
         - pyzmq >=17.1.0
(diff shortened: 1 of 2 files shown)


JoanFM (changes requested):
Do we have docker images downloading the conda package for our executors? if not, I do not see the need for this. Kubernetes as default dependency should not be there for long

JoanFM:
If there is no clear bug about this, I will close this PR. Kubernetes should soon disappear from our standard dependencies

tadejsv:
@JoanFM yes there was a clear bug - conda and pip dependencies have to be in sync, otherwise conda packages do not build.

conda/meta.yaml
```diff
@@ -1,13 +1,13 @@
 {% set name = "jina" %}
-{% set version = "2.0.23" %}
+{% set version = "2.1.5" %}
```
JoanFM, line 2: how is this adapted? conda versions will only exist for this hardcoded versions? we do not have any job to update this?

tadejsv: This is acutally auto-updated in the [feedstock repo](https://github.com/conda-forge/jina-feedstock) by conda-forge bot (see [this PR](https://github.com/conda-forge/jina-feedstock/pull/8) for example. The recipe here is not what is used by conda (you have to copy/paste it to feedstock repo first)

I could add some cd action here, but I feel it is not worth the effort - from 2.0.x till now I never had to make any adjustments, and I don't foresee this being a common thing. Right now I just need to make manual changes when there are changes in the requirements.
 
 What's really important is the script that generates the output.

JoanFM (approved):

Issue #3485, opened 2021-09-27
Add getter and setter for `buffer` in `DocumentArray` for completion


JoanFM:
**Describe the feature**
We have getters and setters for some attributes. (Content and embeddings), but we are missing `buffers` as the only `left` attribute

PR #3550, opened 2021-10-01
feat: add buffers getter setter

bwanglzu:
#3485


tadejsv (dismissed):
LGTM

JoanFM (commented):
Why do we have `Tuple` as annotation when it is clearly a list?

JoanFM (approved):

bwanglzu:
indeed, I tested `tags`, `texts` `buffers` should be list, I updated all the return types.

JoanFM (approved):

PR #3554, opened 2021-10-02
>>>Update neural_ops.py

CHAITANYASAI-del:


JoanFM:
This ia duplicated from #3481

PR #3558, opened 2021-10-02
feat(DocumentArray): sort top_k

### LINKED ISSUE #3467: DocumentArray `sort` should also get `top_k` to allow for more efficient sorting

sauravgarg540:
This PR resolves #3467

- Added `_sort_top_k` method to sort only top_k elements. 
- For `reverse = False`(default), Sorted top_k smallest values them in ascending order and keep it at the top while the rest of the list may not be sorted.
- For `reverse = True`, Sorted top_k biggest values  and keep it at the top while the rest of the list may not be sorted.
- Added unit test case

### DIFF
--- a/jina/types/arrays/document.py
+++ b/jina/types/arrays/document.py
@@ -397,11 +397,82 @@ class DocumentArray(
             hi_idx -= 1
         self._update_id_to_index_map()
 
-    def sort(self, key=None, *args, **kwargs):
+    def _default_key(value):
         """
-        Sort the items of the :class:`DocumentArray` in place.
+        By default use values of arrays to sort on
+
+        :return: returns value in array
+        """
+        return value
+
+    def _swap(self, index_i, index_j):
+        """
+        In-place swap the elements at index_i and index_j.
+
+        :param index_i: index of an element to perform swap on
+        :param index_j: index of an element to perform swap on
+        """
+
+        tmp = jina_pb2.DocumentProto()
+        tmp.CopyFrom(self._pb_body[index_i])
+        self._pb_body[index_i].CopyFrom(self._pb_body[index_j])
+        self._pb_body[index_j].CopyFrom(tmp)
+
+    def _sort_top_k(self, arr, key=_default_key, top_k: int = None, **kwargs):
+        """
+        Extension of sort method to sort top_k elements
+        For ascending orer put sorted top_k smallest element on the top
+        For decending orer put sorted top_k largest element on the top
+        while rest of the collection may not be sorted.
 
         :param key: key callable to sort based upon
+        :param arr: array to be partitioned and sorted
+        :param top_k: number of top k values to be sorted.
+        :param kwargs: keyword arguments to pass to the sorting underlying function
+        """
+
+        if top_k == 1:
+            return
+
+        # Allow only reverse in kwargs
+        check_kwargs = set(kwargs.keys()) - set(['reverse'])
+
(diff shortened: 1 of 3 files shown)


jina/types/arrays/document.py
```diff
@@ -397,11 +397,82 @@ def reverse(self):
...
+                self._swap(i, right)
+                if i <= pivot_index:
+                    left = i
+
+            self._sort_top_k(arr[:top_k], key=key, top_k=top_k - 1, **kwargs)
+
+        else:
+            raise ValueError(f'invalid argument: {check_kwargs}')
 
+    def sort(self, key=None, top_k: int = None, *args, **kwargs):
```
JoanFM, line 471: Please have proper type hinting

jina/types/arrays/document.py
```diff
@@ -397,11 +397,82 @@ def reverse(self):
...
+        :param key: key callable to sort based upon
+        :param arr: array to be partitioned and sorted
+        :param top_k: number of top k values to be sorted.
+        :param kwargs: keyword arguments to pass to the sorting underlying function
+        """
+
+        if top_k == 1:
+            return
+
+        # Allow only reverse in kwargs
```
JoanFM, line 437: In order to only allow reverse in kwargs u just need to add reverse as argument and remove kwargs

JoanFM (changes requested):

jina/types/arrays/document.py
```diff
@@ -397,11 +397,82 @@ def reverse(self):
             hi_idx -= 1
         self._update_id_to_index_map()
 
-    def sort(self, key=None, *args, **kwargs):
```
JoanFM, line 400: What is the point of this method?

JoanFM (changes requested):
I think the easiest and cleanest implementation is to have a heap based algorithm

jina/types/arrays/document.py
sauravgarg540: My idea was to keep the changes to the minimum of the already implemented `sort `method. I agree this method can be updated to accommodate `top_k `and `reverse`.

jina/types/arrays/document.py
sauravgarg540: My idea was to keep the changes to the minimum of the already implemented sort method. I agree this method can be updated to accommodate top_k and reverse.

sauravgarg540:
Hey @JoanFM , I have done some brainstorming on how I can use heaps for this issue, but I was unable to come up with a solution with heaps. That is why in this PR I have created an algo inspired from quickselect.

How would you suggest merging the whole document array(sorted and unsorted) when the heap is used? Also, I think a custom heap needs to be implemented.

Any algo I implement will be slower than timmsort(used by `sort()`). I also don't understand what's the idea behind this issue as when I would call  for sorted top_k elements I also expect only top_k elements as output.

sauravgarg540:
Hey @JoanFM , I see that another PR have been raised which may resolve the issue. I have one comment,

[test_matcharray.py](https://github.com/jina-ai/jina/blob/d36220207842f68db8a5a775c3caf9d067e21663/tests/unit/types/arrays/test_matcharray.py#L63)  uses sort method of DocumentArray. With the other PR, sorting `query.matches` with `top_k` and `reverse=False` will fail.

PR #3562, opened 2021-10-03
feat(client): add prefetch to http client

deepankarm:


JoanFM (commented):
It looks very good, lets see if @jacobowitz can take a close look as well. It would be nice to add unittests for the new classes or make sure we open another issue to make sure we add them properly

jina/peapods/runtimes/gateway/prefetch.py
```diff
@@ -157,3 +139,132 @@ async def prefetch_req(num_req, fetch_to):
                 # this list dries, clear it and feed it with on_recv_task
                 prefetch_task.clear()
                 prefetch_task = [j for j in onrecv_task]
+
+
+class ZmqPrefetchCaller(BasePrefetchCaller):
```
jacobowitz, line 144: Whats the reasoning to further split the BasePrefetchCaller into this ZmqPrefetchCaller? I think that is slightly misleading, because the ZmqPrefetchCaller actually can use either gRPC or ZMQ. 
So for me it would make more sense to have one abstract (BasePrefetchCaller) Prefetcher which can handle all Flow internal traffic (either zmq or GRPC), but has no functionality to communicate those results back to the outside world.

Then we can have concrete Http/Grpc/Websocket Prefetchers implementing the missing communication logic to the outside.
It might even be better to use compositon over inheritance here? So some generic BasePrefetchCaller is just used by some http/grpc/websocket server wrapping code? I think functionally speaking we mix two different things here: Comunication with the `Flow` and the `Client`

jacobowitz (changes requested):
I am not sure about the inheritance concept here, see comment below.

jina/peapods/runtimes/prefetch/base.py
```diff
@@ -100,23 +89,15 @@ async def prefetch_req(num_req, fetch_to):
...
-                    future = get_or_reuse_loop().create_future()
-                    self._message_buffer[next_request.request_id] = future
-                    asyncio.create_task(
-                        self.iolet.send_message(
-                            Message(None, next_request, 'gateway', **vars(self.args))
-                        )
-                    )
-
-                    fetch_to.append(future)
+                    self.handle_request(request=request, fetch_to=fetch_to)
```
jacobowitz, line 100: lets return the task in the `handle_request` and append it to the list here? Otherwise we have to expose the fetch_to list to the child class

jina/peapods/zmq/__init__.py
```diff
@@ -107,11 +107,11 @@ def _register_pollin(self):
             self.poller.register(self.in_connect_sock, zmq.POLLIN)
 
     def pause_pollin(self):
-        """Remove :attr:`in_sock` from the poller """
+        """Remove :attr:`in_sock` from the poller"""
```
jacobowitz, line 110: Lets remove the changes in this file from the PR? I think they are only linting and this PR should not touch this file it appears

jacobowitz (changes requested):
Two minor style comments, the rest looks good. I think that should work :)

jina/peapods/runtimes/prefetch/base.py
deepankarm: done

jina/peapods/zmq/__init__.py
deepankarm: reverted using [git restore](https://stackoverflow.com/a/59756673) :)

jina/clients/base/http.py
```diff
@@ -1,26 +1,68 @@
...
+
+    async def start(self):
+        """Create ClientSession and enter context
+
+        :return: self
+        """
+        import aiohttp
+
+        self.session = aiohttp.ClientSession()
+        await self.session.__aenter__()
```
JoanFM, line 49: to handle this, use the `ExitStack` inheritance pattern? Maybe there is some AsyncExitStack?

jina/peapods/runtimes/prefetch/base.py
```diff
@@ -1,76 +1,66 @@
...
-
-    def _process_message(self, message):
-        if message.request_id in self._message_buffer:
-            future = self._message_buffer.pop(message.request_id)
-            future.set_result(message)
-        else:
-            self.logger.warning(
-                f'Discarding unexpected message with request id {message.request_id}'
-            )
+    def _create_receive_task(self) -> asyncio.Task:
```
JoanFM, line 35: add the `abc` abstract class so that it will never let u instantiate one without this being implemented

JoanFM (changes requested):

jina/clients/base/http.py
deepankarm: I faced issues with `aiohttp.ClientSession()` while reusing the `AsyncExitStack` defined [here](https://github.com/jina-ai/jina/blob/feat-http-client-prefetch/jina/clients/base/http.py#L88), since that is the caller for `HTTPClientlet`. I'll recheck this while refactoring on `WebsocketBaseClient`, as that'd be adapted `aiohttp` in https://github.com/jina-ai/jina/issues/3563

jina/peapods/runtimes/prefetch/base.py
deepankarm: done

jina/peapods/runtimes/prefetch/base.py
```diff
@@ -1,76 +1,69 @@
...
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def receive(self):
+        """Implement `receive` logic for prefetcher
+
+        .. # noqa: DAR202
+        """
+        raise NotImplementedError
```
JoanFM, line 51:
```suggestion
        ...
```

jina/peapods/runtimes/prefetch/base.py
```diff
@@ -1,76 +1,69 @@
...
-                f'Discarding unexpected message with request id {message.request_id}'
-            )
+    @abstractmethod
+    def _create_receive_task(self) -> asyncio.Task:
+        """Start a receive task to be running in the background
+
+        .. # noqa: DAR202
+        :return: asyncio Task
+        """
+        raise NotImplementedError
```
JoanFM, line 43:
```suggestion
        ...
```

jina/peapods/runtimes/prefetch/base.py
```diff
@@ -1,76 +1,69 @@
...
+
+    def handle_request(
+        self, request: 'Request'
+    ) -> Union['asyncio.Task', 'asyncio.Future']:
+        """Handle each request in the iterator
+
+        :param request: current request in the iterator
+        """
+        raise NotImplementedError
```
JoanFM, line 60:
```suggestion
        ...
```

JoanFM (changes requested):

deepankarm:

deepankarm:
recheck

JoanFM (dismissed):

JoanFM (approved):

hanxiao (dismissed):
well done 👍 

can we set default `prefetch=1` in both gateway and client? the place to change is in `--prefetch` of `jina/parsers/peapods/runtimes/remote.py`

deepankarm:
I have reverted setting `prefetch` to its original value, as it has side effects in tests and it should be done in a separate PR.

JoanFM (approved):

PR #3570, opened 2021-10-04
feat(documentarray): heap based top k sorting in DocumentArray

gmastrapas:
Enable more efficient sorting in `DocumentArray` by exposing an optional `top_k` argument in method `sort`. Specifying a `top_k` integer guarantees that only the `top_k` elements will be sorted rather that sorting the entire document list. Uses `heapq.heapify` and `heapq.heappop` (and the maxheap counterparts) from stdlib.


jina/types/arrays/document.py
```diff
@@ -397,18 +406,28 @@ def overriden_key(proto):
...
-            if not overriden:
-                self._pb_body.sort(key=key, *args, **kwargs)
-            else:
-                self._pb_body.sort(key=overriden_key, *args, **kwargs)
+        if top_k is None or top_k >= len(self._pb_body):
+            self._pb_body.sort(key=_key, reverse=reverse)
         else:
-            self._pb_body.sort(*args, **kwargs)
+            if _key is not None:
+                self._pb_body = [(_key(element), element) for element in self._pb_body]
```
JoanFM, line 419: I would prefer not to change `self._pb_body`, I see that u do too to avoid having the `if` else at the other level. But all the magic done later seems to be more confusing. I would separate quite clearly what happens to the actual `_pb_body` and what are the values put in the heap.

JoanFM (changes requested):
Hey @gmastrapas ,

Thank you very much for the contribution, I have some minor comments

jina/types/arrays/document.py
JoanFM: Also, if some exception happens after here, the DocumentArray will stay in a very invalid state

tests/unit/types/arrays/test_documentarray.py
```diff
@@ -332,6 +332,22 @@ def test_da_sort_by_document_interface_in_proto():
...
 
 
+def test_da_sort_topk():
+    da = DocumentArray(
+        [Document(id=i, copy=True, scores={'euclid': 10 - i}) for i in range(10)]
+    )
+    assert da[0].id == '0'
+    assert da[0].scores['euclid'].value == 10
+
+    da.sort(top_k=3, key=lambda d: d.scores['euclid'].value, reverse=True)
```
JoanFM, line 342: This test should validate the up to top_k things are kept, and the rest are not kept in order.

Also it should validate that the set of Documents is still valid

JoanFM (changes requested):

tests/unit/types/arrays/test_documentarray.py
```diff
@@ -332,6 +332,29 @@ def test_da_sort_by_document_interface_in_proto():
     assert da[0].embedding.shape == (1,)
 
 
+def test_da_sort_topk():
```
JoanFM, line 335: what is the behavior when top-k is present but no key is given?

gmastrapas: Both with and without `top_k`, if `key=None` n exception is raised
```
TypeError: '<' not supported between instances of 'DocumentProto' and 'DocumentProto'
```

JoanFM: That was my fear. I think then we should not consider `key` as an Optional `argument`

JoanFM (changes requested):

tests/unit/types/arrays/test_documentarray.py
gmastrapas: Sure I can make it required. We 'll break the API but I think it raises the same Exception in `master` so we 're good, correct?

JoanFM: yes, we are good because `key` would be accepted as a `*kwargs` so good

gmastrapas:
Het @JoanFM can you lend a hand with the failed tests? I'm not getting much from the logs
```
test-flow-with-init-container@2805[E]:Flow is aborted due to ['gateway'] can not be started.
```

Issue #2906, opened 2021-07-09
optimizer TPESampler should automatically construct search space from parameters.yml


cristianmtr:
Currently we have to manually add the space and pass it as a kwarg

RaghavPrabhakar66:
@cristianmtr I would like to take this issue.

maximilianwerk:
Hey @RaghavPrabhakar66, thanks a lot for the initiative. Let me give you some guidance, what needs to be done:

After finishing this ticket, the following special handling for the `GridSampler` should not be necessary anymore: https://github.com/jina-ai/jina/blob/c0b04ffae9f2db818e56e170c5ce4efdb64497bb/tests/integration/optimizers/test_optimizer.py#L71

So the `grid_sampler_search_space` should be build in the `FlowOptimizer.opimize_flow` method. Most probably the best way is to add a `get_values_list` to all classes inheriting from `OptimizationParameter`. For the following classes computing an exhaustive list is straight forward: `IntegerParameter`, `CategoricalParameter`, `ExecutorAlternativeParameter`. For the other classes we

- Either need to implement a deterministic sampling, that samples some values 
- Or raise a `NotImplemented` exception with some speaking text like "GridSampler should only be used with exhaustive parameters (IntegerParameter, CategoricalParameter, ExecutorAlternativeParameter".

I hope this helps. If you have any further questions, please raise them as they appear.

Cheers!

RaghavPrabhakar66:
Hi, Thanks for the guidance. I have some doubts regarding the implementation. I cam across `load_optimization_parameters()` function and it return a list objects of all params. I was wondering if we could use that rather than creating a method in every class. It is based on the assumption that objects returned by `load_optimization_parameters` have `low`, `high`, `step_size` attributes. It could work like this
```python
self.parameters = load_optimization_parameters(self._parameter_yaml)

self.search_space = {}
for parameter in self.parameters:
    self.search_space[parameter.jaml_variable] = [parameter.low, parameter.high, parameter.step_size]
```

What are your thoughts? @maximilianwerk

maximilianwerk:
> It is based on the assumption that objects returned by load_optimization_parameters have low, high, step_size attributes.

This assumption is wrong. Take a `CategoricalParameter`: It only has `choices` as an attribute. It could be something like `model_name` with values `Bert1`, `Bert2` and `MyOtherFancyModel`.

RaghavPrabhakar66:
[...]

PR #3571, opened 2021-10-04
feat(optimizers): automatically infer the parameter search space in FlowOptimizer (#2906)

gmastrapas:
Build attribute `self._search_space` in the `FlowOptimizer` constructor, by calling the `.search_space` property in all the optimization parameters specified in `self.parameters`. The `search_space` property is implemented in all the subclasses of
`OptimizationParameter` except from `UniformParameter` and `LogUniformParameter`.


JoanFM (changes requested):
Hello @gmastrapas,

Thank you very much for this wonderful contribution. I have one question however, is it the case that `optimize_flow` does not work if `sampler` is `GridSampler` and `parameter` is `Uniform` or `LogUniform`.

If that is the case, would you be able to add some tests showing this?

JoanFM:
Also @gmastrapas ,

can u fix the commit message so that it aligns with the expected ones? 

```input: feat(optimizers): automatically infer the parameter search space in FlowOptimizer (#2906)
✖   header must not be longer than 72 characters, current length is 89 [header-max-length]
```

gmastrapas:
> Hello @gmastrapas,
> 
> Thank you very much for this wonderful contribution. I have one question however, is it the case that `optimize_flow` does not work if `sampler` is `GridSampler` and `parameter` is `Uniform` or `LogUniform`.
> 
> If that is the case, would you be able to add some tests showing this?

Yes, that is the case. I wasn't sure how to create the search space for sampled parameters without adding additional stuff (e.g. search space size). I'll add a test case showcasing this

nan-wang:
recheckcla

JoanFM:
Hey @gmastrapas ,

Apparently there seems to be some tests failing, could you take a look?

PR #3572, opened 2021-10-04
ci: fix cd

bwanglzu:
> Error when evaluating 'strategy' for job 'core-test'. (Line: 137, Col: 20): Error reading JToken from JsonReader. Path '',

the error pop up after [this pr](https://github.com/jina-ai/jina/commit/4769b336bb0ff55cb65eb15466d4bacdd0bfc70f).
Hypothesis:
`pre-testbed` failed to generate outputs, thus produce empty output matrix (`core-test` needs `pre-testbed`). And in `core-test`:
```yaml
    strategy:
      fail-fast: false
      matrix:
        python-version: [3.7]
        test-path: ${{fromJson(needs.prep-testbed.outputs.matrix)}}
```
`fromJson` get an empty string.

Previously, `pre-testbed` has 2 needs: `update-schema` and `update-doc`. `update-doc` has been removed in #3526. 

this should be related to env settings. so I added env back to this PR. also some minor clean up such as remove unused `tests` in the bash script.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -264,6 +264,7 @@ jobs:
           JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
 
   k8s-test:
+    needs: update-schema
     runs-on: ubuntu-latest
     needs: update-schema
     steps:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -117,15 +117,14 @@ jobs:
   prep-testbed:
     runs-on: ubuntu-latest
     needs: update-schema
-    with:
-      inputs: '{ "release_token": "${{ env.release_token }}", "triggered_by": "CD"}'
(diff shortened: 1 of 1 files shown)


JoanFM (dismissed):

JoanFM (dismissed):

.github/workflows/cd.yml
```diff
@@ -117,6 +117,9 @@ jobs:
   prep-testbed:
     runs-on: ubuntu-latest
     needs: update-schema
+    with:
+      inputs: '{ "release_token": "${{ env.release_token }}", "triggered_by": "CD"}'
```
deepankarm, line 121: `${{ env.release_token }}` isn't available in the `prep-testbed` job. Please use 
```yaml
env:
```

deepankarm (changes requested):
Why do we need `release_token` with `prep-testbed`?

PR #3581, opened 2021-10-05
test: try to return results

JoanFM:


tests/unit/flow-construct/test_flow_except.py
```diff
@@ -143,25 +143,18 @@ def validate2(x, *args):
 
 @pytest.mark.parametrize('protocol', ['websocket', 'grpc', 'http'])
 def test_no_error_callback(mocker, protocol):
-    def validate2():
```
deepankarm, line 146: Since we have removed the callbacks, this test doesn't make sense anymore right? Let's skip it if it fails.

JoanFM: ah ok let me change this

JoanFM: This test is still valid, we checdk the `on_error` is calle

PR #3590, opened 2021-10-06
refactor: set prefetch to 1, remove on-recv :boom: 

deepankarm:


jina/parsers/peapods/runtimes/remote.py
```diff
@@ -164,14 +164,8 @@ def mixin_prefetch_parser(parser=None):
...
         type=int,
-        default=50,
-        help='The number of pre-fetched requests from the client',
-    )
-    gp.add_argument(
-        '--prefetch-on-recv',
-        type=int,
         default=1,
-        help='The number of additional requests to fetch on every receive',
+        help='The number of pre-fetched requests from the client',
```
hanxiao, line 168:
```suggestion
        help='The number of requests fetched from the client before feeding to the first Executor',
```

PR #3591, opened 2021-10-06
ci(docsbot): reindex and redeploy on cd

deepankarm:
https://www.notion.so/deepankarm-0db932972e624af18cbda187b6129a0d


.github/workflows/cd.yml
```diff
@@ -101,6 +101,13 @@ jobs:
           token: ${{ secrets.JINA_DEV_BOT }}
         env:
           release_token: ${{ secrets.JINA_CORE_RELEASE_TOKEN }}
+      - uses: benc-uk/workflow-dispatch@v1
```
JoanFM, line 104: I think we should check if there are changes in `docs`. Would it be possible?

JoanFM (changes requested):

PR #3597, opened 2021-10-06
docs: add link to flow parallelization

cristianmtr:


JoanFM (commented):
We will add this after PR #3535 is merged, and the concepts are aligned and clarified

PR #3626, opened 2021-10-11
docs: add env to docs

bwanglzu:


tadejsv:
But this is not really a flow feature, it's an executor feature. Passing this argument to the flow is just "syntax sugar" for passing it to all executors in the flow

bwanglzu:
@tadej-redstone agreed..so it is a bit confusing where to add, maybe have a bit more discuss needed

PR #3627, opened 2021-10-11
feat: k8s hubble no cache

florian-hoenicke:
- currently, it can happen that you have an old cache file from the hubble api on your machine. Means k8s won't pull the latest image.
This codechange forces to send a request to the api.


florian-hoenicke:
> How does it work in the `tests`? Can this be bypassed somehow?

We could mock the response from hubble, but then we would not test the integration.

Issue #3598, opened 2021-10-06
Change the way Runtimes are terminated


JoanFM:
**Describe the feature**
To continue with the view that Jina should learn how Kubernetes handles microservices, we have seen that `K8s` kills the runtimes running inside the Containers by emitting `TERMINATE` signals to the Pod as shown here (https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace)

This should be the way `Pea` terminates the signals, avoiding the complex cycle of TERMINATE signals sent to the Runtimes.

The implications would be:

- [ ] Finish ticket #3508 
- [ ] Verify the way SIGTERM, SIGKILL are managed by Gateway and Runtimes can be also kept in Windows
- [ ] Adapt `cancel` staticmethod of `ZEDRuntime`, `GRPCDataRuntime` , `GatewayRuntimes` to emit SIGTERM to the processes (Or maybe emit this from the `Pea`)
- [ ] Adapt `JinaDRuntime` termination logic to communicate to the JinaD endpoint instead of directly to the `Runtime` endpoint
- [ ] Adapt `ContainerRuntime` termination logic to send a TERMINATE signal to the main process of a `Docker` container.

This would align Jina with the practices of infrastructure frameworks

PR #3654, opened 2021-10-12
feat(runtimes): close via SIGTERM

jacobowitz:
Based on https://github.com/jina-ai/jina/pull/3599

Closes https://github.com/jina-ai/jina/issues/3598

There are quite some ugly hacks in here, but not really sure how to make it better:
* For the containers Han introduced some randomness before to avoid conflicts when running the same flow multiple times. I replaced this with port_in to make it deterministic
* We should not send deactivate when the Flow exits (because head peas exit first), I added a flag to the close method which is kind of ugly. But I hope this part of the code is going away soon anyway

### DIFF
--- a/jina/peapods/runtimes/grpc/__init__.py
+++ b/jina/peapods/runtimes/grpc/__init__.py
@@ -145,7 +145,6 @@ class GRPCDataRuntime(BaseRuntime, ABC):
         else:
             # This is to handle threads
             process._stop()
-        process.join()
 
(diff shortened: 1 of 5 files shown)


JoanFM:
@jacobowitz I would break this job in 2 PRs or tasks.

First make the changes and tests proving that all Runtimes can properly handle TERM signals, and afterwards do the change that make use of this feature in Peas.

Otherwise the PR will be too big

PR #3679, opened 2021-10-14
fix: fix windows py38 tornado

JoanFM:


JoanFM:
It would be nice to validate that the tests failed before this change in 3.8 right?

PR #3692, opened 2021-10-16
feat(runtime): return dict will pass to request.parameters

hanxiao:
https://jinaai.slack.com/archives/C018F60RBL5/p1633973734495200

unfortunately, our `@request` decorator is a simple map and it is hard to implement the explicit/enforce as mentioned in the discussion.


JoanFM:
> > adding a new field to request, where only responses are passed.
> 
> what is "where only responses are passed."

Actually it is already coherent with our Document

PR #3705, opened 2021-10-19
feat: add document array embedding filtering and validation logic to match function

winstonww:
Further to the discussion in this [PR](https://github.com/jina-ai/executor-simpleindexer/pull/21), we add the filtering/validation logic to the `match` function so that we still get match results even if some docs indexed have invalid embeddings. We think it might be better to do it here rather than at index time because at index time we do not have information about the `traversal_paths`.

Opening this draft PR to collect feedback on general direction the PR should head. More tests to be added.

Related issue:  jina-ai/internal-tasks#212

### DIFF
--- a/jina/types/arrays/neural_ops.py
+++ b/jina/types/arrays/neural_ops.py
@@ -79,7 +79,7 @@ class DocumentArrayNeuralOpsMixin:
             from .document import DocumentArray
 
             if not isinstance(rhv, DocumentArray):
-                rhv = DocumentArray(lhv)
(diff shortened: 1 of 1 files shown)


winstonww:
I ran the benchmark locally. `2.1.9.main` denotes the performance before the change, and `2.1.9.validate` denotes the performance after the change.

Since docs passed into `match` function all have valid embeddings in the benchmark, the performance difference is insignificant. 

Screen Shot 2021-10-20 at 1 53 50 PM

Issue #3711, opened 2021-10-19
docs: Create better clarification for shards vs replicas concepts


JoanFM:
**Describe the feature**
We want to clarify even more this concepts

PR #3721, opened 2021-10-20
docs(flowtopology): add subsection replica vs shard 

azayz:
Closes https://github.com/jina-ai/jina/issues/3711
Added a subsection describing the differences between Shards and Replicas, also the clarified the confusion between parallel and shards


JoanFM:
@jacobowitz any comments about this?

PR #3742, opened 2021-10-21
refactor: color_axis to channel_axis

### LINKED ISSUE #3738: refactor color_axis to channel_axis
In a typical image search task, we send image with dimensionality of `B, C, H, W` into a deep learning model, where:

- `B` refers to `batch_size`.
- `C` refers to `channels`, e.g. RGB is 3 channel image while grey-scale image only have 1 channel
- `H` is the image height
- `W` is the image width.

By default Jina executors and Finetuner use an argument `channel_axis` which allow user to move the channel axis to a specific axis. For instance `channel_axis=-1` means the image has the shape of `B, H, W, C`.

However, in Jina core, we have a different name `color_axis`. We want to make these two argument has the same name, so we expect someone up us refactor Jina, changing `color_axis` to `channel_axis`, and adjust associated unit tests.

rudranshsharma123:
Fixes #3738 by refactoring color_axis to channel_axis in the files:
- documents/__init__.py
- documents/converters.py
- documents/test_coverters.py

Hopefully, this fixes the issue. Kindly review and let me know if it's good enough


JoanFM:
> Hi, @JoanFM I tried amending the commit message, since this step is very new to me, I am not very sure if I did it correctly. Can you please help me in verifying if it's correct now or not?

It does not seem to have worked.

PR #3756, opened 2021-10-22
feat: try to pass NOBLOCK as flag to zmq send

JoanFM:


niitiikaa:
Check error, its on the wrong location

PR #3816, opened 2021-10-28
feat: add scale to Flow and Pod

JoanFM:
- [x] Scale method to flow & pod.
- [x] Add proper logic in K8sPod
- [x] Add and test scale down (IT) and unit test.
- [x] Add tests while client sending requests integration
- [x] Add tests after rolling update so that it scales the roll-updated Executor
- [x] add k8s integration test.


bwanglzu:
i'll pick up from here

daemon/stores/containers.py
```diff
@@ -225,7 +225,7 @@ async def add(
                         e.message += container_logs
             if id in Dockerizer.containers:
                 self._logger.info(f'removing container {id_cleaner(container.id)}')
-                Dockerizer.rm_container(container.id)
```
JoanFM, line 228: @bwanglzu remember to undo this change, this is only useful for testing locally

jina/peapods/pods/k8s.py
```diff
@@ -128,6 +128,14 @@ def _deploy_runtime(self):
                 ),
             )
 
+        def _scale_runtime(self, replicas: int):
+            k8s_client = kubernetes_client.K8sClients().apps_v1
+            k8s_client.patch_namespaced_deployment_scale(
+                self.dns_name,
+                namespace=self.k8s_namespace,
+                body={"spec": {"replicas": replicas}},
```
JoanFM, line 136: single quotes

tests/integration/scale/test_scale.py
```diff
@@ -160,3 +161,51 @@ def test_scale_fail(flow_with_runtime, pod_params, mocker):
...
+        (2, 3, 1),
+    ],
+    indirect=True,
+)
+@pytest.mark.parametrize('protocol', ['websocket', 'grpc', 'http'])
+def test_scale_with_client(flow_with_runtime, pod_params, protocol):
+    num_replicas, scale_to, shards = pod_params
+    with flow_with_runtime as f:
+        f.protocol = protocol
+        client = Client(
```
JoanFM, line 178: Why is this relevant? This is equivalent to using the `flow`. The relevant thing would be to have the clients send requests in parallel while the `scale` is happening

tests/integration/scale/test_scale.py
```diff
@@ -172,23 +172,30 @@ def test_scale_fail(flow_with_runtime, pod_params, mocker):
...
+        rv = c.index(Document(text=peer_hash), return_results=True)
+        for r in rv:
+            for replica_id in r.docs.get_attributes('tags__replica_id'):
+                ret.append(replica_id)
+        queue.put(ret)
+
+    num_replicas, scale_to, _ = pod_params
+    ret = []
+    queue = multiprocessing.Queue()
+    queue.put(ret)
```
JoanFM, line 189: why are u putting an empty list?

bwanglzu: initialize the empty list to store all the replica_ids

tests/integration/scale/test_scale.py
```diff
@@ -0,0 +1,209 @@
...
+        assert len(r2.docs) == 10
+        for replica_id in r1.docs.get_attributes('tags__replica_id'):
+            replica1_ids.add(replica_id)
+        for replica_id in r2.docs.get_attributes('tags__replica_id'):
+            replica2_ids.add(replica_id)
+    assert replica1_ids == replica2_ids == {0, 1}
+
+
+@pytest.mark.parametrize(
+    'pod_params',
```
JoanFM, line 169: can u make it more explicit and divide in 3 arguments?

JoanFM: also it would be good to test this while downscaling

tests/integration/scale/test_scale.py
```diff
@@ -0,0 +1,209 @@
...
+    def peer_client(port, protocol, peer_hash, queue):
+        c = Client(protocol=protocol, port=port)
+        ret = queue.get()
+        rv = c.index(Document(text=peer_hash), return_results=True)
+        for r in rv:
+            for replica_id in r.docs.get_attributes('tags__replica_id'):
+                ret.append(replica_id)
+        queue.put(ret)
+
+    num_replicas, scale_to, _ = pod_params
```
JoanFM, line 186: why do we add the 3rd param?

tests/integration/scale/test_scale.py
```diff
@@ -0,0 +1,209 @@
...
+
+    with flow_with_runtime as f:
+        f.protocol = protocol
+        port_expose = f.port_expose
+
+        thread_pool = []
+        for peer_id in range(5):
+            t = multiprocessing.Process(
+                target=partial(peer_client, port_expose, protocol, str(peer_id), queue),
+                daemon=True,
```
JoanFM, line 199: no need to have daemon

tests/integration/scale/test_scale.py
JoanFM: Review this communication pattern, u have a queue where u get and put with no control.

I think u should have the clients simply add to the queue, and assert the results from every client outside

tests/integration/scale/test_scale.py
bwanglzu: reusing the previous fixture defined in the test, the third one is shard

JoanFM: ah ok is indirect

tests/integration/scale/test_scale.py
bwanglzu: it's a bit trick since i added it as parameterized fixture,

bwanglzu: >also it would be good to test this while downscaling 
will add

JoanFM:
Test integration scale is leaving zombie processes that are making the pytest timeout

jina/peapods/pods/k8s.py
```diff
@@ -256,6 +256,63 @@ async def wait_restart_success(self, previous_uids: Iterable[str] = None):
...
+                logger.info(
+                    f'🏝️\n\t\tWaiting for "{self.name}" to be scaled, with {self.num_replicas} replicas,'
+                    f'scale to {scale_to}.'
+                )
+                timeout_ns = 1000000000 * _timeout if _timeout else None
+                now = time.time_ns()
+                exception_to_raise = None
+                while timeout_ns is None or time.time_ns() - now < timeout_ns:
+                    try:
+                        api_response = kubernetes_client.K8sClients().apps_v1.read_namespaced_deployment_scale(
```
jacobowitz, line 280: are we sure that `status.replicas` in the response actually contains only the ready replicas?

jacobowitz (commented):
I have one question regarding the use of the k8s api I am not sure about. Otherwise looks good to me!

jacobowitz (approved):
neat feature :+1:

PR #3826, opened 2021-10-29
feat: allow rolling update with uses_with in jinad

JoanFM:


JoanFM:
@deepankarm  it seems that `FastAPI` does not handle `Dicts` properly as `inputs`? Am I doing something wrong?

tests/distributed/test_workspaces/test_remote_workspaces.py
```diff
@@ -101,11 +102,11 @@ def test_upload_multiple_workspaces(replicas, mocker):
         )
     )
     with f:
-        f.index(
-            inputs=(Document(blob=np.random.random([1, 100])) for _ in range(NUM_DOCS)),
-            on_done=response_mock,
+        responses = f.index(
+            inputs=(Document(blob=np.random.random([1, 100])) for _ in range(NUM_DOCS))
```
deepankarm, line 106:
```suggestion
        responses = f.index(
            inputs=(Document(blob=np.random.random([1, 100])) for _ in range(NUM_DOCS)),
            return_results=True
```

daemon/clients/pods.py
```diff
@@ -18,22 +18,26 @@ class AsyncPodClient(AsyncPeaClient):
...
-        :param args: positional args
-        :param kwargs: keyword args
+        :param uses_with: the uses_with to update the Executor
         :return: flow id
         """
+        if dump_path is not None:
+            if uses_with is not None:
+                uses_with['dump_path'] = dump_path
+            else:
+                uses_with = {'dump_path': dump_path}
```
deepankarm, line 34: This seems to be repeated many times. Should we add a helper function?

daemon/api/endpoints/partial/flow.py
```diff
@@ -51,16 +51,18 @@ async def _create(flow: 'FlowModel', ports: Optional[PortMappings] = None):
 )
 async def _update(
     kind: UpdateOperation,
-    dump_path: str,
     pod_name: str,
-    shards: int = None,
+    uses_with: str,
```
deepankarm, line 55: [`uses_with` should be a `Dict[str, str]`](https://fastapi.tiangolo.com/tutorial/body-nested-models/?h=dict#bodies-of-arbitrary-dicts).

deepankarm (changes requested):

daemon/api/endpoints/partial/flow.py
JoanFM: In theory is a Dict of str and anything, would that be possible to implement?

daemon/clients/pods.py
JoanFM: we can do in anothe PR

daemon/clients/flows.py
```diff
@@ -77,7 +77,7 @@ async def update(
         self,
         id: Union[str, 'DaemonID'],
         pod_name: str,
-        dump_path: str,
+        dump_path: Optional[str] = None,
         *,
         uses_with: Dict,
```
JoanFM, line 82: this should also be `Optional` and default to `None`

JoanFM, line 82:
```suggestion
        uses_with: Optional[Dict] = None,
```

deepankarm (approved):

PR #3840, opened 2021-11-01
docs: clearer windows install instructions

alexcg1:
So many of the support requests we get on community Slack are about Windows issues which wouldn't occur if folks actually used WSL2 as recommended. Hence this PR makes it more noticeable and should reduce our support burden.


hanxiao (changes requested):
reject, we have a dedicated windows page
image

alexcg1:
Sure. I just want people to be able to find that page more easily

hanxiao:
reject as it is too shallow.

PR #3843, opened 2021-11-01
docs(tutorial): polish image2image

alexcg1:
Just polishing up the wording a little bit

### DIFF
--- a/docs/datatype/image/image2image.md
+++ b/docs/datatype/image/image2image.md
@@ -1,51 +1,50 @@
 # Search Similar Images
 
-Given an example image can we find similar images without the need of any labels? Leveraging Jina, the advantage that we
-get is that we do not need to use any labels or textual information about the images in order to build similar image
-search.
+Given an example image can we find similar images without the need of any labels? Leveraging Jina, we have the advantage that we
+we don't need to use any labels or textual information about the images in order to build a search for similar images.
 
-In this tutorial we are going to create an image search system that retrieves similar images. The data we are going to
-use is the test split of the [Dogs vs. Cats](https://www.kaggle.com/c/dogs-vs-cats/data?select=test1.zip) datasets. We
-will subsequently refer this dataset as pets dataset. It contains 12.5K images of cats and dogs. Now, we can define our
-problem as selecting an image of cat or dog, we would like to get similar images of cats or dogs respectively.
+In this tutorial we are going to create an image search system that retrieves similar images. We are going to
+use the test split of the [Dogs vs. Cats](https://www.kaggle.com/c/dogs-vs-cats/data?select=test1.zip) dataset, which we
+will subsequently refer to as the pets dataset. It contains 12.5K images of cats and dogs. Now, we can define our
+problem as selecting an image of cat or dog, and getting back images of similar cats or dogs respectively.
 
-We know that Jina searches semantically and this could vary with the neural network that we use for encoding. Since our
-task is to search similar images we will consider visually similar as semantically-related.
+We know that Jina searches semantically and this can vary with the neural network that we use for encoding our images. Since our
+task is to search for similar images we will consider visually similar images as semantically-related.
 
 ## Build the Flow
 
-The solution to the problem entails a simple pipeline that can be subdivided into two steps:  **Index** and **Query**
+The solution to the problem uses a simple pipeline that can be subdivided into two steps:  **Index** and **Query**
 
 ### Index
 
-To search something out of the full data, first we need to index the data. What it means is that we store the embeddings
+To search something out of the full dataset, we first need to index the data. This means that we store the embeddings
 of all the images from the pets dataset in some form of storage. The images can be read as a numpy array which is then
 fed to neural network of our choice. This neural network encodes the input images into some latent space which we call
-as embeddings. We then use **Indexer** to store these embeddings in memory.
+"embeddings". We then use **Indexer** to store these embeddings in memory.
 
 ### Query
 
-Once the data is indexed, i.e. our database is built, we simply need to feed our query which is an image or set of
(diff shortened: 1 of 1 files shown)


alexcg1:
Did we want to explain the indexer here? Instead we're just directing people to our concept docs

```Now, you might wonder what this *Indexer* is, or how to use a neural network of your choice. Worry not, we've got you
covered. In Jina we have three fundamental concepts that you need to know to follow this tutorial. If you
haven't read it yet, head on over to [Jina's basic concepts](https://docs.jina.ai/fundamentals/concepts/) page. 
```

I'd suggest that instead of this, we can just link the first occurrence of Document, Executor and Flow to their respective docs page

alexcg1:
> Did we want to explain the indexer here? Instead we're just directing people to our concept docs
> 
> ```
> covered. In Jina we have three fundamental concepts that you need to know to follow this tutorial. If you
> haven't read it yet, head on over to [Jina's basic concepts](https://docs.jina.ai/fundamentals/concepts/) page. 
> ```
> 
> I'd suggest that instead of this, we can just link the first occurrence of Document, Executor and Flow to their respective docs page

And/or mention that readers should check the basic concepts before they start

makram93:
I think we can remove the explanation. They would have gone through the basics before this tutorial

alexcg1:
> I think we can remove the explanation. They would have gone through the basics before this tutorial

We live in hope

docs/datatype/image/image2image.md
```diff
@@ -1,60 +1,53 @@
...
-Now, one might think what this *Indexer* is, or how to use neural network of our choice. Worry not, we've got you
-covered. In Jina AI, we have three fundamental concepts which is all you need to know to follow this tutorial. If you
-haven't read it yet, head on to [Jina's docs](https://docs.jina.ai/) page and give it a shot. Executor is the
-algorithmic unit in the Jina. It performs a single task on a `Document` or `DocumentArray`.
-
-We have many executors available at [Jina Hub](https://hub.jina.ai) - a marketplace for Executors. You can use any of
-them relevant to your tasks or build one of your own. Coming back to problem, we will use **SimpleIndexer** executor as
-our indexer (the one that stores and retrieves data). This executor also returns us the matching `Document` when we make
+We will use the **SimpleIndexer** Executor as
+our indexer (the one that stores and retrieves data). This Executor also returns the matching `Document` when we make
```
makram93, line 33:
```suggestion
The **Indexer** executor should support vector search along with indexing. For this tutorial we will create our own **Indexer** based on `SimpleIndexer` executor available on Jina Hub. This Executor returns the matching `Document` when we make
```

docs/datatype/image/image2image.md
```diff
@@ -85,19 +78,19 @@ class FlashImageEncoder(Executor):
...
 ```
 
-As one can see, how simple it is to build an `Encoder` executor. We simply inherit the base `Executor` and use decorator
-to define endpoints. As this `request` decorator is empty, it means that this function will be called regardless of the
-endpoints invoked, i.e., on both `/index` and `/search` endpoint. We
-leverage [lightning-flash](https://github.com/PyTorchLightning/lightning-flash) to use pre-trained `ResNet101` model for
-getting the embeddings. Reader can replace this model with any other pre-trained models of their choice. When this
-executor is instantiated, the pre-trained weights are downloaded automatically. The `predict` function takes in
-the `DocumentArray` and extracts embeddings which is then stored in the `embedding` attribute of the
+To build an Encoder Executor we inherit the base `Executor` and use a decorator
```
makram93, line 81:
```suggestion
To build an `Encoder` Executor we inherit the base `Executor` and use a decorator
```

docs/datatype/image/image2image.md
```diff
@@ -85,19 +78,19 @@ class FlashImageEncoder(Executor):
...
 respective `Document`.
 
-Finally, comes the storage/retrieval step. This we accomplish using an **Indexer** executor. We can use any of the
-available indexers on [hub.jina.ai](https://hub.jina.ai) or define our own. To create an **Indexer** we need to have two
-endpoints `/index` and `/search`. For this tutorial we will define a `SimpleIndexer` which is also available on jina
-hub.
+Finally, comes the storage/retrieval step. We do this with the **Indexer** Executor. You can use any of the
+available indexers on [Jina Hub](https://hub.jina.ai) or define your own. To create an **Indexer** you need to have two
+endpoints: `/index` and `/search`. For this tutorial we will define a `SimpleIndexer` which is [also available on jina
+Hub](https://hub.jina.ai/executor/zb38xlt4).
```
makram93, line 93:
```suggestion
Finally, comes the storage/retrieval step. You can use any of the
available indexers on [Jina Hub](https://hub.jina.ai) or define your own. To create an **Indexer** you need to have two
endpoints: `/index` and `/search`.  We can create an Executor which supports vector search and indexing as follows:
```

docs/datatype/image/image2image.md
alexcg1: **Automated message**

## Capitalization

Please capitalize first letters of the following (when used to reference project or component):

- Flow
- Executor
- Document
- Jina
- Jina Hub
- Python
- Docker
- Kubernetes

### Note!

This only applies **outside** a code or CLI context. If you're describing Python, YAML, any programming language, or command line input/output, please use the capitalization you would use **in that context**.

##### Right

```python
from jina.flow import Flow
```

##### Wrong

```python
from Jina.Flow import Flow
```

---

Made with [GitHub saved replies](https://github.com/settings/replies)

alexcg1: Remove "should"

docs/datatype/image/image2image.md
alexcg1: I don't think we need to backtick Encoder or Executor here. We're talking about them as concepts, not code

PR #3844, opened 2021-11-01
feat(proto): adjust proto for star routing

jacobowitz:
This PR makes the necessary changes to our protobuf models for star routing:

1. Remove the RoutingTable completely from the message. This will be an object kept at the gateway locally in the Future.
2. Remove the routes object from the Envelope. Thats not needed anymore as the Gateway can keep track of this locally and add the routes only in the end to the response to the Client.
3. Adjust the CommandMessage: A lot of them are not needed anymore, but ACTIVATE/DEACTIVATE will be used in the future to add/remove peas from a head. Thus we need the networking information to send alongside.
4. Change the gRPC service model to return a Message as the response, instead of the empty workaround

closes https://github.com/jina-ai/internal-tasks/issues/238

### DIFF
--- a/jina/proto/jina.proto
+++ b/jina/proto/jina.proto
@@ -157,28 +157,6 @@ message RouteProto {
     StatusProto status = 5; // the status of the execution
 }
 
-message TargetPodProto{
-    string host = 1; // the host HeadPea of the BasePod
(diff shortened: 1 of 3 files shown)


jacobowitz:
Im merging this for now to unblock me, but please still comment and I will address any issue in a follow up PR

PR #3846, opened 2021-11-01
feat(grpc): refactor grpclet interface

jacobowitz:
* expose polling type via grpclet and connection pool
* change arguments to prepare for removal of port_in
* inject connection pool to the grpclet, so its connections can be managed externally

### DIFF
--- a/jina/peapods/grpc/__init__.py
+++ b/jina/peapods/grpc/__init__.py
@@ -5,8 +5,9 @@ from typing import Optional, Callable
 import grpc
 from google.protobuf import struct_pb2
 
+from jina.enums import PollingType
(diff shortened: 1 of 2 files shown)


jina/peapods/grpc/__init__.py
```diff
@@ -17,59 +18,47 @@
...
     """
 
     def __init__(
         self,
-        args: argparse.Namespace,
-        message_callback: Callable[['Message'], None],
+        port: int,
+        connection_pool: ConnectionPool,
+        host: str = '0.0.0.0',
+        message_callback: Callable[['Message'], None] = None,
```
JoanFM, line 33: Optional

PR #3851, opened 2021-11-02
docs: multiprocessing spawn

deepankarm:


docs/fundamentals/flow/remarks.md
```diff
@@ -57,3 +57,134 @@ class C(Executor):
...
+    ```bash
+    pickle.PicklingError: Can't pickle: it's not the same object
+    AssertionError: can only join a started process
+    ```
+
+    ````
+
+    Inline functions, such as nested or lambda functions are not picklable. Use `functools.partial` instead.
+
+- **Always provide absolute path**
```
JoanFM, line 187: Is this for `spawn` or is this Windows specific?

docs/fundamentals/flow/remarks.md
```diff
@@ -57,3 +57,134 @@ class C(Executor):
             d.text += '!!!'
         return docs
 ```
+
+
+## multiprocessing Spawn
+
+Few cases require to use `spawn` start method for multiprocessing. To enable this, 
+
+- Please set `JINA_MP_START_METHOD=spawn` before starting the Python script.
```
JoanFM, line 66: Maybe we can add the examples provided by the community somehow, or the message they may see when this is required?

docs/fundamentals/flow/remarks.md
deepankarm: This is valid for all `spawn` because we rely on the callstack that is not available in the spawned process.

JoanFM (approved):

PR #3859, opened 2021-11-03
fix: set does not guarantee order

JoanFM:


bwanglzu (approved):
Indeed, thanks for fixing it.

PR #3860, opened 2021-11-03
Chore: Readme logo iteration change

BingHo1013:
The current logo SVG color scheme is not correct for the branding. The waving curve isn't correct and also should be two waves of color not three inside.
image


nan-wang (approved):
LGTM👍

nan-wang:
reopened at #3867

PR #3861, opened 2021-11-03
feat(types): add text_fields to get_vocabulary to doc and da

hanxiao:


JoanFM:
I am not convinced that this should be part of our core. We will always be lacking in terms of specific Analyzers for different languages and tokenizers. I think the message should be to move this to the hub

hanxiao:
it should be, it is part of the Document API.

Also, I'm tired of having low-quality unusable hub executor.

AMSAJ0:
V

PR #3869, opened 2021-11-04
ci: only run commit lint when only changing docs folder

mapleeit:
Two main changes in this PR:

1. Don't run `core-test` or any other test when just changing the docs. This will save a lot of time and resources.
2. Remove *adding-comment when commit lint fails* function because it won't work if the PR is raised by external contributors(in other words, the source branch of the PR comes from a forked repo). That means the value is very limited, and considering the complexity it adds, I would say it's a nice move to remove this functionality. So that this change won't increase a lot modulization concerns, because they are so simple.

### DIFF
--- a/.github/workflows/ci-no-code-related.yml
+++ b/.github/workflows/ci-no-code-related.yml
@@ -1,0 +1,53 @@
+name: CI for Code Un-related Changes
+
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '.github/workflows/**'
+
+jobs:
+  commit-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: find the prev warning if exist
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: 'bad commit message'
+      - name: Delete comment if exist
+        if: ${{ steps.fc.outputs.comment-id != 0 && !github.event.pull_request.head.repo.fork }}
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.issues.deleteComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-no-code-related.yml
```diff
@@ -0,0 +1,53 @@
+name: CI for Code Un-related Changes
```
JoanFM, line 1: why is this needed?

.github/workflows/ci.yml
```diff
@@ -2,6 +2,9 @@ name: CI
 
 on:
   pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '.github/workflows/**'
```
JoanFM, line 7: github workflows should not be added here, changes here may be potential problems

JoanFM (changes requested):

.github/workflows/ci-no-code-related.yml
mapleeit: Commit lint is still needed for the code un-related part.

.github/workflows/ci.yml
mapleeit: Ah. I see. For example, change the test logic in workflows. I'll remove this line.

.github/workflows/ci-no-code-related.yml
JoanFM: Can the `paths` be filtered at each step of the `regular CI`? Our CI is already non-modular enough

JoanFM (dismissed):

.github/workflows/ci-no-code-related.yml
mapleeit: `paths` can not be used on step or job level. For the modular concerns, I have some ideas that can improve it. 

I'll push some changes later.

mapleeit: I did some experiment. Unfortunately, [composite actions](https://docs.github.com/cn/actions/creating-actions/creating-a-composite-action) can't support all the function we need for CI moderation here.

hanxiao (changes requested):
in the new `.github/workflows/ci-no-code-related.yml`, the netlify preview should still be built

mapleeit:
@hanxiao Doesn't deploying stay in `.github/workflows/label-pr.yml`? It won't be affected.

hanxiao:
oh good point

PR #3870, opened 2021-11-04
feat(executor): Executor.from_hub now accepts jinahub scheme 🍾 

hanxiao:
Not going to lie, this is really one of my favorite features recently.

```python
from jina import Executor, DocumentArray, Document

exec = Executor.from_hub('jinahub://DummyHubExecutor')

da = DocumentArray([Document()])
exec.foo(da)
assert da.texts == ['hello']
```

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -721,6 +721,7 @@ class ArgNamespace:
         parser: ArgumentParser,
         warn_unknown: bool = False,
         fallback_parsers: List[ArgumentParser] = None,
+        positional_args: Tuple[str, ...] = None,
     ) -> Namespace:
         """
         Convert dict to a namespace.
@@ -729,9 +730,12 @@ class ArgNamespace:
         :param parser: the parser for building kwargs into a namespace
         :param warn_unknown: True, if unknown arguments should be logged
         :param fallback_parsers: a list of parsers to help resolving the args
+        :param positional_args: some parser requires positional arguments to be presented
         :return: argument list
         """
         args = ArgNamespace.kwargs2list(kwargs)
+        if positional_args:
+            args += positional_args
         p_args, unknown_args = parser.parse_known_args(args)
         if warn_unknown and unknown_args:
             _leftovers = set(unknown_args)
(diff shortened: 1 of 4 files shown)


JoanFM (changes requested):
Can we avoid using `load_config` for this but expose the `load` that encapsulates the previous work done by the runtime before load_config?

`load_config` is already ugly enough. And `passing jinahub://` does not seem to be a config as it is right?

JoanFM (commented):
Forget my previous message, I thought even docker could be used, but of course it cannot

jina/helper.py
```diff
@@ -721,6 +721,7 @@ def kwargs2namespace(
         parser: ArgumentParser,
         warn_unknown: bool = False,
         fallback_parsers: List[ArgumentParser] = None,
+        positional_args: Tuple[str, ...] = None,
```
JoanFM, line 724: what is an example where this is used?

JoanFM:
```suggestion
        positional_args: Optional[Tuple[str, ...]] = None,
```

jina/helper.py
```diff
@@ -721,6 +721,7 @@ def kwargs2namespace(
         parser: ArgumentParser,
         warn_unknown: bool = False,
         fallback_parsers: List[ArgumentParser] = None,
```
JoanFM, line 723:
```suggestion
        fallback_parsers: Optional[List[ArgumentParser]] = None,
```

jina/helper.py
JoanFM: I would keep `positional_args` outside, the user can append them on its own. The name of the function explicitly says `kwargs2Namespace` so why would it have to take `positional_args`?

JoanFM (changes requested):

jina/helper.py
hanxiao: because parser won't work

hanxiao: i would keep it as is as other wise people have to do this aganin

hanxiao: > what is an example where this is used?
in this pR

PR #3872, opened 2021-11-04
feat: refactor connection pool for star routing

jacobowitz:
This refactors the connection pool to unify the K8s and non K8s interface and allows us to use this new connection pool both in the Gateway and in the new heads. I adapted the existing connection pool tests, they should pass.

The new interface now allows us to add and remove pod connections to the pool and messages to pods/heads/shards. It is possible to address all shards, any shard and any head. Its not possible to address all replicas or a specific replica.

I am also adding some labels to you k8s deployments to allow the pool to get those (shard_id, jina_pod_name, pea_type).

The pool changes are unit tested with new tests.

I also did some random clean up (e.g. deleting grpclet)

This closes jina-ai/internal-tasks#248


jina/peapods/pods/k8s.py
```diff
@@ -23,6 +23,8 @@ def __init__(
             tail_port_out: int,
             head_zmq_identity: bytes,
             version: str,
+            pea_type: str,
```
JoanFM, line 26: why are we doing changes at this level?

jina/types/message/__init__.py
```diff
@@ -345,28 +344,6 @@ def _pod_str(r):
         route_str.append('⚐')
         return colored('▸', 'green').join(route_str)
 
-    def add_route(self, name: str, identity: str):
-        """Add a route to the envelope.
-
-        :param name: the name of the pod service
-        :param identity: the identity of the pod service
-        """
-        self._add_route(name, identity, self.envelope)
```
JoanFM, line 354: Why do we remove this?

jina/peapods/networking.py
```diff
@@ -104,67 +93,207 @@ def has_connection(self, address: str) -> bool:
...
+        self._connections = self._ConnectionPoolMap(self._logger)
 
     def send_message(
         self,
         msg: Message,
-        target_address: str,
+        pod: str,
+        head: bool,
+        shard_id: Optional[int] = None,
         polling_type: PollingType = PollingType.ANY,
```
JoanFM, line 230: Why do we have polling type here? Is the connection pool responsible for this? Should it not be handled at `shard` level?

jina/peapods/networking.py
```diff
@@ -311,80 +413,74 @@ def _pod_is_ready(item):
         )
 
     def _process_item(self, item):
-        deployment_name = item.metadata.labels["app"]
+        jina_pod_name = item.metadata.labels["jina_pod_name"]
```
JoanFM, line 416: single quotes

JoanFM (changes requested):

jina/peapods/pods/k8s.py
jacobowitz: We need to know if a kubernetes deployment is a head or a 'worker' pea so we can map it correctly and send to the appropriate one when asked. I think there is no other way to find out for the k8s connection pool otherwise. Same goes for the jina_pod_name and the shard_id

jina/peapods/networking.py
jacobowitz: In the current implementation I do the polling on the connection pool level. Of course we could in theory also do this in the runtime, but I think its fits better here? Especially if want to have eventually things like polling ROLLING or so and go over replicas etc. The runtime may have not much knowledge about those

jina/types/message/__init__.py
JoanFM: Understood. I would do it in another PR however. I am not sure if in the future for monitoring may be useful.

But that is far from a discussion here

jina/peapods/networking.py
JoanFM: Ok

jina/types/message/__init__.py
jacobowitz: yes it doesnt really belong here, but the issue is that the `routes` are already removed from the envelope (its still in the request so we can send it back to the client from the gateway if so desired).
Being removed from the envelope proto means that constructing a Message fails at the moment, which prevents any test etc. Thats why I fixed it here. I can also do it in a separate PR, merge that one and rebase this, but thats quite a detour and I think will happen all the time now with this big refactoring?

PR #3877, opened 2021-11-05
feat: star routing graph gateway

JoanFM:
**Changes added**

Create a `TopologyGraph` class that would be built by the Pod topology, and that would serve as a ComputationalGraph from the Gateway.

It manages the async tasks to be handled by the `gateway` at each request by a client. 

The design principle is that every node in the graph gets the task from their predecessors, and in a push fashion all the tasks are awaited in order.

**TODO**
This is quite ready to be integrated with the new ConnectionPool and it would need to be added to the new Gateway.

- [x] Check better for `hanging pods`


JoanFM:
https://github.com/jina-ai/internal-tasks/issues/249

jina/peapods/runtimes/gateway/graph/topology_graph.py
```diff
@@ -0,0 +1,135 @@
...
+                self.parts_to_send.append(msg)
+                # this is a specific needs
+                if len(self.parts_to_send) == self.number_of_parts:
+                    resp = await connection_pool.send(self.parts_to_send[-1], self.name)
+                    return resp
+
+        def get_leaf_tasks(
+            self,
+            connection_pool,
+            msg_to_send: Optional[str],
```
jacobowitz, line 44: That should be a message proto and not a string?

jina/peapods/runtimes/gateway/graph/topology_graph.py
```diff
@@ -0,0 +1,135 @@
...
+            if previous_task is not None:
+                msg = await previous_task
+            if msg is not None:
+                self.parts_to_send.append(msg)
+                # this is a specific needs
+                if len(self.parts_to_send) == self.number_of_parts:
+                    resp = await connection_pool.send(self.parts_to_send[-1], self.name)
+                    return resp
+
+        def get_leaf_tasks(
```
jacobowitz, line 41: in the graph context the name of the function is of course correct, but may be a bit much detail for a caller? A caller might just do 'send the message to this node in the graph and get me back the task'? Not sure the caller is interested in the fact about leaf tasks etc? But no strong opinion on this

jina/peapods/runtimes/gateway/graph/topology_graph.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    :class TopologyGraph is a class that describes a computational graph of nodes, where each node represents
+        a Pod that needs to be sent messages in the order respecting the path traversal.
+
+    :param graph_description: A dictionary describing the topology of the Pods. 2 special nodes are expected, the name `start-gateway` and `end-gateway` to
+        determine the nodes that receive the very first message and the ones whose response needs to be sent back to the client. All the nodes with no outgoing nodes
+        will be considered to be hanging, and they will be "flagged" so that the user can ignore their tasks and not await them.
+    """
+
+    class _ReqReplyNode:
```
jacobowitz, line 17: The class is part of the interface of the graph as its returned by `origin_nodes`, should it really be private then?

jacobowitz (changes requested):
Should we maybe add a convenience function in the Flow to create the topology?

jina/peapods/runtimes/gateway/graph/topology_graph.py
JoanFM: yes, this was just prototyping. Will change when merging clarified

jina/peapods/runtimes/gateway/graph/topology_graph.py
JoanFM: Yes, it may be true. I just thought since implementation is complex to show what it really does. 

U can propose names

jina/peapods/runtimes/gateway/graph/topology_graph.py
JoanFM: Good point

JoanFM:
> Should we maybe add a convenience function in the Flow to create the topology?

We can try to do this when things start getting integrated.

PR #3888, opened 2021-11-08
perf: match sparse

davidbp:
The `.embeddings` property is not leveraged when using `da1.match(da2)` because it was not adapted to handle sparse arrays.


hanxiao:
this PR needs a complete rework based on the spirit of #3889

davidbp:
I see, this does not make sense anymore, will create a PR for adding scipy inside unravel

PR #3889, opened 2021-11-08
refactor(types): new implementation on ndarray types

hanxiao:
# New implementation of `NdArray`

The new `NdArray` gives unified interface to sparse/dense `.embedding`, `.blob` and unified unravel access to `.embeddings`, `.blobs`. The new `NdArray` stores `type(array)` as well as its side info in Protobuf, and recover them into the original type on demand. As a side-effect of this new design, the efficiency also gets slightly better (observed from `jina hello fashion`)

```python
a = [numpy.ndarray(...),
        torch.tensor(...),
        tf.constant(...), 
        paddle.to_tensor(...),
        torch.tensor(...).to_sparse(),
        tf.sparse.from_dense(...),
        csr_matrix(...),
        bsr_matrix(...),
        coo_matrix(...),
        csc_matrix(...)]

d = Document()
for aa in a:
  d.embedding = aa

print(d.embedding)
```

## Refactoring
- remove all old `jina.types.ndarray`, they are unnecessary
- remove all unnecessary fields in pb

## API Change
- remove `.dict(prettifier)` and `.json(prettifier)` for two reasons, (1) `.embedding/.blob` now returns sparse/dense/tf/torch/scipy array, not just ndarray, the old heuristic does not work. (2) `.dict` and `.json` is in general for serialization, aiming for later recover. Adding prettify logic into it is not a good idea. If insists, have a new `prettify` function.
- remove quantization inside Ndarray proto. Quantization should be done outside the storage, and should be handled by users via Torch/TF/numpy easily. The previous design was unnecessary. 
- remove `get_sparse_embedding`, `get_sparse_blob` as there is no need for that, `.embedding` and `.blob` will directly return sparse if it was stored as sparse.

## Future work

The team can improve `unravel` method (the support faster access `.blobs/.embeddings`) on scipy and Tensorflow, right now the `unravel` only supports PyTorch sparse.


jina/types/ndarray/__init__.py
```diff
@@ -19,34 +33,244 @@ class BaseNdArray(ProtoTypeMixin):
...
+
+    @staticmethod
+    def unravel(protos: Sequence[jina_pb2.NdArrayProto]) -> 'ArrayType':
+        """Unravel many ndarray-like proto in one-shot, by following the shape
+        and dtype of the first proto.
+
+        :param protos: a list of ndarray protos
+        :return: a framework ndarray
+        """
+        first = NdArray(next(iter(protos)))
```
JoanFM, line 84: why not simply access 0? no need to create new object

hanxiao: because it wont work on DAM, as written here https://github.com/jina-ai/jina/blob/master/jina/types/arrays/mixins/content.py#L25

hanxiao: Note that DAM's `self._pb_body` is https://github.com/jina-ai/jina/blob/dd752ff919a9278c8d275710420c00fd551904c4/jina/types/arrays/memmap/__init__.py#L530-L533

PR #3891, opened 2021-11-09
feat(proto): enable sending multiple messages in a list

jacobowitz:
This enables sending mulitple partial requests in a single grpc call by introducing the `MessageListProto` wrapper and a fairly hacky serializer implementation to hide it again from the user.

I think this should work (I adapted the worker runtime and checked the test), but is not ideal:

- We actually only need multiple requests and not multiple messages, this adds unnecessary overhead
- There is too much message/messageProto conversion back and forth with this implementation

I created a follow up issue to implement this better, but this will be significant more work: https://github.com/jina-ai/internal-tasks/issues/259

### DIFF
--- a/jina/peapods/runtimes/worker/__init__.py
+++ b/jina/peapods/runtimes/worker/__init__.py
@@ -4,7 +4,7 @@ import multiprocessing
 import threading
 import time
 from abc import ABC
-from typing import Optional, Union
+from typing import Optional, Union, List
 
 import grpc
 from grpc import RpcError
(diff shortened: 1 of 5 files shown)


jina/peapods/runtimes/worker/__init__.py
```diff
@@ -113,26 +113,28 @@ def wait_for_ready_or_shutdown(
...
-            WorkerRuntime.cancel(self.is_cancel)
-        except (RuntimeError, Exception) as ex:
-            self.logger.error(
-                f'{ex!r}' + f'\n add "--quiet-error" to suppress the exception details'
-                if not self.args.quiet_error
-                else '',
-                exc_info=not self.args.quiet_error,
-            )
-            raise
+        for msg in messages:
```
JoanFM, line 124: what u want is to merge the messages right?

jacobowitz: yes, but not in this pr. I just wanted to add the proto changes here and see that the runtime still works. It doesnt add support for merging in the runtime

PR #3892, opened 2021-11-09
docs: remove compound executor and graph document

Roshanjossey:
Only from toc


hanxiao:
windows was already moved to install
image

docs/advanced/experimental/index.md
```diff
@@ -17,6 +17,5 @@ That said, have fun!
 
 kubernetes
 indexers
-compound-executor
-graph-document
+windows
```
hanxiao, line 20: no need

hanxiao (changes requested):
keep the md, only show remove  `compound-executor, graph-document` in TOC would be enough

Roshanjossey:
Updated to address review comments

PR #3893, opened 2021-11-09
feat: star routing gateway

JoanFM:


jina/peapods/stream/gateway.py
```diff
@@ -1,20 +1,86 @@
...
+            )
+            if len(filtered_partial_responses) > 1:
+                docs = DocumentArray(
+                    [
+                        d
+                        for r in filtered_partial_responses
+                        for d in getattr(r.request, 'docs')
+                    ]
+                )
+                groundtruths = DocumentArray(
```
jacobowitz, line 68: we didnt auto merge the groundtruths before, did we? (only in cases where we had some executor I think?)

jina/peapods/stream/client.py
```diff
@@ -73,6 +80,40 @@ async def _receive(self):
                     future.cancel()
                 self.request_buffer.clear()
 
+    def _handle_request(self, request: 'Request') -> 'asyncio.Future':
+        """
+        For zmq & grpc data requests from gateway, for each request in the iterator, we send the `Message`
```
jacobowitz, line 85: there is no zmq anymore

jacobowitz (changes requested):
Looks good! 
One improvement could be made in the BaseStreamer I think: We dont need the result_queue/callback mechanism anymore, because the sending Future from the new Graph will hold the result and can be used to retrieve it and yielding it to the client

jina/peapods/stream/client.py
JoanFM: I will review docstrings later

jina/peapods/stream/gateway.py
JoanFM: yes, which seems a little bit dangerous, We can revisit this later

PR #3895, opened 2021-11-09
docs: add configurations to hide sidebar and nav when loaded in iframe

Roshanjossey:
I'll merge this once I test embedding this in learning portal

### DIFF
--- a/docs/_static/iframe.js
+++ b/docs/_static/iframe.js
@@ -1,7 +1,8 @@
 function inIframe() {
     try {
       return window.self !== window.top;
-    } catch (Exception) {
+    } catch (err) {
+      console.log("Iframe error: ", err)
       return true;
     }
   }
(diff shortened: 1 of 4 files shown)


Roshanjossey:
Tested in https://iframe.vercel.app/?src=https://docs-configure-embed-version--jina-docs.netlify.app&width=1400&height=1000

Looks good

docs/_static/iframe.js
```diff
@@ -0,0 +1,11 @@
+function inIframe() {
+    try {
+      return window.self !== window.top;
+    } catch (Exception) {
```
nomagick, line 4: I think in Javascript we `catch (err)`.
It can't match by class.

Roshanjossey: updated

mapleeit (approved):

nomagick (approved):

PR #3897, opened 2021-11-09
feat: star routing gateway all protocols

JoanFM:


tests/unit/peapods/runtimes/container/mwu-encoder/mwu_encoder.py
```diff
@@ -19,6 +19,8 @@ def close(self) -> None:
         import pickle
 
         os.makedirs(self.workspace, exist_ok=True)
+        print(f' \n\n\n\n\n\nworkspace {self.workspace}')
```
JoanFM, line 22:
```suggestion
```

PR #3898, opened 2021-11-09
feat: add scipy sparse support to unravel

davidbp:
Allow DocumentArray to work with scipy sparse matrices.


hanxiao:
from its title, this PR seems unnecessary as it is handled by #3898

davidbp:
Correct, you already did it.

PR #3899, opened 2021-11-09
feat(head): new head runtime for star routing

jacobowitz:
This PR adds a new runtime for driving the head in Star Routing. This closes https://github.com/jina-ai/internal-tasks/issues/242

I also upgrade our protobuf version in this PR, because I wanted to use the 'new' `optional` feature.


jina/parsers/peapods/runtimes/head.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def mixin_head_parser(parser):
+    """Mixing in arguments required by head peas and runtimes into the given parser.
+    :param parser: the parser instance to which we add arguments
+    """
+
+    gp = add_arg_group(parser, title='Head')
+
+    gp.add_argument(
```
JoanFM, line 11: can u prepare the args to have multiple addresses? so that even if we add it later we have less breaking changes?

jina/proto/jina.proto
```diff
@@ -285,6 +285,16 @@ message DocumentArrayProto {
     repeated DocumentProto docs = 1; // a list of Documents
 }
 
+/**
+ * Represents an entity (like a Pea)
```
JoanFM, line 289:
```suggestion
 * Represents an entity (like an ExecuorRuntime) 
```
Let's prepare for the separation of concerns, proto does not know about Peas and pods

jina/proto/jina.proto
```diff
@@ -285,6 +285,16 @@ message DocumentArrayProto {
...
+ */
+message RelatedEntity {
+    string id = 1; // unique id of the entity, like the name of a pea
+    string address = 2; // address of the entity, could be an IP address
```
JoanFM, line 293: both port and address seem redundant?

jina/parsers/peapods/runtimes/head.py
jacobowitz: why multiple addresses? `Uses_before`/`Uses_after` should be always single instances I think? At least with the current design that was what I intended. If you need to scale them, you need to scale the head.

jina/proto/jina.proto
jacobowitz: address should be IP, domain etc - maybe I should it make more clear in the comment. This does not include the port

PR #3905, opened 2021-11-10
feat(worker): merge docs in worker runtime

jacobowitz:
This closes https://github.com/jina-ai/internal-tasks/issues/264

One can see the disadvantage of the multi message approach here. The DataRequestHandler gets multiple messages and we arbitrarily manipulate the first message and return it.
Ideally this should be multiple requests and we always return the only existing message (to preserve meta information)

### DIFF
--- a/tests/unit/peapods/runtimes/worker/test_worker_runtime.py
+++ b/tests/unit/peapods/runtimes/worker/test_worker_runtime.py
@@ -3,7 +3,7 @@ import multiprocessing
 import os
 import time
 from multiprocessing import Process
-from threading import Event, Thread
+from threading import Event
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


tests/unit/peapods/runtimes/worker/test_worker_runtime.py
```diff
@@ -43,9 +39,45 @@ def start_runtime(args, handle_mock, cancel_event):
...
+    assert WorkerRuntime.wait_for_ready_or_shutdown(
+        timeout=5.0, ctrl_address=f'{args.host}:{args.port_in}', shutdown_event=Event()
+    )
+
+    response = GrpcConnectionPool.send_messages_sync(
+        [_create_test_data_message(), _create_test_data_message()],
+        f'{args.host}:{args.port_in}',
     )
-    assert handle_mock.is_set()
+    assert response
```
JoanFM, line 78: let's assert after the terminate and join so that we make sure all ends well

JoanFM (changes requested):

PR #3907, opened 2021-11-10
feat(k8s): add scale to k8s pod

bwanglzu:
part of #3816 as an independent PR.


bwanglzu:
closing since this will be part of #3816

PR #3910, opened 2021-11-11
test: test runtimes fast slow request

JoanFM:


jacobowitz (approved):
good that this still works as expected :)

PR #3918, opened 2021-11-11
feat(pea): adapt peas for star routing

jacobowitz:
This adapts the pea for star routing and closes https://github.com/jina-ai/internal-tasks/issues/272

I removed some of the more tidious mocking tests in the pea tests and added three new ones:
* Check that all gateway runtimes start correctly in a pea
* Check that all non gateway runtimes start correctly in a pea
* Check that a failing Executor is handled correctly

I also noticed that I broke a lot of the new tests with the previous ZMQ purging PR and fixed those tests again.
I also remove the unnecessary SCALE from protobuf model.

I also assume in the typing of the runtime_cls that its a AsyncNewLoopruntime, which is not correct due to jinad/container, but we want to kill it soon so I hope its okay.


cli/autocomplete.py
```diff
@@ -56,22 +56,15 @@
...
+            '--timeout-ctrl',
             '--uses',
             '--uses-with',
             '--uses-metas',
             '--uses-requests',
             '--py-modules',
             '--port-in',
-            '--port-out',
-            '--hosts-in-connect',
             '--host-in',
```
JoanFM, line 67: is host in relevant? should it only be called host?

jina/peapods/peas/__init__.py
```diff
@@ -194,9 +195,13 @@ def terminate(self):
         This method calls :meth:`terminate` in :class:`threading.Thread` or :class:`multiprocesssing.Process`.
         """
         if hasattr(self.worker, 'terminate'):
-            self.logger.debug(f' terminating the runtime process')
+            self.logger.debug(f'terminating the runtime process')
             self.worker.terminate()
             self.logger.debug(f' runtime process properly terminated')
+        else:
+            self.logger.debug(f'canceling the runtime thread')
+            self.runtime_cls.cancel(cancel_event=self.cancel_event)
```
JoanFM, line 203: now all the runtimes should be equivalent and simply set cancel event  and thats end of the story? or is it still good to have it in the runtime?

JoanFM (changes requested):
Can we add tests also to see what happens if GatewayRuntime or HeadRuntime fail to start?

cli/autocomplete.py
jacobowitz: yes I think it should be named host, port-in should also be port instead I think. But lets tidy this up in the end?

jina/peapods/peas/__init__.py
jacobowitz: true, setting the cancel event directly would also work. I have no strong opinion here, but letting the runtime decide that looks a bit cleaner to me?

JoanFM: yes, it can be considered cleaner, but maybe it is more readable if done here

jina/peapods/peas/__init__.py
```diff
@@ -323,30 +304,18 @@ async def async_wait_start_success(self):
...
         self.logger.debug('waiting for ready or shutdown signal from runtime')
-        terminated = False
-        if self.is_ready.is_set() and not self.is_shutdown.is_set():
+        if not self.is_shutdown.is_set():
             try:
-                self.logger.debug(f' Cancel runtime')
-                self._cancel_runtime()
                 self.logger.debug(f' Wait to shutdown')
+                self.terminate()
+                time.sleep(0.1)
```
JoanFM, line 317: I wonder if this sleep is necessary anymore

jina/peapods/runtimes/worker/__init__.py
```diff
@@ -83,7 +83,7 @@ def is_ready(ctrl_address: str, **kwargs) -> bool:
 
         try:
             GrpcConnectionPool.send_message_sync(ControlMessage('STATUS'), ctrl_address)
-        except RpcError:
+        except RpcError as e:
```
JoanFM, line 86: revert or log

jina/peapods/peas/__init__.py
jacobowitz: sure, changed it

jacobowitz:
added the new test and addressed the other comments as well

PR #3922, opened 2021-11-12
feat: star routing container pea

JoanFM:


jina/peapods/peas/factory.py
```diff
@@ -0,0 +1,37 @@
...
+from . import BasePea, Pea
+from .container import ContainerPea
+
+from ...hubble.helper import is_valid_huburi
+from ...hubble.hubio import HubIO
+
+
+class PeaFactory:
+    """
+    A PeaFactory is a factory class, abstracting the PeaFactory creation
```
jacobowitz, line 13:
```suggestion
    A PeaFactory is a factory class, abstracting the Pea creation
```

jina/peapods/peas/factory.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    @staticmethod
+    def build_pea(
+        args: 'Namespace',
+    ) -> BasePea:
+        """Build an implementation of a `BasePod` interface
```
jacobowitz, line 20:
```suggestion
        """Build an implementation of a `BasePea` interface
```

jina/peapods/runtimes/asyncio.py
```diff
@@ -125,22 +131,48 @@ def activate(**kwargs):
...
+        Check if status is ready.
+
+        :param ctrl_address: the address where the control message needs to be sent
+        :param kwargs: extra keyword arguments
+
+        :return: True if status is ready else False.
+        """
+
+        try:
+            GrpcConnectionPool.send_message_sync(ControlMessage('STATUS'), ctrl_address)
```
jacobowitz, line 146: I am not sure that really belongs here? the grpc server is created by the child class implementations.

jina/peapods/runtimes/gateway/grpc/__init__.py
```diff
@@ -40,6 +40,7 @@ async def async_setup(self):
         jina_pb2_grpc.add_JinaRPCServicer_to_server(self.streamer, self.server)
         bind_addr = f'{__default_host__}:{self.args.port_expose}'
         self.server.add_insecure_port(bind_addr)
+        self.logger.debug(f' Start server bound at {bind_addr}')
```
jacobowitz, line 43:
```suggestion
        self.logger.debug(f' Start server bound to {bind_addr}')
```

jacobowitz (changes requested):
Some smaller comments, otherwise looks good.
I was just wondering if we should add a test trying to create ContainerPeas from within a container? Or is this a challenge we leave for jinad?

jina/peapods/runtimes/asyncio.py
JoanFM: But it is the only way ContainerPea can avoid knowing runtime_cls. I agree this should be done as discussed via the container itself, but not possible now

JoanFM:
> Some smaller comments, otherwise looks good. I was just wondering if we should add a test trying to create ContainerPeas from within a container? Or is this a challenge we leave for jinad?

This should be a challenge for JinaD yes I'd say.

jacobowitz (approved):
:rocket:

PR #3923, opened 2021-11-12
feat(tests): integration tests for peas

jacobowitz:
This closes https://github.com/jina-ai/internal-tasks/issues/265

This adds integration tests for peas (basically the same as for runtimes, just with peas).
I am also changing the PeaRoleType enum. I think the new definition is clearer.

### DIFF
--- a/jina/peapods/peas/__init__.py
+++ b/jina/peapods/peas/__init__.py
@@ -360,11 +360,3 @@ class BasePea:
 
         .. #noqa: DAR201"""
         return self.args.pea_role
-
-    @property
-    def _is_inner_pea(self) -> bool:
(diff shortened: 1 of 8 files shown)


jina/peapods/peas/__init__.py
```diff
@@ -363,8 +363,8 @@ def role(self) -> 'PeaRoleType':
 
     @property
     def _is_inner_pea(self) -> bool:
```
JoanFM, line 365: We can remove this I think, this is only used in a test

jacobowitz: sure, removed and changed in the test (which is not working anyway of course atm)

PR #3940, opened 2021-11-16
feat: star routing pods

jacobowitz:
This PR adapts the Pod level for Star Routing and closes https://github.com/jina-ai/internal-tasks/issues/273. It also adapts the pod unit tests for this.
CompoundPod is removed as a concept and its functionality is merged into the regular Pod.

I also changed CI to only run the tests we are expecting to work. This change should be reverted once we refactor the broken tests, but now its handy to make sure we dont break tests supposed to work.

What is not part of this PR:
- [Adapting K8s pods](https://github.com/jina-ai/internal-tasks/issues/273)
- Adapt the mermaid string. Ideally this would be done here, but its a bit hard to debug without the Flow, so I would rather want to do this as part of the Flow adaption.

what will not be done here: k8s, plotting


JoanFM (changes requested):

tests/unit/peapods/runtimes/gateway/grpc/test_grpc_gateway_runtime.py
```diff
@@ -134,222 +135,252 @@ async def task_wrapper():
...
+                responses = []
+                req = request_generator(
+                    '/', DocumentArray([Document(text='client0-Request')])
+                )
+                async for resp in runtime.streamer.Call(request_iterator=req):
+                    responses.append(resp)
+                return responses
+
+            responses = asyncio.run(_test())
+        assert len(responses) > 0
```
JoanFM, line 169: These tests need to be redefined. U need to put the `responses` in some `queue` ans assert the response from the main pytest process

JoanFM (changes requested):

tests/unit/peapods/pods/test_pods.py
```diff
@@ -192,24 +193,30 @@ def foo(self, docs: DocumentArray, **kwargs):
...
+            assert 2 == len(response_texts)
+            assert all(text in response_texts for text in ['10', 'client'])
+
+            await pod.rolling_update(uses_with={'param': 20})
+            response_texts = await _send_requests(pod)
+            assert 2 == len(response_texts)
+            assert all(text in response_texts for text in ['20', 'client'])
+            assert '10' not in response_texts
+
+        p = Process(target=run_async_test)
```
JoanFM, line 216: isn't it weird to call a function defined in a different context? I do not understand

jina/peapods/peas/helper.py
```diff
@@ -92,6 +93,7 @@ def update_runtime_cls(args, copy=False) -> 'Namespace':
         _args.runtime_cls not in gateway_runtime_dict.values()
         and _args.host != __default_host__
         and not _args.disable_remote
+        and not ip_address(_args.host).is_loopback
```
JoanFM, line 96: what is this?

jina/peapods/pods/__init__.py
```diff
@@ -333,15 +255,14 @@ class Pod(BasePod):
     """
 
     class _ReplicaSet:
-        def __init__(self, pod_args: Namespace, args: List[Namespace]):
+        def __init__(
+            self, pod_args: Namespace, args: List[Namespace], head_args: Namespace
+        ):
             self.pod_args = copy.copy(pod_args)
+            self.head_args = head_args
```
JoanFM, line 262: if a `ReplicaSet` needs to represent a set of Peas that can be also Heads, then I think `args` and `head_args` should be the same.

JoanFM: or they should be Optional

jina/peapods/pods/__init__.py
```diff
@@ -509,21 +445,17 @@ def update_pea_args(self):
...
     def first_pea_args(self) -> Namespace:
-        """Return the first non-head/tail pea's args
+        """Return the first worker pea's args
 
 
         .. # noqa: DAR201
         """
         # note this will be never out of boundary
-        return self.peas_args['peas'][0]
+        return self.peas_args['peas'][0][0]
```
JoanFM, line 458: why does it have depth of 2, this dict should contain `peas`, `head`, `uses_before`, `uses_after`?

jina/peapods/pods/__init__.py
```diff
@@ -84,7 +83,7 @@ def _fix_exception_context(new_exc, old_exc):
 
 
 class BasePod(ExitFIFO):
```
JoanFM, line 85: I think we can safely use `ExitStack` since the order does not matter anymore? I would put the `workers` at the top of the stack however

jina/peapods/pods/__init__.py
JoanFM: Ah is because u have `shards`

JoanFM: Is this property needed? should we try to remove?

jina/peapods/peas/helper.py
jacobowitz: This is checking if the host parameter is actually localhost (aka on the same machiene), in this case we should not use jinad as we only need it for remote entities.

jina/peapods/pods/__init__.py
jacobowitz: you mean `self.peas_args`? This is used everywhere, in particular to store the shard/replica information. At the moment we need it, we could refactor like moving to a distinct class or so, but we will need sth like this.

jina/peapods/pods/__init__.py
jacobowitz: `ReplicaSet` can not  have heads, the head args are passed here because we need the address for activating/deactivating

tests/unit/peapods/pods/test_pods.py
jacobowitz: I can slightly improve the structure here, but I need to run the test code (I think at least the request sending) in a different process, otherwise we have multiple grpc clients running concurrently in the same process which messes up things

tests/unit/peapods/runtimes/gateway/grpc/test_grpc_gateway_runtime.py
jacobowitz: I agree that design is pretty hacky now, but it works as I am checking the exit code and any test failure will yield a non zero exit code. But it will hide the actual problem in the test.

tests/unit/peapods/peas/container/test_container_pea.py
```diff
@@ -89,6 +89,11 @@ def test_failing_executor(fail_start_docker_image_built):
     assert pea._container.id not in containers
 
 
+@pytest.mark.slow
+@pytest.mark.skipif(
+    'GITHUB_WORKFLOW' in os.environ,
+    reason='for some reason this fails persistently in CI',
+)
 def test_pass_arbitrary_kwargs(monkeypatch, mocker):
```
JoanFM, line 97: I changed this test in a recent PR

jina/peapods/pods/__init__.py
```diff
@@ -757,140 +632,164 @@ async def rolling_update(
...
+                tasks.append(
+                    asyncio.create_task(
+                        self.shards[shard_id].rolling_update(
+                            dump_path=dump_path, uses_with=uses_with
+                        )
+                    )
+                )
+
+            for future in asyncio.as_completed(tasks):
+                _ = await future
```
deepankarm, line 650:
```suggestion
            await asyncio.gather(tasks):
```

jina/peapods/pods/__init__.py
```diff
@@ -757,140 +632,164 @@ async def rolling_update(
...
+        try:
+            import asyncio
+
+            for shard_id in self.shards:
+                tasks.append(
+                    asyncio.create_task(self.shards[shard_id].scale(replicas=replicas))
+                )
+
+            for future in asyncio.as_completed(tasks):
+                _ = await future
```
deepankarm, line 674: same

jina/peapods/pods/__init__.py
```diff
@@ -757,140 +632,164 @@ async def rolling_update(
...
-                head_args=parsed_args.get('head', None),
-                tail_args=parsed_args.get('tail', None),
-            )
+    @staticmethod
+    def _set_uses_before_after_args(args: Namespace, type: str) -> Namespace:
+
+        _args = copy.deepcopy(args)
+        _args.pea_role = PeaRoleType.WORKER
+        _args.identity = random_identity()
+        _args.host = '127.0.0.1'
```
deepankarm, line 731:
```suggestion
        _args.host = __default_host__
```

jina/peapods/pods/__init__.py
```diff
@@ -757,140 +632,164 @@ async def rolling_update(
...
-            if getattr(args, 'uses_after', None):
-                self.is_tail_router = True
-                parsed_args['tail'] = self._copy_to_tail_args(args, args.polling)
-            parsed_args['peas'] = self._set_peas_args(
-                args,
-                head_args=parsed_args.get('head', None),
-                tail_args=parsed_args.get('tail', None),
-            )
+    @staticmethod
+    def _set_uses_before_after_args(args: Namespace, type: str) -> Namespace:
```
deepankarm, line 726: Better to avoid `type` as it is a python keyword

jina/peapods/peas/helper.py
jacobowitz: yes, but the runtime is actually not removed here, right? I mean not properly at least. So my code will do weird stuff without this check?

jacobowitz: ah actually I will remove the check completely, you are right

jacobowitz:
rebased

PR #3943, opened 2021-11-16
fix(evaluate): fix length and hash check on evaluate mixin

hanxiao:


jina/types/arrays/mixins/evaluation.py
```diff
@@ -48,6 +50,13 @@ def evaluate(
         metric_name = metric_name or metric_fn.__name__
         results = []
         for d, gd in zip(self, other):
+            if hash_fn(d) != hash_fn(gd):
```
JoanFM, line 53: I am not sure this is a wanted feature? why would I care about this if I just only want to add matches?

hanxiao: because this feature requires left&right to be fully aligned: on the length, and on the semantic of length. Otherwise, why would evaluation on completely irrelevant matches make sense? just because they have the same length?

it's like saying two ndarrays are same if their shapes are the same.

JoanFM (changes requested):
I would remove the hash check at the root level, I do not see any added value to this check, only potential annoyance to user

hanxiao:
> I would remove the hash check at the root level, I do not see any added value to this check, only potential annoyance to user

reject, I argue the other way

jina/types/arrays/mixins/evaluation.py
JoanFM: imagine I have my matches with id under tags[id], why the root document and the groundtruth also need this? There may not even exist this concept for the root documents

hanxiao: > imagine I have my matches with id under tags[id],

whatever the user store the id is irrelevant, docs are considered as identical if they hash to the same via `hash_fn`. In this way this argument is not valid as storing `tags[id]` is no differ than storing in `.id`. `hash_fn` requires the a "semantic identical identifier" to be applied on *ALL* documents, regardless their *nested* structures, which is a pretty valid requirement. 

Imagine the powerfulness of the `traverse_flat` and our Document structure that allows you to build nested structures in an arbitrary way. I would be super-surprised that a "semantic identical identifier" can be applied only to a certain level but not the other level, applied only to root but not matches, applied only to matches but not chunks. applied only to matches of matches but not matches. Inconsistency knows no scope, this can go on and on. 

Consistency is universal.

Issue #3947, opened 2021-11-17
traversal paths are checked twice when using traverse_flat and will become buggy after deprecation of list of string syntax 


JoanFM:
**Describe the bug**
When the list of string syntax of traversal_paths becomes deprecated, traverse_flat will not work because _check_traversal_path_type is called twice (the second time always receives the old syntax).
Currently, even if traverse_flat receives the new syntax, it will always through a DeprecationWarning

PR #3951, opened 2021-11-18
fix: check traversal_paths once

alaeddine-13:
closes https://github.com/jina-ai/jina/issues/3947

### DIFF
--- a/jina/types/arrays/mixins/traverse.py
+++ b/jina/types/arrays/mixins/traverse.py
@@ -18,16 +18,13 @@ if TYPE_CHECKING:
 def _check_traversal_path_type(tp):
     is_deprecate_scheme = True
     if isinstance(tp, str):
-        tp = tp.split(',')
         is_deprecate_scheme = False
-
-    if not (
-        tp
-        and not isinstance(tp, str)
-        and isinstance(tp, Sequence)
-        and all(isinstance(p, str) for p in tp)
-    ):
-        raise TypeError('`traversal_paths` needs to be `Sequence[str]`')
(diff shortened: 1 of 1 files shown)


jina/types/arrays/mixins/traverse.py
```diff
@@ -48,13 +48,15 @@ class TraverseMixin:
     def traverse(
         self: 'T',
         traversal_paths: str,
+        check_traversal_paths: bool = True,
```
JoanFM, line 51: Why is this needed? I do not like adding this to the public interface, this should not be cared by the user

JoanFM (changes requested):

jina/types/arrays/mixins/traverse.py
```diff
@@ -18,16 +18,13 @@
...
         is_deprecate_scheme = False
-
-    if not (
-        tp
-        and not isinstance(tp, str)
-        and isinstance(tp, Sequence)
-        and all(isinstance(p, str) for p in tp)
-    ):
-        raise TypeError('`traversal_paths` needs to be `Sequence[str]`')
+    elif isinstance(tp, Sequence) and all(isinstance(p, str) for p in tp):
```
JoanFM, line 22: no need to check `is_deprecate_warning`. Just raise the `warning` here and that's it

JoanFM (changes requested):
Looks good, just a minor thing

jina/types/arrays/mixins/traverse.py
alaeddine-13: changed

hanxiao (approved):

Issue #3954, opened 2021-11-18
reduce overhead coming from set_attributes


alaeddine-13:
set_attributes is dynamic and slower than directly setting each attribute.
We should rely less on it and probably optimize it

hanxiao:
we should make it private

PR #3955, opened 2021-11-18
perf: use less set_attributes

alaeddine-13:
closes https://github.com/jina-ai/jina/issues/3954


JoanFM (commented):
Seems good, if we find tests failing, and we find them legit, we can check if `kwargs` exist and set attributes only if present

JoanFM (approved):

bwanglzu:
seems we break some tests, even hello world examples.

JoanFM (approved):

PR #3956, opened 2021-11-18
perf: small improve on uuid generation in doc

bwanglzu:
test settings: 1 million documents generation:

```python
from jina import Document

[Document() for _ in range(1000000)]
```

feature branch: 21239777 function calls (21233402 primitive calls) in 16.524 seconds

master branch: 23239839 function calls (23233463 primitive calls) in 19.815 seconds

we saved 2million function calls and roughly 17% time saving.

old & new uuid looks a bit different: 

D0364F55-4D62-490A-AB1D-9749A233AC7A

### DIFF
--- a/jina/types/document/__init__.py
+++ b/jina/types/document/__init__.py
@@ -1,4 +1,5 @@
 import json
+import uuid
 import mimetypes
 from typing import (
     Dict,
@@ -23,7 +24,7 @@ from ..score import NamedScore
 from ..score.map import NamedScoreMapping
 from ..struct import StructView
 from ...excepts import BadDocType
-from ...helper import random_identity, typename
+from ...helper import typename
 from ...proto import jina_pb2
 
 if TYPE_CHECKING:
@@ -265,7 +266,7 @@ class Document(AllMixins, ProtoTypeMixin):
             ) from ex
 
         if self._pb_body.id is None or not self._pb_body.id:
-            self.id = random_identity(use_uuid1=True)
+            self.id = uuid.uuid1().hex
 
         if kwargs:
             # check if there are mutually exclusive content fields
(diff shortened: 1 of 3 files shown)


davidbp:
Just note that the output will be formatted  differently. From `'707a71ee-4883-11ec-a415-787b8ab3f5de'` to '707a71ee488311eca415787b8ab3f5de'. To me it still makes sense.

jina/types/document/__init__.py
```diff
@@ -265,7 +266,7 @@ def _update_doc(d: Dict):
             ) from ex
 
         if self._pb_body.id is None or not self._pb_body.id:
-            self.id = random_identity(use_uuid1=True)
```
JoanFM, line 268: please remove this function from the `codebase` if it nos used

JoanFM (changes requested):

bwanglzu:
we `str` conversion, it is:
```
21239914 function calls (21233528 primitive calls) in 17.544 seconds
```

tadejsv:
Can you also test by using `uuid4`? This is actually the one used now

hanxiao:
nice catch!

but then i would suggest refactoring `random_identity` to use `.hex`, instead of making this exception for Document constructor.

JoanFM:
> nice catch!
> 
> but then i would suggest refactoring `random_identity` to use `.hex`, instead of making this exception for Document constructor.

I would do that, but at the same time keep the `.hex` here, saving the function call and the `if else` can be good

bwanglzu:
okay now I break a lot of distributed test

JoanFM:
> okay now I break a lot of distributed test

no pain, no gain

bwanglzu:
> > okay now I break a lot of distributed test
> 
> no pain, no gain

:sob:

JoanFM (approved):

PR #3960, opened 2021-11-19
refactor: refactor streamer result handling

JoanFM:
**Changes introduced**

- [x] Has a single `RequestStreamer` instead of multiple implementations.
- [x] Streamer is responsible on handling the requests and responses and how they are awaited, cancelled, etc... but not about the handling of those. This is given by the caller. 
- [x] Gateway logic remains in the `gateway` runtime `submodule`.

**TODO**

- [x] Add unittests to streamer


jacobowitz (approved):
to me it looks much cleaner now, but I think @deepankarm should have the final say here probably

PR #3964, opened 2021-11-19
feat(k8s): Adapt K8s Pod for StarRouting

jacobowitz:
This PR is part of https://github.com/jina-ai/internal-tasks/issues/286
Its adapting the K8sPod for the StarRouting changes. This enables the Head Deployment in K8s to also contain uses_before/uses_after container if necessary.
The unit test is adapter and added to CI. There is also an integration test, but that one will not be added due to lack of k8s support in the star routing ci.


jacobowitz:
incorporated https://github.com/jina-ai/jina/pull/3936

jina/peapods/networking.py
```diff
@@ -388,11 +396,13 @@ async def task_wrapper(new_messages, stub):
                 except AioRpcError as e:
                     if e.code() != grpc.StatusCode.UNAVAILABLE or i == 2:
                         raise
+                    elif e.code() == grpc.StatusCode.UNAVAILABLE and i == 2:
```
JoanFM, line 399: this code is not reachable right? `i==2` will always activate the first `if`

jina/peapods/runtimes/head/__init__.py
```diff
@@ -37,7 +37,12 @@ def __init__(
         """
         super().__init__(args, cancel_event, **kwargs)
 
+        if args.name is None:
+            args.name = ''
         self.name = args.name
+        self.pod_name = (
```
JoanFM, line 43: isn't there a better way to handle this? What problem is it handling?

JoanFM (changes requested):

jina/peapods/networking.py
jacobowitz: True, will fix it. I wanted to print the log message here

jina/peapods/runtimes/head/__init__.py
jacobowitz: I don't like it either but the problem is that we need to know the pod name in a k8s setting for the pool. The pool observes all changes in a k8s namespace, no matter it's position in the Flow. Then the head runtime needs to be able to send a message to a specific pod  which is identified by name. 
To make it less guess work, I could pass it down explicitly and not derive if from the name of the head?

JoanFM: I would be more explicit yes

jacobowitz: I changed it now so that I use the env variable JINA_POD_NAME instead. This one is always set in K8s (unless the user provides their own template and removes it).
In a none K8s settings this value acually does not matter and it can be whatever static string, I default to worker in this case. But we happen to set the env variable in this case to sth else, it also will work.

jacobowitz:
merged feat-star-routing into these to solve the conflicts in the test script with the new tests being added

Issue #3617, opened 2021-10-08
Resource temporarily unavailable: '/home/kavan/.jina/disk_cache.db' when you run flow in thread 


Kavan72:
When i try to run flow in multiple threads then i'm getting this error

```
        encoder@13543[E]:Error while pulling jinahub://TransformerSentenceEncoder: 
FileNotFoundError(2, 'No such file or directory')
        encoder@13543[E]:Error while pulling jinahub://TransformerSentenceEncoder: 
FileNotFoundError(2, 'No such file or directory')
        encoder@13543[E]:Error while pulling jinahub://TransformerSentenceEncoder: 
FileNotFoundError(2, 'No such file or directory')
        encoder@13543[E]:Error while pulling jinahub://TransformerSentenceEncoder: 
FileNotFoundError(2, 'No such file or directory')
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/media/kavan/WORK/Job/jina_rnd/venv/lib/python3.8/site-packages/jina/hubble/helper.py", line 285, in wrapper
    cache_db = shelve.open(
  File "/usr/lib/python3.8/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.8/shelve.py", line 227, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.8/dbm/__init__.py", line 95, in open
    return mod.open(file, flag, mode)
_gdbm.error: [Errno 11] Resource temporarily unavailable: '/home/kavan/.jina/disk_cache.db'
```

During handling of the above exception, another exception occurred:

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/media/kavan/WORK/Job/jina_rnd/main.py", line 13, in encode
    with self.flow:
  File "/media/kavan/WORK/Job/jina_rnd/venv/lib/python3.8/site-packages/jina/flow/base.py", line 1097, in __enter__
    return self.start()
  File "/media/kavan/WORK/Job/jina_rnd/venv/lib/python3.8/site-packages/jina/flow/base.py", line 1141, in start
    self.enter_context(v)
  File "/usr/lib/python3.8/contextlib.py", line 425, in enter_context
[...]

PR #3966, opened 2021-11-20
fix: hubio disk cache for multithreading

Kavan72:
adding filelock for hubio's `disk_cache.db` to resolve issue #3617


JoanFM (approved):
LGTM! Thanks for the contribution

PR #3982, opened 2021-11-23
refactor(daemon): use context to set envs

bwanglzu:
use `context` to pass `envs` to YAML configuration, tests adjusted to new style envs setting in YAML.


bwanglzu:
seems i break a lot of other tests

tests/daemon/unit/api/flow_with_env.yml
```diff
@@ -0,0 +1,14 @@
+jtype: Flow
+with:
+  name: my-flow
+  grpc_data_requests: true
+  port_expose: 12345
+  protocol: http
+  env:
+    key1: ${{ context_var1 }}  # this should be replaced
+    key2: $context_var2  # this should not be replaced
+    key3: ${ context_var3 }  # this should not be replaced
```
deepankarm, line 10: User's Flow yaml might include the first 2 syntaxes. Is there any reason, the 2nd one shouldn't be replaced?

bwanglzu: it is not supported by `load_config` while testing. maybe I can allow 2nd case happen

deepankarm: But the 2nd syntax is available via an environment variable?

bwanglzu: i created a minimum example:

```python
from jina.jaml import JAML
from pprint import pprint

with open('toy.yml')as fp:
    a = JAML.load(fp)
    b = JAML.expand_dict(a, context={'context_var1': 'MMM', 'context_var2': 'WWW'})
pprint(b)
```

with YAML:
```yaml
jtype: Flow
with:
  name: my-flow
  grpc_data_requests: true
  port_expose: 12345
  protocol: http
  random_key: ${{ context_var1 }}
  random_key2: $context_var2
executors:
  - name: local_replicas
    port_in: 45678
    replicas: 2

```

bwanglzu: essentially, flow.load_config is calling this `expand_dict` where we pass a `context`

deepankarm (approved):

PR #3983, opened 2021-11-23
fix: filelock dependency not required

JoanFM:


jina/hubble/helper.py
```diff
@@ -287,10 +282,18 @@ def wrapper(*args, **kwargs):
...
+                help_text=f'FileLock is needed to guarantee non-concurrent access to the'
+                f'cache_file {cache_file}',
+            ):
+                import filelock
+
+                file_lock = filelock.FileLock(f'{cache_file}.lock', timeout=-1)
 
             cache_db = None
-            with filelock.FileLock(f'{cache_file}.lock', timeout=-1):
+            with file_lock:
```
cristianmtr, line 296: Should we even allow a nullcontext? Maybe make required=True?

JoanFM: required True means an exception will be raised and failed.

cristianmtr: Yes, that was my point. Shouldn't make this required? What happens if you don't have the actual package and just have a nullcontext?

JoanFM: in a very rare case where Flows are run in multiprocess or multithreaded way

PR #4002, opened 2021-11-26
docs: refactor indexers page

cristianmtr:
Some clean-up / updates of indexers page in docs, before writing the benchmark section

PR #4003, opened 2021-11-26
docs: indexers benchmark

cristianmtr:
TODO

- [x] find right place in docs. Maybe on same page as Indexers. Then wait for #4002 to be merged
- [x] include table with results
- [x] include plots. As `iframe`? Can we include html files somehow?


hanxiao:
let me restructure the TOC a bit

hanxiao:
nested
image

davidbp (commented):
When you talk about 1k query vectors it is not clear if this is 'done as a batch' (one function call that processes 1k query vector) or if query vectors are processes one at a time. If I were a user probably I would like to know

- What's the expected time  between `.search` and getting a result for a single query?
- Whats the expected throughput of queries I can expect using `X` resources in `delta` time.

docs/advanced/experimental/indexers-benchmark.md
```diff
@@ -0,0 +1,53 @@
...
+In this page we will show the performance capabilities of the different indexers available in Jina Hub.
+The purpose is to compare how they each scale, both with the amount of Documents indexed and with the embedding dimensionality.
+We have restricted our selection to the following
+
+- [SimpleIndexer](https://hub.jina.ai/executor/zb38xlt4)
+- [FaissSearcher](https://hub.jina.ai/executor/gilkzt3f)
+- [HNSWSearcher](https://hub.jina.ai/executor/jdb3vkgo)
+
+`SimpleIndexer` is based on doing exhaustive search across all the Documents.
+However, Both FAISS and HNSW are based on the approximate nearest neighbor approach.
```
maximilianwerk, line 12:
```suggestion
However, both FAISS and HNSW are based on the approximate nearest neighbor approach. 
```

maximilianwerk (commented):
I like it so far. Obviously the results data is needed.

In the final version, I'd put the results at the top and put the methodology behind. Most views will be for the results and thus they should be at the top and easiest accessible.

cristianmtr:
@maximilianwerk moved results to top. 

@hanxiao @davidbp added results now too. Can you check again?

cristianmtr:
Not sure why the deployment doesn't include the subpage. Locally it is there

davidbp (dismissed):
I see now the benchmarks.

cristianmtr:
> I see now the benchmarks.

In the netlify deployment? I still don't. Tried a hard refresh but nothing.

maximilianwerk:
> > I see now the benchmarks.
> 
> In the netlify deployment? I still don't. Tried a hard refresh but nothing.

me neither. weird.

maximilianwerk (commented):
I find it confusing, that the x-axis has varying scales in all the plots. This makes it much harder to get fast conclusions. Having a changing y-axis is OK. Opinions?

Hippopotamus0308:
> > I find it confusing, that the x-axis has varying scales in all the plots. This makes it much harder to get fast conclusions. Having a changing y-axis is OK. Opinions?
> 
> I'd need to recreate them and the code from @Hippopotamus0308 is not yet in the benchmarks repo. We can either wait and recreate them with fixed axes, or merge it as it is now.

@cristianmtr I've added to the pr in https://github.com/jina-ai/benchmark_indexers/pull/33, please check it.

cristianmtr:
@maximilianwerk 

I tried to render them with fixed x_range but they are not all readable

ex.

image
image

maximilianwerk:
> @maximilianwerk
> 
> I tried to render them with fixed x_range but they are not all readable
> 
> ex.

Ok then don't do that and keep it as they are :)

docs/advanced/experimental/indexers-benchmark.md
```diff
@@ -0,0 +1,60 @@
...
+The tables are sorted by `time_search`.
+
+```{include} bench/gist.md
+```
+
+```{include} bench/sift.md
+```
+
+```{include} bench/sift1bil.md
+```
```
maximilianwerk, line 19: In the tables inside, precision and recall is always identical.

maximilianwerk: Ah the reason is, because they are both: from the top 100 and in the top 100. :D This looks a bit weird.

maximilianwerk: This might be the reason, they don't care about precision in the [ann benchmark](http://ann-benchmarks.com/)

cristianmtr: Yeah, we should remove it then. The formula seems correct me to in the [code](https://github.com/jina-ai/benchmark_indexers/blob/57276a44267e193127b65954b07fa6f529e0bd23/src/evaluator.py#L20)

maximilianwerk (commented):
apart from this and the precision/recall, I am good with merging.

docs/advanced/experimental/indexers-benchmark.md
maximilianwerk: If you take 100 correct Documents and check how many of them are in the top 100, precision and recall are the same. Because it is both `number of correct documents / 100`.

maximilianwerk (approved):
LGTM

PR #4006, opened 2021-11-29
feat: add reducer module

alaeddine-13:
closes: https://github.com/jina-ai/internal-tasks/issues/300


hanxiao:
by reading the code, im guessing it also works on recursive structure, tho no test covers that

jina/types/arrays/mixins/reduce.py
```diff
@@ -0,0 +1,47 @@
...
+class ReduceMixin:
+    """A mixing that provides reducing logic for :class:`DocumentArray` or :class:`DocumentArrayMemmap`"""
+
+    def reduce(self: 'T', da: 'T'):
+        """
+        Reduces da and the current DocumentArray into one DocumentArray in-place. Changes are applied to the current
+        DocumentArray
+        :param da: DocumentArray
+        """
+        for doc in da:
```
JoanFM, line 19: I am missing the sorting

jina/types/arrays/mixins/reduce.py
```diff
@@ -0,0 +1,47 @@
...
+if TYPE_CHECKING:
+    from ..document import DocumentArray
+    from ...document import Document
+    from ....helper import T
+
+
+class ReduceMixin:
+    """A mixing that provides reducing logic for :class:`DocumentArray` or :class:`DocumentArrayMemmap`"""
+
+    def reduce(self: 'T', da: 'T'):
```
hanxiao, line 13:
```suggestion
    def reduce(self: 'T', da: 'T') -> 'T':
```
 and `return self`

jina/types/arrays/mixins/reduce.py
```diff
@@ -0,0 +1,47 @@
...
+        :param doc1: first Document
+        :param doc2: second Document
+        """
+        if len(doc2.matches) > 0:
+            doc1.matches.reduce(doc2.matches)
+
+        if len(doc2.chunks) > 0:
+            doc1.chunks.reduce(doc2.chunks)
+
+    def reduce_mat(self: 'T', da_matrix: List['T']) -> 'T':
```
hanxiao, line 38:
```suggestion
    def reduce_all(self: 'T', others: List['T']) -> 'T':
```

jina/types/arrays/mixins/reduce.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        Reduces doc1 and doc2 into one Document in-place. Changes are applied to doc1
+        :param doc1: first Document
+        :param doc2: second Document
+        """
+        if len(doc2.matches) > 0:
+            doc1.matches.reduce(doc2.matches)
+
+        if len(doc2.chunks) > 0:
+            doc1.chunks.reduce(doc2.chunks)
```
hanxiao, line 36: so we still dont have a conclusion on how to reduce on other properties right? what about following the logic of `MergeFrom`: https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.MergeFrom

> Singular fields that are set in the specified message overwrite the corresponding fields in the current message. Repeated fields are appended. Singular sub-messages and groups are recursively merged.

alaeddine-13: I suggest the following logic: all fields that are set except matches and chunks overwrite the corresponding fields in the current document.
Content attributes (text, blob, ...) are ignored and only `content` is considered

alaeddine-13: I applied the logic from MergeFrom using the method `update`
However, after some thoughts, I think it's better to always keep the attributes that are set on the left-most document.
This will avoid redundantly setting the attributes each time we reduce and will save time
@JoanFM your thoughts ?

alaeddine-13: reduction on property attributes is applied with priority to the left-most document arrays

jina/types/document/mixins/attribute.py
```diff
@@ -139,13 +139,15 @@ def update(
         self: T,
         source: T,
         fields: Optional[List[str]] = None,
+        exclude_fields: Optional[List[str]] = None,
```
jacobowitz, line 142: Do you use this anywhere in this PR?

jacobowitz (changes requested):
PR looks good! :) Can you maybe describe the reduce logic somewhere? I mean the business logic that Docs different to each other will be concatenated and Docs with same id are reduced in a way that the first Document will keep its properties if they are set in both?

I think the description can go either in the Mixin class Pydoc or in the function documentations?

jina/types/document/mixins/attribute.py
alaeddine-13: ah yeah I added it but then I no more need it. Just removed it from the PR

tests/unit/types/arrays/mixins/test_reduce.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+    da1.reduce_all([da2, da3])
+    assert da1[0].text == 'doc1'
+    assert da1[0].id == 'r'
+
+    # chunks and matches merged, not overridden
+    assert da1[0].chunks[0].id == 'c1'
+    assert da1[0].chunks[1].id == 'c2'
+    assert da1[0].matches[0].id == 'm3'
+    assert da1[0].tags == {'a': 'b'}
```
cristianmtr, line 209: Can you add an integration test with shards?

cristianmtr (changes requested):

jina/types/document/mixins/attribute.py
alaeddine-13: > PR looks good! :) Can you maybe describe the reduce logic somewhere? I mean the business logic that Docs different to each other will be concatenated and Docs with same id are reduced in a way that the first Document will keep its properties if they are set in both?
> 
> I think the description can go either in the Mixin class Pydoc or in the function documentations?

I added a description to the Mixin docstring.
Do you reduce should also be documented in https://docs.jina.ai/fundamentals/document/documentarray-api/ ?
I'm thinking we just need to document it [here](https://docs.jina.ai/fundamentals/flow/topology/) at the end of [this PR](https://github.com/jina-ai/jina/pull/4030) because we don't expect the user to use reduce by himself, rather, should know how it is applied in the Flow by default

tests/unit/types/arrays/mixins/test_reduce.py
jacobowitz: an integration test would be great, but I think it should rather go in here https://github.com/jina-ai/jina/issues/4015

This PR is just adding the Mixin, so unit testing this code is fine for me. Once we actually integrate it, lets do more sophisticated testing

alaeddine-13: Integrating in the flow is in-progress in this [PR](https://github.com/jina-ai/jina/pull/4030/)
I added there an integration test with shards, can you check it there ?

tests/unit/types/document/test_document.py
```diff
@@ -651,6 +651,16 @@ def test_doc_update_given_fields_and_source_has_more_attributes(test_docs):
     assert doc1.chunks != doc2.chunks
 
 
+def test_doc_update_exclude_fields(test_docs):
```
jacobowitz, line 654: this needs to go now I think

jacobowitz (changes requested):

tests/unit/types/document/test_document.py
alaeddine-13: done

jacobowitz (approved):
Cool feature! This will be also helpful later in `feat-star-routing` :+1:

cristianmtr (approved):

PR #4027, opened 2021-12-03
feat(executor): to async or not to async

deepankarm:
```python
import time
import asyncio
from jina import Executor, Flow, Document, DocumentArray, requests


class Ex(Executor):
    @requests(on='/foo')
    def foo(self, docs: DocumentArray, **kwargs):
        print(f'{docs[0].text} reached foo. time.sleep for 1sec')
        time.sleep(1)
        print(f'done sleeping in foo {docs[0].text}')

    @requests(on='/async_foo')
    async def async_foo(self, docs: DocumentArray, **kwargs):
        print(f'{docs[0].text} reached async_foo. await asyncio.sleep for 1sec')
        await asyncio.sleep(1)
        print(f'done sleeping in async_foo {docs[0].text}')

e = Ex()
e('foo')

f = Flow().add(uses=Ex)
```

------

### normal def

```python

with f:
    with TimeContext('foo:  '):
        f.post(
            on='/foo',
            inputs=DocumentArray(Document(text=f'id-{i}') for i in range(5)),
            request_size=1,
        )

```
```
         Flow@299125[I]:🎉 Flow is ready to use!
        🔗 Protocol:            GRPC
        🏠 Local access:        0.0.0.0:60171
        🔒 Private network:     192.168.1.28:60171
foo:   ...      id-0 reached foo. time.sleep for 1sec
done sleeping in foo id-0
id-1 reached foo. time.sleep for 1sec
done sleeping in foo id-1
id-2 reached foo. time.sleep for 1sec
done sleeping in foo id-2
id-3 reached foo. time.sleep for 1sec
done sleeping in foo id-3
id-4 reached foo. time.sleep for 1sec
done sleeping in foo id-4
foo:   takes 5 seconds (5.03s)
```

### async def

```python
with f:
    with TimeContext('async_foo:  '):
        f.post(
            on='/async_foo',
            inputs=DocumentArray(Document(text=f'id-{i}') for i in range(5)),
            request_size=1,
        )

```

```
           Flow@299459[I]:🎉 Flow is ready to use!                                                  
        🔗 Protocol:            GRPC
        🏠 Local access:        0.0.0.0:34223
        🔒 Private network:     192.168.1.28:34223
async_foo:   ...        executor0/rep-0@299487[E]:TypeError('The return type must be DocumentArray / DocumentArrayMemmap / Dict / `None`, but getting <coroutine object Ex.async_foo at 0x7f9891766200>')
 add "--quiet-error" to suppress the exception details
  File "/home/deepankar/repos/new/jina/jina/peapods/runtimes/worker/__init__.py", line 84, in Call
    return self._handle(messages)
  File "/home/deepankar/repos/new/jina/jina/peapods/runtimes/worker/__init__.py", line 115, in _handle
    return self._data_request_handler.handle(messages=messages)
  File "/home/deepankar/repos/new/jina/jina/peapods/runtimes/request_handlers/data_request_handler.py", line 173, in handle
    f'The return type must be DocumentArray / DocumentArrayMemmap / Dict / `None`, '
TypeError: The return type must be DocumentArray / DocumentArrayMemmap / Dict / `None`, but getting <coroutine object Ex.async_foo at 0x7f9891766200>
RuntimeWarning: coroutine 'Ex.async_foo' was never awaited (raised from /home/deepankar/repos/new/jina/jina/peapods/runtimes/worker/__init__.py:96)
```

### normal def:

```python


with f:
    with TimeContext('foo: '):
        f.post(
            on='/foo',
            inputs=DocumentArray(Document(text=f'id-{i}') for i in range(5)),
            request_size=1,
        )
```

```
           Flow@298500[I]:🎉 Flow is ready to use!                                                  
        🔗 Protocol:            GRPC
        🏠 Local access:        0.0.0.0:41329
        🔒 Private network:     192.168.1.28:41329
foo:   ...      id-0 reached foo. time.sleep for 1sec
id-1 reached foo. time.sleep for 1sec
id-2 reached foo. time.sleep for 1sec
id-3 reached foo. time.sleep for 1sec
id-4 reached foo. time.sleep for 1sec
done sleeping in foo id-0
done sleeping in foo id-1
done sleeping in foo id-2
done sleeping in foo id-3
done sleeping in foo id-4
foo:   takes 1 second (1.02s)
```

### async def:

```python
with f:
    with TimeContext('async_foo: '):
        f.post(
            on='/async_foo',
            inputs=DocumentArray(Document(text=f'id-{i}') for i in range(5)),
            request_size=1,
        )
```

```
           Flow@298500[I]:🎉 Flow is ready to use!                                                  
        🔗 Protocol:            GRPC
        🏠 Local access:        0.0.0.0:57209
        🔒 Private network:     192.168.1.28:57209
async_foo:   ...        id-0 reached async_foo. await asyncio.sleep for 1sec
id-1 reached async_foo. await asyncio.sleep for 1sec
id-2 reached async_foo. await asyncio.sleep for 1sec
id-3 reached async_foo. await asyncio.sleep for 1sec
id-4 reached async_foo. await asyncio.sleep for 1sec
done sleeping in async_foo id-0
done sleeping in async_foo id-1
done sleeping in async_foo id-2
done sleeping in async_foo id-3
done sleeping in async_foo id-4
async_foo:   takes 1 second (1.02s)

```


JoanFM:
@deepankarm the example u put in the documentations should not be way faster with async io?

jina/executors/__init__.py
```diff
@@ -198,6 +198,24 @@ def __call__(self, req_endpoint: str, **kwargs):
...
+            return await self.__acall_endpoint__(req_endpoint, **kwargs)
+        elif __default_endpoint__ in self.requests:
+            return await self.__acall_endpoint__(__default_endpoint__, **kwargs)
+
+    async def __acall_endpoint__(self, req_endpoint, **kwargs):
+        func = self.requests[req_endpoint]
+        if iscoroutinefunction(func):
+            return await (await func(self, **kwargs))  # 2 awaits ugly
+        else:
+            return await run_in_threadpool(func, self, **kwargs)
```
JoanFM, line 217: why is this needed, shouldnt it simply fail? Running in thread is undesired

JoanFM: Here we just simply do asyncio.run()

JoanFM (changes requested):
I think it should be async run, but why is the speedup not felt?

jina/executors/__init__.py
deepankarm: The event loop is already running inside the Executor. So we cannot do `asyncio.run` or `loop.run_until_complete`. Anyway, doing that will block the event loop. The non async defs are `run_in_executor` to keep the event loop unblocked.

deepankarm: > why is the speedup not felt?

?? Before the PR, sending 5 requests with used to take 5 secs. Now it takes one (both normal def and async def). The event loops remains unblocked throughout.

deepankarm:
> @deepankarm the example u put in the documentations should not be way faster with async io?

Yes it is. From 5secs to 1sec for 5 requests.

jina/executors/__init__.py
JoanFM: ok not understood. But in the non async def it takes 1 sec because u run it on a thread right?

JoanFM:
> > @deepankarm the example u put in the documentations should not be way faster with async io?
> 
> Yes it is. From 5secs to 1sec for 5 requests.

Did not understand, but I think the call to the non-async should block the io loop

JoanFM:
The non-async calls should block the io loop, otherwise they are running in different threads. We have in mind to do multiprocessing at Runtime level,  but multithreading I do not think it should be there

deepankarm:
Managed in https://github.com/jina-ai/jina/pull/4102

Issue #4015, opened 2021-12-02
integrate reduce logic in the runtimes


alaeddine-13:
use DocumentArray.reduce_mat in order to reduce the DocumentArray matrix into one when shards are used

PR #4030, opened 2021-12-06
feat: integrate reduce runtimes

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/4015


JoanFM:
@alaeddine-13 please do not merge the addition in the types with the executor part. They shouuld be independent commits, because the second one will not be relevant for `star-routing`

JoanFM (changes requested):
Please, first merge the `types` part with no `Executor` involved

tests/integration/reduce_shards/test_reduce.py
```diff
@@ -0,0 +1,76 @@
...
+                assert f'c-{shard}-{chunk}' in chunks
+
+
+@pytest.mark.parametrize('n_shards', [3, 5])
+@pytest.mark.parametrize('n_docs', [3, 5])
+def test_uses_after_no_reduce(n_shards, n_docs):
+    search_flow = Flow().add(
+        uses=MyExecutor,
+        shards=n_shards,
+        uses_after='Executor',
```
cristianmtr, line 60: How can the results be the same with uses_after='Executor' as they are with the Reducer Executor?

cristianmtr (changes requested):

tests/integration/reduce_shards/test_reduce.py
alaeddine-13: It shouldn't be, that's a WIP code that I didn't finish. I'll fix this and properly assert that reduce wasn't executed

alaeddine-13:
> @alaeddine-13 please do not merge the addition in the types with the executor part. They shouuld be independent commits, because the second one will not be relevant for `star-routing`

should be okay now

jina/peapods/runtimes/request_handlers/data_request_handler.py
```diff
@@ -22,7 +23,9 @@ def _get_docs_matrix_from_message(
     field: str,
 ) -> List['DocumentArray']:
     if partial_request is not None:
-        result = [getattr(r, field) for r in reversed(partial_request)]
+        result = [(r.pod_name, getattr(r, field)) for r in reversed(partial_request)]
+        result.sort(key=itemgetter(0))
```
JoanFM, line 27: why is this needed?

jina/proto/jina.proto
```diff
@@ -314,6 +314,7 @@ message RequestProto {
     message DataRequestProto {
         repeated DocumentProto docs = 1; // a list of Documents to query
         repeated DocumentProto groundtruths = 2; // a list of groundtruth Document you want to evaluate it with
+        string pod_name = 3; // sender pod name
```
JoanFM, line 317: do not add this here, why is this needed?

JoanFM (changes requested):
Do not change protobuf specifciation for this

jina/peapods/runtimes/request_handlers/data_request_handler.py
alaeddine-13: This is important to have deterministic results after reducing
It turns out that the document arrays coming from different shards do not have a deterministic order: the one that completes first will be returned first.
As reduce's result (and also match merger) depend on the order of the DocumentArrays, there has to be a way to keep the order deterministic.
I thought about using the pod_name to do so. It's also easy to understand for the user

jina/proto/jina.proto
alaeddine-13: https://github.com/jina-ai/jina/pull/4030#discussion_r767439206

jina/peapods/runtimes/request_handlers/data_request_handler.py
JoanFM: different shards share pod name from the user point of view. This is a weak reason to alter the proto structure so heavily

alaeddine-13: Removed.
Results still depend on the order and are undeterministic. Should be solved in star routing

jina/proto/jina.proto
alaeddine-13: removed

JoanFM (changes requested):
We need to add tests that this work in K8s

jina/peapods/runtimes/request_handlers/data_request_handler.py
```diff
@@ -121,7 +128,7 @@ def handle(
             msg.envelope.header.exec_endpoint not in self._executor.requests
             and __default_endpoint__ not in self._executor.requests
         ):
-            self.logger.debug(
+            self.logger.info(
```
JoanFM, line 131:
```suggestion
            self.logger.debug(
```

JoanFM (changes requested):
Small thing

JoanFM (approved):

PR #4041, opened 2021-12-07
fix(feat): add inv norm to image sprite gen

makram93:
When using `plot_image_sprites` func it will be helpful if we also add `inverse_normalize` as param, which will inverse the normalisation process before plotting it as a sprite.

This can come handy when we preprocess the docs before feeding it to the model and then would like to plot sprites of the matches directly.

### DIFF
--- a/jina/types/arrays/mixins/plot.py
+++ b/jina/types/arrays/mixins/plot.py
@@ -5,7 +5,7 @@ import tempfile
 import threading
 import warnings
 from math import sqrt, ceil, floor
-from typing import Optional
+from typing import Optional, Tuple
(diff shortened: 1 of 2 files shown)


hanxiao:
Reject, as user can always do `set_image_blob_inv_normalization()` beforehand.

Note that this mindset - nesting an existing function to another function - is not appreciated, as there is no boundary for that. One can argue that `plot_image_sprite` should contain *every possible* preprocessing function because it is handy.

Similar discussion: https://jinaai.slack.com/archives/C018F60RBL5/p1636480462201600

PR #4051, opened 2021-12-09
feat(proto): remove message proto

jacobowitz:
This PR removes the Message proto object. This this PR we will only send Request protos around. The envelope was not providing much value anymore without ZMQ so it made sense to remove the Message. This is preperation to unify gatewat/executor interfaces.
I also separate DataRequest/ControlRequest on a proto level and for GRPC interfaces.

I disabled the latency tracking test, it appears to be broken with the latest docarray seperation on master, we probably have to rebase. I would suggest lets complete this PR and then rebase feat-star-routing against master and reenable the test.

There are two immediate follow up tasks for this PR necessary:

1. Replace the list of DataRequests for the GRPC interface with a streaming approach (https://github.com/jina-ai/internal-tasks/issues/329)
2. Add the hacks for serilization of requests again to avoid (de-)serilization if not needed (https://github.com/jina-ai/internal-tasks/issues/302)


jina/proto/jina.proto
```diff
@@ -281,68 +224,74 @@ message RelatedEntity {
...
-    google.protobuf.Struct parameters = 5; // extra kwargs that will be used in executor
+    HeaderProto header = 1; // header contains meta info defined by the user
 
-    repeated RouteProto routes = 6; // status info on every routes
+    google.protobuf.Struct parameters = 2; // extra kwargs that will be used in executor
 
-    StatusProto status = 7; // status info
+    repeated RouteProto routes = 3; // status info on every routes
 
+    repeated DocumentProto docs = 4; // the docs in this request
```
JoanFM, line 258: May we want to wrap `docs` and `groundtruths` together inside `data` as it used to?

jina/clients/base/__init__.py
```diff
@@ -61,7 +61,7 @@ def check_input(inputs: Optional['InputType'] = None, **kwargs) -> None:
             # it is a function
             inputs = inputs()
 
-        kwargs['data'] = inputs
+        kwargs['docs'] = inputs
```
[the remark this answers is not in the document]
jacobowitz: thats separate now

jina/proto/jina.proto
jacobowitz: hmm I have no strong opinion here, but my impression is that most people are rather interested in the docs and only a small minority in groundtruth.
Anyway I added a small hack for backwards compatibility. Trying to access `data` now will return a named tuple wrapping `docs` and `groundtruths` 
Its tested here https://github.com/jina-ai/jina/pull/4051/files#diff-a1b83d7177090e17771956ad8ec7a3c5cac4c9cc795a6e65e16a8d96b32c1f4fR81

JoanFM (approved):

PR #4065, opened 2021-12-16
fix(hub): disk_cache_offline to return chached mark

numb3r3:
Now, `fetch_meta` is called twice when `force=True`. This would pollute the count of pulls of each executor in Hubble system. 

In this PR, the decorator `disk_cache_offline` is revised to return a tuple consisting of *returned value* and *bool value* (indicate whether the value is get from cache).


mapleeit:
FYI. Previous implement will cause inaccurate (x2) statistics of Executor usage in some circumstances.

JoanFM (approved):

PR #4066, opened 2021-12-16
docs(CONTRIBUTING): add instructions to build docs locally

Roshanjossey:


mapleeit (approved):

alexcg1:
Can we put this in README.md under `docs/` and redirect users there? I don't want to bloat out `CONTRIBUTING.md` too much, otherwise folks will never be able to plow through the whole thing

Issue #4021, opened 2021-12-03
integrate reduce logic in StarRouting


JoanFM:

JoanFM:
Same as #4015 but in star routing branch to make sure is not forgotten

PR #4070, opened 2021-12-16
feat: add reduce to star routing

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/4021


jina/peapods/runtimes/head/__init__.py
```diff
@@ -235,7 +247,7 @@ async def _handle_data_request(
         elif len(worker_results) > 1:
             DataRequestHandler.replace_docs(
                 response_request,
-                docs=get_docs_from_request(worker_results, field='docs'),
+                docs=docs,
```
jacobowitz, line 250: Lets move the other code here? I mean the reduce logic inside the elif. Maybe even inside the DataRequestHandler?
If it is before the if we may do unnecessary work (ok actually you are checking correctly in your if, but the place is still a bit surprising to me ;))

jacobowitz (changes requested):
Now that I see the code I would say lets to the reducing inside the DataRequestHandler?

jina/peapods/runtimes/request_handlers/data_request_handler.py
```diff
@@ -112,6 +119,11 @@ def handle(self, requests: List['DataRequest']) -> DataRequest:
         :param requests: The messages to handle containing a DataRequest
         :returns: the processed message
         """
+        docs = get_docs_from_request(
```
jacobowitz, line 122: I think there is no need to move this code up, lets leave it where it was before? at line 133?

jina/peapods/runtimes/head/__init__.py
```diff
@@ -209,6 +211,12 @@ async def _handle_data_request(
                 requests, pod='uses_before'
             )
             requests = [response]
+        elif len(requests) > 1:
+            docs_matrix = get_docs_matrix_from_request(requests, field='docs')
```
jacobowitz, line 215: I would suggest to move this and the following line to the DataRequestHandler as a static function? Then there is less doc related logic here in this runtime and we dont have to copy those two lines again in line 244/245

jacobowitz (changes requested):
just some minor style comments, looks great otherwise!

jina/peapods/runtimes/request_handlers/data_request_handler.py
alaeddine-13: changed

jina/peapods/runtimes/head/__init__.py
alaeddine-13: I refactored this, can you re-check ?

jacobowitz (approved):
looks good :+1: 
Lets wait for CI and if its green we can merge

jina/types/request/data.py
```diff
@@ -22,14 +22,14 @@ class _DataContent:
         def __init__(self, content: 'jina_pb2.DataRequestProto.DataContentProto'):
             self._content = content
 
-        @property
+        @cached_property
```
JoanFM, line 25: why we need this?

JoanFM (changes requested):

jina/types/request/data.py
alaeddine-13: it's needed to make sure request objects and DocumentArrays extracted from the request point to the same protobuf object. This helps perform reduce in-place without having to copy back to the request object.
Actually we noticed that sometimes Documents extracted from the request object aren't views to the same protobuf so we wanted to enforce the pattern in all cases

JoanFM: then this will need to be reviewed with the new docArray

jacobowitz (approved):
Looks good! I checked the failing tests and they are unrelated to your change. Actually they surfaced a bug in k8s when the connection pool is disabled. I will tackle this separately

jacobowitz:
K8s follow up bug https://github.com/jina-ai/internal-tasks/issues/346

alaeddine-13:
I see that tests/k8s/test_k8s.py::test_flow_with_needs is related to this PR
Actually the test fails because default reduce is applied right before custom merging.

PR #4071, opened 2021-12-17
feat: retry url to blob

florian-hoenicke:
When mapping a large DocumentArray from URLs to Blobs it sometimes happens that the URL request fails.
This pr aims to let the user configure how often it should try fetching the URL and how long it should wait between the tries.


JoanFM:
Hey @florian-hoenicke ,

We decided to avoid adding this complexity and let the user take care of these edge cases.

PR #4075, opened 2021-12-17
feat(DocumentArray): get items by boolean array

davidbp:
DocumentArray does not currently support slicing with boolean arrays.
This  can be a bottleneck in standard python code, where "vectorized expressions" (that return a vector of booleans) are frequently used.
 
Example:

Consider the following `DocumentArray` and pandas `DataFrame`.

```python
import numpy as np
import pandas as pd

n = 10_000
np.random.seed(123)

values = np.random.random(n)
da = DocumentArray(Document(weight=w) for w in values)
df = pd.DataFrame({'weights':values})
```

Users that want to slice a `DocumentArray` curretnly use option `(I)` but the proposed approach is 12x faster.

```python
# (I) current version
%timeit DocumentArray([d for d in da if d.weight>0.04])
# 30.8 ms ± 1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# (II) proposed version
%timeit da[values>0.04]
# 2.38 ms ± 102 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Other frameworks, such as pandas, offer a similar API as this PR aims to achieve

```python
# (I) Similar API to this PR
df[df['weights']>0.04]

# (II) Similar API of what we currently have
pd.DataFrame({'weights':[value for value in df['weights'] if value>0.04]})
```


JoanFM:
I think is an interesting feature. Lets keep as draft for after the refactoring

davidbp:
In my humble opinion it could be discussed to drive this sort of API further. 

Instead of doing `da.get_attributes('weight')` we could simply do  `da['weight']` and get a numpy array with the weights in each document of `da`. A lot of people is familiar with pandas and numpy. So getting inspired by their API might lower the difficulty of new users using Document Array.

JoanFM:
> In my humble opinion it could be discussed to drive this sort of API further.
> 
> Instead of doing `da.get_attributes('weight')` we could simply do `da['weight']` and get a numpy array with the weights in each document of `da`. A lot of people is familiar with pandas and numpy. So getting inspired by their API might lower the difficulty of new users using Document Array.

may u open a ticket in our Internal tasks repo to discuss this. Seems like an interesting approach

JoanFM:
TO be open in new repo when open

PR #4083, opened 2021-12-21
docs: update python source code in remarks

davidbp:
Refactor/Fix the source code to be rendered properly in the docs


davidbp:
Source code of the remarks section is still broken

JoanFM (approved):

PR #4087, opened 2021-12-22
fix(k8s): fix static cluster address config

jacobowitz:
Fix https://github.com/jina-ai/internal-tasks/issues/346

There is a bug when a Flow is used in K8s without the K8sGrpcConnectionPool, but just with the static GrpcConnectionPool.
There are two things wrong:
1. The Gateway is configured to send to the worker deployments, not to the head deployment
2. The Head is not configuring its static worker addresses correctly, because the shard_id is interpreted as a string, but it should be an integer.

This was surfaced by a failing K8s test, this test should work now in this PR


jacobowitz:
k8s tests are passing, the other failures are unrelated (mostly port reuse)

PR #4091, opened 2021-12-23
feat(grpc): support connect grpc server via ssl tunnel

mapleeit:


jina/peapods/networking.py
```diff
@@ -451,6 +451,50 @@ async def task_wrapper(requests, stubs):
...
+        :param options: A list of options to pass to the grpc channel
+
+        :return: A async grpc channel
+        """
+        if options == None:
+            options = GrpcConnectionPool.get_default_grpc_options()
+
+        port = address.split(':')[-1]
+
+        if port == '443':
```
JoanFM, line 491: what is this?

mapleeit: Using SSL channel if the port is 443, which is dedicated port for HTTPS.

JoanFM: it should be an option and not a hack like this I would say

mapleeit: addressed

jina/peapods/networking.py
```diff
@@ -451,6 +451,39 @@ async def task_wrapper(requests, stubs):
...
+        """
+
+        secure_channel = grpc.secure_channel
+        insecure_channel = grpc.insecure_channel
+
+        if asyncio:
+            secure_channel = grpc.aio.secure_channel
+            insecure_channel = grpc.aio.insecure_channel
+
+        if options == None:
```
JoanFM, line 479: options is

jina/peapods/networking.py
```diff
@@ -562,18 +601,25 @@ def get_default_grpc_options():
         ]
 
     @staticmethod
-    async def send_request_async(request: Request, target: str, timeout=1.0) -> Request:
+    async def send_request_async(
+        request: Request,
+        target: str,
+        timeout=1.0,
```
JoanFM, line 607: please add proper typehints.

JoanFM (commented):
For the args u will need to add it to the corresponding parser. We may be better to take from ot

PR #4092, opened 2021-12-23
feat(grpc): support connect grpc server via ssl tunnel

mapleeit:
Replacement of #4091


jina/peapods/networking.py
```diff
@@ -451,6 +451,39 @@ async def task_wrapper(requests, stubs):
...
+        """
+
+        secure_channel = grpc.secure_channel
+        insecure_channel = grpc.insecure_channel
+
+        if asyncio:
+            secure_channel = grpc.aio.secure_channel
+            insecure_channel = grpc.aio.insecure_channel
+
+        if options == None:
```
JoanFM, line 479:
```suggestion
        if options is None:
```

jina/peapods/networking.py
```diff
@@ -562,18 +601,25 @@ def get_default_grpc_options():
         ]
 
     @staticmethod
-    async def send_request_async(request: Request, target: str, timeout=1.0) -> Request:
+    async def send_request_async(
+        request: Request,
+        target: str,
+        timeout=1.0,
```
JoanFM, line 607:
```suggestion
        timeout: float=1.0,
```

jina/peapods/networking.py
```diff
@@ -562,18 +601,25 @@ def get_default_grpc_options():
...
+        https=False,
```
JoanFM, line 608:
```suggestion
        https: bool =False,
```

JoanFM (changes requested):
This will require changes in parser and proper tests

jacobowitz (changes requested):
I would add tests, but I think otherwise it looks good.

This test only adds https support for the client side (the runtimes will continue to expose only insecure channels). But I think for the use case here that is good enough because TLS termination happens before. So for me this approach here is valid!

mapleeit:
Ok. I'll add the tests then.

mapleeit:
I added test for static connection pool methods. 

But it's not easy to add test for Client since we don't have a https Flow. I can keep it as it is originally if you want. @jacobowitz @JoanFM

urica12:
Great

jina/peapods/networking.py
```diff
@@ -451,6 +451,48 @@ async def task_wrapper(requests, stubs):
...
+            secure_channel = grpc.aio.secure_channel
+            insecure_channel = grpc.aio.insecure_channel
+
+        if options is None:
+            options = GrpcConnectionPool.get_default_grpc_options()
+
+        if https:
+            credentials = grpc.ssl_channel_credentials()
+
+            if root_certificates is not None:
```
jacobowitz, line 487: I think its okay to pass None as root_certificates to ssl_channel_credentials(), see https://grpc.github.io/grpc/python/_modules/grpc.html#ssl_channel_credentials

so I would simplify line 485 to 488 to the single line
credentials = grpc.ssl_channel_credentials(
                    root_certificates=root_certificates
                )
no first call without args and no if needed?

jacobowitz (changes requested):
Commit message linting failed, you need to rewrite some of your messages according to the commit style guide. Otherwise I only have one minor coding style suggestion, the rest looks good to me. Nice feature :+1:

mapleeit:
Comments addressed. @jacobowitz

jacobowitz (changes requested):
commit linting is failing (I think commit messages should not be capitalized after the colon, so no feat: Update ...).

Otherwise the PR looks good to me!

mapleeit:
Commit lint fixed

jacobowitz (approved):
neat addition :+1:

Issue #4090, opened 2021-12-22
jinahub+docker:// doesn't work inside flow in star routing


alaeddine-13:
installing jina from star-routing branch and running
```
from jina import DocumentArray
from docarray.document.generators import from_files

docs_array = DocumentArray(from_files('*.jpg'))
from jina import Flow

f = Flow().add(uses='jinahub+docker://ImageTorchEncoder', uses_with={'model_name': 'resnet50'})
with f:
    f.post('/index', docs_array)
    f.block()
```
fails

PR #4093, opened 2021-12-23
fix: fix Flow uses executor from jinahub+docker 

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/4090

### DIFF
--- a/tests/unit/peapods/peas/test_pea_factory.py
+++ b/tests/unit/peapods/peas/test_pea_factory.py
@@ -1,0 +1,18 @@
+from jina.peapods.peas.factory import PeaFactory
+
+from jina.hubble.hubio import HubIO
+from jina.parsers import set_pea_parser
+
+
+def test_container_pea(mocker, monkeypatch):
+    mock = mocker.Mock()
+
+    def _mock_pull(self):
+        return 'docker://jinahub/dummy_executor'
+
+    monkeypatch.setattr(HubIO, 'pull', _mock_pull)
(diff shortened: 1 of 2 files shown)


JoanFM (approved):

tests/unit/peapods/peas/test_pea_factory.py
```diff
@@ -0,0 +1,18 @@
+from jina.peapods.peas.factory import PeaFactory
+
+from jina.hubble.hubio import HubIO
+from jina.parsers import set_pea_parser
+
+
+def test_container_pea(mocker, monkeypatch):
+    mock = mocker.Mock()
+
+    def _mock_pull(self):
```
JoanFM, line 10: this was not the observed problem rigth? the problem was `jinahub+docker` ?

alaeddine-13: The problem was that we correctly pull the image name from jinahub but we don't set it in the pea args.
Therefore we can safely mock pull and check that 'jinahub+docker://DummyExecutor' becomes 'docker://jinahub/dummy_executor'

alaeddine-13: The problem was that we correctly pull the image name from jinahub but we don't set it in the pea args.
Therefore we can safely mock pull and check that 'jinahub+docker://DummyExecutor' becomes 'docker://jinahub/dummy_executor'

PR #4096, opened 2021-12-23
feat(daemon): daemon flow on star routing

bwanglzu:


tests/distributed/test_remote_flows/test_remote_flow.py
```diff
@@ -79,6 +80,7 @@ async def test_remote_jinad_flow_async(async_jinad_client, flow_envs):
     assert remote_flow_args['protocol'] == PROTOCOL
     assert await async_jinad_client.flows.delete(flow_id)
     assert await async_jinad_client.workspaces.delete(workspace_id)
+    time.sleep(0.5)
```
jacobowitz, line 83: await asyncio.sleep(0.5)
no? Otherwise you block the loop which is probably not what you want?

deepankarm:
```
    JinaDClient@5174[E]:connection to server failed: ClientConnectorError(ConnectionKey(host='0.0.0.0', port=8000, is_ssl=False, ssl=None, proxy=None, proxy_auth=None, proxy_headers_hash=None), ConnectionRefusedError(111, 'Connection refused'))
```
This means JinaD is not started before running actual tests. Please make sure the server is started for the correct test.

deepankarm (approved):

PR #4099, opened 2021-12-23
fix: dont reduce if uses exist

jacobowitz:
Before this change we are always reducing if there is no uses_before.
With this PR we also check for `uses` and dont reduce if a meaningful `uses` exists.
I've adjusted some tests, there might be some more tests I need to adjust for this changed design.


jacobowitz:
The failing flow test is duplicate port assignment

PR #4100, opened 2021-12-27
fix: star routing fixes

alaeddine-13:
A few fixes after reviewing the tutorials and docs
closes: https://github.com/jina-ai/internal-tasks/issues/342

### DIFF
--- a/jina/types/request/__init__.py
+++ b/jina/types/request/__init__.py
@@ -24,6 +24,8 @@ class Request(ProtoTypeMixin):
     """
 
     def __getattr__(self, name: str):
+        if name == 'non_empty_fields':
+            return self.non_empty_fields
         return getattr(self.proto, name)
 
     def add_exception(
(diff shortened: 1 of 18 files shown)


jina/types/request/__init__.py
```diff
@@ -24,6 +24,8 @@ class Request(ProtoTypeMixin):
     """
 
     def __getattr__(self, name: str):
+        if name == 'non_empty_fields':
```
JoanFM, line 27: can u add a test for this?

JoanFM (changes requested):

jina/types/request/__init__.py
alaeddine-13: added

alaeddine-13: non_empty_fields was completely removed from Request in https://github.com/jina-ai/jina/pull/4102

jacobowitz (approved):
nice docs improvements :+1:

PR #4102, opened 2021-12-27
feat(executor): enable async executors

jacobowitz:
Based on the work here https://github.com/jina-ai/jina/pull/4027
See more extensive description in the base PR

Closes https://github.com/jina-ai/internal-tasks/issues/357

The previous PR was extended by these items:

- [x] Add tests
- [x] return await (await func(self, **kwargs)) -> fixed
- [x] adjust docs
- [x] set max_workers to 1

### DIFF
--- a/docs/fundamentals/executor/executor-api.md
+++ b/docs/fundamentals/executor/executor-api.md
@@ -6,7 +6,7 @@
 - An `Executor` should subclass directly from `jina.Executor` class.
 - An `Executor` class is a bag of functions with shared state (via `self`); it can contain an arbitrary number of
   functions with arbitrary names.
-- Functions decorated by `@requests` will be invoked according to their `on=` endpoint.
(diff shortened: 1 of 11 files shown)


docs/fundamentals/executor/executor-api.md
```diff
@@ -216,17 +240,20 @@ processing.
...
-def foo(docs: DocumentArray,
+async def foo(docs: DocumentArray,
         parameters: Dict,
         docs_matrix: List[DocumentArray],
         groundtruths: Optional[DocumentArray],
         groundtruths_matrix: List[DocumentArray]) -> Optional[DocumentArray]:
     pass
 ```
 
+If the function is using `async` in its signature, it will be used as a coroutine and the regular `asyncio` functionality is available inside the function.
```
JoanFM, line 254: I would say that this means that it will not block the event loop?

JoanFM (approved):
I would still review a little the documentation, but I guess is something we will have to do soon in general

PR #4110, opened 2021-12-28
fix: Doc import and add test cases

makram93:
Fix the import for Document and add test cases for sliding_window


JoanFM (changes requested):
This will need to wait for January 15th in the new repo

PR #4111, opened 2021-12-28
feat: polling per endpoint

jacobowitz:
Closes https://github.com/jina-ai/internal-tasks/issues/358

This PR adds the possibility to configure polling also per endpoint. A user can specify e.g. for an endpoint `/custom_endpoint` the polling to be ANY/ALL, independently from the general polling of the Jina Pod.

This functionality is show cased in one of the new `Flow` tests:
```
Flow().add(
        uses=DynamicPollingExecutorDefaultNames,
        shards=2,
        polling=polling,
        endpoint_polling={'/custom': 'ALL', '/search': 'ANY'},
)
```

By default `/index` has Polling ANY and `/search` has Polling `ALL`. 

From a user perspective it might be a bit confusing to have both `polling` and `endpoint_polling` as an option. If we want to unify this, we could remove polling, but then we need to add some kind of wildcard (like `/*`) to the endpoint configuration to be able to configure a polling for all endpoints not explicitly listed. I think that may make sense, but I suggest to do this in a follow up issue if we want this.

### DIFF
--- a/docs/fundamentals/flow/topology.md
+++ b/docs/fundamentals/flow/topology.md
@@ -218,7 +218,11 @@ This is helpful in two situations:
 
 Then splitting the load across two or more machines yields better results.
 
-For Shards, you can define which shard (instance) will receive the request from its predecessor. This behaviour is called `polling`. By default `polling` is set to `ANY`, which means only one shard will receive a request. If `polling` is to `ALL` it means that all Shards will receive a request.
+For Shards, you can define which shard (instance) will receive the request from its predecessor. This behaviour is called `polling`. `ANY` means only one shard will receive a request and `ALL` means that all Shards will receive a request. Polling can be configured per endpoint (like `/index` and Executor).
+By default the following `polling` is applied:
+- `ANY` for endpoints at `/index`
+- `ALL` for endpoints at `/search`
+- `ANY` for all other endpoints
 
 When you shard your index, the request handling usually differs between index and search requests:
 
@@ -231,9 +235,14 @@ For searching, you probably need to send the search request to all Shards, becau
 ```python Usage
 from jina import Flow
 
-index_flow = Flow().add(name='ExecutorWithShards', shards=3, polling='any')
-search_flow = Flow().add(name='ExecutorWithShards', shards=3, polling='all')
+flow = Flow().add(name='ExecutorWithShards', shards=3, polling='any', endpoint_polling={'/custom': 'ALL', '/search': 'ANY'})
 ```
+
+The example above will result in a Flow having the Executor `ExecutorWithShards` with the following polling options configured
+- `/index` has polling `ANY` (the default value is not changed here)
+- `/search` has polling `ANY` as it is explicitly set (usually that should not be necessary)
+- `/custom` has polling `ALL`
+- all other endpoints will have polling `ANY`
 
 ### Merging search results
 
--- a/docs/fundamentals/flow/topology.md
+++ b/docs/fundamentals/flow/topology.md
@@ -218,7 +218,8 @@ This is helpful in two situations:
 
 Then splitting the load across two or more machines yields better results.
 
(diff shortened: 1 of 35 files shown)


jina/peapods/runtimes/head/__init__.py
```diff
@@ -150,6 +164,13 @@ async def process_data(self, requests: List[DataRequest], context) -> DataReques
             )
             raise
 
+    def _extract_endpoint(self, context):
+        metadata = context.invocation_metadata()
+        for key, value in metadata:
+            if key == 'endpoint':
+                return value
+        return None
```
alaeddine-13, line 172:
```suggestion
        return dict(context.invocation_metadata()).get('endpoint')
```

jina/peapods/runtimes/head/__init__.py
```diff
@@ -49,6 +49,19 @@ def __init__(
             k8s_namespace=args.k8s_namespace,
         )
 
+        self._endpoint_polling = {}
```
alaeddine-13, line 52: maybe we can bring this line up: `self.polling = args.polling if hasattr(args, 'polling') else PollingType.ANY`

tests/integration/gateway_clients/test_clients_gateways.py
```diff
@@ -81,7 +81,9 @@ def complete_graph_dict():
 
 
 class DummyNoDocAccessMockConnectionPool:
-    def send_requests_once(self, requests, pod: str, head: bool) -> asyncio.Task:
+    def send_requests_once(
+        self, requests, pod: str, head: bool, endpoint: str = None
```
alaeddine-13, line 85:
```suggestion
        self, requests, pod: str, head: bool, endpoint: Optional[str] = None
```
or maybe not so important inside tests

alaeddine-13 (commented):
I just left some style comments.
I tested locally reduce and with an executor that has many endpoints on the same method:
```
class Exec(Executor):
    @requests(on=['/search', '/index', '/foo'])
    def search_index(self, **kwargs):
        pass
```
Do you think it's helpful to include tests for both cases ?

tests/integration/gateway_clients/test_clients_gateways.py
jacobowitz: for the test not needed I think, specifically because its always None here actually and not uses as a string

jacobowitz:
@alaeddine-13 
thx for the feedback! I've incorporate the style suggestions (except the one in the test), I think they all make the code better readable and less error prone. 

Adding another test with reduce should not be needed I think as you dont rely explictly on the polling?

alaeddine-13:
> explictly on the polling

correct

docs/fundamentals/flow/topology.md
```diff
@@ -218,7 +218,11 @@ This is helpful in two situations:
 
 Then splitting the load across two or more machines yields better results.
 
-For Shards, you can define which shard (instance) will receive the request from its predecessor. This behaviour is called `polling`. By default `polling` is set to `ANY`, which means only one shard will receive a request. If `polling` is to `ALL` it means that all Shards will receive a request.
+For Shards, you can define which shard (instance) will receive the request from its predecessor. This behaviour is called `polling`. `ANY` means only one shard will receive a request and `ALL` means that all Shards will receive a request. Polling can be configured per endpoint (like `/index`) and Executor.
+By default the following `polling` is applied:
+- `ANY` for endpoints at `/index`
+- `ALL` for endpoints at `/search`
+- `ANY` for all other endpoints
```
alaeddine-13, line 225: maybe we should mention that `/index` will always be 'any' and `/search` will always be `all` regardless of the `polling` value, unless you modify them in `endpoint_polling` ?

alaeddine-13 (approved):
LGTM, just a simple documentation suggestion

docs/fundamentals/flow/topology.md
jacobowitz: i tried to say that with this paragraph, but I will make it a bit more clear

jacobowitz: I've made it more explicit now in the docs. I think its still confusing, but it will be hard to explain in the docs I think.
I think we should probably merge the two options into a single one, I've created a follow up issue for this https://github.com/jina-ai/internal-tasks/issues/359

PR #4112, opened 2021-12-29
feat(daemon): remote peas on star routing

bwanglzu:


tests/distributed/test_remote_peas/test_remote_peas.py
```diff
@@ -270,9 +275,9 @@ async def test_psuedo_remote_peas_shards(gateway, head, worker, polling):
         worker_host = __docker_host__ if is_remote(head) else HOST
```
jacobowitz, line 275: I think its never correct to use __docker_host__ here no? The head is not running as a container as far as I understand so __docker_host__ will never be resolved correctly? You should use that rather to communicate from inside a container to the host

jacobowitz (changes requested):

tests/distributed/test_remote_peas/test_remote_peas.py
```diff
@@ -185,23 +189,31 @@ async def async_inputs():
...
-@pytest.mark.parametrize('gateway', ['local'])
-@pytest.mark.parametrize('head', ['local', 'remote'])
-@pytest.mark.parametrize('worker', ['local', 'remote'])
-async def test_psuedo_remote_peas_topologies(gateway, head, worker):
+@pytest.mark.parametrize(
+    'gateway, head, worker',
+    [
+        ('local', 'local', 'local'),
+        ('local', 'local', 'remote'),
+        # ('local', 'remote', 'remote'),  TODO bring it back.
```
jacobowitz, line 197: why is this excluded? this should work? Only remote head and local worker is a problem, both being remote should work.

jacobowitz (changes requested):

jacobowitz:
fixed a small issue in the test setup, preventing jinad from kicking in. Also rebased against latest feat-star-routing

jacobowitz:
I think the tests should pass now, at least locally it looks fine. It takes ages though. So I hope the tests doesnt time out

PR #4129, opened 2022-01-05
feat: change Flow return flattened DocumentArray when no callback is specified

JohannesMessner:


tests/unit/flow-orchestrate/test_docarray_return.py
```diff
@@ -0,0 +1,83 @@
...
+def test_flatten_docarrays():
+    f = Flow().add(uses=SimplExecutor)
+    with f:
+        results = f.post(
+            on='/index',
+            inputs=[Document() for _ in range(100)],
+            request_size=10,
+            return_results=True,
+        )
+        assert isinstance(results, DocumentArray)
```
JoanFM, line 29: do the asserts outside the context manager, otherwise an error would not be raised and test would pass

JoanFM (changes requested):

JoanFM:
Closing as duplicated

PR #4130, opened 2022-01-06
feat: support jinahub+sandbox

mapleeit:
@jacobowitz @JoanFM 

I drafted this PR based on the discussion we have yesterday. Please tell me if I missed something. 

- [x] Fetching sandbox address dynamically.
- [x] Use production environment
- [x] Reduce the invoking time  
- [x] Documentation

---

Close: https://github.com/jina-ai/hubble/issues/245

### DIFF
--- a/docs/advanced/experimental/sandbox.md
+++ b/docs/advanced/experimental/sandbox.md
@@ -43,11 +43,11 @@ If the Jina version of Gateway is later than the latest released Jina version (t
 
 ## Version Consistency
 
-The Jina version inside the Sandbox will be the same as the one in the Gateway. The Gateway means the place where the Flow was run.
+The Jina version inside the Sandbox will be the same as the one in the place where the Flow was run. For example, if you run the Flow in your local machine, then it's the version of Jina in your local pip packages.
 
-## Mixed with local Executors
+## Mixed with non-sandbox Executors
 
-It can also be mixed with local Executors.
+It can also be mixed with non-sandbox Executors. For example, pure Python class.
 
 ```python
 from jina import Flow, Document, DocumentArray, Executor, requests
--- a/docs/advanced/experimental/sandbox.md
+++ b/docs/advanced/experimental/sandbox.md
@@ -6,8 +6,8 @@ Jina Sandbox allows you to try out different Jina Executors online and see how t
 
 ## Prerequisites
 
-- Knowledge about ref`Jina Executor <executor>`
-- Knowledge about ref`Jina Hub <hub>`
+- Knowledge about Jina Executor
+- Knowledge about Jina Hub
 
 ## Start a Flow using Jina Sandbox
 
--- a/docs/advanced/experimental/sandbox.md
+++ b/docs/advanced/experimental/sandbox.md
@@ -1,0 +1,69 @@
+(sandbox)=
+
+# Jina Sandbox
+
+Jina Sandbox allows you to try out different Jina Executors online and see how they work together without downloading all the Executors to your local machine.
+
+## Prerequisites
+
+- Knowledge about ref`Jina Executor <executor>`
+- Knowledge about ref`Jina Hub <hub>`
+
+## Start a Flow using Jina Sandbox
+
+```python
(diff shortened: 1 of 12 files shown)


jina/peapods/networking.py
```diff
@@ -35,13 +36,18 @@ def add_connection(self, address: str):
         :param address: Target address of this connection
         """
         if address not in self._address_to_connection_idx:
+            parsed_address = urlparse(address)
+            use_https = parsed_address.scheme == "https"
```
JoanFM, line 40: use single quote please

jina/peapods/networking.py
```diff
@@ -35,13 +36,18 @@ def add_connection(self, address: str):
         :param address: Target address of this connection
         """
         if address not in self._address_to_connection_idx:
+            parsed_address = urlparse(address)
```
JoanFM, line 39: Sometimes addresses are names when docker compose or kubernetes is used. Are we sure this will not raise an exception there?

JoanFM (commented):
Looks quite good, seems like the way to go

jina/peapods/networking.py
mapleeit: addressed

jina/peapods/networking.py
mapleeit: addressed

mapleeit:
@JoanFM I don't have too much faith on the test result. I tried two times. All the failed ones in the 2nd test result are successful in the 1st test result.

Failed in 2nd test. https://github.com/jina-ai/jina/actions/runs/1712028062/attempts/2

- core-test (3.7, tests/integration/scale/)
- core-test (3.7, tests/distributed/test_topologies_docker/)

But succeeded in 1st test. https://github.com/jina-ai/jina/actions/runs/1712028062/attempts/1

docs/advanced/experimental/sandbox.md
```diff
@@ -0,0 +1,69 @@
...
+If all these three factors are matched, then it will reuse the existing sandboxes.
+
+```{admonition} Caution
+:class: caution
+If the Jina version of Gateway is later than the latest released Jina version (things happens when you are in the master branch of jina repository), then the sandbox will always be created instead of reused.
+```
+
+## Version Consistency
+
+The Jina version inside the Sandbox will be the same as the one in the Gateway. The Gateway means the place where the Flow was run.
```
JoanFM, line 46: This is not correct, I would not mention the `Gateway`. I would mention the `driver` program where the Flow lives

docs/advanced/experimental/sandbox.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Mixed with local Executors
+
+It can also be mixed with local Executors.
```
JoanFM, line 50: Do not call local, call them own executors or non-sandboxed. It is not accurate to say local, because it can be merged with any kind of Executor

jina/flow/base.py
```diff
@@ -1031,7 +1031,7 @@ def start(self):
                 os.environ[k] = str(v)
 
         for k, v in self:
-            if not getattr(v.args, 'external', False):
+            if not getattr(v.args, 'external', False) and not v.is_sandbox:
```
JoanFM, line 1034: Can u perhaps add a `property` in `BasePod` 

```python
@property 
def external(self)
    return getattr(self.args, 'external', False) or self.is_sandbox
```

and use it in this check?

JoanFM (changes requested):
Some comments

docs/advanced/experimental/sandbox.md
mapleeit: addressed

docs/advanced/experimental/sandbox.md
mapleeit: addressed

jina/flow/base.py
mapleeit: addressed

JoanFM (approved):

JoanFM (approved):

PR #4132, opened 2022-01-06
feat: enable daemon tests again

jacobowitz:
enable more daemon tests again

TODO: create follow up task to remove cancel event setting


jacobowitz:
all distributed/daemon tests are passing now :tada:

PR #4137, opened 2022-01-07
feat: flow post return DocumentArray

JohannesMessner:


JoanFM (changes requested):
Do not add result_as_docarray option in Client

tests/unit/flow-orchestrate/test_flow_routing.py
```diff
@@ -32,10 +32,10 @@ def test_expected_messages_routing():
     )
 
     with f:
-        results = f.post(on='/index', inputs=[Document(text='1')], return_results=True)
+        docs = f.post(on='/index', inputs=[Document(text='1')], return_results=True)
         # there merge executor actually does not merge despite its name
-        assert len(results[0].docs) == 2
-        assert results[0].docs[0].text == 'merged'
+        assert len(docs) == 2
```
JoanFM, line 37: try moving outside context manager

JoanFM, line 37: try moving outside context manager

jina/clients/base/__init__.py
```diff
@@ -27,7 +27,12 @@ class BaseClient(ABC):
     :param kwargs: additional parameters that can be accepted by client parser
     """
 
-    def __init__(self, args: Optional['argparse.Namespace'] = None, **kwargs):
+    def __init__(
+        self,
+        args: Optional['argparse.Namespace'] = None,
+        results_as_docarray=False,
```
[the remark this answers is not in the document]
JohannesMessner: @jacobowitz suggested I do this as the easiest solution, and I think it makes sense. Because to change the behaviour in Flow I changed PostMixin, which then also affects Client, but we want that to remain unchanged. So as I see it, either we split the Mixin somehow (which might defeats it purpose a little bit), or we make this optional on the Client level. That way the Flow can set this property for its internal Client. So that was the idea - is there a better way?

tests/unit/flow-orchestrate/test_flow_routing.py
JohannesMessner: Ok but a lot of tests have the assert inside the context manager; I left it where it is and just changed it. Should I go through all the existing tests and move the asserts outside the context manager, while I'm at it already?

JoanFM: Just move them as u see, do not focus on this, but if u see it while changing and remember is nice to change

jina/clients/base/__init__.py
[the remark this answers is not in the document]
JoanFM: Ok, then keep it okey

JoanFM (changes requested):

jina/clients/__init__.py
```diff
@@ -22,6 +22,7 @@ def Client(
     port: Optional[int] = None,
     protocol: Optional[str] = 'GRPC',
     proxy: Optional[bool] = False,
+    results_as_docarray: Optional[bool] = False,
```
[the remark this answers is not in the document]
JohannesMessner: It is supposed to be a False default, actually. Is that denoted differently?

tests/distributed/test_workspaces/test_remote_workspaces.py
```diff
@@ -215,7 +215,7 @@ def docker_compose(request):
 @pytest.mark.parametrize('docker_compose', [compose_yml], indirect=['docker_compose'])
 def test_upload_simple_non_standard_rootworkspace(docker_compose):
     f = (
-        Flow()
+        Flow(port_expose=exposed_port)
```
JoanFM, line 218: where is this exposed port?

JohannesMessner: It is defined up top (line 17) and then just used between each Flow and Client. Is that what you mean?

jina/clients/mixin.py
```diff
@@ -25,22 +26,22 @@ def post(
         continue_on_error: bool = False,
         return_results: bool = False,
         **kwargs,
-    ) -> Optional[List['Response']]:
+    ) -> Optional['DocumentArray']:
```
JoanFM, line 29: This type hint should be:

```python
Optional[Union['DocumentArray', List['Response']]]
```

JoanFM (changes requested):

jina/clients/__init__.py
JoanFM, line 25:
```suggestion
    results_as_docarray: bool = False,
```

JoanFM (approved):

PR #4139, opened 2022-01-07
fix: random port assignment

jacobowitz:
Changing the way we assign ports from a Flow to avoid double allocations (both within a Flow and between a Flow and JinaD).
Also increasing the number of ports used in CI to avoid issues there.


jina/orchestrate/peas/jinad.py
```diff
@@ -59,7 +58,7 @@ def __call__(
         self.is_cancelled = is_cancelled
         self.pea_id = None
         self._logger = JinaLogger('RemotePea', **vars(args))
-        run_async(self._run)
+        asyncio.run(self._run())
```
JoanFM, line 61: I am not sure this would work, u are passing the result

jacobowitz: the result of an `async def` is a courine which should be executed by asyncio.run(). I dont see why we are using our custom logic here? I also doubt that my change will help but this section of the code is a bit phishy and I try to remove the strange things here

JoanFM (approved):

PR #4154, opened 2022-01-11
fix(document): fix import statement

pete-sk:
Fixed wrong number of dots when importing Document in image.py


JoanFM:
Hey @pete-sk ,

Thanks for the contribution,

However, we are in the process of migrating to use our new project https://github.com/jina-ai/docarray as a replacement of this internal docarray, so this fix would not be released and therefore is better to wait for this change and see if it is fixed then.

Closing as explained here

PR #4165, opened 2022-01-12
refactor: use absolute imports in daemon

JoanFM:


daemon/helper.py
```diff
@@ -5,7 +5,7 @@
 
 import aiohttp
 
-from .excepts import PartialDaemon400Exception
+from daemon.excepts import PartialDaemon400Exception
 
 if TYPE_CHECKING:
     from .models import DaemonID
```
JoanFM, line 11:
```suggestion
    from daemon.models import DaemonID
```

deepankarm (approved):

PR #4168, opened 2022-01-12
refactor: create orchestrate and serve modules

JoanFM:


jacobowitz (dismissed):
cool, I think it makes a lot of sense to change the packages like this :+1:

PR #4177, opened 2022-01-17
test: switch test from threading to multiprocessing

JohannesMessner:


JoanFM (changes requested):
Make sure that the test can fail, and that the asserts stay meaningful

tests/integration/concurrent_clients/test_concurrent_clients.py
```diff
@@ -21,33 +21,42 @@ def ping(self, docs: DocumentArray, **kwargs):
...
-            process_pool.append(t)
+            p.start()
+            process_pool.append(p)
 
-        for t in process_pool:
-            t.join()
+        for p in process_pool:
+            p.join()
+
+        while not pqueue.empty():
```
JoanFM, line 60: assert the length of `pqueue` before entering the while

JoanFM (changes requested):

tests/integration/concurrent_clients/test_concurrent_clients.py
JohannesMessner: according to the doc, `qsize()` is only approximate; it probably doesn't matter because all the processes are `join()`ed, but I'm doing the length check manually to be safe.

JoanFM (approved):

PR #4189, opened 2022-01-19
fix(runtime): worker grpc proxy env

numb3r3:
The `GrpcConnectionPool` cannot work with http and https proxy

related discussion: https://github.com/jina-ai/jina/pull/1348#discussion_r530721752


JoanFM (approved):

sanchezsm91:
I would like to be unsubscribed. Thank you

Issue #4194, opened 2022-01-20
module 'docarray' has no attribute '__version__' when upgrading jina from 2.6.4 to 2.6.5.dev60


mapleeit:
**Describe the bug**

Reproduces steps:

### 1. create a new pyenv
`pyenv virtualenv test1`

### 2. use this new env
`pyenv activate test1`

### 3. check jina
`jina -v` -> no jina, good

### 4. install 2.6.4
`pip install jina==2.6.4`
`jina -v` -> 2.6.4, good

### 5. upgrade jina 
`pip install -U --pre jina`
`jina -v` -> **throw error, werid**

```
Traceback (most recent call last):
  File "/Users/miaozhaofeng/.pyenv/versions/test1/bin/jina", line 8, in <module>
    sys.exit(main())
  File "/Users/miaozhaofeng/.pyenv/versions/3.9.5/envs/test1/lib/python3.9/site-packages/cli/__init__.py", line 107, in main
    from cli import api
  File "/Users/miaozhaofeng/.pyenv/versions/3.9.5/envs/test1/lib/python3.9/site-packages/cli/api.py", line 1, in <module>
    from jina.peapods.runtimes.head import HeadRuntime
  File "/Users/miaozhaofeng/.pyenv/versions/3.9.5/envs/test1/lib/python3.9/site-packages/jina/__init__.py", line 68, in <module>
    __docarray_version__ = _docarray.__version__
AttributeError: module 'docarray' has no attribute '__version__'
```

### 6. check docarray package

```
❯ ls /Users/miaozhaofeng/.pyenv/versions/3.9.5/envs/test1/lib/python3.9/site-packages/docarray
__pycache__ document    proto       score
array       math        resources   types.py
```

❌ **No `__init__.py`!** WHY

**Describe how you solve it**

---

**Environment**

**Screenshots**

jacobowitz:
I understood what happens if you do the steps you described:

1. `docarray` gets installed into your site-packages in your venv as a new dependency
2. Jina 2.6.4 gets uninstalled which means removing the previous docarray folder. Unfortunately this docarray folder is also in your site-packages. Effectively this will destroy the new `docarray` installation.

I will try to come up with some workaround in Jina setup

jacobowitz:
We decided that this cant be easily fixed. The recommend workaround is to not upgrade Jina, but rather install in a new venv oder uninstall the old jina version first.

PR #4203, opened 2022-01-21
fix: rescue docarray in setup

jacobowitz:
This closes https://github.com/jina-ai/jina/issues/4194

Upgrading from Jina 2.x to 3.x is broken because pip will install first the new `docarray` pip package and then remove the old docarray package (which was part of jina repo). Because both are installed in the same folder, this will leave the new `docarray` package in some invalid state.
This change to the setup introduces a hack to rescue those cases: My code is checking if `docarray` can be imported correctly (by checking access to the version, this fails for master now). If this fails `docarray` is reinstalled via pip.

In non upgrading cases we will not enter the exception handler and my hack will not do anything (when installing jina in a new venv).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,11 +74,32 @@ def register_ac():
         pass
 
 
+def rescue_docarray():
+    """Upgrading from 2.x to 3.x is broken (https://github.com/jina-ai/jina/issues/4194)
+    This function checks if docarray is broken and if so attempts to rescue it
+    """
+    try:
+        import docarray as docarray
+
+        __docarray_version__ = docarray.__version__
+
+    except AttributeError:
+        # Being here means docarray is not installed correctly, attempt to reinstall it
+        # as recommended by pip https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
+        import subprocess
+
+        subprocess.check_call(
+            [sys.executable, '-m', 'pip', 'uninstall', '--yes', 'docarray']
+        )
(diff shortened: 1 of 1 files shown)


mapleeit (approved):
Python package system is so strange, from a NodeJSer's perspective. 

Good hack. ; )

setup.py
```diff
@@ -74,11 +74,32 @@ def add_ac(k, v):
...
+        import docarray as docarray
+
+        __docarray_version__ = docarray.__version__
+
+    except AttributeError:
+        # Being here means docarray is not installed correctly, attempt to reinstall it
+        # as recommended by pip https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
+        import subprocess
+
+        subprocess.check_call(
```
JoanFM, line 91: doesnt pip have a python package? This for instance would not work on my environment becauee I use pip3.7 for instance

hanxiao: @numb3r3 could you throw some light here? i think you did something similar in `hub.uses(from_source)`

jacobowitz: I did this specifically because Pip says so: https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program

numb3r3: same method is utilized in `hub.uses(...)`. `[sys.executorable, '-m', 'pip']` always use the executable pip associated with the current python version you are using.

numb3r3 (approved):

Issue #4145, opened 2022-01-10
Check external pods in K8s


JoanFM:
**Describe the feature**
After #4089 and #4130 are merged. We should make sure the user experience of connecting to a sandbox external Executor is good

jacobowitz:
Make sure that:

- External pods are not part of the ressources created by us
- Gateway can communicate with the external pod

PR #4223, opened 2022-01-27
feat: use external pods from k8s

jacobowitz:
This closes https://github.com/jina-ai/jina/issues/4145

This PR adds the possibility to use external Pods from a Flow in K8s. Before this PR, we would always generate K8s yaml files for the external Pods.
With this change those yamls for the external Pod are no longer generated. Also the Gateway pod address list logic is updated to cater for those external Pods. External Pods always need to be announced to the Gateway. It does not matter if the K8sConnectionPool is used or not. The gateway will always communicate to the head of the external Pod via a fixed address, configured in the Flow.

### DIFF
--- a/jina/orchestrate/flow/base.py
+++ b/jina/orchestrate/flow/base.py
@@ -445,7 +445,9 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
 
         return graph_dict
 
-    def _get_k8s_pod_addresses(self, k8s_namespace: str) -> Dict[str, List[str]]:
+    def _get_k8s_pod_addresses(
+        self, k8s_namespace: str, k8s_connection_pool: bool
+    ) -> Dict[str, List[str]]:
         graph_dict = {}
         from jina.serve.networking import K8sGrpcConnectionPool
         from jina.orchestrate.pods.config.helper import to_compatible_name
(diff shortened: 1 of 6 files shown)


jina/orchestrate/flow/base.py
```diff
@@ -445,22 +445,33 @@ def _get_pod_addresses(self) -> Dict[str, List[str]]:
...
                 continue
-            pod_k8s_address = (
-                f'{to_compatible_name(v.head_args.name)}.{k8s_namespace}.svc'
-            )
-            graph_dict[node] = [
-                f'{pod_k8s_address}:{K8sGrpcConnectionPool.K8S_PORT_IN}'
-            ]
 
-        return graph_dict
+            if hasattr(v.args, 'external') and v.args.external:
```
JoanFM, line 459: A Pod has an external Property that accounts for sandbox as well, better to use that one

jina/orchestrate/flow/base.py
```diff
@@ -1679,13 +1690,17 @@ def to_k8s_yaml(
         k8s_namespace = k8s_namespace or self.args.name or 'default'
 
         for node, v in self._pod_nodes.items():
+            if hasattr(v.args, 'external') and v.args.external:
```
JoanFM, line 1693: same here

jina/orchestrate/pods/config/k8s.py
```diff
@@ -354,7 +352,9 @@ def to_k8s_yaml(
             .. # noqa: DAR201
             .. # noqa: DAR101
         """
-        if self.name == 'gateway':
+        if hasattr(self.args, 'external') and self.args.external:
+            return []
```
JoanFM, line 356: here as well

tests/k8s/test_k8s.py
```diff
@@ -587,3 +589,58 @@ async def test_flow_connection_pool(logger, k8s_connection_pool, docker_images,
...
+
+@pytest.mark.asyncio
+@pytest.mark.timeout(3600)
+@pytest.mark.parametrize(
+    'docker_images',
+    [['jinaai/jina']],
+    indirect=True,
+)
+async def test_flow_with_external_pod(logger, docker_images, tmpdir):
+    class DocReplaceExecutor(Executor):
```
JoanFM, line 602: just for completion, can we add a test where the external Pod is in Kubernetes?

JoanFM (changes requested):

tests/k8s/test_k8s.py
jacobowitz: I could add this, but I am not sure its providing much value. This test would actually test more the external pod itself being in K8s which is not what this feature is about (its about the Flow using a given external Pod). 

From the Flow/Gateway perspective it should not matter whatsever where the external Pod is located, we just connect to any given address.

jina/orchestrate/pods/config/k8s.py
jacobowitz: thats not a Pod, but a K8sPodConfig, I think it has no `external` property and it also does not make much sense to add here. An external K8sPodConfig should never be created in the first place (and it is not with the current code), this check is just to double check this behaviour.

JoanFM: Ah yes, sure! then this check is not needed at all in theory right?
I would just have an `assert` at instantiation time

tests/k8s/test_k8s.py
JoanFM: It would document, ensure that the pattern works

Issue #3946, opened 2021-11-17
Executor: endpoints should merge parameters from shards


cristianmtr:
**Describe the feature**

As an Executor developer, I would like that, when I return a dictionary (e.g. on a `/status`-type endpoint) and have `parallel > 1`, this dictionary is not overridden by the last shards, but instead is **merged** somehow. 

**Your proposal**

Instead of returning just a dictionary, return a dictionary of dictionaries perhaps, with pea_ids as keys. The default dictionary could be duplicated as the top-level entries too, to maintain backwards-compatibility.

Something like, perhaps:

```
{
  'last_sync': '2021-11-17T15:05:14.137362',
  'psql_docs': 10.0,
  'pea_id': 0.0,
  'hnsw_docs': 3.0,
  'pea_0': {
    'last_sync': '2021-11-17T15:05:14.137362',
    'psql_docs': 10.0,
    'pea_id': 0.0,
    'hnsw_docs': 3.0
  },
  ...
  'pea_2': {
    'last_sync': '2021-11-17T15:05:14.137362',
    'psql_docs': 10.0,
    'pea_id': 2.0,
    'hnsw_docs': 3.0
  },          
}

```

---

Reproduce scenario with:

start docker PSQL:

`docker run -e POSTGRES_PASSWORD=123456  -p 127.0.0.1:5432:5432/tcp postgres:13.2`                                                                                                                                                                        

and then run:

```python
import numpy as np
from jina import DocumentArray, Flow, Document


def get_documents(nr=10, index_start=0, emb_size=7):
    random_batch = np.random.random([nr, emb_size]).astype(np.float32)
    for i in range(index_start, nr + index_start):
        d = Document()
        d.id = f'aa{i}'  # to test it supports non-int ids
        d.embedding = random_batch[i - index_start]
        yield d


def main():
    emb_size = 10

    docs = DocumentArray(get_documents(emb_size=emb_size))

    with Flow().add(
            uses='jinahub://HNSWPostgresIndexer/test-cristian',
            uses_with={
                'dim': emb_size
            },
            parallel=3,
            polling='all',
            install_requirements=True,
            uses_after='jinahub+docker://MatchMerger/v0.2'
    ) as f:
        # test for empty sync from psql
        result = f.post('/status', None, return_results=True)
        result = dict(result[0].parameters)
        print(result)
        f.post('/index', docs)
        f.post('/sync')
        search_docs = DocumentArray(
[...]

PR #4224, opened 2022-01-28
feat: change the way parameters are returned from Executors

samsja:
… parameters but merge them

### DIFF
--- a/jina/clients/mixin.py
+++ b/jina/clients/mixin.py
@@ -11,16 +11,19 @@ if TYPE_CHECKING:
 
 
 def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
+    key_result = '__results__'
+
     if parameters:
-        if 'results' in parameters.keys():
-            if not isinstance(parameters['results'], dict):
+
+        if key_result in parameters.keys():
+            if not isinstance(parameters[key_result], dict):
                 warnings.warn(
-                    'It looks like you passed a dictionary with the key `results` to `parameters`.'
+                    'It looks like you passed a dictionary with the key `{key_result}` to `parameters`.'
                     'This key is reserved, so the associated value will be deleted.'
                 )
-                parameters.update({'results': dict()})
+                parameters.update({key_result: dict()})
     else:
-        parameters = {'results': dict()}
+        parameters = {key_result: dict()}
 
     return parameters
 
--- a/jina/clients/mixin.py
+++ b/jina/clients/mixin.py
@@ -1,5 +1,6 @@
 from functools import partialmethod
 from typing import Optional, Dict, List, AsyncGenerator, TYPE_CHECKING, Union
+import warnings
 
 from jina.helper import run_async
 
@@ -11,13 +12,15 @@ if TYPE_CHECKING:
 
 def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
     if parameters:
-        if "results" in parameters.keys():
-            raise KeyError(
-                f"key results can't be used in `parameters` because it is use internally to return the results"
-            )
-        parameters.update({"results": dict()})
+        if 'results' in parameters.keys():
+            if not isinstance(parameters['results'], dict):
+                warnings.warn(
+                    'It looks like you passed a dictionary with the key `results` to `parameters`.'
+                    'This key is reserved, so the associated value will be deleted.'
+                )
+                parameters.update({'results': dict()})
     else:
-        parameters = {"results": dict()}
+        parameters = {'results': dict()}
 
     return parameters
 
--- a/jina/clients/mixin.py
+++ b/jina/clients/mixin.py
@@ -18,7 +18,7 @@ def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
         if key_result in parameters:
             if not isinstance(parameters[key_result], dict):
                 warnings.warn(
-                    'It looks like you passed a dictionary with the key `{key_result}` to `parameters`.'
+                    f'It looks like you passed a dictionary with the key `{key_result}` to `parameters`.'
                     'This key is reserved, so the associated value will be deleted.'
                 )
                 parameters.update({key_result: dict()})
--- a/jina/clients/mixin.py
+++ b/jina/clients/mixin.py
@@ -9,6 +9,19 @@ if TYPE_CHECKING:
     from jina import DocumentArray
 
 
+def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
+    if parameters:
+        if "results" in parameters.keys():
+            raise KeyError(
+                f"key results can't be used in `parameters` because it is use internally to return the results"
+            )
+        parameters.update({"results": dict()})
+    else:
+        parameters = {"results": dict()}
+
+    return parameters
+
+
 class PostMixin:
     """The Post Mixin class for Client and Flow"""
 
@@ -70,6 +83,8 @@ class PostMixin:
         if (on_always is None) and (on_done is None):
             return_results = True
 
+        parameters = _include_results_field_in_param(parameters)
+
         return run_async(
             _get_results,
             inputs=inputs,
@@ -125,6 +140,9 @@ class AsyncPostMixin:
         c = self.client
         c.show_progress = show_progress
         c.continue_on_error = continue_on_error
+
+        parameters = _include_results_field_in_param(parameters)
+
         async for r in c._get_results(
             inputs=inputs,
             on_done=on_done,
--- a/jina/clients/mixin.py
+++ b/jina/clients/mixin.py
@@ -15,7 +15,7 @@ def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
 
     if parameters:
 
-        if key_result in parameters.keys():
+        if key_result in parameters:
             if not isinstance(parameters[key_result], dict):
                 warnings.warn(
                     'It looks like you passed a dictionary with the key `{key_result}` to `parameters`.'
--- a/jina/serve/runtimes/request_handlers/data_request_handler.py
+++ b/jina/serve/runtimes/request_handlers/data_request_handler.py
@@ -121,7 +121,7 @@ class DataRequestHandler:
         return requests[0]
 
     @staticmethod
-    def replace_docs(request, docs):
+    def replace_docs(request: List['DataRequest'], docs: 'DocumentArray') -> None:
         """Replaces the docs in a message with new Documents.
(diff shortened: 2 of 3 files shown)


jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -164,6 +173,20 @@ def get_docs_matrix_from_request(
...
+    def get_parameters_dict_from_request(
+        requests: List['DataRequest'],
+    ) -> List['Dict']:
+        """
+        Returns a parameters dict from a list of DataRequest objects.
+        :param requests: List of DataRequest objects
+        :return: parameters matrix: list of parameters (Dict) objects
+        """
+        return {
+            f"request-{i}": getattr(request, "parameters")
```
JoanFM, line 186: please use single quotes

JoanFM (changes requested):
Can you please describe what is the behavior being implemented?

samsja:
Response parameters are now reduced and merge when using an executors with multiple shards. Similar to what is happening when reducing requests docs.

old behavior : Only the last request parameters were return as discussed in https://github.com/jina-ai/jina/issues/3946

new behavior : All of the requests parameters are return in dictionary where the keys is 'request-i' i being the number of the request.

Ex:
```python

class ExecutorStatus(Executor):
    @requests
    def endpoint(self, docs: DocumentArray, **kwargs):
       return {'shard_id': self.runtime_args.shard_id}

flow = Flow(port_expose=exposed_port).add(
        uses=ExecutorStatus, shards=2, polling='all'
    )

    with flow as f:
        da = DocumentArray([Document() for _ in range(5)])
        resp = Client(port=exposed_port).post('/status', inputs=da, return_results=True)

print(resp[0].parameters)

>>> 
{ 
    "request-0" : {"shard_id":0.0} ,  
    "request-0" : {"shard_id":1.0} 
}
 ```
 
 Why not use a list to return the parameters ? :
 
 * Protobuff require the parameters field to be a dictionary ( could it be change somehow ?) with str-like keys 
 * to preserve compatibility, resp.paremeters use to be a dictionary

**Comment by JoanFM** [comment]
> new behavior : All of the requests parameters are return in dictionary where the keys is 'request-i' i being the number of the request.

The `request-id` as key is not needed. Because this from the `Client` can be automatically extracted.

I am not sure how u add 2 values to the same dictionary?

**Comment by JoanFM** [comment]
What may be interesting is to return the `runtime` name as the `key` so that it is automatically inferred

File: jina/serve/runtimes/request_handlers/data_request_handler.py [hunk 2]
```diff
@@ -164,6 +181,26 @@ def get_docs_matrix_from_request(
...
+    ) -> 'Dict':
+        """
+        Returns a parameters dict from a list of DataRequest objects.
+        :param requests: List of DataRequest objects
+        :return: parameters matrix: list of parameters (Dict) objects
+        """
+
+        parameters = requests[0].parameters
+        if "results" not in parameters.keys():
+            parameters["results"] = dict()
```
samsja, line 196:
```suggestion
            parameters['results'] = dict()
```

jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -164,6 +181,26 @@ def get_docs_matrix_from_request(
...
+        # we only merge the results and make the assumption that the others params does not change during execution
+
+        for req in requests:
+            parameters["results"].update(req.parameters.get("results", dict()))
```
samsja, line 200:
```suggestion
            parameters['results'].update(req.parameters.get('results', dict()))
```

samsja (commented):
should use single quote

jina/clients/mixin.py
```diff
@@ -9,6 +10,21 @@
     from jina import DocumentArray
 
 
+def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
+    if parameters:
+        if 'results' in parameters.keys():
```
JoanFM, line 15: Use a protected `key`. `__RESULTS__` . Otherwise it is a too common word

JoanFM (changes requested):

samsja:
new behaviour : 

```python

class ExecutorStatus(Executor):
    @requests
    def endpoint(self, docs: DocumentArray, **kwargs):
       return {'shard_id': self.runtime_args.shard_id}

flow = Flow(port_expose=exposed_port).add(
        uses=ExecutorStatus, shards=2, polling='all'
    )

    with flow as f:
        da = DocumentArray([Document() for _ in range(5)])
        resp = Client(port=exposed_port).post('/status', inputs=da, return_results=True)

print(resp[0].parameters)

>>> 
{ "results":
    {
        "pea-name-0" : {"shard_id":0.0} ,  
        "pea-name-1" : {"shard_id":1.0} 
   }
}
 ```

File: tests/integration/reduce/test_reduce.py [hunk 5]
```diff
@@ -195,3 +208,27 @@ def test_uses_before_no_reduce_real_executor_uses():
...
+
+    with flow as f:
+        da = DocumentArray([Document() for _ in range(5)])
+        resp = Client(port=exposed_port).post(
+            '/status', parameters={"foo": "bar"}, inputs=da, return_results=True
+        )
+
+    # assert no reduce happened
+
+    assert resp[0].parameters["foo"] == "bar"
```
JoanFM, line 227: single quotes

tests/integration/reduce/test_reduce.py
```diff
@@ -195,3 +208,27 @@ def test_uses_before_no_reduce_real_executor_uses():
...
+    assert len(resp[0].parameters["results"]) == n_shards
+
+    print(resp[0].parameters)
```
JoanFM, line 230: remove print from test

tests/integration/reduce/test_reduce.py
```diff
@@ -195,3 +208,27 @@ def test_uses_before_no_reduce_real_executor_uses():
...
+
+    for _, param in resp[0].parameters["results"].items():
```
JoanFM, line 232: single quote

JoanFM (changes requested):
Use a different key than `results`

jina/clients/mixin.py
samsja: the issue with this option is that when the user is going to access the result he will have to do 
`paramaters["__results__"]` so he will probably think that he is doing something wrong

jina/clients/mixin.py
```diff
@@ -9,6 +10,24 @@
...
 
+def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
+    key_result = '__results__'
+
+    if parameters:
+
+        if key_result in parameters.keys():
+            if not isinstance(parameters[key_result], dict):
+                warnings.warn(
+                    'It looks like you passed a dictionary with the key `{key_result}` to `parameters`.'
```
cristianmtr, line 21: not an f-string

jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -121,6 +129,15 @@ def replace_docs(request, docs):
         """
         request.data.docs = docs
 
+    @staticmethod
+    def replace_parameters(request, parameters):
```
cristianmtr, line 133: type annotations please

tests/integration/reduce/test_reduce.py
```diff
@@ -195,3 +208,23 @@ def test_uses_before_no_reduce_real_executor_uses():
...
+    )
+
+    with flow as f:
+        da = DocumentArray([Document() for _ in range(5)])
+        resp = Client(port=exposed_port).post(
+            '/status', parameters={'foo': 'bar'}, inputs=da, return_results=True
+        )
+
+    assert resp[0].parameters['foo'] == 'bar'
+    assert len(resp[0].parameters['__results__']) == n_shards
```
cristianmtr, line 226: Why this string key `__results__`? Why not just 

```python
for param in resp[0].parameters:
...
```

cristianmtr (changes requested):

jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -95,17 +96,24 @@ async def handle(self, requests: List['DataRequest']) -> DataRequest:
...
+                DataRequestHandler.replace_docs(requests[0], return_data)
+            elif isinstance(return_data, dict):
                 params = requests[0].parameters
-                params.update(r_docs)
+                results_key = '__results__'
+
+                if not results_key in params.keys():
+                    params[results_key] = dict()
+
+                params[results_key].update({self.args.name: return_data})
```
jacobowitz, line 109: I think we also need `DataRequestHandler.replace_docs(requests[0], docs)` ? Otherwise one can not change docs in place when returning params.

tests/integration/reduce/test_reduce.py
```diff
@@ -195,3 +208,23 @@ def test_uses_before_no_reduce_real_executor_uses():
...
+
+    for _, param in resp[0].parameters['__results__'].items():
```
jacobowitz, line 228: can you please assert also on the text of the document being `doc.text = 'exec-status'`?

jacobowitz (changes requested):

jina/serve/runtimes/request_handlers/data_request_handler.py
samsja: it seems that none of the static method of the DataRequestHandler are typed annotated, should I add type annotation for the method that are out of the scope of this PR as well ? ( for replace_docs and replace_parameters)

samsja: If the type replace_docs and merge_routes should not be added in this PR I can revert (https://github.com/jina-ai/jina/pull/4224/commits/c9a3267fb87df50d605c1472f2c6585424ef5e76)

tests/integration/reduce/test_reduce.py
samsja: good catch once the test are failing now. However it seems that this bug was here before.

jina/serve/runtimes/request_handlers/data_request_handler.py
samsja: You are right something need to be done otherwise the docs are not replace as soon as the executor return a dict. Nevertheless it seems that this bug is not new, it is the failing on current master as well

jacobowitz:
I think tests/unit/orchestrate/flow/flow-construct/test_flow.py::test_flow_change_parameters is broken?

Issue #4054, opened 2021-12-10
uses_requests tries to bind uses_after Executor too


cristianmtr:
```python
from jina import Flow

f = Flow(protocol='http').add(
    uses='jinahub://SimpleIndexer',
    uses_after='jinahub://MatchMerger',
    uses_requests={'/foo': 'update'}, # SI has update, MatchMerger doesn't
)

with f:
     f.block()
```

```
  File "/home/cristian/envs/hub20/lib/python3.8/site-packages/jina/executors/decorators.py", line 60, in arg_wrapper
    f = func(self, *args, **kwargs)
  File "/home/cristian/.jina/hub-packages/mruax3k7/match_merger.py", line 18, in __init__
    super().__init__(**kwargs)
  File "/home/cristian/envs/hub20/lib/python3.8/site-packages/jina/executors/decorators.py", line 60, in arg_wrapper
    f = func(self, *args, **kwargs)
  File "/home/cristian/envs/hub20/lib/python3.8/site-packages/jina/executors/__init__.py", line 97, in __init__
    self._add_requests(requests)
  File "/home/cristian/envs/hub20/lib/python3.8/site-packages/jina/executors/__init__.py", line 118, in _add_requests
    _func = getattr(self.__class__, func)
AttributeError: type object 'MatchMerger' has no attribute 'update'
```

PR #4228, opened 2022-01-31
fix: uses_request being applied to uses_after and uses_before

JohannesMessner:


tests/integration/v2_api/test_override_requests.py
```diff
@@ -25,3 +25,36 @@ def foo(self, docs, **kwargs):
...
+
+    class OtherExecutor(Executor):
+        @requests(on='/bar')
+        def bar(self, docs, **kwargs):
+            for doc in docs:
+                doc.text = 'bar called'
+
+    with Flow(port_expose=exposed_port).add(
+        uses=FooExecutor,
+        uses_requests={'/foo': 'foo'},
```
cristianmtr, line 45: Is the `bar` available on `/bar`? Can you add a test to assert this?

cristianmtr (changes requested):

tests/integration/v2_api/test_override_requests.py
cristianmtr: No, all good for this PR, from what I can tell. Just wondering in terms of general usage. I would expect that OtherExecutor still has its endpoint bound to /bar

JohannesMessner: I agree, might be worth a change

samsja:
https://github.com/jina-ai/jina/issues/4054

tests/integration/v2_api/test_override_requests.py
JohannesMessner: Re-read your comment, understood it this time, and added a test accordingly ;)

cristianmtr: Oh so it does preserve the OtherExecutor.bar binding. Good

cristianmtr (approved):

PR #4229, opened 2022-01-31
fix: parsing of env vars of the form '{ var }'

JohannesMessner:


cristianmtr (approved):

jina/jaml/__init__.py
```diff
@@ -49,17 +82,45 @@ class DummyClass:
...
+    An environment variable `var` is accessed through the following syntax:
+
+    .. highlight:: yaml
+    .. code-block:: yaml
+
+        ${{ env.var }}
+
+    Note the mandatory spaces before and after the variable denotation.
+
+    Context variables can be accessed using the following syntax:
```
cristianmtr, line 109: What is a context var? Isn't this the same 

```
${{ env.var }}
# and
${{ var }}
```?

File: jina/jaml/__init__.py [hunk 2]
```diff
@@ -494,8 +600,8 @@ def load_config(
         :class:`BaseExecutor`, :class:`BaseDriver` and all their subclasses.
 
         Support substitutions in YAML:
-            - Environment variables: `${{ENV.VAR}}` (recommended), ``${{VAR}}``.
-            - Context dict (``context``): ``${{VAR}}``(recommended).
+            - Environment variables: ``${{ ENV.VAR }}`` (recommended), ``$VAR`` (deprecated).
```
cristianmtr, line 603: If it's deprecated, why are we keeping it?

cristianmtr: General question. can be raised in slack or in an alignment. How long are we going to keep it around?

jina/jaml/__init__.py
```diff
@@ -49,17 +82,45 @@ class DummyClass:
...
+
+    Or, if you want to be explicit:
+
+    .. highlight:: yaml
+    .. code-block:: yaml
+
+        ${{ context.context_var }}
 
-    To substitute the value based on a dict,
+    These context variables are passed as a dict:
```
cristianmtr, line 123: I've always passed env vars directly, without the need to `ENV.` prefix. Will it not work from now on? In what situation would you want context variables and how would you pass them to a Flow?

cristianmtr (changes requested):

jina/jaml/__init__.py
JohannesMessner: No, the latter refers exclusively to context variables, which can be passed to Flow.load_config(), Executor.load_config(), Jaml.parse() as a dict. Slightly more info is below. We didn't know about these either until we found them during testing.

jina/jaml/__init__.py
JohannesMessner: We are trying to align with GitHub actions syntax, in which both work, but the first version is encouraged. Han told us to deprecate the second version, my guess is it will stick around for as long as it does within github actions.

jina/jaml/__init__.py
JohannesMessner: Passing env vars like this: ${{ var }} should never have worked in the first place, we believe. So now you either do ${{ env.var }} or $var; the latter is discouraged but should work perfectly fine. You should be able to pass a dict containing context variables to Flow.load_config(context={'context_var' : 1}) . Not sure what the use cases for context vars are, though.

samsja:
The yaml parsing now strictly respect the github actions rules
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
https://docs.github.com/en/actions/learn-github-actions/environment-variables

jina/jaml/__init__.py
```diff
@@ -49,17 +82,45 @@ class DummyClass:
     An expression can be any combination of literal values, references to a context, or functions.
     You can combine literals, context references, and functions using operators.
 
-    You need to use specific syntax to tell Jina to evaluate an expression rather than treat it as a string.
+    You need to use specific syntax to tell Jina to evaluate an expression rather than treat it as a string,
+    which is based on GitHub actions syntax, and looks like this:
```
JoanFM, line 86: is there a link to Github syntax explained?

JoanFM (approved):

PR #4243, opened 2022-02-07
feat: change the way parameters are returned from Executors

samsja:
follow up on this [PR](https://github.com/jina-ai/jina/pull/4224) which is somehow broken on the github side


samsja:
the test named `test_flow_change_parameters` has been deleted because we don't allow anymore to change the parameters within the executors context

JoanFM (approved):

PR #4255, opened 2022-02-08
Update gpu guide for jina 3

JohannesMessner:


JohannesMessner:
I was not able to verify that the container parts still work, since i don't have a local gpu, google colab can't use docker, and i don't have AWS access (and have never used it either...). But i don't see why those parts should be broken by jina 3.

docs/tutorials/gpu-executor.md
```diff
@@ -26,6 +18,102 @@ This tutorial assumes you are already familiar with basic Jina concepts, such as
...
+from jina import DocumentArray, requests
+
+class MyGPUExec(Executor):
+    def __init__(self, device: str = 'cpu', *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.device = device
+
+
+    @requests
+    def encode(self, docs: Optional[DocumentArray], **kwargs):
```
JoanFM, line 47: docs are not `Optional`

docs/tutorials/gpu-executor.md
```diff
@@ -26,6 +18,102 @@ This tutorial assumes you are already familiar with basic Jina concepts, such as
...
+        with torch.inference_mode():
+          ## generate random embeddings
+          embeddings = torch.rand((len(docs),5), device=self.device)
```
JoanFM, line 50: Can we have some example that actually use a `model` that uses device? Here I just have to believe they are created but I cannot feel

JoanFM (changes requested):

docs/tutorials/gpu-executor.md
```diff
@@ -140,12 +228,9 @@ class SentenceEncoder(Executor):
     @requests
     def encode(self, docs: Optional[DocumentArray], **kwargs):
```
JoanFM, line 229: remove `Optional` from `docs`. They are not `Optional` anymore

docs/tutorials/gpu-executor.md
```diff
@@ -26,6 +18,102 @@ This tutorial assumes you are already familiar with basic Jina concepts, such as
...
+Jina enables you to use GPUs like you normally would in a Python script, or in a Docker 
+container - it does not impose any additional requirements or configuration.
+
+
+Let's take a look at a minimal working example, written in PyTorch.
+
+```python
+
+import torch
+from jina import DocumentArray, requests
```
JoanFM, line 38: from docarray import DocumentArray
from jina import Executor, requests

JoanFM (changes requested):

docs/tutorials/gpu-executor.md
JohannesMessner: That is done further down in the tutorial, using a sentence encoder, i kept that more or less as it was in the previous version. This first part is more intended for people that already know how to use their model on GPU, and just want to quickly see how to integrate that with jina. The second part is more elaborate.

JoanFM (approved):

PR #4262, opened 2022-02-09
feat(sandbox): using given port

mapleeit:
- [x] add unit test

### DIFF
--- a/jina/hubble/hubio.py
+++ b/jina/hubble/hubio.py
@@ -612,6 +612,7 @@ with f:
         console = Console()
 
         host = None
+        port = None
         try:
             json_response = requests.get(
                 url=get_hubble_url_v2() + '/rpc/sandbox.get',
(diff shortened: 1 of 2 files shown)


jina/hubble/hubio.py
```diff
@@ -613,19 +613,26 @@ def deploy_public_sandbox(uses: str):
 
         host = None
         try:
-            res = requests.get(
+            json = requests.get(
```
JoanFM, line 616: can we have a different name, it can be confused with the package

JoanFM: maybe json_response?

JoanFM (changes requested):
There should be some tests regarding this behavior

jina/hubble/hubio.py
mapleeit: addressed

mapleeit:
Good job @delgermurun 

The tests have been added, I think it's ready to go, now. @JoanFM

JoanFM (approved):

PR #4266, opened 2022-02-09
docs: rewrite flow section

jacobowitz:
# docs: rewrite flow section

## Description

WIP: This will rewrite the Flow section of the docs for 3.0

TODO:

- [x] Add section for sharding/replica/polling?
- [ ] Fix images
- [ ] Check links
- [ ] Check overlap with Executor section
- [ ] Polish language, code style etc 
- [x] Clarify overlap with docker/k8s how tos
- [ ] Clarify where uses_before/after should go, probably to shards?


docs/fundamentals/flow/flow-api.md
```diff
@@ -1,80 +1,255 @@
...
+
+    client.post('/endpoint')  # empty
 ```
-````
-## Visualize a Flow
 
+Especially during indexing a Client often sends thousands of Documents to a `Flow`. Those Documents are internally batched into a `Request`. The size of these batches can be controlled with the `request_size` keyword. The default `request_size` is 100 `Documents`. The optimal size will depend on your use case.
 ```python
-from jina import Flow
```
JoanFM, line 155: let's import `Document` from `docarray`

docs/fundamentals/flow/flow-api.md
```diff
@@ -1,80 +1,255 @@
...
 
-f = Flow().add().plot('f.svg')
+with Flow() as f:
+    f.post('/', (Document() for _ in range(100)), request_size=10)
```
JoanFM, line 158: let's create a `DocumentArray`

PR #4267, opened 2022-02-09
docs: external executors tutorial

JohannesMessner:


docs/tutorials/external-executor.md
```diff
@@ -0,0 +1,155 @@
...
+docs = DocumentArray([Document(text='Embed me please!') for _ in range(5)])
+
+def print_embedding(resp):
+    doc = resp.docs[0]
+    print(f'{doc.text} has been embedded to shape {doc.embedding.shape}')
+
+with f:
+    docs = f.index(inputs=docs, on_done=print_embedding)
+
+>>> TODO paste results
```
JohannesMessner, line 97: can't try this because huggingface seems still down

docs/tutorials/external-executor.md
```diff
@@ -0,0 +1,155 @@
...
+Now we can run the CLI command again, this time using our custom Executor:
+
+```bash
+jina executor --uses my-exec.yml --port-in 12345
+```
+
+Now that your executor is up and running, we can tap into it just like before:
+
+```python
+from jina import Flow, Document, DocumentArray
```
JohannesMessner, line 140: this part doesn't get rendered as python, idk why

docs/tutorials/external-executor.md
```diff
@@ -0,0 +1,155 @@
+# Remote executors in Flows
```
JoanFM, line 1: Do not merge `Remote` and `External`. The keyword here needs to be `External`

JoanFM (changes requested):

JoanFM (changes requested):
The focus of this documentation seems wrong.

It is good to mention about `remote` executors as `external` ones, but the main focus is to explain how to `connect` External Executors from a Flow, and how maybe 2 Flows can share an `external` executor.

The key is to explain the `external` keyword

docs/tutorials/external-executor.md
```diff
@@ -0,0 +1,155 @@
...
+Then, adding the Executor is a simple call to `Flow.add()`:
+
+```python
+from jina import Flow
+
+exec_host, exec_port = 'localhost', 12345
+f = Flow().add(host=exec_host, port_in=exec_port, external=True)
+```
+
+After that, the external Executor will behave just like a local one. And you can even add the same Executor to multiple
```
JohannesMessner, line 27: This works, but is it safe/recommended?

JohannesMessner:
The docker hub part still needs to be tested once jina 3 executors are available from there

docs/tutorials/external-executor.md
```diff
@@ -0,0 +1,164 @@
...
+docs = DocumentArray([Document(text='Embed me please!') for _ in range(5)])
+
+def print_embedding(resp):
+    doc = resp.docs[0]
+    print(f'{doc.text} has been embedded to shape {doc.embedding.shape}')
+
+with f:
+    f.index(inputs=docs, on_done=print_embedding)
+
+>>> **TODO run once Jina 3 Executors are available from the hub**
```
JohannesMessner, line 104: Waiting for the hub to have Jina 3 executors

docs/tutorials/external-executor.md
JoanFM: I will check no worries

PR #4298, opened 2022-02-14
docs: rephrase dockerize section

JohannesMessner:
WIP, do not merge yet

### DIFF
--- a/docs/fundamentals/executor/containerize-executor.md
+++ b/docs/fundamentals/executor/containerize-executor.md
@@ -1,45 +1,47 @@
 # Dockerize your Executor
 
-Once you have understood what an `Executor` is and how it can be used inside a `Flow`, you may be interested om how to wrap this Executor into a container
+Once you have understood what an `Executor` is and how it can be used inside a `Flow`, you may be interested in how to wrap this Executor into a container
 so that you can isolate its dependencies and make it ready to run in the cloud or in Kubernetes.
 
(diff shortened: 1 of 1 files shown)


docs/fundamentals/executor/containerize-executor.md
```diff
@@ -1,45 +1,47 @@
...
 ```
 
-## Jina Executor CLI as entrypoint
+## Setting Jina Executor CLI as entrypoint
 
-When a containerized Executor is run inside a Flow, what Jina does under the hood is to basically do `docker run` with some extra arguments. 
+When a containerized Executor is run inside a Flow,
+under the hood Jina basically executes `docker run` with some extra arguments.
 
 This means that Jina assumes that what runs inside the `container` is the same as it would run in a regular OS process. Therefore, you need to make sure that
```
JohannesMessner, line 46: I am not sure what this sentence means (again, I'm still a docker idiot). Would this be a fair rephrasing?

```suggestion
This means that Jina assumes that what runs inside the `container` runs just like it would run in a regular OS process. Therefore, you need to make sure that
```

JoanFM: Yes it would be correct

JohannesMessner:
Overall this feels a lot like a HowTo. Should it be moved to there?

JoanFM (approved):

PR #4305, opened 2022-02-14
docs: restructure content and layout

hanxiao:


docs/fundamentals/executor/executor-api.md
```diff
@@ -114,10 +117,13 @@ class MyExecutor(Executor):
...
 
     @requests(on='/index')
-    def bar(self, **kwargs):
-        print(kwargs)
+    async def bar(self, **kwargs):
+        await asyncio.sleep(1.0)
+        print(f'Calling bar') 
 ```
 
+For I/O bound Executors it can be helpful to utilize Python `async def`. This means we can wait for multiple pending Executor function calls concurrently.
```
JoanFM, line 125: This can be confusing I think, since most commonly, no concurrent code is run in Executors

docs/how-to/deploy.md
```diff
@@ -0,0 +1,8 @@
+# Deploy on the Cloud
+
+```{toctree}
+:maxdepth: 1
+
+docker-compose
```
JoanFM, line 6: Docker compose is not being deployed in the cloud actually.

Docker compose provides just better and smoother local packaging and running, but is technically not deploying in the cloud

JoanFM (changes requested):
I have a couple of comments

docs/fundamentals/executor/executor-api.md
hanxiao: i actually don't know what is this section is about, I personally would never use the code snippet as show in that subsection.

docs/how-to/deploy.md
hanxiao: sure i will just call it Deploy

PR #4309, opened 2022-02-14
docs: unify docarray import

bwanglzu:


JoanFM (changes requested):
Please remove the changes from `datatypes` since are not included and may confuse when real refactoring happens

PR #4319, opened 2022-02-14
feat(gateway): graphql endpoint

deepankarm:
# GraphQL support for jina

## Description
graphql

### DIFF
--- a/jina/orchestrate/flow/base.py
+++ b/jina/orchestrate/flow/base.py
@@ -142,6 +142,7 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
         native: Optional[bool] = False,
         no_crud_endpoints: Optional[bool] = False,
         no_debug_endpoints: Optional[bool] = False,
+        no_graphql_endpoint: Optional[bool] = False,
         polling: Optional[str] = 'ANY',
         port_expose: Optional[int] = None,
         port_in: Optional[int] = None,
@@ -203,6 +204,7 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
 
                   Any executor that has `@requests(on=...)` bind with those values will receive data requests.
         :param no_debug_endpoints: If set, /status /post endpoints are removed from HTTP interface.
+        :param no_graphql_endpoint: If set, /graphql endpoint is removed from HTTP interface.
         :param polling: The polling strategy of the Deployment and its endpoints (when `shards>1`).
               Can be defined for all endpoints of a Deployment or by endpoint.
               Define per Deployment:
--- a/jina/orchestrate/flow/base.py
+++ b/jina/orchestrate/flow/base.py
@@ -13,50 +13,50 @@ import uuid
 from collections import OrderedDict
 from contextlib import ExitStack
 from typing import (
-    Optional,
-    Union,
-    Tuple,
-    List,
-    Set,
-    Dict,
-    overload,
-    Type,
     TYPE_CHECKING,
+    Dict,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    Union,
+    overload,
 )
 
-from jina.orchestrate.flow.builder import allowed_levels, _hanging_deployments
 from jina import __default_host__, helper
 from jina.clients import Client
 from jina.clients.mixin import AsyncPostMixin, PostMixin
 from jina.enums import (
-    FlowBuildLevel,
     DeploymentRoleType,
+    FlowBuildLevel,
     FlowInspectType,
     GatewayProtocolType,
 )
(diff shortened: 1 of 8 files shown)


jina/serve/runtimes/gateway/http/models.py
```diff
@@ -216,7 +214,7 @@ class Config:
 
 
 def _get_example_data():
-    from jina import DocumentArray, Document
+    from jina import Document, DocumentArray
```
JoanFM, line 217:
```suggestion
    from docarray import Document, DocumentArray
```

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -227,6 +228,88 @@ async def _render_custom_swagger_html(req: Request) -> HTMLResponse:
...
+            from docarray import DocumentArray
+            from docarray.document.strawberry_type import (
+                Base64,
+                JSONScalar,
+                NdArray,
+                StrawberryDocument,
+            )
+            from strawberry.fastapi import GraphQLRouter
+
+            @strawberry.input
```
JoanFM, line 245: This needs to go in `docarray`

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -227,6 +228,88 @@ async def _render_custom_swagger_html(req: Request) -> HTMLResponse:
...
+                    description='A dictionary of parameters to be sent to the executor.',
+                )
+                exec_endpoint: str = strawberry.field(
+                    default='/search',
+                    description='The endpoint string, by convention starts with `/`. '
+                    'All executors bind with `@requests(on="/foo")` will receive this request.',
+                )
+
+            @strawberry.type
+            class Query:
```
JohannesMessner, line 288: Should this be `class Mutation` instead?
If I understand the GraphQL docs (https://graphql.org/learn/queries/#mutations) correctly, if server-side side effects can occur, a mutation should be used, not a query. Since we can target any executor endpoint, and these endpoints can do arbitrary stuff (including changing state, writing data etc), should we maybe only accept mutations instead of queries?
If i understand correctly, query is like http get, and mutation is like http post.

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -227,6 +228,88 @@ async def _render_custom_swagger_html(req: Request) -> HTMLResponse:
...
+                Jina HTTP request model.
+                """
+
+                data: Optional[List[StrawberryDocumentInput]] = strawberry.field(
+                    default=None,
+                    description='Data to send, a list of dict/string/bytes that can be converted into a list of `Document` objects',
+                )
+                target_executor: Optional[str] = strawberry.field(
+                    default=None,
+                    description='A regex string represent the certain pods/deployments request targeted.',
```
JohannesMessner, line 275:
```suggestion
                    description='A regex string representing the specific pods/deployments targeted by the request.',
```
```suggestion
                    description='A regex string represent the certain pods/deployments request targeted.',
```

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -227,6 +228,88 @@ async def _render_custom_swagger_html(req: Request) -> HTMLResponse:
...
+                )
+                parameters: Optional[JSONScalar] = strawberry.field(
+                    default=None,
+                    description='A dictionary of parameters to be sent to the executor.',
+                )
+                exec_endpoint: str = strawberry.field(
+                    default='/search',
+                    description='The endpoint string, by convention starts with `/`. '
+                    'All executors bind with `@requests(on="/foo")` will receive this request.',
```
JohannesMessner, line 284: Maybe this is clearer, but maybe it is actually more confusing to the user. I'm not sure myself.
```suggestion
                exec_endpoint: str = strawberry.field(
                    default='/search',
                    description='The endpoint string, by convention starts with `/`. '
                    'All executors bind with `@requests(on=exec_endpoint)` will receive this request.',
```
```suggestion
                exec_endpoint: str = strawberry.field(
                    default='/search',
                    description='The endpoint string, by convention starts with `/`. '
                    'All executors bind with `@requests(on="/foo")` will receive this request.',
```

jina/orchestrate/flow/base.py
```diff
@@ -1490,6 +1492,14 @@ def _get_address_table(self, address_table):
                     attrs='underline',
                 )
             )
+            address_table.append(
```
JoanFM, line 1495: check if the graphql endpoint is there right?

JoanFM (changes requested):

jina/orchestrate/flow/base.py
JohannesMessner: I think it is always there when the gateway uses http, which we check for above

PR #4338, opened 2022-02-15
docs: tweaks to 'what is neural search'

JohannesMessner:


hanxiao:
wrong base?

docs/get-started/neural-search.md
```diff
@@ -4,23 +4,24 @@ The core idea of neural search is to leverage state-of-the-art deep neural netwo
...
 Neural search creates a new way to comprehend the world. It is creating new doors that lead to new businesses. 
 
 ## Seize tomorrow today
 
-Has neural search been solved and is it widely applicable? No. Compared to traditional symbolic search, a neural search system:
+Has neural search been solved and is it widely applicable? Not quite yet - but we're working on it. Compared to traditional symbolic search,
+building a neural search system can seem daunting for the following reasons:
 - takes much more time to develop due to the complexity of AI and system engineering;
 - suffers from a fragmented tech stack and [glue code](https://en.wikipedia.org/wiki/Glue_code);
 - is computationally demanding and can be very inefficient;
```
JohannesMessner, line 27: I tried to make this sound a bit more optimistic/positive, but this still sounds a bit like building neural search is not easy, and that we haven't solve it yet, either. Is that the message we want to send? How confident are we in Jina as it is right now; are we still cautious or do we tell people that we essentially solved the problem ? @alexcg1 @hanxiao any thoughts?

hanxiao: i think it is good. this basically serves as enlightening section for people who have not heard of "neural search" before

PR #4344, opened 2022-02-15
docs: adjust docs to new .post() behaviour

JohannesMessner:


docs/fundamentals/flow/flow-api.md
```diff
@@ -224,17 +224,26 @@ executors:
 
 ## Working with the Python Client
 The most convenient way to work with the `Flow` API is the Python Client. It enables you to send `Documents` to the `Flow` API in a number of different ways shown below:
+
+```{admonition} Caution
+:class: caution
+`Flow` provides a `.post()` method that follows the same interface as `client.post()`. Due to reasons of performance and
```
JoanFM: These are not the reasons. However, when deploying in the Cloud the Flow interface will not be present

JoanFM (commented):
Make sure that `request_size` parameter is explained and considered also

docs/fundamentals/flow/client.md
```diff
@@ -0,0 +1,283 @@
...
+```
+>>> ['Hi there!']
+>>> None
+```
+
+````
+
+## Async Python Client
+
+There is also an async version of the Python Client so that it can easily be used from `asyncio` context:
```
JoanFM, line 259: Explain that the `async_client` allows a `coroutine` as input and returns an `async_iterator`

docs/fundamentals/flow/flow-api.md
```diff
@@ -363,5 +240,4 @@ with Flow() as f:
     f.to_k8s_yaml('flow_k8s_configuration')
 ```
 This will generate the necessary Kubernetes configuration files for all the `Executors` of the `Flow`. The generated folder can be used directly with `kubectl` to deploy the `Flow` to an existing Kubernetes cluster.
-More in depth information can be found in {ref}`this How-TO <docker-compos>`.
-
+More in depth information can be found in {ref}`this How-TO <docker-compos>`.
```
JoanFM, line 243: this HOW-TO should be better linked and with another name perhaps?

JoanFM (changes requested):

docs/fundamentals/flow/client.md
```diff
@@ -0,0 +1,294 @@
...
+    client.post('/endpoint', doc_gen)  # Document generator
+
+    client.post('/endpoint', DocumentArray([d1, d2]))  # DocumentArray
+
+    client.post('/endpoint')  # empty
+```
+
+## Batching Requests
+
+Especially during indexing, a Client often sends thousands of Documents to a `Flow`.
```
JoanFM, line 40:
```suggestion
Especially during indexing, a Client can send up to thousands or millions of Documents to a `Flow`.
```

docs/fundamentals/flow/client.md
```diff
@@ -0,0 +1,294 @@
...
+from docarray import Document, DocumentArray
+from jina import Flow, Client
+
+with Flow() as f:
+    client = Client(port=f.port_expose)
+    client.post('/', DocumentArray(Document() for _ in range(100)), request_size=10)
+```
+
+## Targeting a specific Executor
+Usually a `Flow` will send each request to all Executors with matching Endpoints as configured. But the `Client` also allows you to only target a specific Executor in a `Flow` using the `target_executor` keyword. The request will then only be processed by the Executor with the provided name. Its usage is shown in the listing below.
```
JoanFM, line 56:
```suggestion
Usually a `Flow` will send each request to all Executors with matching endpoints as configured. But the `Client` also allows you to only target a specific Executor in a `Flow` using the `target_executor` keyword. The request will then only be processed by the Executor with the provided name. Its usage is shown in the listing below.
```

JoanFM (commented):
Should we also explain the `return_responses` or we should not? Also we are missing docs about the `Response` object right?

JohannesMessner:
> Should we also explain the return_responses or we should not?
It is explained under 'returning results from .post()', does it need more explanation?

> Also we are missing docs about the Response object right?
I think we said we would give some broad overview of it without going into detail, which is done in the callback section. Should it get its own section? Personally I don't think that that is necessary, but maybe it is

docs/fundamentals/flow/flow-api.md
JohannesMessner: yes, thinking about a tweak

JoanFM:
> > Should we also explain the return_responses or we should not?
> 
> It is explained under 'returning results from .post()', does it need more explanation?
> 
> > Also we are missing docs about the Response object right?
> 
> I need to change that, it should say `DataRequest` and not `Response`. But I don't think we wanted to only give a broad overview, without explaining every detail

It needs to be clear what can be found in `the response` object (not sure now what is the exact class). But the user needs to understand the attributes

docs/fundamentals/flow/client.md
```diff
@@ -0,0 +1,314 @@
...
+    client = Client(port=f.port_expose)
+    client.post('/endpoint', d1)  # Single document
+
+    client.post('/endpoint', [d1, d2])  # a list of Document
+
+    client.post('/endpoint', doc_gen)  # Document generator
+
+    client.post('/endpoint', DocumentArray([d1, d2]))  # DocumentArray
+
+    client.post('/endpoint')  # empty
```
bwanglzu, line 35: i suggest us to make it clear. `Flow` is on the server side, client is in the client side, it makes no sense to put client inside of the context manager

JoanFM: yes, let's create the `client` outside`

JohannesMessner: makes sense, doing this

docs/fundamentals/flow/client.md
```diff
@@ -0,0 +1,332 @@
...
+>>> None
+```
+
+````
+
+## Async Python Client
+
+There also exists an async version of the Python Client so that it can easily be used from an `asyncio` context.
+
+While the standard `Client` is also asynchronous under the hood, its async version exposes this fact to the outside world,
```
JoanFM, line 307: what is meant by this?

JoanFM: I would mention, it allows to `asyncrhonously iterate over responses as they come`

JohannesMessner: > what is meant by this?

That the async version doesn't hide its async-ness, unlike the normal client. To be honest, I don't know all the details about this async client myself, i just kept this more or less unchanged from Tobi's version. Changed it slightly, let me know if it makes more sense to you.

JohannesMessner: Otherwise feel free to suggest and apply changes yourself of course

JoanFM: @jacobowitz what is meant with client also being async?

JoanFM (approved):

PR #4346, opened 2022-02-15
feat: impose docs and parameters as attribute for request

samsja:
# add more docstring in the Request object

## Description

When looking at the documentation, especially the docstring, of the `Request` classes there is no way to know that it has a `docs` and `parameters` attributes which are essential. This is because these attribute are access "on the fly" by overloading the `__getattr__` method and hunder the hood accessing the proto attributes which can't be found from the documentation. Even by looking at mother classes of `Request` you can't find it because it is defined in the proto files which are out of the scope of the documentation. 

This very simple PR make these two attributes part of the `Request` class, by hard-coded their existences. 

* It make things clearer in the documentation: a user can quickly use the `Request` without reading the whole tutorial because he know that there is a `docs` and `parameters` attributes
* more important: it makes this attributes searchable in in the documentation, i.e before this PR they were no way to know that the classes `Request` was linked to the `docs``by searching request and docs key-word. No it is possible and docQA can even benefice from it

### DIFF
--- a/jina/types/request/__init__.py
+++ b/jina/types/request/__init__.py
@@ -33,14 +33,22 @@ class Request(ProtoTypeMixin):
         """Get the documents attached with the Request
         :return: documents attached to the Request
         """
-        return self.docs
+        return self.proto.docs
+
+    @docs.setter
+    def docs(self, docs: 'DocumentArray'):
+        self.proto.docs = docs
 
     @property
     def parameters(self) -> Dict:
         """Get the parameters attached with the Request
         :return: parameters attached to the Request.
         """
-        return self.parameters
+        return self.proto.parameters
+
+    @parameters.setter
+    def parameters(self, parameters: Dict):
+        self.proto.parameters = parameters
 
     def add_exception(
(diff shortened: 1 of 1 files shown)


jacobowitz (commented):
I see the point, but it assumes that we actually want to advertise `request.docs`. I am not sure about if we want this? In reality the docs are in `request.data.docs` and `request.docs` is just a convenience shortcut. 
I have no particular opinion on that though.

samsja:
> I see the point, but it assumes that we actually want to advertise `request.docs`. I am not sure about if we want this? In reality the docs are in `request.data.docs` and `request.docs` is just a convenience shortcut. I have no particular opinion on that though.

As long as we talk about `request.docs` in the tutorial of the docs we assume that it exists so we should somewhere enforce this behavior. Maybe this should not be assumed for any kind of request, but then we should create a response class which inherit from request with this enforce behavior.

JoanFM (changes requested):
If u do this, then the `attributes` need to be `private`. Otherwise we will have problem setting them.

So I would suggest:

```put docs and parameters as private and writ also the setter of the property```

PR #4350, opened 2022-02-16
fix: fix override of empty docs

JoanFM:


bwanglzu (approved):
LGTM!

jacobowitz (approved):
I was running the image test case and it seems to fix the issue as far as I can see

PR #4351, opened 2022-02-16
docs: add explain on return

hanxiao:
# Pull Request Title

## Description

Please describe what your PR is doing and why. Are there any parts which need extra attention during review? Are there any dependencies from other PRs or projcts? Is this a breaking change?

Is this PR ready or work in progress (WIP)? Ready means it can be reviewed and merged from the author's perspective. If the PR is WIP: Make it a draft PR and state open questions and TODO items.

Closes # (issue)

### DIFF
--- a/docs/fundamentals/executor/executor-api.md
+++ b/docs/fundamentals/executor/executor-api.md
@@ -194,11 +194,11 @@ class MyExecutor(Executor):
 
 ### Method return
 
-Every Executor method can return 3 types: 
+Every Executor method can `return` in 3 ways: 
 
 - If you return a `DocumentArray` object, then it will be sent over to the next Executor.
 - If you return `None` or if you don't have a `return` in your method, then the original `doc` object (potentially mutated by your function) will be sent over to the next Executor.
-- If you return a `dict` object, then it will be considered as a set of parameters and sent to the next Executor. You can leverage this feature to {ref}`send config from Executor to Executor<pass-parameters>`. The original `doc` object (potentially mutated by your function) will be sent over to the next Executor.
+- If you return a `dict` object, then it will be considered as a result and passed on behind `parameters['__results__']`. The original `doc` object (potentially mutated by your function) will be sent over to the next Executor.
   
 
 ### Example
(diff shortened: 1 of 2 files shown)


docs/fundamentals/executor/executor-api.md
```diff
@@ -147,7 +147,7 @@ class MyExecutor(Executor):
     @requests
     async def foo(
         self, docs: DocumentArray, parameters: Dict, docs_matrix: List[DocumentArray]
-    ) -> Optional[Union[DocumentArray, Dict]]:
+    ) -> Union[DocumentArray, Dict, None]:
```
JoanFM, line 150:
```suggestion
    ) -> Optional[Union[DocumentArray, Dict]]:
```

This is by definition Optional and is much more standard

hanxiao: actually it is the oppsite, if you look `typing`, `pydantic` or dataclass, they all get translated into `Union[..., None]`

hanxiao: Here it is better map to what I later said, "there are 3 ways of returning", `Union[1,2,3] -> 3 ways`

docs/fundamentals/executor/executor-api.md
```diff
@@ -193,6 +192,17 @@ class MyExecutor(Executor):
 ```
 ````
 
+### Method return
+
+Every Executor method can return 3 types: 
+
+- If you return a `DocumentArray` object, then it will be sent over to the next Executor.
+- If you return `None` or if you don't have a `return` in your method, then the original `doc` object (potentially mutated by your function) will be sent over to the next Executor.
+- If you return a `dict` object, then it will be considered as a set of parameters and sent to the next Executor. You can leverage this feature to {ref}`send config from Executor to Executor<pass-parameters>`. The original `doc` object (potentially mutated by your function) will be sent over to the next Executor.
```
JohannesMessner, line 201: I don't think this works quite as described, because the `dict` will be put behind `parameters['__results__'][pea_id]`, so the next executor will still use essentially the same parameters

JoanFM: yes, sending config from Executor to Executor I would not consider recommended

docs/fundamentals/executor/executor-api.md
```diff
@@ -193,6 +192,17 @@ class MyExecutor(Executor):
 ```
 ````
 
+### Method return
+
+Every Executor method can return 3 types:
```
JohannesMessner, line 197:
```suggestion
Every Executor method can `return` in 3 ways: 
```

PR #4356, opened 2022-02-16
docs: complete containerization example

jacobowitz:
I think we should create the DA necessary for the example directly here. Otherwise the user needs to look this up in the previous example, which is awkward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,9 @@ You can containerize the Executors and play them in a sandbox thanks to [Hub](ht
     from docarray import DocumentArray
     from jina import Flow
 
-    index_data = DocumentArray.pull('demo-leftda', show_progress=True)
+    index_data = DocumentArray.pull(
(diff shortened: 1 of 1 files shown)


JoanFM (approved):

README.md
```diff
@@ -251,8 +251,11 @@ You can containerize the Executors and play them in a sandbox thanks to [Hub](ht
 
 4. In particular, sandbox hosts your Executor on Jina Cloud and allows you to play with it from local:
     ```python
+    from docarray import DocumentArray
     from jina import Flow
 
+    index_data = DocumentArray.pull('demo-leftda', show_progress=True)
```
hanxiao, line 257:
```suggestion
    index_data = DocumentArray.pull('demo-leftda', show_progress=True)   # write something like, you should have it tho by following the tutorial, but more polite
```

PR #4364, opened 2022-02-17
fix: broken link in docstring

bwanglzu:
fixed topology link error, affected docs when checking `make linkcheck`


JoanFM:
And in docs does this happen?

JoanFM (approved):

PR #4373, opened 2022-02-18
feat: add conditions to flow topology

JoanFM:


bwanglzu (commented):
how do you think replace:
 `graph_condition` -> `graph_mask` and `condition` -> `mask`, let's say you want to only keep `Document` with `key=5`, I'll do `mask key!=5`. 
 I feel this way is more natural.

JoanFM:
> how do you think replace: `graph_condition` -> `graph_mask` and `condition` -> `mask`, let's say you want to only keep `Document` with `key=5`, I'll do `mask key!=5`. I feel this way is more natural.

Let's see how the `Query Language` in `DocumentArray` ends up being and we can decide? I think this would heavily depend on how it is implemented there. What do u think @bwanglzu ?

JohannesMessner:
The actual code should be ready, so feel free to review.
However, it is not ready to merge yet, since:
- QL feature is not really released in DocArray, so can't link to their docs
- Tutorial is referenced in the docs but not done yet

docs/fundamentals/flow/create-flow.md
```diff
@@ -453,9 +453,172 @@ This will get you the following output:
...
+[]
+```
+````
+
+This feature is useful to prevent some specialized Executors from processing certain Documents.
+It can also be used to build *switch-like nodes*, where some Documents pass through one parallel branch of the Flow,
+while other Documents pass through a different branch.
+
+Also note that whenever a Document does not satisfy the condition of an Executor, it will not even be sent to that Executor.
+This means that you can not only use this feature to build complex logic, but also to minimize your networking overhead.
```
JoanFM, line 616: this will not remain true right?

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -25,6 +26,9 @@ def _handle_request(request: 'Request') -> 'asyncio.Future':
 
         request_graph = copy.deepcopy(graph)
         # important that the gateway needs to have an instance of the graph per request
+        request_doc_ids = request.data.docs[
```
JoanFM, line 29: We need to detect if there is a condition in the `graph` before doing it.

Doing serialization is rather `expensive` and I would rather avoid this if not needed

JoanFM (commented):
Just a minor thing

docs/fundamentals/flow/create-flow.md
JohannesMessner: Added a clarification

PR #4375, opened 2022-02-18
fix: missing import in fashion example

maloyan:


JoanFM (approved):

JoanFM:
Thank you very much for the contribution @maloyan

Issue #4367, opened 2022-02-17
Unify port_expose and port_in


jacobowitz:
We have `port_expose` and `port_in` arguments for historical reasons, but we don't have the need anymore to distinguish between the two. Every Jina Pod only exposes a single port.
Thus we should unify the two arguments into a new argument, which is only called `port`. This needs to be done in a backwards compatible fashion, so make sure to deprecate the old arguments, but dont remove them.

This feature will go in the next release, so dont update the docs before the next release.

What needs to be done in this issue:

- [x] Add the new `port` argument and make sure it is used everywhere instead of `port_in` and `port_expose`. Make sure that docker-compose and K8s configuration generation is correct and does not generate the deprecated `port_in` and `port_expose`.
- [x] Make `port_in` and `port_expose` set the new `port` argument only and deprecate them.
- [x] Update all the tests defining port arguments
- [x] Prepare a separate PR against the docs to replace the port usage there. But only merge this AFTER the next release (to not update the docs prematurely) (Make sure that the docs state from which version the usage changed)

PR #4382, opened 2022-02-21
refactor: unify port args

jacobowitz:
# Unify port arguments

## Description

This PR unifies the `port_expose` and `port_in` arguments into a new `port` argument. Every jina instance only exposes a single port so there is no need for multiple port arguments. This also simplifies the docker-compose and k8s yaml files a bit. Before we head unnecessary parameters there.
I also change the default K8s port assignments. Before we used 8080 for port_expose and 8081 worker port_in. With this change it is always 8080 for `port. The ports for uses_before/uses_after are changed to 8081/8082.

The failing jinad test seems unrelated

Closes https://github.com/jina-ai/jina/issues/4367


JoanFM:
Make sure to add `docs` and make sure to say it works from `3.1`

jacobowitz:
rebased against master, need to double check the tests now

JoanFM (approved):

PR #4388, opened 2022-02-22
optimize create hub root

mapleeit:
- fix(hubble): do not create hub root folder globally
- docs(executor): add explanation for environment variables


docs/fundamentals/executor/hub/index.md
```diff
@@ -35,6 +35,15 @@ The Hub architecture looks like the following:
 :align: center
 ```
 
+## Environment Variables
+
+A list of environment variables which takes effects during Jina Hub operations. e.g. `jina hub push`
+
+### `JINA_HUB_ROOT`
+
+**Define the place where the Executor package cache lives.** Default value is `Path.home().joinpath('.jina', 'hub-packages')`
```
JoanFM, line 44: Explain that you may want to `change` to make sure the right permissions are there

JoanFM (changes requested):

mapleeit:
@JoanFM Comments addressed. PTAL

PR #4390, opened 2022-02-22
docs: update google analytics tracking ID

Roshanjossey:
This is done to differenciate traffic from docs2.jina.ai and
docs.jina.ai


alexcg1:
I'd suggest change the id for docs2. That way moving forwards we can compare doc.jina.ai with the performance we previously had with docs2 (pre-v3 launch). Gives us more data to work with

Roshanjossey:
I changed it for docs.jina.ai

I figured it made sense with the current naming we have in analytics. This will clear the confusion we have right now

JoanFM (approved):

PR #4399, opened 2022-02-24
refactor: clean up helloworld

nan-wang:
# Pull Request Title

## Description

- clean up docstrings
- update the YAML APIs

Closes # (issue)

### DIFF
--- a/jina/helloworld/multimodal/flow-index.yml
+++ b/jina/helloworld/multimodal/flow-index.yml
@@ -2,7 +2,7 @@ jtype: Flow
 version: '1'
 py_modules:
   - ${{ ENV.PY_MODULE }}
-deployments:
+executors:
   - name: segment
     uses: Segmenter
     uses_metas:
--- a/jina/helloworld/multimodal/flow-index.yml
+++ b/jina/helloworld/multimodal/flow-index.yml
@@ -1,64 +1,39 @@
 jtype: Flow
 version: '1'
-executors:
+py_modules:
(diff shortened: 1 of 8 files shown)


jina/helloworld/multimodal/flow-index.yml
```diff
@@ -1,64 +1,39 @@
 jtype: Flow
 version: '1'
-executors:
+py_modules:
+  - ${{ ENV.PY_MODULE }}
+deployments:
```
JoanFM, line 5: this should remain executors, we do not expose deployments

JoanFM (changes requested):

jina/helloworld/multimodal/flow-index.yml
```diff
@@ -1,64 +1,39 @@
...
   - name: textIndexer
-    uses:
-      jtype: DocVectorIndexer
-      with:
-        index_file_name: "text.json"
-      metas:
-        workspace: ${{ ENV.HW_WORKDIR }}
-        py_modules:
-          - ${{ ENV.PY_MODULE }}
+    uses: DocVectorIndexer
```
JoanFM, line 15: I prefer to have the style of `using` `with` and `metas` than the `uses_with` and `uses_metas`. 

I prefer to have each Executor as its own YAML config to be honest. `uses_with` and `uses_metas` is kind of a way to override but the other one should be preferred

JoanFM (changes requested):

PR #4403, opened 2022-02-24
feat(gateway): add graphql support

JohannesMessner:
# GraphQL support for Jina

Adds a single endpoint for GraphQL, through which Executor endpoints can be reached.
Also adds GraphQL support to the Client.

Depends on this: https://github.com/jina-ai/docarray/pull/141

Migrated from #4319

Please look extra carefully at the async client, honestly I don't know what I'm doing in that regard, so it might be nonsense.


JoanFM (commented):
We need to investigate if there is any easy way for the `Python client` to send this schema

docs/fundamentals/flow/flow-api.md
```diff
@@ -230,6 +231,58 @@ executors:
...
+:class: seealso
+
+This article does not serve as the introduction to GraphQL.
+If you are not already familiar with GraphQL, we recommend you learn more about GraphQL from the official [GraphQL documentation](https://graphql.org/learn/).
+You may also want to learn about [Strawberry](https://strawberry.rocks/), the library that powers Jina's GraphQL support.
+````
+
+Jina Flows that use the HTTP protocol provide a GraphQL API out of the box, which is located behind the '/graphql' endpoint.
+
+This means that you can access the Flow from any GraphQL client, like for example, `sgqlc`.
```
JoanFM, line 246: I would also explain here the benefit. About limiting the schema or defining the schema. Compare the result with one or the other I'd say?

tests/integration/graphql/test_graphql.py
```diff
@@ -77,13 +77,33 @@ def test_id_only(req_type):
...
+def test_data_list(req_type):
+    texts = 'abcd', 'efgh', 'ijkl'
+    response = graphql_query(
+        '''
+        %s {
+            docs(data: [{text: "%s"}, {text: "%s"}, {text: "%s"}]) { 
+                text
+            } 
+        }
+    '''
```
JoanFM, line 96: can u use `f'{}` style better? the codebase is more used to that syntax

JohannesMessner: I would love to, but f-strings don't seem to like strings that have `"` inside of them. Maybe there is a way, but couldn't get it to work.

JohannesMessner:
Please look extra carefully at the async client, honestly I don't know what I'm doing in that regard, so it might be nonsense.

jina/clients/base/http.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 from contextlib import nullcontext, AsyncExitStack
+from sgqlc.endpoint.http import HTTPEndpoint as SgqlcHTTPEndpoint
```
JoanFM, line 3: This import needs to be lazy, we are not sure this will be installed in every client.

jina/clients/base/http.py
```diff
@@ -19,6 +20,45 @@
...
+
+        :param mutation: the GraphQL mutation as a single string.
+        :param variables: variables to be substituted in the mutation. Not needed if no variables are present in the mutation string.
+        :param timeout: HTTP request timeout
+        :param headers: HTTP headers
+        :return: dict containing the optional keys ``data`` and ``errors``, for response data and errors.
+        """
+        proto = 'https' if self.args.https else 'http'
+        graphql_url = f'{proto}://{self.args.host}:{self.args.port}/graphql'
+        endpoint = SgqlcHTTPEndpoint(graphql_url)
```
JoanFM, line 40: Do import lazy, and make use of `ImportExtensions` to give proper message when dependency not found

jina/clients/base/http.py
```diff
@@ -19,6 +20,45 @@
 class HTTPBaseClient(BaseClient):
     """A MixIn for HTTP Client."""
 
+    def _graphql_mutate(
+        self,
```
JoanFM, line 24: These implementations should be directly inside the `mutate` in the `Mixins`

JoanFM (changes requested):

jina/clients/mixin.py
```diff
@@ -28,6 +30,50 @@ def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
...
+
+class AsyncMutateMixin:
+    """The GraphQL Mutation Mixin for Client and Flow"""
+
+    async def mutate(
+        self,
+        mutation: str,
+        variables: dict = None,
+        timeout: float = None,
+        headers: dict = None,
```
JoanFM, line 62: have proper hinting, `Opional` should be the case

JoanFM (changes requested):

extra-requirements.txt
```diff
@@ -81,3 +82,4 @@ trimesh:                    cicd
 paddlepaddle==2.2.0:        cicd
 onnx:                       cicd
 onnxruntime:                cicd
+sgqlc:                      cicd, graphql
```
JohannesMessner, line 85: Is it a good idea to introduce this tag, or should i revert this?

JoanFM: Let's not add it for the moment, as long as ImportExtensions is properly handled, it is okey

JoanFM (changes requested):

jina/clients/mixin.py
```diff
@@ -29,6 +31,68 @@ def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
...
+            proto = 'https' if self.args.https else 'http'
+            graphql_url = f'{proto}://{self.args.host}:{self.args.port}/graphql'
+            endpoint = SgqlcHTTPEndpoint(graphql_url)
+            res = endpoint(
+                mutation, variables=variables, timeout=timeout, extra_headers=headers
+            )
+            return res
+
+
+class AsyncMutateMixin(MutateMixin):
```
deepankarm, line 64: Should it inherit from another mixin?

jina/clients/mixin.py
```diff
@@ -29,6 +31,68 @@ def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
...
+    """The async GraphQL Mutation Mixin for Client and Flow"""
+
+    async def _async_mutate(
```
deepankarm, line 67: Where is this used?

jina/clients/mixin.py
```diff
@@ -29,6 +31,68 @@ def _include_results_field_in_param(parameters: Optional['Dict']) -> 'Dict':
...
+    ):
+        """Perform a GraphQL mutation, asynchronously
+
+        :param mutation: the GraphQL mutation as a single string.
+        :param variables: variables to be substituted in the mutation. Not needed if no variables are present in the mutation string.
+        :param timeout: HTTP request timeout
+        :param headers: HTTP headers
+        :return: dict containing the optional keys ``data`` and ``errors``, for response data and errors.
+        """
+        return await asyncio.get_event_loop().run_in_executor(
```
deepankarm, line 91: Let's use `jina.helper.get_or_reuse_loop()` here instead of `asyncio.get_event_loop()`.

jina/orchestrate/flow/base.py
```diff
@@ -1497,6 +1499,14 @@ def _get_address_table(self, address_table):
                     attrs='underline',
                 )
             )
+            address_table.append(
```
deepankarm, line 1502: Add this according to `no_graphql_endpoint`

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -227,6 +228,72 @@ async def _render_custom_swagger_html(req: Request) -> HTMLResponse:
...
+                    request_generator(**req_generator_input)
+                )
+                return DocumentArray.from_dict(response['data']).to_strawberry_type()
+
+            @strawberry.type
+            class Mutation:
+                @strawberry.mutation
+                async def docs(
+                    self,
+                    data: Optional[List[StrawberryDocumentInput]] = None,
```
deepankarm, line 271: Shouldn't these be defined as a `StrawberryRequestModel` of type `input` & used here?

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -227,6 +228,72 @@ async def _render_custom_swagger_html(req: Request) -> HTMLResponse:
...
+                    target_executor: Optional[str] = None,
+                    parameters: Optional[JSONScalar] = None,
+                    exec_endpoint: str = '/search',
+                ) -> List[StrawberryDocument]:
+                    return await get_docs_from_endpoint(
+                        data, target_executor, parameters, exec_endpoint
+                    )
+
+            @strawberry.type
+            class Query:
```
deepankarm, line 281: Why do we still need a `Query`?

jina/clients/mixin.py
JoanFM: this can be removed I think

jina/orchestrate/flow/base.py
JoanFM: `no_graphql_endpoint` then needs to be passed to the `Flow` and not only the `gateway` @JohannesMessner this should be seen in the `parser`

jina/clients/mixin.py
JohannesMessner: I thought about this, too, but without inheritance it is completely duplicate and easy to forget on if you change the other. The alternative is having a base class from which both mixins inherit, but that seems more convoluted

jina/orchestrate/flow/base.py
JohannesMessner: That created some issues for @JoanFM and we decided to remove it. I wasn't able to replicate that problem, so I don't know.

jina/serve/runtimes/gateway/http/app.py
JohannesMessner: With strawberry you always have to have a Query, othws it doesn't work. Not sure if that's just strawberry or GraphQL in general, but we're stuck with it either way. So now query is not mentioned anywhere in the docs, but if someone tries it just does the same thing as mutation.

jina/serve/runtimes/gateway/http/app.py
JohannesMessner: Yes we could couple all inputs into one input type, but I see two issues with that:

1. From the user perspective, they now have to put all of their inputs into this type when defining their mutation. This creates unnecessary nesting. So instead of being able to do
```
mutation {docs(data: {...}, execEndpoint: "/foo", ...)
                ...            
                }
```
they would have to to something like this:
```
mutation {docs(inputType: {data: {...}, execEndpoint: "/foo", ...})
                ...            
                }
```

2. I think the above makes sense if you are grouping things together that are conceptually part of the same object, like in the strawberry example, the x an y coordinates of a point. But our inputs don't belong to the same object, so I believe that they should be separate and on the top level.

jina/orchestrate/flow/base.py
JoanFM: The problem was at which level the `args` was described. If you fix that, it would be good for me

jina/serve/runtimes/gateway/http/app.py
deepankarm: Makes sense, then.

JoanFM (dismissed):

docs/fundamentals/flow/flow-api.md
```diff
@@ -259,6 +261,9 @@ f = Flow(cors=True, protocol='http')
 
 ## Generate deployment configuration
 
+=======
```
JoanFM, line 264: fix this merging issue

extra-requirements.txt
```diff
@@ -72,3 +73,10 @@ portforward>=0.2.4:         cicd
 kubernetes>=18.20.0:        perf, standard, devel
 pytest-kind==21.1.3:        test
 pytest-lazy-fixture:        test
+datasets:                   cicd
```
JoanFM, line 76: wrong merging

jina/clients/mixin.py
```diff
@@ -1,9 +1,12 @@
 from functools import partialmethod, wraps
+import asyncio
+from sgqlc.endpoint.http import HTTPEndpoint as SgqlcHTTPEndpoint
```
JoanFM, line 3: this `import` is dangerous. This should be hidden with the `ImportExtensions` in a lazy way

JoanFM, line 3:
```suggestion
```

JoanFM (changes requested):

JohannesMessner:
I hate rebasing 🙃

extra-requirements.txt
```diff
@@ -73,10 +73,3 @@ portforward>=0.2.4:         cicd
 kubernetes>=18.20.0:        perf, standard, devel
 pytest-kind==21.1.3:        test
 pytest-lazy-fixture:        test
-datasets:                   cicd
-av:                         cicd
-trimesh:                    cicd
-paddlepaddle==2.2.0:        cicd
-onnx:                       cicd
-onnxruntime:                cicd
-sgqlc:                      cicd
```
JoanFM: sgqlc does indeed need to be added

JoanFM (changes requested):

extra-requirements.txt
JohannesMessner: f*#$k

docs/fundamentals/flow/flow-api.md
```diff
@@ -238,6 +238,8 @@ with Flow(prefetch=2).add(uses=MyExecutor) as f:
     f.post(on='/', inputs=requests_generator)
 ```
 
+<<<<<<< HEAD
```
JoanFM, line 241:
```suggestion
```

docs/fundamentals/flow/flow-api.md
```diff
@@ -238,6 +238,8 @@ with Flow(prefetch=2).add(uses=MyExecutor) as f:
...
+<<<<<<< HEAD
```
JoanFM, line 242:
```suggestion
```

JoanFM (approved):

Issue #4391, opened 2022-02-23
`on_error` broken for Client with unreachable server


deepankarm:
```python
from jina import Client, DocumentArray


def on_error(response):
    print('on_error called')
    print(response.to_dict())


for protocol in ['http', 'grpc', 'websocket']:
    print(f'\n\n\n\nprotocol: {protocol}')
    try:
        Client(host='0.0.0.0', protocol=protocol, port=12345).post(
            '/blah',
            inputs=DocumentArray.empty(10),
            on_error=on_error,
        )
    except Exception as e:
        print(f'protocol {protocol} raised \n{e}')
```

**HTTP** 
- No exceptions raised. 
- `on_error` is not invoked.

**gRPC**
- Exception raised.
- `on_error` is not invoked.

**Websocket**
- No exceptions raised. 
- `on_error` is not invoked.

----

```
protocol: http
     HTTPClient@1909863[E]:Error while fetching response from HTTP server ClientConnectorError(ConnectionKey(host='0.0.0.0', port=12345, is_ssl=False, ssl=None, proxy=None, proxy_auth=None, proxy_headers_hash=None), ConnectionRefusedError(111, 'Connection refused'))




protocol: grpc
     GRPCClient@1909863[E]:gRPC error: StatusCode.UNAVAILABLE failed to connect to all addresses
the ongoing request is terminated as the server is not available or closed already
protocol grpc raised 
<AioRpcError of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "failed to connect to all addresses"
        debug_error_string = "{"created":"@1645593433.556564109","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3134,"referenced_errors":[{"created":"@1645593433.556563329","description":"failed to connect to all addresses","file":"src/core/lib/transport/error_utils.cc","file_line":163,"grpc_status":14}]}"
>




protocol: websocket
WebSocketClient@1909863[E]:Error while streaming response from websocket server ClientConnectorError(ConnectionKey(host='0.0.0.0', port=12345, is_ssl=False, ssl=None, proxy=None, proxy_auth=None, proxy_headers_hash=None), ConnectionRefusedError(111, 'Connection refused'))
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f9159c2e710>
```

PR #4410, opened 2022-02-25
fix: fix client error 4391

samsja:
fixing https://github.com/jina-ai/jina/issues/4391

New behavior:
if an exception is raised then :

- [x] if there is an on_error callback then the callback is called and the original exception is passed to the on_error
- [x] Otherwise we just throw an ClientError which wrap the actual exception
- [x] docs

### DIFF
--- a/docs/fundamentals/executor/executor-in-flow.md
+++ b/docs/fundamentals/executor/executor-in-flow.md
@@ -325,7 +325,7 @@ class MyExecutor(Executor):
 f = Flow().add(uses=MyExecutor)
 
 
-def print_why(resp):
+def print_why(resp, exception):
     print(resp.status.description)
 
 
--- a/jina/clients/base/grpc.py
+++ b/jina/clients/base/grpc.py
@@ -6,7 +6,7 @@ import grpc
 
 from jina.clients.base import BaseClient
 from jina.clients.helper import callback_exec, callback_exec_on_error
-from jina.excepts import BadClient, BadClientInput
+from jina.excepts import BadClient, BadClientInput, BaseJinaException
 from jina.logging.profile import ProgressBar
 from jina.proto import jina_pb2_grpc
 from jina.serve.networking import GrpcConnectionPool
@@ -68,19 +68,8 @@ class GRPCBaseClient(BaseClient):
             my_code = rpc_ex.code()
             my_details = rpc_ex.details()
             msg = f'gRPC error: {my_code} {my_details}'
-            if on_error or on_always:
-                if on_error:
-                    callback_exec_on_error(on_error, rpc_ex, self.logger)
-                if on_always:
-                    callback_exec(
-                        response=None,
-                        on_error=None,
-                        on_done=None,
-                        on_always=on_always,
-                        continue_on_error=self.continue_on_error,
-                        logger=self.logger,
-                    )
-            else:
+
+            try:
                 if my_code == grpc.StatusCode.UNAVAILABLE:
                     self.logger.error(
                         f'{msg}\nthe ongoing request is terminated as the server is not available or closed already'
@@ -100,3 +89,22 @@ class GRPCBaseClient(BaseClient):
                     ) from rpc_ex
                 else:
                     raise BadClient(msg) from rpc_ex
+
+            except (
+                grpc.aio._call.AioRpcError,
+                BaseJinaException,
+            ) as e:  # depending on if there are callbacks we catch or not the exception
+                if on_error or on_always:
+                    if on_error:
+                        callback_exec_on_error(on_error, e, self.logger)
+                    if on_always:
+                        callback_exec(
+                            response=None,
+                            on_error=None,
+                            on_done=None,
+                            on_always=on_always,
+                            continue_on_error=self.continue_on_error,
+                            logger=self.logger,
+                        )
+                else:
+                    raise e
--- a/jina/clients/base/grpc.py
+++ b/jina/clients/base/grpc.py
@@ -68,8 +68,18 @@ class GRPCBaseClient(BaseClient):
             my_code = rpc_ex.code()
             my_details = rpc_ex.details()
             msg = f'gRPC error: {my_code} {my_details}'
-            if on_error:
-                callback_exec_on_error(on_error, rpc_ex, self.logger)
+            if on_error or on_always:
+                if on_error:
+                    callback_exec_on_error(on_error, rpc_ex, self.logger)
+                if on_always:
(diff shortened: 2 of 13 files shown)


tests/unit/clients/python/test_on_err.py
```diff
@@ -43,3 +49,42 @@ def validate(x):
...
+
+    on_error = OnError()
+
+    Client(host='0.0.0.0', protocol=protocol, port=12345).post(
+        '/blah',
+        inputs=DocumentArray.empty(10),
+        on_error=on_error,
+    )
+
+    assert on_error.is_called
```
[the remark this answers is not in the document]
samsja: was not None you mean ? If there are no `exception` then on_error won't be called

jina/clients/base/grpc.py
```diff
@@ -68,22 +68,25 @@ async def _get_results(
...
-            elif (
-                my_code == grpc.StatusCode.UNKNOWN
-                and 'asyncio.exceptions.TimeoutError' in my_details
-            ):
-                raise BadClientInput(
-                    f'{msg}\n'
-                    'often the case is that you define/send a bad input iterator to jina, '
-                    'please double check your input iterator'
-                ) from rpc_ex
+            if on_error:
```
JoanFM, line 71: I would still suggest to print the `log errors`

JoanFM (changes requested):
Please make sure to update `documentation` stating the change of behavior from `3.1`

jina/clients/helper.py
```diff
@@ -81,3 +81,31 @@ def callback_exec(
...
+
+    >>> def on_error(response, e = None):
+    >>>      print(e)
+    >>>      print(response.docs[0])
+    >>>
+    >>> except aiohttp.ClientError as e:
+    >>> callback_exec_on_error(on_error, e, logger)
+    """
+
+    on_error_wrap = lambda resp: on_error(resp, exception)
```
JoanFM, line 109: for better backward compatibility, I would suggest to inspect if `on_error` accepts 2 parameters, if only 1, pass only resp.

JoanFM (changes requested):

jina/clients/helper.py
```diff
@@ -81,3 +81,31 @@ def callback_exec(
...
+    :param logger: a logger instance
+
+    Example:
+
+    >>> def on_error(response, e = None):
+    >>>      print(e)
+    >>>      print(response.docs[0])
+    >>>
+    >>> except aiohttp.ClientError as e:
+    >>> callback_exec_on_error(on_error, e, logger)
```
nan-wang, line 106: The example codes is not complete. Please make a good example or leave it blank.
```suggestion
   .. highlight:: yaml
   .. code-block:: yaml
         def on_error(response, e = None):
              print(e)
              print(response.docs[0])
  
          except aiohttp.ClientError as e:
         callback_exec_on_error(on_error, e, logger)
```

jina/clients/helper.py
```diff
@@ -81,3 +81,31 @@ def callback_exec(
...
+    exception: Exception,
+    logger: JinaLogger,
+    response: Optional = None,
+) -> None:
+    """Execute the on_error callback with the response, Use when anerror outside the responde status was thrown.
+
+    :param response: the response
+    :param on_error: the on_error callback
+    :param exception: the exception with was thrown and led to the call of on_error
+    :param logger: a logger instance
```
nan-wang, line 97: the order of the docstrings is wired

nan-wang (dismissed):

jina/clients/helper.py
samsja: Do you mean that I should add the import etc ?

jina/clients/helper.py
```diff
@@ -81,3 +83,29 @@ def callback_exec(
...
+
+    if len(signature(on_error).parameters) == 1:
+        on_error_wrap = on_error
+        warnings.warn(
+            "on_error callback taking only the response parameters is deprecated. Please add one parameter "
+            "to handle additional optional Exception as well",
+            DeprecationWarning,
+        )
+    else:
+        on_error_wrap = lambda resp: on_error(resp, exception)
```
JoanFM, line 109: I would like this to be done only once? Can we wrap the `on_error` method right when the `Client.post` is done? Like this we will do this only once for all the requests and not once per request?

What I mean, override the `on_error` by the wrap` thing that adds the extra `kwargs`

JoanFM (changes requested):

jina/clients/mixin.py
```diff
@@ -167,3 +170,23 @@ async def post(
...
+
+    num_args = len(signature(on_error).parameters)
+    if num_args == 1:
+        warnings.warn(
+            "on_error callback taking only the response parameters is deprecated. Please add one parameter "
+            "to handle additional optional Exception as well",
+            DeprecationWarning,
+        )
+
+        @wraps(on_error)
```
JoanFM, line 185: I am not sure u need this extra decoration? you are already wrapping it by returning `_fn`

jina/excepts.py
```diff
@@ -1,93 +1,97 @@
...
     """Raised when the docker version is incompatible"""
 
 
-class DaemonInvalidDockerfile(FileNotFoundError, BaseJinaExeception):
+class DaemonInvalidDockerfile(FileNotFoundError, BaseJinaException):
     """Raised when invalid dockerfile is passed to JinaD"""
 
 
-class NoContainerizedError(Exception, BaseJinaExeception):
+class NoContainerizedError(Exception, BaseJinaException):
```
JoanFM, line 96: Are we sure? not sure if it is safe from multi inheritance p.o.v

jina/clients/mixin.py
JoanFM: Ok I see, this way it is more transparent!

jina/excepts.py
samsja: I added a deprecation warning so the old one is still usable

JoanFM (approved):

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -325,7 +325,7 @@ class MyExecutor(Executor):
 f = Flow().add(uses=MyExecutor)
 
 
-def print_why(resp):
+def print_why(resp, exception: None):
```
JoanFM, line 328: exception: None seems wrong

JoanFM (changes requested):

JoanFM (approved):

PR #4430, opened 2022-03-03
fix: dont use latest docarray

jacobowitz:


jacobowitz:
will be fixed in docarray

PR #4433, opened 2022-03-04
feat: Linkerd as default service mesh

jacobowitz:
# feat: use Linkerd as default service mesh

## Description

This adds the necessary Linkerd annotations to our YAML templates. Also all code around the `K8sGrpcConnectionPool` is removed. This includes the code itself, unnecessary arguments, K8s resources and tests.

- [x] Remove K8sGrpcConnection pool and all logic around it
- [x] Fix PyKind + LinkerD setup in CI
- [x] Add Linkerd annotation
- [x] Adjust the docs and explain service mesh usage
- [ ] ~~Optional: Give the user a choice betwen istio and linkderd~~

Closes https://github.com/jina-ai/jina/issues/4438


JoanFM:
@jacobowitz we will be able to remove `k8s` as core requirement then also right?

jacobowitz:
I changed the dependency on `kubernetes` to `test`, because we need it actually there for some test setup.

tests/k8s/conftest.py
```diff
@@ -19,10 +21,44 @@ def __init__(self, kind_cluster: KindCluster, logger: JinaLogger):
...
+        if returncode is not None and returncode != 0:
+            raise Exception(f"Installing linkerd failed with {returncode}")
+
+        self._log.info('check linkerd status')
+        out = subprocess.check_output(
+            [f'{Path.home()}/.linkerd2/bin/linkerd', 'check'],
+            env=os.environ,
+        )
+
+        print(f'linkerd check yields {out.decode() if out else "nothing"}')
```
jacobowitz, line 58: fyi: thats intentional, because the output hier is some specially formatted, multi line output. Our logger does not print it correctly so I use print here. Its just to understand what the test is doing so I think its not a big deal

docs/how-to/kubernetes.md
```diff
@@ -219,6 +219,18 @@ Once the Flow is deployed on Kubernetes, you can use all the native Kubernetes t
 
 You can use this to [add or remove replicas](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#scaling-a-deployment), to run [rolling update](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#updating-a-deployment) operations, etc ...
 
+## Integration with Linkerd
+Jina's Kubernetes support comes with a built-in integration for [Linkerd](https://linkerd.io/) as the service mesh for connecting all the Jina services in Kubernetes. All Jina deployment resources are annotated with the necessary [Linkerd annotations](https://linkerd.io/2.11/tasks/adding-your-service/#meshing-a-service-with-annotations). So we recommend every Jina user to [install Linkerd](https://linkerd.io/2.11/getting-started/#step-3-install-the-control-plane-onto-your-cluster) into the Kubernetes cluster they are using Jina in.
+
```
JohannesMessner, line 224: This has less to do with your PR and is more of a general question: It seems to me like k8s (and now also service meshes) is really what we recommend to the pros to use, so seems like a really important part of Jina.

Therefore I think that most of the k8s related info should go into the user guide, and quite prominently i would say, instead of being a bit hidden away in this how-to.

JoanFM: I would agree @JohannesMessner , any recommendation or suggestion?

JohannesMessner: I would say we either create a page under `Flow/Deploy`, so on the same level as `create a Flow`, or if we have more content we could even create a section on the same level as `Flow` and `Executor`.
image


This can be done independently of this PR, I will create a ticket and discuss with the unofficial docs task force ;) @samsja @cristianmtr

samsja: I agree with @JohannesMessner we need to highlight the k8s feature in the docs by putting it in the right section

docs/how-to/kubernetes.md
```diff
@@ -219,6 +219,18 @@ Once the Flow is deployed on Kubernetes, you can use all the native Kubernetes t
...
+Linkerd will automatically load balance traffic between all replicas of an Executor and offers also [additional features](https://linkerd.io/2.11/features/).
```
cristianmtr, line 225:
```suggestion
Linkerd will automatically load balance traffic between all replicas of an Executor. It also has [additional features](https://linkerd.io/2.11/features/).
```

cristianmtr (changes requested):

docs/how-to/kubernetes.md
JoanFM: Agree, feel free to discuss in the scout of Documentation, but yes agree on the idea

jacobowitz: very true, the K8s section is not really a how to anymore. But lets do in a separate issue

jacobowitz:
rebased due to the compression PR

JoanFM (approved):

PR #4445, opened 2022-03-07
feat(hubble): fetch image only when required

### LINKED ISSUE #4170: Don't fetch docker image if only use `jinahub://`
**Describe the feature**
Right now, it will always try to get the docker image even using `jinahub://MyExecutor`. This can be optimized.

**Your proposal**
Only fetching the source code when using `jinahub://MyExecutor`.
Feching the docker image when using `jinahub+docker://MyExecutor`

delgermurun:
Closes #4170


jina/hubble/hubio.py
```diff
@@ -674,9 +678,10 @@ def deploy_public_sandbox(args: Union[argparse.Namespace, Dict]) -> str:
                     headers=get_request_header(),
                 ).json()
 
-                host = json_response.get('data', {}).get('host', None)
-                port = json_response.get('data', {}).get('port', None)
-                livetime = json_response.get('data', {}).get('livetime', '15 mins')
+                data = json_response.get('data') or {}
```
JoanFM, line 681:
```suggestion
                data = json_response.get('data', {})
```

JoanFM (changes requested):

jina/hubble/hubio.py
```diff
@@ -559,13 +559,16 @@ def _get_prettyprint_usage(self, console, executor_name, usage_kind=None):
     def fetch_meta(
         name: str,
         tag: str,
+        *,
         secret: Optional[str] = None,
+        image_required=True,
```
JoanFM, line 564:
```suggestion
        image_required: bool = True,
```

jina/hubble/hubio.py
delgermurun: this is intended change.
In case of error, `json_response` is `{'data': None, ...}`, then `data` variable becomes `None` instead of `{}`.

JoanFM: okk

JoanFM (approved):

PR #4460, opened 2022-03-10
docs(clients): update docstrings

nan-wang:
# Pull Request Title

## Description

update docstrings for `jina/clients`

Closes # (issue)


jina/clients/mixin.py
```diff
@@ -184,20 +183,23 @@ async def post(
         continue_on_error: bool = False,
         **kwargs,
     ) -> AsyncGenerator[None, 'Response']:
```
JoanFM, line 185: Can u check if the type hint is correct?

PR #4461, opened 2022-03-10
fix: graphql dependencies

JohannesMessner:
BREAKING CHANGE:
`no_graphql_endpoint` is removed in favor of `expose_graphql_endpoint`. Now the defaut is not to have a graphql endpoint


extra-requirements.txt
```diff
@@ -73,4 +73,5 @@ portforward>=0.2.4:         cicd
 kubernetes>=18.20.0:        perf, standard, devel
 pytest-kind==21.1.3:        test
 pytest-lazy-fixture:        test
-sgqlc:                      cicd
\ No newline at end of file
+sgqlc:                      cicd, graphql
+docarray>=0.8.8:            cicd, graphql
```
JoanFM, line 77: remove this, otherwise `docarray` will start getting crazy

PR #4464, opened 2022-03-10
fix(hubble): use commit id instead of serial number

delgermurun:
`sn` no longer exists instead now we use commits to manage the executor version.

Closes https://github.com/jina-ai/hubble/issues/299


JoanFM (approved):

numb3r3 (approved):
LGTM 👍

PR #4465, opened 2022-03-11
refactor: delete daemon args from pods

samsja:


jina/orchestrate/pods/__init__.py
```diff
@@ -340,6 +339,7 @@ def __init__(self, args: 'argparse.Namespace'):
                 else None,
                 'runtime_cls': self.runtime_cls,
                 'jaml_classes': JAML.registered_classes(),
+                'daemon': True,
```
JoanFM, line 342: There is some `_get_worker` or `process` in every kind of `Pod`, needs to be done also for `ContainerPod` at least

JoanFM (changes requested):

jina/orchestrate/pods/__init__.py
```diff
@@ -149,7 +149,7 @@ def __init__(self, args: 'argparse.Namespace'):
             getattr(args, 'runtime_backend', RuntimeBackendType.THREAD),
             events_list=[self.is_ready, self.is_shutdown],
         )
-        self.daemon = self.args.daemon
+        self.daemon = True
```
JoanFM, line 152: how is this used? can u set it to true in the `worker`?

JoanFM (changes requested):

jina/orchestrate/pods/__init__.py
samsja: it should be deleted

JoanFM (approved):

PR #4475, opened 2022-03-14
feat(executor): add serve function

hanxiao:
on developing remote Executor, I have to quickly make an Executor "served", I hope there is a shortcut for doing the following work:

```python
from jina import Flow

f = Flow(port=12345).add(uses=MyExec)
with f:
    f.block()
```

so now I can:

```python
from jina import Executor

class MyExec(Executor):
  ...

MyExec().serve()
```

Please improve doc or whatever needs. I wont have time to revisit this PR, hope to see it merge soon.

### DIFF
--- a/docs/fundamentals/executor/executor-api.md
+++ b/docs/fundamentals/executor/executor-api.md
@@ -315,7 +315,54 @@ print(docs.embeddings.shape)
 ```
 ````
 
+## Serve Executor without Flow
 
+Executors can be served - and remotely accessed - directly, without the need to instantiate a Flow manually.
+This is especially useful when debugging an Executor in a remote setting.
+
+An Executor can be served using the `.serve()` class method:
+
+````{tab} Serve Executor
+
+```python
+from jina import Executor, requests
+from docarray import DocumentArray, Document
+
+
+class MyExec(Executor):
+    @requests
+    def foo(self, docs: DocumentArray, **kwargs):
+        docs[0] = 'executed MyExec'  # custom logic goes here
+
+
+MyExec.serve(port=12345)
+```
+
+````
+
+````{tab} Access served Executor
+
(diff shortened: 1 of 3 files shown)


jina/serve/executors/__init__.py
```diff
@@ -317,6 +317,35 @@ def from_hub(
             uses_requests=uses_requests,
         )
 
+    def serve(
+        self,
```
JoanFM, line 321: I would not have this as a method of the instance, but a `class` method

jina/serve/executors/__init__.py
```diff
@@ -317,6 +317,35 @@ def from_hub(
...
+        :param uses_requests: dictionary of parameters to overwrite from the default config's requests field
+        :param stop_event: a threading event or a multiprocessing event that once set will resume the control Flow
+            to main thread.
+        :param kwargs: other kwargs accepted by the Flow, full list can be found in INSERT_DOC_URL
+
+        """
+        from jina import Flow
+
+        f = Flow(**kwargs).add(
+            uses=type(self),
```
JoanFM, line 341: therefore u could pass `cls`

JoanFM (changes requested):

docs/fundamentals/executor/executor-api.md
```diff
@@ -315,7 +315,54 @@ print(docs.embeddings.shape)
 ```
 ````
 
+## Serve Executor without Flow
```
JoanFM, line 318: I would call it `serve Executor standalone`

JoanFM (changes requested):
Please add at least small test about the feature

JoanFM (approved):

JohannesMessner (approved):

PR #4506, opened 2022-03-17
docs: explain host parameter for client and tweak client docs page

JohannesMessner:


docs/fundamentals/flow/client.md
```diff
@@ -19,19 +20,37 @@ Starting the Flow:
 ```python
 from jina import Flow
 
-port = 12345
+HOST = '0.0.0.0'
+PORT = 12345
+PROTOCOL = 'grpc'  # one of 'grpc', 'http', 'websocket'
 
-with Flow(port=port) as f:
+with Flow(host=HOST, port=PORT, protocol=PROTOCOL) as f:
```
JoanFM, line 27: I think the `host` parameter does not make much sense to be honest. Since this would expose `JinaD`. What should be explained is that u can run the `Flow` in another machine and call it from a client in another? @jacobowitz does this make sense?

JoanFM (changes requested):

docs/fundamentals/flow/client.md
JohannesMessner: Yes i just checked with tobi on this, will change it

JohannesMessner: i will not mention the host parameter on the Flow side

JoanFM (approved):

Issue #4474, opened 2022-03-14
Allow custom docker image for gateway


deepankarm:
**Describe the feature**
Allow an environment variable to pass custom images for gateway during `flow.to_k8s_yaml()` & `flow.to_docker_compose_yaml()`

PR #4508, opened 2022-03-17
feat: custom gateway image

jacobowitz:
# Custom Gateway image in K8s/Docker-Compose

## Description

This PR allows the user to set the environment variable `JINA_GATEWAY_IMAGE` to some custom Jina Docker image. It will be used for the gateway image when generating docker-compose or K8s configurations.
Existing tests have been slightly extended to cover this case.

Closes https://github.com/jina-ai/jina/issues/4474


jina/orchestrate/deployments/config/docker_compose.py
```diff
@@ -51,11 +51,12 @@ def get_gateway_config(
             import os
 
             test_pip = os.getenv('JINA_K8S_USE_TEST_PIP') is not None
```
deepankarm, line 53: This shouldn't be used anymore, right? Rather set `JINA_GATEWAY_IMAGE` in CI?

jacobowitz: good point, will change

JoanFM (approved):

Issue #4411, opened 2022-02-25
self.workspace and self.metas.workspace


cristianmtr:
**Describe the bug**

In this repo https://github.com/jina-ai/tmp-testing-simpleindex if you don't pass uses_metas:workspace as some value, self.workspace is None, but `self.metas.workspace` is `''` 

**Describe how you solve it**

they should both hold the same value
they should both default to something, not just `None`

JohannesMessner:
I looked into this a bit, and to me it looks like this:
`self.metas.workspace` seems useless and confusing, since there is `self.workspace`, and you can just pass a workspace like this: `f.add(..., worskpace='work/space/')`. In that case `self.workspace` would be set, but `self.metas.workspace` would not be set.
Regarding the default, if `self.workspace` is not given, it defaults to the Flow's workspace. This in turn does not have a default.

So I propose:
1. Remove `metas.workspace` from the documentation, deprecate it, and tell people to just pass `workspace`. Maybe also remove it from the code, @JoanFM do you think that is safe? Do we ever need `metas.workspace` instead of just `workspace`? If so we can automatically set it, and still tell people not to set it manually.
2. Set a default for `flow.workspace`, based on `flow.name`.

JohannesMessner:
Is the problem that sometimes the user wants to define a workspace in an Executor yaml, and then they need `metas.workspace`?

JoanFM:
The problem here is with `shards`. That we try to solve the `workspace` individually for each `shard`. It is quite a mess of code, and to be honest, I think is the weakest part of the abstraction when it comes to `local`-`cloud` behavior

PR #4511, opened 2022-03-17
refactor: clean up workspace parameter, add default workspace

JohannesMessner:
**Summary of new behaviour:**
- You can set an Executor's workspace by passing **`f.add(..., workspace=...)`**, or doing the equivalent in CLI. This has always been possible, but is now documented and recommended. Under the hood, this puts it into `runtime_args.workspace`, and of course `self.workspace`.
- **`self.metas.workspace`** sticks around for compatibility, but is **deprecated**
- If no workspace is provided, a **default workspace** will be used, located under `$HOME/.jina/executor-workspace`
- This default workspace is defined in the **environment variable** `JINA_DEFAULT_WORKSPACE_BASE`, which is set at installation, and can be changed.

**ToDo's:**

~~- [x] add default workspace for flow and executor in flow~~
- [x] add default workspace *only* for executor, not for Flow
- [x] create env var for default workspace
- [x] remove `metas.workspace` from documentation
- [x] deprecate `metas.workspace`
- [x] documentation

Discussion here: https://github.com/jina-ai/jina/issues/4411#issuecomment-1071068158


JoanFM (changes requested):
Would be cool to have a unit test for this value

tests/unit/serve/executors/test_executor.py
```diff
@@ -331,3 +337,22 @@ def test_serve(served_exec):
     docs = Client(port=PORT).post(on='/foo', inputs=DocumentArray.empty(5))
 
     assert docs.texts == ['foo' for _ in docs]
+
+
+def test_set_workspace():
```
JoanFM, line 342: use `tmpdir` fixture to make sure test uses a temporary directory and does not pollute user system (ours)

tests/unit/serve/executors/test_executor.py
```diff
@@ -331,3 +337,22 @@ def test_serve(served_exec):
...
+    _workspace = 'worki'
+    complete_workspace = os.path.abspath(os.path.join(_workspace, 'WorkspaceExec', '0'))
+    with Flow().add(uses=WorkspaceExec, workspace=_workspace) as f:
+        resp = f.post(on='/foo', inputs=Document())
+        assert resp[0].text == complete_workspace
```
JoanFM, line 347: move assert outside context manager to make sure it is really done

tests/unit/serve/executors/test_executor.py
```diff
@@ -331,3 +337,22 @@ def test_serve(served_exec):
...
+    with Flow().add(uses=WorkspaceExec, uses_metas={'workspace': _workspace}) as f:
+        resp = f.post(on='/foo', inputs=Document())
+        assert resp[0].text == complete_workspace
+
+
+def test_default_workspace():
```
JoanFM, line 353: same here

JoanFM (changes requested):

tests/unit/serve/executors/test_executor.py
```diff
@@ -331,3 +338,35 @@ def test_serve(served_exec):
...
+    assert resp[0].text == complete_workspace
+
+
+def test_default_workspace():
+    with Flow().add(uses=WorkspaceExec) as f:
+        resp = f.post(on='/foo', inputs=Document())
+    assert resp[0].text
+
+    result_workspace = resp[0].text
+    remove_test_dirs(result_workspace)  # delete created directories
```
JohannesMessner, line 359: Here I can't use `tmpdir`, since the Executor creates a default workspace, so I need to do cleanup manually.
But this makes me think, do we want to create a dir every time we use and Executor? having a default `workspace` would do that

JoanFM: i would avoid creating it if not required.

And for the tmpdir, u can use tmpdir and then u do not need to remove it. It is very dangerous in tests to add remove folder instructions

jina/resources/completions/jina.bash
```diff
@@ -21,5 +21,7 @@ complete -F _jina jina
 # session-wise fix
 ulimit -n 4096
 export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES
+# default workspace for Executors
+export JINA_DEFAULT_WORKSPACE_BASE="${HOME}/.jina/executor-workspace"
```
JohannesMessner, line 26: If a user changes this in their `.bashrc`, then the next install will overwrite that back to this value. Is that a problem?

jacobowitz (dismissed):
I think now its much cleaner, nice :+1:

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -61,9 +60,15 @@ The YAML API will ignore `.runtime_args` during save and load as they are not st
...
-Each `Executor` has a special `workspace` that is reserved for that specific `Executor` instance. The `workspace` property that can be used to extract the `path` to this workspace.
+Each `Executor` has a special *workspace* that is reserved for that specific Executor instance.
+The `.workspace` property contains the path to this workspace.
 
-This `workspace` is generated using the workspace specified in `metas` or `runtime_args` workspace, plus adding special suffixes in case of sharded `Executor`s.
+This `workspace` is generated using the workspace passed when adding the Executor: `flow.add(..., workspace='path/to/workspace/')`.
+The value that is passed there is added to the `runtime_args`, and the final `workspace` is generated by appending `'/<executor_name>/<shard_id>/'`,
+as described above.
+
+If the user provided no workspace, a the Executor uses a default workspace, which is defined in the `JINA_DEFAULT_WORKSPACE_BASE`
```
cristianmtr, line 70:
```suggestion
If the user hasn't provided a workspace, the Executor uses a default workspace, which is defined in the `JINA_DEFAULT_WORKSPACE_BASE`
```

jina/resources/completions/jina.bash
JohannesMessner: I'm now explaining this in the docs, should be ok

JoanFM (approved):

Issue #4503, opened 2022-03-17
Remove `Head` entity from `Deployments` without shards


JoanFM:
**Describe the feature**
With the inclusion of `linkerd` as default `service mesh` we can go and remove `head` which would imply better `QPS` and no extra hops. This will be done for all Deployments without shards. For Deployments with shards, the head will be kept.

- [ ] Remove RollingUpdate/Scale feature from `Flow`/`Deployment`/`Pod`/`JinaD`
- [ ] Implement reducing after multiple needs in the Gateway
- [ ] Set all worker Pods in the Gateway (Outside K8s)
- [ ] Make the gateway send to worker Pods
- [ ] Dont create heads for Deployments with no shards
- [ ] Change plotting
- [ ] Change `Architecture Overview` in the docs

JoanFM:
Block by #4433

jacobowitz:
Removing the `head` Pod for Deployments without shards will remove some functionality outside a K8s deployment and shift some responsibilities to the Gateway. 

**Features to be removed:**

1. **RollingUpdate**: Has to be removed, because the `Head` is handling the dynamic addition/removal of workers. This is not implemented in the Gateway and we are not planning to do so. In theory this could be rescued, but it does not seem to be worth it really. Within K8s it will still be possible with your regular K8s tools.
2. **Scale**: Same reasoning as RollingUpdate

Gateway assumes these responsibilities:

2. Reducing for an Executor with multiple needs (unless disabled by config). This means the Gateway will touch `Documents` and depend on DocArray.

Other necessary Changes:
* The worker Pods need to be set in the Gateway on creation
* The plotting needs to be updated

jacobowitz:
Completing this issue also means that External Deployment cant leverage multiple replicas without additional measures (Kubernetes/LoadBalancers/...)

jacobowitz:
I created a follow up ticket to get the external Executors right. We can close this issue before, but we should wait with a release until we have the follow up issue implemented. https://github.com/jina-ai/jina/issues/4556
Also I think the Deployment as a first class citizen issue should be done before a release: https://github.com/jina-ai/jina/issues/4502

PR #4517, opened 2022-03-18
feat: remove head for non sharded deployments

jacobowitz:
# Remove heads for non sharded Deployments

## Description

This PR removes the head for non sharded deployments. This means if shards are set to 1 for a Deployment, there will be no head created. In these cases the Gateway communicates directly with the workers. 
Also reducing requests in case of multiple `needs` is moved to the Gateway now as we cant rely on the fact to due it at the head anymore.

This is a breaking change :boom: It breaks uses_before/uses_after for non sharded Deploments. It also changes the way we interact with external Executors. Additionally `rolling_update` and `scale` are removed from the Flow.

Before we do a release we should do the follow up issues around (external) Deployments as first class citizens.

- [x]  Remove RollingUpdate/Scale feature from Flow/Deployment/Pod/JinaD
- [x]  Set all worker Pods in the Gateway (Outside K8s)
- [x]  Make the gateway send to worker Pods
- [x]  Dont create heads for Deployments with no shards
- [x]  Implement reducing after multiple needs in the Gateway
- [x]  Adapt tests
- [x]  Change docs

Closes [#4503 (issue)](https://github.com/jina-ai/jina/issues/4503)


daemon/stores/deployments.py
```diff
@@ -1,72 +1,12 @@
-from typing import TYPE_CHECKING, Dict, Optional
+from typing import TYPE_CHECKING
 
-from jina.helper import colored
-from daemon.stores.base import BaseStore
 from daemon.stores.pods import PodStore
 
 if TYPE_CHECKING:
```
JoanFM, line 5: remove this if

jina/orchestrate/deployments/__init__.py
```diff
@@ -107,14 +85,14 @@ def head_host(self) -> str:
         """Get the host of the HeadPod of this deployment
         .. # noqa: DAR201
         """
-        return self.head_args.host
+        return self.head_args.host if self.head_args else None
 
     @property
     def head_port(self):
```
JoanFM, line 91: we would need then another property for `ports()`? Should we have inheritance in Deployments?

JoanFM (commented):
I do not see where the Head logic has been migrated to the `Gateway`

jacobowitz:
> I do not see where the Head logic has been migrated to the `Gateway`

Its in the topology graph [here](https://github.com/jina-ai/jina/pull/4517/files#diff-0df9ebcdeb893e3c1f72b4a2a4e1c5473230e58ae5748cc4913c1720129f60b4R79)

jina/orchestrate/deployments/__init__.py
jacobowitz: you mean we should expose sth like `worker_ports`? Yeah that makes sense. I dont think we need inheritance though? Every deployment should have worker_ports. Just their usage in the Gateway and other places differ, but thats not the issue of the `Deployment` class?

JoanFM:
> > I do not see where the Head logic has been migrated to the `Gateway`
> 
> Its in the topology graph [here](https://github.com/jina-ai/jina/pull/4517/files#diff-0df9ebcdeb893e3c1f72b4a2a4e1c5473230e58ae5748cc4913c1720129f60b4R79)
> 
> Its only for the needs reducing. For shards and polling it will still be happening at the heads

I mean, where is the logic of the ConnectionPool where multiple `connections` are used? Where is the ROUND-ROBIN logic?

jacobowitz:
> I mean, where is the logic of the ConnectionPool where multiple `connections` are used? Where is the ROUND-ROBIN logic?

This logic in the ConnectionPool has not been changed. The Gateway is using the ConnectionPool as the head has done before. The only change is that we tell the Gateway now about all the workers instead of configuring the head. So the Gateway will add all workers to the ConnectionPool. This configuration is done [here](https://github.com/jina-ai/jina/pull/4517/files#diff-683724080977b02f2d25f165d53e871967021a0473235d7e12aad9466f6ec9aeR493).

jacobowitz:
rebased and added ports property to the Deployment

jina/orchestrate/deployments/__init__.py
jacobowitz: added a new `ports` property which can be used from the Flow and other places (like tests etc)

PR #4500, opened 2022-03-16
docs: explain https support

JohannesMessner:
Still ToDo:

- [x] wait for tls to be implemented with all protocols: https://github.com/jina-ai/jina/issues/4492
- [x] align naming with this pr: https://github.com/jina-ai/jina/pull/4498
- [x] rename `https` -> `tls`, or whatever keyword the PRs above land on

PR #4522, opened 2022-03-21
feat: add grpc tls support on gateway

samsja:
This pr add tls suppport for the grpc gateway 

- [x] allow tls on grpc server (gateway)
- [x] refactor to have a common API for http and gtpc tls support
- [x] (skip) OPTIONAL : intra pod communication using tls

PS: websocket and http gateway already handle tls


hanxiao:
it is not clear to me if websocket is included, if not, better include it. If it is already included, then please add it to PR's body

samsja:
> it is not clear to me if websocket'a TLS is included: if not, better include it. If it is already included, then please add it to PR's body

This pr is only about grpc as websocket is already handle both on server and on client side

JoanFM (commented):
Documentation required

tests/unit/serve/runtimes/gateway/grpc/test_grpc_tls.py
```diff
@@ -0,0 +1,56 @@
...
+    return f'{cur_dir}/cert/server.key'
+
+
+def test_grpc_ssl_with_flow(cert_pem, key_pem):
+    with Flow(
+        protocol='grpc',
+        ssl_certfile=cert_pem,
+        ssl_keyfile=key_pem,
+    ) as f:
+        os.environ['JINA_LOG_LEVEL'] = 'ERROR'
```
JoanFM, line 32: If u change this env variable, put it in a `fixture` that after running puts it back to the original value:

```python
@pytest.fixture
def error_log_level():
    old_env = os.environ.get('JINA_LOG_LEVEL')
    os.environ['JINA_LOG_LEVEL'] = 'ERROR'
    yield
    os.environ['JINA_LOG_LEVEL'] = old_env
```

JoanFM (changes requested):

tests/unit/serve/runtimes/gateway/grpc/test_grpc_tls.py
samsja: Should I changed it there as well ? (this is were I originally took the code) https://github.com/jina-ai/jina/blob/dc8a99addde6f16637cc8439fed4d6208bc6384e/tests/unit/serve/runtimes/gateway/http/test_app.py#L207

samsja:
> Documentation required

there is a pending PR that will add documentation for everything related to `tls`. https://github.com/jina-ai/jina/pull/4500
I think that it is better to don't add documentation for this PR and do it only in the documentation PR

tests/unit/serve/runtimes/gateway/grpc/test_grpc_tls.py
JoanFM: yes

JoanFM (approved):

Issue #4521, opened 2022-03-21
bug in transmitting blob via http 


hanxiao:
a bug in the core on http server, the blob reconstruct is wrong with http protocol (b'aGVsbG8='), my guess is base64 is not applied.

```python
from docarray import Document

from jina import Executor, requests, Flow


class MyExec(Executor):

    @requests
    def foo(self, docs, **kwargs):
        print(docs[0].blob)


for p in ('grpc', 'websocket', 'http'):
    f = Flow(protocol=p).add(uses=MyExec)
    with f:
        f.post('/', Document(blob=b'hello'))
```

the result:
```
/Users/hanxiao/.pyenv/versions/3.7.9/bin/python /Users/hanxiao/Documents/jina/toy31.py
⠦ 2/2 waiting  to be ready...
                                               
 🔗          Protocol:                     0   
 🏠      Local access:         0.0.0.0:50846   
 🔒   Private network:   192.168.0.102:50846   
 🌐    Public address:  87.191.159.105:50846   
b'hello'
⠧ 2/2 waiting  to be ready...
                                               
 🔗          Protocol:                     2   
 🏠      Local access:         0.0.0.0:51807   
 🔒   Private network:   192.168.0.102:51807   
 🌐    Public address:  87.191.159.105:51807   
  gateway/rep-0@19003[I]:client 127.0.0.1:56109 connected with subprotocol bytes
b'hello'
⠧ 2/2 waiting  to be ready...
                                                       
 🔗          Protocol:                             1   
 🏠      Local access:                 0.0.0.0:59276   
 🔒   Private network:           192.168.0.102:59276   
 🌐    Public address:          87.191.159.105:59276   
 💬        Swagger UI:   http://localhost:59276/docs   
 📚             Redoc:  http://localhost:59276/redoc   
b'aGVsbG8='
Process finished with exit code 0
```

PR #4524, opened 2022-03-22
fix: blob en/decoding over http

JohannesMessner:
This PR becomes obsolete once upstream fix is merged: https://github.com/jina-ai/docarray/pull/223

Issue: #4521 

**TODO**

- [x] fix the bug
- [x] write test
- [x] fix now failing tests...

### DIFF
--- a/jina/serve/runtimes/gateway/http/models.py
+++ b/jina/serve/runtimes/gateway/http/models.py
@@ -276,8 +276,10 @@ class JinaRequestModel(BaseModel):
         :return: self as Python dict
         """
         d = super().dict(*args, **kwargs)
-        if isinstance(self.data, list) and isinstance(
-            self.data[0], PydanticDocument
+        if (
+            self.data
+            and isinstance(self.data, list)
+            and isinstance(self.data[0], PydanticDocument)
         ):  # list of PydanticDocuments == PydanticDocumentArray
             docs_dicts = DocumentArray.from_pydantic_model(self.data).to_dict()
             d['data'] = docs_dicts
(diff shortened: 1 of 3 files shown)


jina/serve/runtimes/gateway/http/models.py
```diff
@@ -266,6 +267,20 @@ class JinaRequestModel(BaseModel):
...
+
+        :param *args: args passed to BaseModel.dic()
+        :param *kwargs: kwargs passed to BaseModel.dic()
+        :return: self as Python dict
+        """
+        d = super().dict(*args, **kwargs)
+        if self.data is not None:
+            docs_dicts = DocumentArray.from_pydantic_model(self.data).to_dict()
+            d['data'] = docs_dicts
+        return d
```
JohannesMessner, line 282: It feels a bit iffy here to create a DocumentArray just to then convert it to dict, but I can't really think of a nicer way.
Context: Here `self.data` is a `PydanticDocumentArray`, and in that the blob is always encoded *twice*. To handle properly, `from_pydantic_model` is used instead of a built in conversion do dict.

jina/serve/runtimes/gateway/http/models.py
```diff
@@ -238,7 +239,36 @@ def _get_example_data():
...
 
-class JinaRequestModel(BaseModel):
+class PydanticToDictModel(BaseModel):
+    """
+    Base class for models that contain PydanticDocument or PydanticDocarray and need to be converted to dict.
+    """
+
+    def dict(self, *args, **kwargs):
+        """
+        Generate a dictionary representation of the model, optionally specifying which fields to include or exclude.
```
samsja, line 249: not sure to see in the code how to specify which field to include/exclude is it in the super().dict() method ?

JohannesMessner: Exactly, the docstring is just copied from the same method in `super()`. This method itself just replaces one filed in the dict provided by super

JoanFM (commented):
It is hard for me to understand the original problem we are trying to fix here

JohannesMessner:
> It is hard for me to understand the original problem we are trying to fix here

Issue was originally raised here https://github.com/jina-ai/jina/issues/4521, essentially when using http and sending a `blob`, it would not encode and decode it properly, resulting in different (wrong) data being received by the executors.

JohannesMessner:
This PR becomes obsolete once upstream fix is merged: https://github.com/jina-ai/docarray/pull/223
Keeping it open in the meantime

PR #4527, opened 2022-03-22
docs: remove misplaced text

JohannesMessner:


docs/api.md
```diff
@@ -1,12 +1,5 @@
 # Python API
 
-To generate docs, please use:
-
-```bash
-cd docs
-bash makedocs.sh local-only
-```
-
 There are four packages shipped with Jina:
```
JohannesMessner, line 3: I don't think this belongs in the public documentation at all, but rather in `CONTRIBUTING.md`. The same snipped is already there, so I'm just removing here.

samsja (approved):
lgtm

Issue #4488, opened 2022-03-16
showing http in the log with grpc gateway


hanxiao:
MWU to reproduce

```python
import os
os.environ['JINA_LOG_LEVEL'] = 'DEBUG'
from jina import Executor

Executor.serve(port=34567)
```

why ` ... deployment executor0/heads/0 to http://0.0.0.0:55078/` in the log?

```
executor0/head@15653[D]:Start listening on 0.0.0.0:55078
executor0/rep-0@15654[D]:Start listening on 0.0.0.0:59451
  gateway/rep-0@15655[D]:Adding connection for deployment executor0/heads/0 to http://0.0.0.0:55078/
  gateway/rep-0@15655[D]: Start server bound to 0.0.0.0:34567
 executor0/head@15653[D]:recv ControlRequest 073b3777abf64206873ae369f9fc9a3e STATUS
  gateway/rep-0@15655[D]:recv ControlRequest f6de619cf7ce4f9bbf48446e25f3bf4e STATUS
  gateway/rep-0@15649[D]:ready and listening
 executor0/head@15649[D]:ready and listening
executor0/rep-0@15649[D]:ready and listening
 executor0/head@15653[D]:recv ControlRequest 2a7590e7943a476ca299573e55dc31ec ACTIVATE
 executor0/head@15653[D]:Adding connection for deployment worker/shards/0 to 0.0.0.0:59451
⠧ 2/2 waiting  to be ready...
🎉 Flow is ready to use!
                                               
 🔗          Protocol:                     0   
 🏠      Local access:         0.0.0.0:34567   
 🔒   Private network:     192.168.3.6:34567   
 🌐    Public address:  87.191.159.105:34567
```

PR #4541, opened 2022-03-24
fix: deployment protocol

JohannesMessner:
return correct protocol when calling `.protocol` on a deployment.
Fixes https://github.com/jina-ai/jina/issues/4488

### DIFF
--- a/jina/orchestrate/deployments/__init__.py
+++ b/jina/orchestrate/deployments/__init__.py
@@ -448,8 +448,6 @@ class Deployment(BaseDeployment):
         """
         :return: the protocol of this deployment
         """
-        if self.is_sandbox:
-            return 'https'
         protocol = getattr(self.args, 'protocol', 'grpc')
         return str(protocol) + ('s' if self.tls_enabled else '')
 
--- a/jina/orchestrate/deployments/__init__.py
+++ b/jina/orchestrate/deployments/__init__.py
@@ -5,18 +5,17 @@ from argparse import Namespace
 from contextlib import ExitStack
 from typing import Dict, List, Optional, Set, Union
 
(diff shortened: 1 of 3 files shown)


jina/orchestrate/deployments/__init__.py
```diff
@@ -436,10 +446,12 @@ def external(self) -> bool:
     @property
     def protocol(self):
         """
-        :return: the protocol of this deployment, https or http
+        :return: the protocol of this deployment
         """
-        protocol = getattr(self.args, 'protocol', 'http')
-        return 'https' if self.is_sandbox else protocol
+        if self.is_sandbox:
+            return 'https'
```
JohannesMessner, line 452: I couldn't find a way around this @jacobowitz

jacobowitz: how does the log look like for a sandbox executor then? I guess it is still printing `https://<sandboxaddress>` then?

tests/integration/issues/github_4488/test_deployment_protocol.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    with f:
+        for node, v in f._deployment_nodes.items():
+            p = v.protocol.lower()
+            if node == 'gateway':
+                assert p == protocol + ('s' if tls else '')
+            elif node == 'executor0':
+                assert p == 'grpc'
+            elif node == 'executor1':
+                assert p == 'https'
```
jacobowitz, line 50: I think we should try to fix this, otherwise we are still in a non consistent state here?

jacobowitz (changes requested):

jacobowitz (approved):

PR #4548, opened 2022-03-25
docs: refactor executor in flow

cristianmtr:
# docs: refactor Executor in Flow section

## Description

Refactor Executor in Flow page

Closes https://github.com/jina-ai/internal-tasks/issues/402


docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
-This `workspace` is generated using the workspace passed when adding the Executor: `flow.add(..., workspace='path/to/workspace/')`.
-The value that is passed there is added to the `runtime_args`, and the final `workspace` is generated by appending `'/<executor_name>/<shard_id>/'`,
-as described above.
+````{tab} via YAML
 
-If the user hasn't provided a workspace, the Executor uses a default workspace, which is defined in the `JINA_DEFAULT_WORKSPACE_BASE`
-environment variable.
+```yaml
+  jtype: MyExecutor
+  uses_with:
```
JoanFM, line 77: it would be interesting to see here how u can pass some `with` parameters directly in the `MyExecutor`

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -146,7 +135,7 @@ with f:
 
 ```yaml
 jtype: MyExecutor
-with:
```
JoanFM, line 149: we need an example using `with` simply

JoanFM (changes requested):

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
+````{dropdown} executor.py
+:open:
 
-class MyExecutor(Executor):
-    def __init__(self, foo: str, bar: int, **kwargs):
-        super().__init__(**kwargs)
-        self.bar = bar
-        self.foo = foo
-```
+This is a basic Executor, defined in its own file. Notice that there are no {ref}`request bindings <executor-requests>`. These are configured via the Python and YAML syntax.
```
JohannesMessner, line 15: I would just make it a bit clearer that the 'old' way of defining a request mapping is still valid, and also mention the advantages of doing it this way:

```suggestion
This is a basic Executor, defined in its own file. Notice that there are no {ref}`request bindings <executor-requests>`. If an Executor is used in a Flow, these bindings can be configured via the Python or YAML Flow syntax, so each Flow can use the Executor in its own way. However, the standard way of adding `@requests` in the Executor itself would still work just fine.
```

JohannesMessner (changes requested):
I like it a lot, seems a lot more approachable now. Just two small suggestions :)

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
 
-- `name`: Name given to the `Executor`. This is dynamically adapted from the `name` in `metas` and depends on some additional arguments like `shard_id`. 
-- `replicas`: Number of replicas of the same `Executor` deployed with the `Flow`.
-- `shards`: Number of shards of the same `Executor` deployed with the `Flow`.
-- `shard_id`: Identifier of the `shard` corresponding to the given `Executor` instance.
-- `workspace`: Path to be used by the `Executor`. Note that the actual workspace directory used by the Executor is obtained by appending `'/<executor_name>/<shard_id>/'` to this value.
-- `py_modules`: Path to the modules needed to import the `Executor`. This is another way to pass `py-modules` to the `Executor` from the `Flow`
+Python API-specific options:
 
+- `uses` can be a class or string. Defines the Executor to load. You can also run Executors available on [Jina Hub](https://hub.jina.ai) with a {ref}`special syntax <use-hub-executor>`.
```
alexcg1, line 69:
```suggestion
- `uses` can be a class or string that defines the Executor to load. You can also run Executors available on [Jina Hub](https://hub.jina.ai) with a {ref}`special syntax <use-hub-executor>`.
```

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
 
-````{admonition} Note
-:class: note
-The YAML API will ignore `.runtime_args` during save and load as they are not statically stored
 ````
 
-### Requests
-By default, an `Executor` object contains `.requests` as an attribute when loaded from the `Flow`. This attribute is a `Dict` describing the mapping between Executor methods and network endpoints: It holds endpoint strings as keys, and pointers to `function`s as values.
```
alexcg1, line 60: Why is `function`s backticked here?

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
-````{admonition} Caution
-:class: caution
-After you install jina, the `JINA_DEFAULT_WORKSPACE_BASE` environment variable will be set in your `.bashrc`, `.zshrc`, or
-`.fish` file.
+YAML-specific options:
 
-To change the default Executor workspace on your system, you can change the value of this environment variable.
-However, if you directly edit the corresponding command in your `.bashrc` (or `.zshrc`/`.fish`) file, your changes will be reverted the next time
-you install jina on your system.
+- `jtype` is a string. Defines the class name, interchangeable with bang mark `!`;
```
alexcg1, line 94:
```suggestion
- `jtype` is a string that defines the class name, interchangeable with bang mark `!`;
```

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
-  name: MyExecutor
-  description: "MyExecutor does a thing to the stuff in your Documents"
-  py_modules:
-    - executor.py
-requests:
-  /index: MyExecutor_index_method
-  /search: MyExecutor_search_method
-  /random: MyExecutor_other_method
-```
+Common arguments to both YAML and Python API:
```
alexcg1, line 98:
```suggestion
Common arguments for both YAML and Python API:
```

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -1,136 +1,125 @@
...
 
-- `jtype` is a string. Defines the class name, interchangeable with bang mark `!`;
-- `with` is a map. Defines `kwargs` of the class `__init__` method
-- `metas` is a dictionary. It defines the meta information of that class. It contains the following fields:
+- `uses_with` is a key-value map. Defines `kwargs` of the class `__init__` method
+- `uses_metas` is a key-value map. It defines the meta information of that class. It contains the following fields:
```
alexcg1, line 101: Please fix the "Defines" text for each entry, as suggested in prior comments

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -377,19 +433,20 @@ NotImplementedError: no time for it
...
 
-## Gracefully closing an Executor
+````
+
+## Graceful shutdown of an Executor
 
-You might need to execute some logic when your executor's destructor is called. For example, you want to
-persist data to the disk (e.g. in-memory indexed data, fine-tuned model,...). To do so, you can overwrite the
-method `close` and add your logic.
+You might need to execute some logic when your executor's destructor is called.
```
alexcg1, line 440:
```suggestion
You might need to execute some logic when your Executor's destructor is called.
```

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -427,12 +489,14 @@ closing...
 Process finished with exit code 0
 ```
 
+````
+
 ## Multiple DocumentArrays as input
 
-We have seen that `Executor` methods can receive 3 types of parameters: `docs`, `parameters` and `docs_matrix`.
+You have seen that `Executor` methods can receive 3 types of parameters: `docs`, `parameters` and `docs_matrix`.
```
alexcg1, line 496:
```suggestion
You have seen that `Executor` methods can receive three types of parameters: `docs`, `parameters` and `docs_matrix`.
```

alexcg1 (changes requested):

docs/fundamentals/executor/executor-in-flow.md
cristianmtr: We wanted to keep it consistent across examples, using `uses_X` in both. In Python you can't use the short form `with`

docs/fundamentals/executor/executor-in-flow.md
cristianmtr: We wanted to keep it consistent across examples, using uses_X in both. In Python you can't use the short form with

docs/fundamentals/executor/executor-in-flow.md
```diff
@@ -243,31 +233,107 @@ with flow2:
...
+        self.bar = bar
+        self.foo = foo
 ```
 
+This is important because when an `Executor` is instantiated in the context of a Flow, Jina is adding extra arguments.
+Some of these `arguments` can be used when developing the internal logic of the Executor.
+
+These `special` arguments are `workspace`, `requests`, `metas`, `runtime_args`.
+
+### `metas`
```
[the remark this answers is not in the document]
cristianmtr: I'd say the order should be

- workspace
- requests
- metas
- r.. args (these are never configured by a user)

docs/fundamentals/executor/executor-in-flow.md
cristianmtr: Fixed

cristianmtr:
@JoanFM @alexcg1 @JohannesMessner Can you re-check?

alexcg1 (approved):
On mobile now so I only see URL changes. If that's all it is, consider it approved

JohannesMessner (approved):

JoanFM (approved):

PR #4550, opened 2022-03-28
refactor: remove jinad

jacobowitz:
# Remove JinaD

## Description

This PR removes JinaD. This should include all code, tests and CI integrations.


deepankarm:
image

Dockerfiles/debianx.Dockerfile
```diff
@@ -71,13 +71,4 @@ RUN cd /jina && \
 
 ENTRYPOINT ["jina"]
 
-FROM jina AS jina_daemon
```
JoanFM, line 74: then maybe we can remove this file?

jacobowitz: we are still using it in CI. I am not completely sure what is the difference between this file and the PIP one. But we are using this e.g. in the ARM64 test I believe. Maybe for other things as well.

jacobowitz: e.g. [here](https://github.com/jina-ai/jina/blob/master/.github/workflows/ci.yml#L142)

JoanFM (approved):

PR #4557, opened 2022-03-29
fix: pin black version

jacobowitz:
We need to pin black to a new version [due this upstream issue](https://github.com/psf/black/pull/2966)


JoanFM (dismissed):

jacobowitz:
I blacked some files just to generate a real diff on python code. Otherwise the Black action would fail (we should fix this, but lets do this in a separate PR)

Issue #4555, opened 2022-03-29
Two jina flow can run on the same port


samsja:
**Describe the bug**

You can launch two flow on the same port, meaning two gateway serving on the same port without any of the two complaining.

**Step to reproduce**

In `toy.py`
```python
from jina import Flow, requests, Executor

class Dummy(Executor):

    @requests
    def foo(self,docs,**kwargs):
        ...

if __name__ == '__main__':

    with Flow(port=8888).add(uses=Dummy) as f:
        f.block()
```

Then do two times in two terminal

```bash
python toy.py
```
None of them complain, the second call should say that the port is already used

**Environment**
jina version 3.2.9

hanxiao:
let's give it a higher priority, i think it can be really confusing for users, especially multiple-same-port-gateway gives weird empty results, as we shown in CLIP-as-service

PR #4562, opened 2022-03-30
fix: check if port is occupied

samsja:
fix https://github.com/jina-ai/jina/issues/4555


jina/helper.py
```diff
@@ -1589,3 +1590,10 @@ def _parse_url(host):
         port = None
 
     return scheme, host, port
+
+
+async def is_port_free(host, port):
```
JoanFM, line 1595: why is it async?

JoanFM (changes requested):

tests/integration/gateway/test_gateway.py
```diff
@@ -0,0 +1,30 @@
+import time
+from threading import Thread
+
+from jina import Flow
+
+
+def test_borrow_same_port(port_generator):
```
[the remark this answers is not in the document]
samsja: yes you are right I will add a delay on the first one then

jina/helper.py
samsja: my bad it used to await co routine inside but not anymore

samsja:
Block this PR for now as the failing behavior is not consistent

JoanFM (dismissed):

tests/integration/gateway/test_gateway.py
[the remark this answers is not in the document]
samsja: done

jina/orchestrate/flow/base.py
```diff
@@ -471,6 +473,12 @@ def _add_gateway(
 
         kwargs.update(self._common_kwargs)
         args = ArgNamespace.kwargs2namespace(kwargs, set_gateway_parser())
+
+        if not (
+            is_port_free(__default_host__, args.port)
```
jacobowitz, line 478: I don't think this is correct. You are checking if the port defined for the Flow on localhost is available. But the Flow will not necessarily run on localhost!
You may build a Flow on one machine, generate K8s configs and deploy them elsewhere. Your could would fail in this case when port is occupied on the building machine, which is wrong.

I think you rather need to check this when the Flow starts. I dont see how you can do this on build time as it is impossible to say where the Flow will be started.

tests/integration/gateway/gateway_clients/test_streaming.py
```diff
@@ -1,22 +1,22 @@
-import time
-import os
 import asyncio
 import copy
 import multiprocessing
-import pytest
-
-from typing import Dict
+import os
```
jacobowitz, line 4: these changes here are probably isort, but lets try to not change imports on files you are making no other changes?

tests/integration/gateway/gateway_non_blocking/test_gateway_non_blocking.py
```diff
@@ -1,8 +1,8 @@
-import pytest
+import time
 
-from jina import Flow, Executor, requests, DocumentArray, Document
+import pytest
 
-import time
+from jina import Document, DocumentArray, Executor, Flow, requests
```
jacobowitz, line 5: same as above

tests/integration/issues/github_4555/test_port.py
```diff
@@ -0,0 +1,21 @@
+import time
+from threading import Thread
+
+import pytest
+
+from jina import Flow
+from jina.excepts import PortAlreadyUsed
+
+
+def test_two_flow_using_on_same_port_failing(port_generator):
```
jacobowitz, line 10: Please also add a test where you start `f1` first, build Flow `f2` (this is supposed to work!) and expect a failure on starting `f2`

jacobowitz (changes requested):
I think the check on Flow build needs to go, the check on starting the gateway looks good. We may want to do the same even for head/workers?

jina/orchestrate/flow/base.py
samsja: Arf, you are completly right:

To add some context: I wanted initially to only check if the port is free before starting the grpc server. However because of some flakyness in the healt check it was not working proprely (there is still this check btw). Testing at parsing time is a hack see this pr for more info about the flakyness/bug : https://github.com/jina-ai/jina/pull/4565

I hack would be to check at parsing time only when run locally. 

The actual true solution is to fix the problem of the pr above, but we decide with @JoanFM that is was not worth it investigate as we move more and more to use k8s

jacobowitz: The issue with this is that it is basically impossible to know upfront if a Flow will run locally or not? You can't anticipate if the user will eventually start the Flow locally or generate some configs.
I think I dont really get the reasoning why you cant just move the check from build time to start time? If you just have this check on `Flow.start()` it should behave the same no? The only difference is that we can be sure then that the Flow is indeed running locally.

JoanFM: yes, it should be moved to `start`

tests/integration/gateway/gateway_clients/test_streaming.py
samsja: That's strange that it has been caught in the pre-commit then :thinking: I will remove it

jacobowitz (approved):
nice, looks good now :)

PR #4564, opened 2022-03-30
feat: add jina now

florian-hoenicke:
The purpose of this pr is to let the user run jina now from the jina package installation.
In case jina-now is not installed it will ask you to install it first


samsja:
What about just do what @nomagick suggested ? It is as fast and we do not have cyclic import or dependency 

Yes jina  doesn’t require the plugins but detects if some runnable jina-{argv[1]}  exists in PATH, if true then run this runnable with the rest of the arguments.

JoanFM (changes requested):
This has to be discussed, it is nowhere clear that this is the direction we want to go. Blocked for now

PR #4565, opened 2022-03-30
Fix sync worker

samsja:
This PR is here to showcase a bug that I encounter while working on https://github.com/jina-ai/jina/pull/4562 to fix https://github.com/jina-ai/jina/issues/4555

How to reproduce:

in `toy.py`
```python
from jina import Flow
if __name__ == '__main__':
    with Flow(port=8888).add(uses=Dummy) as f:
        f.block()
```

do two times
```bash
python toy.py
```

on the second call you will see :

```    bash
raise PortAlreadyUsed(f'port:{self.args.port}')
jina.excepts.PortAlreadyUsed: port:8888
 🔗         Protocol                 GRPC   
 🏠     Local access         0.0.0.0:8888   
```

I excepted that the second flow will throw a Runtime Error but it does not and stay healthy:

Possible explanation:

The heath check is passing because there is actually a gateway serving on port 8888 (from the first call of toy)
Even if the health check is passing it there should be a signal that should be send to the main process from the runtime process to say that something went wrong during setup.

It might not be worth it to investigate as we focus more on k8s now and that this issue is only about local serving with the process runtime.

There might still be a bug somewhere though


hanxiao:
> It might not be worth it to investigate as we focus more on k8s now and that this issue is only about local serving with the process runtime.

but in Jina 2.x we don't have this problem, why do we have it now?

samsja:
> > It might not be worth it to investigate as we focus more on k8s now and that this issue is only about local serving with the process runtime.
> 
> but in Jina 2.x we don't have this problem, why do we have it now?

it was probably already here in 2.x the issue is with the health check that return true because an other gateway is serving on the same port

PR #4571, opened 2022-04-01
fix: start DataRequestHandler before grpc server

### LINKED ISSUE #4570: WorkerRuntime starts gRPC Server before DataRequestHandler is ready
The WorkerRuntime calls `super().init()` before `DataRequestHandler(args, self.logger)`. The super init will in turn call `async_setup()`, which will initialize the grpc server. So this server will be ready BEFORE the DataRequestHandler.

This is bad, because a STATUS request will succeed signalling the runtime being ready when it may in fact not be. DataRequests will get stuck though.

jacobowitz:
# Start DataRequestHandler before grpc server

## Description

We need to start the DataRequestHandler before the grpc Server. We may have Executors which are slow to init (downloading models etc) and we should not indicate to ready before the DataRequestHandler has been started.
This works atm with STATUS requests, but on K8s we are just doing plain TCP connects to the gRPC server to check if the pod is ready. These connects should fail while the Excutor is still initializing.

Closes #4570 ([issue](https://github.com/jina-ai/jina/issues/4570))

### DIFF
--- a/tests/unit/serve/runtimes/worker/test_worker_runtime.py
+++ b/tests/unit/serve/runtimes/worker/test_worker_runtime.py
@@ -308,14 +308,14 @@ async def test_worker_runtime_slow_init_exec():
 
     # try to connect a TCP socket to the gRPC server
     # this should only succeed after the Executor is ready, which should be after 5 seconds
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    connected = False
-    while not connected:
-        try:
-            s.connect((args.host, args.port))
-            connected = True
-        except ConnectionRefusedError:
-            time.sleep(0.2)
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        connected = False
+        while not connected:
+            try:
+                s.connect((args.host, args.port))
+                connected = True
+            except ConnectionRefusedError:
+                time.sleep(0.2)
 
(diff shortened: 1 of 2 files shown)


JohannesMessner (approved):
We should tell the community member that raised this that his issue will be fixed in the next release

tests/unit/serve/runtimes/worker/test_worker_runtime.py
```diff
@@ -275,5 +274,69 @@ async def task_wrapper(adress, messages_received):
...
+    # try to connect a TCP socket to the gRPC server
+    # this should only succeed after the Executor is ready, which should be after 5 seconds
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    connected = False
+    while not connected:
+        try:
+            s.connect((args.host, args.port))
+            connected = True
+        except ConnectionRefusedError:
+            time.sleep(0.2)
```
samsja, line 318: I had some issue with pytest when I don't close the socket. Here it might not change anything but it might prevent future but in the test maybe
```suggestion
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        connected = False
        while not connected:
            try:
                s.connect((args.host, args.port))
                connected = True
            except ConnectionRefusedError:
                time.sleep(0.2)
```

jacobowitz: good point, added it!

samsja (approved):
lgtm

PR #4579, opened 2022-04-04
feat: add support for cli plugins

JohannesMessner:
**TODO**

- [x] automatically support commands if the start with 'jina-'
- [x] give sensible warning about uninstalled jina packages
- [x] enable suggestions for uninstalled jina packages
- [x] centralize checking for new versions of jina packages
- [x] move 'database' of known plugins/packages to github

**Adding a Jina project to the core CLI as a plugin**

Basic usage:
- The project defines its own CLI and command
- The command associated with the project need to look like `jina-whatever`. If it does, and it is installed on the user's system, then `jina whatever` will also work automatically, **no need to configure anything, no need to wait for a core release**
- Any changes to the plugin CLI will be reflected automatically in the core CLI

Slightly-more-advanced stuff:
- For these features, the new **project's name, pip-package,  and plugin sub-command must be added to a json** on the Jina master branch. This way this info is always accessible and not tied to core release cycles.
- When the user uses the plugin, the core CLI automatically checks if the plugin has a newer version released on pip, and alerts the user
- If the user tries to use the plugin through the core CLI, but the plugin is not installed, the core CLI tells the user what to install

**Known issues, will not work in this first iteration**

- Autocompletion with a plugin is not trivial. There are a few ideas, but none of them are perfect, so it is probably better to wait out on this.

### DIFF
--- a/cli/__init__.py
+++ b/cli/__init__.py
@@ -141,18 +141,19 @@ def _is_latest_version(package='jina', suppress_on_error=True):
 def _try_plugin_command():
     """Tries to call the CLI of an external Jina project."""
     argv = sys.argv
-    if len(argv) < 2:
+    if len(argv) < 2:  # no command given
+        return False
+
+    from .autocomplete import ac_table
+
+    if argv[1] in ac_table['commands']:  # native command can't be plugin command
         return False
 
     def _cmd_exists(cmd):
         return shutil.which(cmd) is not None
 
     subcommand = argv[1]
-    cmd = (
-        PLUGIN_INFO[subcommand]['internal-command']
-        if subcommand in PLUGIN_INFO
-        else 'jina-' + subcommand
-    )
+    cmd = 'jina-' + subcommand
     if _cmd_exists(cmd):
         if subcommand in PLUGIN_INFO:
             import threading
--- a/cli/__init__.py
+++ b/cli/__init__.py
@@ -5,6 +5,10 @@ import sys
 
 from packaging.version import Version, parse
 
+KNOWN_PLUGINS_URL = (
+    'https://raw.githubusercontent.com/jina-ai/jina/unified-cli/cli/known_plugins.json'
+)
+
 
 def _get_run_args(print_args: bool = True):
     from jina.helper import get_rich_console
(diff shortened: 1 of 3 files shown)


cli/__init__.py
```diff
@@ -91,31 +93,43 @@ def _quick_ac_lookup():
             exit()
 
 
-def _is_latest_version(suppress_on_error=True):
+def _parse_latest_release_version(resp):
+    # credit: https://stackoverflow.com/a/34366589
+    import json
+
+    latest_release_ver = parse('0')
```
JoanFM, line 96: what does this do?

JohannesMessner: In case there is no release version, e.g. Now only has devel versions.Then this will be used as a fallback of sorts

cli/__init__.py
```diff
@@ -1,6 +1,14 @@
 import os
+import shutil
+import subprocess
 import sys
 
+from packaging.version import Version, parse
+
+KNOWN_PLUGINS_URL = (
+    'https://raw.githubusercontent.com/jina-ai/jina/master/cli/known_plugins.json'
+)
```
JohannesMessner, line 10: I believe we could make this accessible through the `https://api.jina.ai/` API instead, but I don't know how to set that up

cli/__init__.py
JoanFM: I there are no release versions I think it should not work

cli/__init__.py
```diff
@@ -1,6 +1,14 @@
 import os
+import shutil
+import subprocess
 import sys
 
+from packaging.version import Version, parse
+
+KNOWN_PLUGINS_URL = (
```
JoanFM, line 8: if this is linked to master, why not keep it here statically?

cli/__init__.py
JohannesMessner: With this solutions, if there is no release version and this fallback kicks in, nothing will happen, so I  think that is what we want (the user will not be alerted). Without this, things can break.

cli/__init__.py
JohannesMessner: So that no release is necessary to update this, only a merge. But as discussed in the daily, I will make it static to keep it simple.

cli/known_plugins.py
```diff
@@ -0,0 +1,6 @@
+plugin_info = {
```
JoanFM, line 1: this file is still needed?

JoanFM (changes requested):

cli/known_plugins.py
JohannesMessner: I could put this dict in the `__init__` file, but I thought for the people that want to add their plugin it is nicer if it is in this separate file.

JoanFM: ok, but is not extracted from github right?

JohannesMessner: No, imported and used locally: `from .known_plugins import plugin_info`

JoanFM (approved):

PR #4581, opened 2022-04-05
fix(executor): runtime args are now passed directly without any parsing

samsja:
done with @alaeddine-13 
The `JinaLoader` now allow to pass `runtime_args` to the init of the Executor without parsing them to a yaml before

### DIFF
--- a/jina/jaml/__init__.py
+++ b/jina/jaml/__init__.py
@@ -168,7 +168,7 @@ class JAML:
         :param stream: the stream to load
         :param substitute: substitute environment, internal reference and context variables.
         :param context: context replacement variables in a dict, the value of the dict is the replacement.
-        :param runtime_args: Optional runtime_args to be passed to the created object without coming from the JAML file
+        :param runtime_args: Optional runtime_args to be directly passed without being parsed into a yaml config
         :return: the Python object
 
         """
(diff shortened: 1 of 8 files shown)


JoanFM (approved):

jina/jaml/__init__.py
```diff
@@ -559,10 +569,13 @@ def _from_yaml(cls, constructor: FullConstructor, node):
         :param node: the node to traverse
         :return: the parser associated with the class
         """
-        data = constructor.construct_mapping(node, deep=True)
+        data = constructor.construct_mapping(node, deep=True)  # todo update
```
alaeddine-13, line 572:
```suggestion
        data = constructor.construct_mapping(node, deep=True)
```

jina/jaml/__init__.py
```diff
@@ -559,10 +569,13 @@ def _from_yaml(cls, constructor: FullConstructor, node):
...
         from jina.jaml.parsers import get_parser
 
-        return get_parser(cls, version=data.get('version', None)).parse(cls, data)
+        return get_parser(cls, version=data.get('version', None)).parse(
+            cls, data, runtime_args=constructor.runtime_args
+        )
+        # return get_parser(cls, version=data.get('version', None)).parse(cls, data)
```
alaeddine-13, line 578:
```suggestion
```

jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -47,6 +47,7 @@ def _load_executor(self):
                     'py_modules': self.args.py_modules,
                 },
                 extra_search_paths=self.args.extra_search_paths,
+                kwargs={'runtime_args': {}},
```
alaeddine-13, line 50: do we still need this ?

samsja: Yes, otherwise there are some errors in the parsing :thinking: it is a bit hacky but I don't think that it is worth the effort to investigate what is going on

jacobowitz (changes requested):
Can you please add a test showcasing the original issue? Just some mock runtime args, which cant be serialized to yaml?

samsja:
> Can you please add a test showcasing the original issue? Just some mock runtime args, which cant be serialized to yaml?

done

jacobowitz (approved):
that's very useful indeed!

PR #4582, opened 2022-04-05
fix: increase pytest timeout to 30

jacobowitz:
# Increase CD pytest timeout to 30 min

## Description

Recently CD have been flaky. It seems that random tests timeout. I checked and the pytest timeout is actually set to 15 minutes only. [In CI it is 30 minutes](https://github.com/jina-ai/jina/blob/master/.github/workflows/ci.yml#L292), so we should set the same value here.


JoanFM (approved):

samsja (approved):

jacobowitz:
black failure is [this problem ](https://github.com/jina-ai/jina/issues/4583)

Issue #4556, opened 2022-03-29
Support Gateway to Gateway communication


jacobowitz:
We decided that `standalone` Executors will always be started together with a Gateway. See also this [issue](https://github.com/jina-ai/jina/issues/4502) for more details.

This means that all external Executors in a `Flow` will be fronted by a Gateway (Sandbox executors being an exception for now). So the Gateway needs to be able to send its requests to another Gateway. This is not possible at the moment and should be implemented as part of this issue. For now we will only support gRPC here. So the external Executor's Gateway needs to expose gRPC.

Issue #4502, opened 2022-03-17
Expose Executor as a Service as First class citizen in Documentation.


JoanFM:
**Describe the feature**
Right now we have added `Executor.serve` method and we are seeing increasing interest in serving a single `Executor` as an isolated service.

After `Head` is removed as a result of the `service mesh` introduction, this will be even more performant, and `gateway` remains a useful entity in the picture. 

I think code-wise perspective, I belive the only needed thing is to add a method `Executor.to_k8s_yaml()` and `Executor.to_docker_compose_yaml` that under the hood create a simple `Flow` and call its `to_*` method.

The main work would be in documentation and storytelling about exposing this as a first-class option of `Jina`.

JoanFM:
Blocked by #4503

JohannesMessner:
Seems like a good thing to do.
I see some conceptual overlap with external Executors, and we probably need to be careful to make the distinction clear.

JoanFM:
> Seems like a good thing to do. I see some conceptual overlap with external Executors, and we probably need to be careful to make the distinction clear.

Good point, did not think about it, but it would be the biggest blocker

jacobowitz:
Let's restrict this for now to gRPC? Otherwise it will be more complicatd to integrate as external Executor

JohannesMessner:
> Let's restrict this for now to gRPC? Otherwise it will be more complicatd to integrate as external Executor

Makes sense. Remind me, did we land on `serve`d executors being the only recommended way of doing external Execs from now on?

JohannesMessner:
> Also we may want to think about exposing serve as part of CLI? 
Agree

jacobowitz:
We somehow should be able to deploy this with and without a gateway?

PR #4585, opened 2022-04-05
feat: add grpc reflection support

jacobowitz:
This PR allows external Executor to be fronted by our gRPC Gateway. This enables us to deploy external Executors together with a Gateway (as `Executor.serve()` is doing.). In theory this also allows us to integrate complete Flows as external Executors into another Flow. See the tests for some examples.

Implementation details:
I've added support for gRPC reflection with this PR. When we add a new connection, we first check the available gRPC endpoints using reflection. This way we can figure out automatically if the GrpcConnctionPool is used to target a Worker/Head or Gateway. This is necessary because they have different gRPC interfaces (streaming and non streaming). There is no configuration necessary to enable this. The reflection call is done lazily before the first request is send to the target.
 
- [x] Support reflection via grpc. Find out if the target is gateway or head/worker. Create the right stubs and hide this in the GrpcConnectionPool
- [x] Add test case for the g2g scenario
- [x] Check on parsing that merging is not required at the external Executor
- [x] Mention merge limitations in docs

Closes https://github.com/jina-ai/jina/issues/4556

Unblocks: https://github.com/jina-ai/jina/issues/4502

### DIFF
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -855,10 +855,7 @@ class GrpcConnectionPool:
 
     @staticmethod
     def create_async_channel_stub(
-        address,
-        tls=False,
-        root_certificates: Optional[str] = None,
-        summary: Optional['Summary'] = None
+        address, tls=False, root_certificates: Optional[str] = None, summary=None
     ) -> Tuple[_ConnectionStubs, grpc.aio.Channel,]:
         """
         Creates an async GRPC Channel. This channel has to be closed eventually!
@@ -866,6 +863,7 @@ class GrpcConnectionPool:
         :param address: the address to create the connection to, like 127.0.0.0.1:8080
         :param tls: if True, use tls for the grpc channel
         :param root_certificates: the path to the root certificates for tls, only u
+        :param summary: Optional Prometheus summary object
 
         :returns: DataRequest/ControlRequest stubs and an async grpc channel
         """
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 import ipaddress
 import os
+from collections import defaultdict
 from typing import Dict, List, Optional, Tuple
 from urllib.parse import urlparse
 
@@ -30,11 +31,12 @@ class ReplicaList:
     Maintains a list of connections to replicas and uses round robin for selecting a replica
     """
 
-    def __init__(self):
+    def __init__(self, summary):
         self._connections = []
         self._address_to_connection_idx = {}
         self._address_to_channel = {}
         self._rr_counter = 0
+        self.summary = summary
 
     def add_connection(self, address: str):
         """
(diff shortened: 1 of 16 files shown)


jina/serve/networking.py
```diff
@@ -28,11 +31,12 @@ class ReplicaList:
     Maintains a list of connections to replicas and uses round robin for selecting a replica
     """
 
-    def __init__(self):
+    def __init__(self, summary):
         self._connections = []
         self._address_to_connection_idx = {}
         self._address_to_channel = {}
         self._rr_counter = 0
+        self.summary = summary
```
JoanFM, line 39: why are u adding this, shouldn't this be part of another PR?

jina/serve/networking.py
```diff
@@ -145,13 +146,105 @@ class GrpcConnectionPool:
...
+            self._initialized = False
+
+        # This has to be done lazily, because the target endpoint may not be available
+        # when a connection is added
+        def _init_stubs(self):
+            available_services = GrpcConnectionPool.get_available_services(self.address)
+            stubs = defaultdict(lambda: None)
+            for service in available_services:
+                # skip the reflection service, it will not be used anymore
+                if service != 'grpc.reflection.v1alpha.ServerReflection':
```
JoanFM, line 174: maybe is better to filter this when `get_available_services` is obtained

jina/serve/networking.py
```diff
@@ -145,13 +146,105 @@ class GrpcConnectionPool:
...
+            self.control_stub = stubs['jina.JinaControlRequestRPC']
+            self.data_list_stub = stubs['jina.JinaDataRequestRPC']
+            self.single_data_stub = stubs['jina.JinaSingleDataRequestRPC']
+            self.stream_stub = stubs['jina.JinaRPC']
+            self._initialized = True
+
+        async def send_requests(
+            self, requests: List[Request], metadata, compression
+        ) -> Tuple:
+            if not self._initialized:
```
JoanFM, line 185: can't we avoid this and simply initialize at `__init__` time? I prefer a more stateless approach

jina/serve/networking.py
```diff
@@ -790,21 +855,15 @@ async def send_request_async(
 
     @staticmethod
     def create_async_channel_stub(
```
JoanFM, line 857: it is awkward for a public method to return a private class' instance

JoanFM (changes requested):

jina/serve/networking.py
jacobowitz: the metrics PR was merged already, but I am changing the ConnectionPool a bit and thus I need the prometheus summary object elswhere.
So I am passing it done to the new object introduced by this PR

jina/serve/networking.py
jacobowitz: unfortunately not. The lazy init here is connecting to the server's grpc reflection endpoint, but the __init__ is called on adding connections. At this time the server might not be ready yet. This is why I am delaying this until sending of the first data request. Then the server should be ready (or the request would fail anyway).

jina/serve/networking.py
jacobowitz: made the class public and added necessary docstrings

PR #4608, opened 2022-04-13
fix: sandbox tls

jacobowitz:
# Fix gRPC reflection using TLS

## Description

gRPC reflection is not respecting the TLS setttings. This makes all encrypted gRPC connections fail. This happens e.g. for sandbox.
This PR changes the logic so that we just reuse the existing channel for the reflection. This way I don't have to worry about the settings

### DIFF
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -902,12 +902,16 @@ class GrpcConnectionPool:
         response = reflection_stub.ServerReflectionInfo(
             iter([ServerReflectionRequest(list_services="")])
         )
+        service_names = []
         async for res in response:
-            return [
-                service.name
-                for service in res.list_services_response.service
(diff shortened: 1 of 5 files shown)


jina/serve/networking.py
```diff
@@ -166,8 +166,10 @@ def __init__(self, address, channel, summary):
 
         # This has to be done lazily, because the target endpoint may not be available
         # when a connection is added
-        def _init_stubs(self):
-            available_services = GrpcConnectionPool.get_available_services(self.address)
+        async def _init_stubs(self):
+            available_services = await GrpcConnectionPool.get_available_services(
+                self.channel
```
jacobowitz, line 171: this is the important change as we use the existing `self.channel` and not create a new one from `address`

mapleeit (approved):
I can understand the important change and the test passed. So looks good to me. 

But others can have a precise review since I'm not a Python expert.

Issue #4612, opened 2022-04-14
fix: hubble CI display message


samsja:
CI in hubble are poluted by rich status. @deepankarm @delgermurun 

![image](https://user-images.githubusercontent.com/55492238/163327958-6c63b925-21b4-40dc-a6de-b5737617ec0c.png)

samsja:
Seems like we still have the issue because of the `force terminal` option.

We could disable it, the only issue that we do need it to be rendered correctly in pycharm.

I need to clarify the issue and make it reproducable.

@deepankarm please add some screenshot and link to the issue in the CI

PR #4614, opened 2022-04-14
fix: remove force interactive in rich console

samsja:
fixing https://github.com/jina-ai/jina/issues/4612


samsja:
@deepankarm @delgermurun can you check that it is fix now in the ci ?

jacobowitz (approved):

PR #4635, opened 2022-04-19
feat: add process request seconds for executor

samsja:
Add `process_request_seconds` (on requests method) as a prometheus metrics to the Executor.

The goal of this metric is to monitor how much time each method decorated with the `@requests` takes to be processed

The metrics name is the same for all of the decorated method but there is a label for filtering.

Screenshot from 2022-04-19 15-29-08

### DIFF
--- a/jina/serve/executors/decorators.py
+++ b/jina/serve/executors/decorators.py
@@ -1,12 +1,23 @@
 """Decorators and wrappers designed for wrapping :class:`BaseExecutor` functions. """
-
+import contextlib
 import functools
 import inspect
 from functools import wraps
-from typing import Callable, Union, List, Optional, Dict, Sequence, TYPE_CHECKING
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    ContextManager,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Union,
+)
 
-from jina.serve.executors.metas import get_default_metas
+from prometheus_client import Summary
+
 from jina.helper import convert_tuple_to_list, iscoroutinefunction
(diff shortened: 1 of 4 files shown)


jina/serve/executors/decorators.py
```diff
@@ -129,6 +147,15 @@ def __set_name__(self, owner, name):
 
             setattr(owner, name, self.fn)
 
+        def _get_summary(
+            self, summary: Optional['Summary'], method_name: str
+        ) -> ContextManager:
+            return (
+                summary.labels(method_name).time()
```
jacobowitz, line 154: Can we also add the executor name as a label?

jina/serve/executors/decorators.py
```diff
@@ -104,15 +115,22 @@ def __init__(self, fn):
...
+                        return await fn(self2, *args, **kwargs)
 
                 self.fn = arg_wrapper
             else:
 
                 @functools.wraps(fn)
-                def arg_wrapper(*args, **kwargs):
-                    return fn(*args, **kwargs)
+                def arg_wrapper(
+                    self2, *args, **kwargs
```
jacobowitz, line 129: maybe lets rename self2 to sth else? Is this conceptually a `self`? Its just `executor_instance` or sth no??

jacobowitz (changes requested):
this is pretty cool!

jina/serve/executors/decorators.py
samsja: Great idea, I will add it

jacobowitz (approved):
this will be very helpful!

jacobowitz (approved):

PR #4638, opened 2022-04-19
feat: add receiving request second to gateway

samsja:


jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -1,112 +1,155 @@
 import asyncio
 import copy
-from typing import TYPE_CHECKING, Callable, List
+import time
+from typing import TYPE_CHECKING, Callable, List, Optional
 
 from docarray import DocumentArray
+from prometheus_client import Summary
```
JoanFM, line 7: this should be hidden

jina/serve/runtimes/head/__init__.py
```diff
@@ -10,6 +11,7 @@
 
 import grpc
 from grpc_reflection.v1alpha import reflection
+from prometheus_client import Summary
```
JoanFM, line 14: same here

JoanFM (changes requested):
Hide all the prometheus client imports so that they are only imported if the feature is activated

JoanFM (approved):

PR #4639, opened 2022-04-20
docs(sandbox): sandbox is serverless now

mapleeit:
# Update documentation about sandbox's lifecycle

## Description

Sandbox won't be deleted if there is no traffic but scale down to 0.

### DIFF
--- a/docs/how-to/sandbox.md
+++ b/docs/how-to/sandbox.md
@@ -31,7 +31,7 @@ This starts a Flow that only has one Executor, and sends a Document to it. The D
 
 ## Sandbox Lifecycle
 
-The sandbox will not be removed immediately after the Flow is closed. It will be kept alive until there is no traffic during this certain period. The default period is currently 15 minutes.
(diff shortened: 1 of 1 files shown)


docs/how-to/sandbox.md
```diff
@@ -31,7 +31,7 @@ This starts a Flow that only has one Executor, and sends a Document to it. The D
 
 ## Sandbox Lifecycle
 
-The sandbox will not be removed immediately after the Flow is closed. It will be kept alive until there is no traffic during this certain period. The default period is currently 15 minutes.
+Sandbox is serverless. It will not be removed immediately after the Flow is closed but keep alive for several minutes. If there is no traffic, it will automatiaclly scale down to 0 to save resources, but it will restart again and give the response whenever there is a new request has been sent to the same Sandbox.
```
JoanFM, line 34:
```suggestion
Sandbox is serverless. It will not be removed immediately after the Flow is closed but kept alive for several minutes. If there is no traffic, it will automatically scale down to 0 to save resources, but it will restart again and give the response whenever a new request is sent to the same Sandbox.
```

JoanFM (changes requested):

JoanFM (approved):

PR #4641, opened 2022-04-20
fix: remove prometheus import from the top of the files

samsja:
We want to be able to remove prometheus from core dependency if needed. Therefore we need to remove the import from the top of the file

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -163,22 +163,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ 3.7 ]
-        test-path: ${{fromJson(needs.prep-testbed.outputs.matrix)}}
+        core: ['', 'true']
+        perf: ['', 'true']
+        exclude: 
+          - core: 'true'
+            perf: 'true'
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python 3.7
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.7
       - name: Prepare enviroment
         run: |
           python -m pip install --upgrade pip
           python -m pip install wheel
           pip install --no-cache-dir .
+        env:
+          JINA_PIP_INSTALL_CORE: ${{ matrix.core }}
+          JINA_PIP_INSTALL_PERF: ${{ matrix.perf }}
       - name: Test basic import
         run: python -c 'from jina import Executor,requests'
-
   docker-image-test:
     needs: update-schema
     runs-on: ubuntu-latest
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -157,6 +157,28 @@ jobs:
           flags: ${{ steps.test.outputs.codecov_flag }}
           fail_ci_if_error: false
 
+  import-test:
+    runs-on: ubuntu-latest
+    needs: [ commit-lint, lint-flake-8, code-injection ]
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ 3.7 ]
+        test-path: ${{fromJson(needs.prep-testbed.outputs.matrix)}}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Prepare enviroment
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install wheel
+          pip install --no-cache-dir .
+      - name: Test basic import
+        run: python -c 'from jina import Executor,requests'
+
   docker-image-test:
     needs: update-schema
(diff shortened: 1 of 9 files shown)


jina/serve/executors/__init__.py
```diff
@@ -124,7 +122,12 @@ def _add_runtime_args(self, _runtime_args: Optional[Dict]):
             self.runtime_args = SimpleNamespace()
 
     def _init_monitoring(self):
-        if hasattr(self.runtime_args, 'metrics_registry'):
+        if (
+            hasattr(self.runtime_args, 'metrics_registry')
```
JoanFM, line 126: can't we make sure `metrics_registry` is always in the `runtime_args`?

JoanFM (changes requested):

jina/serve/executors/__init__.py
samsja: Yes we should you are right

JoanFM: then let's do it here

JoanFM (changes requested):

jina/serve/executors/__init__.py
```diff
@@ -124,7 +122,12 @@ def _add_runtime_args(self, _runtime_args: Optional[Dict]):
             self.runtime_args = SimpleNamespace()
 
     def _init_monitoring(self):
-        if hasattr(self.runtime_args, 'metrics_registry'):
+        if not hasattr(self.runtime_args, 'metrics_registry'):
```
JoanFM, line 125: this should not be done here. This should be handled in the setting of these `arguments`. `_init_monitoring` should not alter the `runtime_args`

JoanFM (changes requested):

jina/serve/executors/__init__.py
```diff
@@ -124,7 +122,12 @@ def _add_runtime_args(self, _runtime_args: Optional[Dict]):
...
+            self.runtime_args.metrics_registry = None
+
+        if self.runtime_args.metrics_registry:
+            from prometheus_client import Summary
+
```
JoanFM, line 130: use ImportExtensions to guide the user in case not added

JoanFM (approved):

.github/workflows/ci.yml
```diff
@@ -304,10 +304,33 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Prepare enviroment
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install wheel
+          pip install --no-cache-dir .
+      - name: Test basic import
+        run: python -c 'from jina import Executor,requests'
```
mapleeit, line 327:
```suggestion
  import-test:
    runs-on: ubuntu-latest
    needs: [ commit-lint, lint-flake-8, code-injection ]
    strategy:
      fail-fast: false
      matrix:
        core: ['', 'true']
        perf: ['', 'true']
        exclude: 
          - core: 'true'
            perf: 'true'
    steps:
      - uses: actions/checkout@v2
      - name: Set up Python 3.7
        uses: actions/setup-python@v2
        with:
          python-version: 3.7
      - name: Prepare enviroment
        run: |
          python -m pip install --upgrade pip
          python -m pip install wheel
          pip install --no-cache-dir .
        env:
          JINA_PIP_INSTALL_CORE: ${{ matrix.core }}
          JINA_PIP_INSTALL_PERF: ${{ matrix.perf }}
      - name: Test basic import
        run: python -c 'from jina import Executor,requests'
```

.github/workflows/cd.yml
```diff
@@ -157,6 +157,28 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Prepare enviroment
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install wheel
+          pip install --no-cache-dir .
+      - name: Test basic import
+        run: python -c 'from jina import Executor,requests'
+
```
mapleeit, line 181:
```suggestion
  import-test:
    runs-on: ubuntu-latest
    needs: [ commit-lint, lint-flake-8, code-injection ]
    strategy:
      fail-fast: false
      matrix:
        core: ['', 'true']
        perf: ['', 'true']
        exclude: 
          - core: 'true'
            perf: 'true'
    steps:
      - uses: actions/checkout@v2
      - name: Set up Python 3.7
        uses: actions/setup-python@v2
        with:
          python-version: 3.7
      - name: Prepare enviroment
        run: |
          python -m pip install --upgrade pip
          python -m pip install wheel
          pip install --no-cache-dir .
        env:
          JINA_PIP_INSTALL_CORE: ${{ matrix.core }}
          JINA_PIP_INSTALL_PERF: ${{ matrix.perf }}
      - name: Test basic import
        run: python -c 'from jina import Executor,requests'
```

mapleeit:
@samsja @JoanFM I optimized the `import-test` a little bit to cover the case of use `jina[core]` and `jina[perf]`, can you review if it makes sense?

mapleeit:
btw, the two failed tests can be ignored. We need to release a patch to fix that.

mapleeit (approved):

PR #4644, opened 2022-04-20
refactor: remove runtime backend cli arg

JoanFM:
# Pull Request Title

## Description

Remove Runtime Backend since THREAD Is not a good choice and should not be supported


samsja (approved):
lgtm

PR #4646, opened 2022-04-20
feat: pass environment to head and gateway

JoanFM:
# Pull Request Title

## Description

Allow to pass the environments passed to Executors to the Head, and the one from Flow to the Gateway


jacobowitz (changes requested):
lets add a test for this feature?

JoanFM:
> lets add a test for this feature?

How would u do so?

jacobowitz:
> > lets add a test for this feature?
> 
> How would u do so?

Write a test with a Flow where the envs are set differently for the Gateway and the head and then assert on that?

JoanFM:
> > > lets add a test for this feature?
> > 
> > 
> > How would u do so?
> 
> Write a test with a Flow where the envs are set differently for the Gateway and the head and then assert on that?

but I have no access to the env variables on those Processes right?

PR #4650, opened 2022-04-21
fix: set the multiprocessing start_method spawn

numb3r3:
# Pull Request Title

## Description

Failed to set the multiprocessing start_method `spawn` via `JINA_MP_START_METHOD`

```python
# example.py

from jina import Flow

if __name__ == '__main__':
    with Flow().add() as f:
        f.block()
```

Start the flow service
```
JINA_MP_START_METHOD=spawn python example.py
```
it raises the following error
```
RuntimeError: context has already been set
```

### DIFF
--- a/jina/__init__.py
+++ b/jina/__init__.py
@@ -39,13 +39,18 @@ _os.environ['OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'
 
 # JINA_MP_START_METHOD has higher priority than os-patch
 _start_method = _os.environ.get('JINA_MP_START_METHOD', None)
-
 if _start_method and _start_method.lower() in {'fork', 'spawn', 'forkserver'}:
     from multiprocessing import set_start_method as _set_start_method
 
-    _set_start_method(_start_method.lower())
-    _warnings.warn(f'multiprocessing start method is set to `{_start_method.lower()}`')
-    _os.environ.pop('JINA_MP_START_METHOD')
+    try:
+        _set_start_method(_start_method.lower())
+        _warnings.warn(
+            f'multiprocessing start method is set to `{_start_method.lower()}`'
+        )
+    except Exception as e:
+        _warnings.warn(
+            f'failed to set multiprocessing start_method to `{_start_method.lower()}`: {e!r}'
+        )
 elif _sys.version_info >= (3, 8, 0) and _platform.system() == 'Darwin':
     # DO SOME OS-WISE PATCHES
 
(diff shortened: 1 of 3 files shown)


jina/__init__.py
```diff
@@ -55,6 +60,7 @@ def _warning_on_one_line(message, category, filename, lineno, *args, **kwargs):
 
     _set_start_method('fork')
```
JoanFM, line 61: u have to do the same here

jina/__init__.py
```diff
@@ -39,13 +39,18 @@ def _warning_on_one_line(message, category, filename, lineno, *args, **kwargs):
...
-    _set_start_method(_start_method.lower())
-    _warnings.warn(f'multiprocessing start method is set to `{_start_method.lower()}`')
-    _os.environ.pop('JINA_MP_START_METHOD')
+    try:
+        _set_start_method(_start_method.lower())
+        _warnings.warn(
+            f'multiprocessing start method is set to `{_start_method.lower()}`'
+        )
+    except Exception as e:
+        _warnings.warn(
```
JoanFM, line 51: Please guide the user to say that most likely it was already set

JoanFM (changes requested):
Please consider comments

jina/__init__.py
numb3r3: I tested the case again and found it is not necessary to apply the try-catch here.

jina/__init__.py
numb3r3: the exception `{e!r}` explains the context is already set

jina/__init__.py
JoanFM: because u removed the `pop` of the env variable.

JoanFM (dismissed):

numb3r3:
since the issue https://github.com/grpc/grpc/issues/29044, we need pin grpcio < 1.44.0.

JoanFM:
> since the issue [grpc/grpc#29044](https://github.com/grpc/grpc/issues/29044), we need pin grpcio < 1.44.0.

Or do this https://github.com/grpc/grpc/issues/29044#issuecomment-1062773527

jina/__init__.py
numb3r3: I think the macos patch above takes effect only when `JINA_MP_START_METHOD` is not set. Hence, try-catch does not make sense (the exception should be raised in this case).

numb3r3:
From https://github.com/grpc/grpc/issues/29044#issuecomment-1063434288
> The poll strategy is generally less performant on linux than either of the epoll options, changing that default would hurt a lot of folks.

JoanFM (approved):

Issue #4630, opened 2022-04-14
Clean up timeout handling


jacobowitz:
Clean up timeout handling: Make sure its consistent and configurable throughout the code base. There should be timeouts everywhere

PR #4660, opened 2022-04-21
feat: add timeout_send

jacobowitz:
This PR adds a timeout for sending data requests over grpc. This applies to gateway and head, both using the ConnectionPool.
The parameter is 'timeout-send' and needs to be specified in ms. It is disabled by default (otherwise it conflicts somewhat with disabled prefetch.

This closes https://github.com/jina-ai/jina/issues/4630


jina/parsers/orchestrate/pod.py
```diff
@@ -37,6 +37,13 @@ def mixin_pod_parser(parser):
         'forever',
     )
 
+    gp.add_argument(
+        '--timeout-send',
```
JoanFM, line 41: this should go into gateway and Head right?

jina/serve/networking.py
```diff
@@ -181,7 +182,11 @@ async def _init_stubs(self):
             self._initialized = True
 
         async def send_requests(
-            self, requests: List[Request], metadata, compression
+            self,
+            requests: List[Request],
+            metadata,
+            compression,
+            timeout: Optional[float] = 1.0,
```
JoanFM, line 189: let's have the same default in the library as in the parsers for sanity and clarity

JoanFM (changes requested):

jina/serve/runtimes/asyncio.py
```diff
@@ -148,7 +148,9 @@ def is_ready(ctrl_address: str, **kwargs) -> bool:
         """
 
         try:
-            GrpcConnectionPool.send_request_sync(ControlRequest('STATUS'), ctrl_address)
+            GrpcConnectionPool.send_request_sync(
+                ControlRequest('STATUS'), ctrl_address, timeout=1.0
```
JoanFM, line 152: why is here `timeout` hardcoded to 1?

jina/parsers/orchestrate/runtimes/head.py
```diff
@@ -46,3 +46,10 @@ def mixin_head_parser(parser):
         default=False,
         help='Disable the built-in reduce mechanism, set this if the reduction is to be handled by the Executor connected to this Head',
     )
+
+    gp.add_argument(
+        '--timeout-send',
+        type=int,
+        default=-1,
```
JoanFM, line 53: can it be declared None?

JoanFM (changes requested):

jina/serve/runtimes/asyncio.py
jacobowitz: We use this part during the ready check of the Pods, which is controlled by the `timeout_ready`. The way it is implemented atm is that we try the STATUS request multiple times until `timeout_ready` is hit. What we use as the timeout for the STATUS request itself here does not really matter. I think 1 second is okay, it could also be 0.5 or 2 seconds.

jina/parsers/orchestrate/runtimes/head.py
jacobowitz: I checked and it seems to be working. Good point, `None` is much better. Also the user never has to set `None` as it is the default. The user just has to provide now a positive integer if they want to use it. Otherwise leave as is.

jina/serve/runtimes/asyncio.py
JoanFM: Understand

JoanFM (approved):

PR #4672, opened 2022-04-21
feat(monitoring): count the number of processed documents

samsja:


jina/serve/executors/__init__.py
```diff
@@ -250,10 +251,20 @@ async def __acall__(self, req_endpoint: str, **kwargs):
...
     async def __acall_endpoint__(self, req_endpoint, **kwargs):
         func = self.requests[req_endpoint]
-        if iscoroutinefunction(func):
-            return await func(self, **kwargs)
-        else:
-            return func(self, **kwargs)
+
+        _summary = (
+            self._summary_method.labels(
+                func.__name__, self.__class__.__name__, req_endpoint
```
jacobowitz, line 257: I am not sure, but maybe we should remove the method name as a label now that we always also expose endpoint? Its kind of redundant now? I would think that endpoint/executor tuple should be good enough?

jacobowitz (commented):
I just have one question, otherwise it looks good to me!

jina/serve/executors/__init__.py
JoanFM: Yes makes sense

samsja: You are right I will do

jacobowitz (approved):
lgtm

PR #4676, opened 2022-04-24
feat: immutable tags for executors

nomagick:
Goals:

- Allow user to push executor with immutable tags to Jina Hub using the --immutable-tag option.
- Document this new feature inside the docs

Tested locally.
Code changed by this pr is not covered in automated tests.

Initially suggested by @deepankarm via https://github.com/jina-ai/hubble/issues/371
https://jinaai.slack.com/archives/C023A7RAPGR/p1649245281097999

### DIFF
--- a/docs/fundamentals/executor/hub/push-executor.md
+++ b/docs/fundamentals/executor/hub/push-executor.md
@@ -68,9 +68,11 @@ If you want to create a new tag for an existing Executor, you can also add the `
 jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> -t TAG <path_to_executor_folder>
 ```
 
-If you want to create an immutable tag, you can leverage the `--immutable-tag` option. 
+### Immutable tags
+
+If you want to create immutable tags, you can leverage the `--immutable-tag` option. 
 After being pushed for the first time, the immutable tags can not be pushed again.
 
 ```bash
-jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG> <path_to_executor_folder>
+jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG_1> --immutable-tag <IMMUTABLE_TAG_2> <path_to_executor_folder>
 ```
--- a/docs/fundamentals/executor/hub/push-executor.md
+++ b/docs/fundamentals/executor/hub/push-executor.md
@@ -68,7 +68,8 @@ If you want to create a new tag for an existing Executor, you can also add the `
 jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> -t TAG <path_to_executor_folder>
 ```
 
-If you want to create an immutable tag, that could not be modified again after push, you can leverage the `--immutable-tag` option:
+If you want to create an immutable tag, you can leverage the `--immutable-tag` option. 
+After being pushed for the first time, the immutable tags can not be pushed again.
 
 ```bash
 jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG> <path_to_executor_folder>
--- a/docs/fundamentals/executor/hub/push-executor.md
+++ b/docs/fundamentals/executor/hub/push-executor.md
@@ -70,7 +70,10 @@ jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> -t TA
 
 ### Immutable tags
 
-If you want to create immutable tags, you can leverage the `--immutable-tag` option. 
+If you don’t want some tags to be later overwritten to keep a stable, consistent behavior, 
+immutable tags are the exact thing you are looking for.
+
+You can leverage the `--immutable-tag` option to create immutable tags. 
 After being pushed for the first time, the immutable tags can not be pushed again.
 
 ```bash
--- a/docs/fundamentals/executor/hub/push-executor.md
+++ b/docs/fundamentals/executor/hub/push-executor.md
@@ -67,3 +67,9 @@ If you want to create a new tag for an existing Executor, you can also add the `
 ```bash
 jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> -t TAG <path_to_executor_folder>
 ```
+
+If you want to create an immutable tag, that could not be modified again after push, you can leverage the `--immutable-tag` option:
+
+```bash
+jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG> <path_to_executor_folder>
+```
--- a/jina/parsers/hubble/push.py
+++ b/jina/parsers/hubble/push.py
@@ -50,6 +50,12 @@ One can later fetch a tagged Executor via `jinahub[+docker]://MyExecutor/gpu`
     )
 
     gp.add_argument(
+        '--immutable-tag',
+        action='append',
+        help='A list of immutable tags. Like tag but immutable after first push. ',
+    )
+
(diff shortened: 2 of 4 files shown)


mapleeit:
Is this PR ready for review? I saw it's still WIP @nomagick

mapleeit (dismissed):

nomagick:
> Is this PR ready for review? I saw it's still WIP
Yes I was waiting for the tests

mapleeit (dismissed):

docs/fundamentals/executor/hub/push-executor.md
```diff
@@ -67,3 +67,10 @@ If you want to create a new tag for an existing Executor, you can also add the `
 ```bash
 jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> -t TAG <path_to_executor_folder>
 ```
+
+If you want to create an immutable tag, you can leverage the `--immutable-tag` option. 
+After being pushed for the first time, the immutable tags can not be pushed again.
+
+```bash
+jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG> <path_to_executor_folder>
```
mapleeit, line 75:
```suggestion
### Immutable tags

If you want to create immutable tags, you can leverage the `--immutable-tag` option. 
After being pushed for the first time, the immutable tags can not be pushed again.

```bash
jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG_1> --immutable-tag <IMMUTABLE_TAG_2> <path_to_executor_folder>
```

jina/parsers/hubble/push.py
```diff
@@ -49,6 +49,12 @@ def dir_path(string):
 ''',
     )
 
+    gp.add_argument(
+        '--immutable-tag',
+        action='append',
+        help='A list of immutable tags. Like tag but immutable after first push. ',
```
mapleeit, line 55:
```suggestion
        help='A list of immutable tags. Like tag but immutable after first push.',
```

mapleeit (dismissed):

docs/fundamentals/executor/hub/push-executor.md
```diff
@@ -67,3 +67,12 @@ If you want to create a new tag for an existing Executor, you can also add the `
 ```bash
 jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> -t TAG <path_to_executor_folder>
 ```
+
+### Immutable tags
```
JoanFM, line 71: Can we have a description of why is this desirable? I cannot see the point of the feature with this description

JoanFM (changes requested):

hanxiao (changes requested):
do we have a better name for it? plus i also think the necessity of such feature is chanllegable. We have many meta fields in `manifest.yml`, why `tags` is more special than `name`, `author`, `url`, ...? will we also add `--immutable-name`, `-immutable-author`, `--immutable-url` to the CLI?

mapleeit:
> do we have a better name for it? plus i also think the necessity of such feature is chanllegable. We have many meta fields in `manifest.yml`, why `tags` is more special than `name`, `author`, `url`, ...? will we also add `--immutable-name`, `-immutable-author`, `--immutable-url` to the CLI?

`tags` indeed is more special than other meta info. **It will affect how people use this Executor, others will not.** We are following docker tags' strategy. Tags can always be overwritten. This strategy has several benefits:

- Allow `latest`, `v1` can change as time goes, users don't need to change the calling codes all the times if they want to use latest version or major version. 

But it also has drawbacks:

- We can't make sure that the content any of tags are always the same. 
- People could overwrite the existing tag by accident.

@deepankarm requested this feature in the past, what's your scenarios? Do you still need it?

docs/fundamentals/executor/hub/push-executor.md
mapleeit: @nomagick Please click the "Resolve conversation" if you think this comment has been addressed.

JoanFM (approved):

deepankarm:
> > do we have a better name for it? plus i also think the necessity of such feature is chanllegable. We have many meta fields in `manifest.yml`, why `tags` is more special than `name`, `author`, `url`, ...? will we also add `--immutable-name`, `-immutable-author`, `--immutable-url` to the CLI?
> 
> `tags` indeed is more special than other meta info. **It will affect how people use this Executor, others will not.** We are following docker tags' strategy. Tags can always be overwritten. This strategy has several benefits:
> 
> * Allow `latest`, `v1` can change as time goes, users don't need to change the calling codes all the times if they want to use latest version or major version.
> 
> But it also has drawbacks:
> 
> * We can't make sure that the content any of tags are always the same.
> * People could overwrite the existing tag by accident.
> 
> @deepankarm requested this feature in the past, what's your scenarios? Do you still need it?

This feature is absolutely helpful. Whenever we release a new version of [docsQA](https://github.com/jina-ai/docsQA), we want to make sure someone doesn't overwrite another version of this Executor with the same tag. The only way to make sure of that is, by making the `tag` immutable. This is unrelated to metadata.

docs/fundamentals/executor/hub/push-executor.md
```diff
@@ -67,3 +67,15 @@ If you want to create a new tag for an existing Executor, you can also add the `
...
+
+If you don’t want some tags to be later overwritten to keep a stable, consistent behavior, 
+immutable tags are the exact thing you are looking for.
+
+You can leverage the `--immutable-tag` option to create immutable tags. 
+After being pushed for the first time, the immutable tags can not be pushed again.
+
+```bash
+jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG_1> --immutable-tag <IMMUTABLE_TAG_2> <path_to_executor_folder>
```
deepankarm, line 80: My assumption for the usage of `immutable` was :point_down: . This would make sure, a certain tag is immutable. 
```bash
jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --tag v1 --immutable 
```

Not :point_down: . Any reason we want `tag` & `immutable-tag` both?

```bash
jina hub push [--public/--private] --force-update <NAME> --secret <SECRET> --immutable-tag <IMMUTABLE_TAG_1> --immutable-tag <IMMUTABLE_TAG_2> <path_to_executor_folder>
```

nomagick: Tag is an array, so there could be multiple `--tag`. 
I'm not familiar with `argparse`, Is there a proper way to distinguish which `--tag` the `--immutable` is for when there's multiple `--tag`?

deepankarm: I'd say if `--immutable` is passed, we mark all tags immutable.

mapleeit: @deepankarm There could be some situaction we want to use them both at the same time. For example a common scenerio, I want v1.2.3 immutable but v1 mutable. Then I can write 

`jina hub push --tag v1 --immutable-tag v1.2.3 .`

hanxiao:
can we change the name to `--readonly`?

mapleeit:
> can we change the name to `--readonly` / **`--protected` as opposed to the current `--public`/`--private`**? without the suffix `tag` in `--readonly-tag` and `--protected-tag`?
> 
> 1. `immutable` is very low-level term
> 2. you are basically protecting a version of the executor, you are not protecting the literal tag

We (me, @nomagick @delgermurun) discussed the naming part. 

We think it makes sense to change it from `--immutable-tag` to `--protected-tag`. But can't remove the `-tag` suffix. `--tag` and `--protected-tag` are a pair. Removing the `-tag` suffix will be confusing.

Issue #4626, opened 2022-04-14
Graceful Pod termination


jacobowitz:
Handle Pod termination more graceful: Check if the server implementations can be improved, but this is mainly on the Caller side. Make sure that the Caller can detect cancelled grpc calls (because the remote endpoint terminated) and retry them. Dont retry other errors (executor problems etc)

PR #4682, opened 2022-04-25
feat: graceful shutdown

jacobowitz:
This PR makes the termination of a Pod in Kubernetes Graceful. I also enabled the tests, which were failing before. With this PR there will be no requests lost if a Kubernetes Pod is terminated. Outside of Kubernetes this does not work, because we have no way of removing the dead replica.

The change is actually simple: When a Pod is terminated, its outstanding grpc calls will be cancelled by the server. This is a specific gRPC error. This error is caught now and the request is retried. The retry will then hit one of the remaining Pods. 

This closes [4626](https://github.com/jina-ai/jina/issues/4626)

### DIFF
--- a/jina/serve/runtimes/worker/__init__.py
+++ b/jina/serve/runtimes/worker/__init__.py
@@ -100,8 +100,8 @@ class WorkerRuntime(AsyncNewLoopRuntime, ABC):
         self.logger.debug('Cancel WorkerRuntime')
 
         # 0.5 gives the runtime some time to complete outstanding responses
-        # this should be handled better, 0.5 is a rather random number
-        await self._grpc_server.stop(0.5)
+        # this should be handled better, 1.0 is a rather random number
+        await self._grpc_server.stop(1.0)
         self.logger.debug('Stopped GRPC Server')
 
     async def async_teardown(self):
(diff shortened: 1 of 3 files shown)


jina/serve/runtimes/worker/__init__.py
```diff
@@ -100,8 +100,8 @@ async def async_cancel(self):
         self.logger.debug('Cancel WorkerRuntime')
 
         # 0.5 gives the runtime some time to complete outstanding responses
-        # this should be handled better, 0.5 is a rather random number
-        await self._grpc_server.stop(0.5)
+        # this should be handled better, 1.0 is a rather random number
+        await self._grpc_server.stop(1.0)
```
JoanFM, line 104: there is no way to handle this better? even without timeout?

jacobowitz: I dont think there is. What this does is "Stop accepting new requests" -> "wait for grace period" -> "cancel non completed requests".
We have no control over this otherwise and I think the general mechanism is fine. Just the grace period is a random value. Its not needed for this PR, we can also leave it as it was. I was just thinking 1s is a bit better than 0.5s.

JoanFM (approved):

samsja (approved):

PR #4684, opened 2022-04-25
fix: fix configuration of addresses in docker compose

JoanFM:
Goals:
Problem when using `to_docker_compose` of a Flow, the gateway deployment addresses to the Executor replicas is not properly set.

This PR fixes it.

Logic is:

- Local and Docker-compose: Gateway connects directly to each replicas doing the load balancing
- K8s: Gateway connects to a single address corresponding to the LinkerD or lstio load balancer


jacobowitz (approved):
lgtm

PR #4689, opened 2022-04-25
fix(dependency): prometheus client should be a core dependency

samsja:
Prometheus client should be a core dependency otherwise we can't monitor executor running in k8s


JoanFM (approved):

nan-wang:
this sounds a bit wired for me. As the `prometheus` is used with `k8s`, I am supposing that `prometheus-client` should be the same level of `kubernetes`.

samsja:
@nan-wang we changed it to perf recently. However `prometheus` and the monitoring feature are independent from `k8s`, it can be used without it

PR #4695, opened 2022-04-26
fix: fix dependency

samsja:
Goals: delete useless dependency from core to keep it as minimal as possible

- [x] remove `prometheus_client` from core dependency
- [x] remove `packaging` from core dependency

### DIFF
--- a/cli/__init__.py
+++ b/cli/__init__.py
@@ -3,8 +3,6 @@ import shutil
 import subprocess
 import sys
 
-from packaging.version import Version, parse
-
 
 def _get_run_args(print_args: bool = True):
     from jina.helper import get_rich_console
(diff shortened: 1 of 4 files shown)


JoanFM (approved):

samsja:
Please don't merge we still need to take care of the packaging things ( @JohannesMessner )

JohannesMessner:
Like suggested, I changed to using `distutils` for version parsing.
But note that **distutils will be deprecated and removed**: https://peps.python.org/pep-0632/

Also, version checking for plugins seems to not work when using multiprocessing, but this is a different issue.

JohannesMessner:
I will remove distutils and replace it with packaging, instead of the other way around. Low prio for now, will return to this.

JohannesMessner:
Looks like `distutils` is already gone on master. Closing.

PR #4698, opened 2022-04-26
fix: parsing requirements of executor

numb3r3:
Goals:
To allow the following URI or VCS requirements can be parsed correctly:
```
git+https://github.com/huggingface/transformers.git
git+https://github.com/patil-suraj/vqgan-jax.git
git+https://github.com/borisdayma/dalle-mini.git
```

### DIFF
--- a/tests/unit/hubble/test_helper.py
+++ b/tests/unit/hubble/test_helper.py
@@ -97,6 +97,7 @@ def test_install_requirements():
             [],
             ['dev'],
         ),
+        ('http://pypi.python.org/packages/source/p/jina/jina.tar.gz', 'jina', [], []),
     ],
 )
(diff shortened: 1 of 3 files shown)


tests/unit/hubble/test_helper.py
```diff
@@ -83,6 +83,18 @@ def test_install_requirements():
     )
 
 
+def test_parse_requirements():
```
JoanFM, line 86: I think a better test should be built.

JoanFM (changes requested):

tests/unit/hubble/test_helper.py
numb3r3: a stronger test is added.

JoanFM (approved):

PR #4715, opened 2022-04-28
fix: fix grpcio-reflection version

JoanFM:
Goals:
Fix the conda recipe building


nan-wang:
Unpin the version restriction in #4650 . 
This is an upstream  issue and it is fixed. https://github.com/grpc/grpc/issues/29044

PR #4716, opened 2022-04-28
fix: remove version check from cli

JoanFM:


hanxiao:
btw this is the same as moving `cli` module completely under `jina` module, where the latter is tbh more clear

PR #4720, opened 2022-04-28
feat: allow user not to print version check print

JoanFM:


jina/__init__.py
```diff
@@ -128,7 +128,7 @@ def _warning_on_one_line(message, category, filename, lineno, *args, **kwargs):
     _os.path.dirname(_sys.modules['jina'].__file__), 'resources'
 )
 
-if _os.environ.get('JINA_CHECK_VERSION', None):
+if _os.environ.get('JINA_CHECK_VERSION', None) == 'True':
```
hanxiao, line 131: best practice on checking env is the following
```python
'JINA_CHECK_VERSION' in os.environ
```

Issue #4725, opened 2022-04-29
Client.encode() does not close file descriptors used.


MrBeeMovie:
I am calling .encode() multiple times and it seems as though file descriptors are being opened but not closed which results in this error after running for a period of time.

```
E0429 15:10:27.741673000 4681657856 wakeup_fd_pipe.cc:39]              pipe creation failed (24): Too many open files
E0429 15:10:27.744256000 4681657856 ev_poll_posix.cc:1415]             Skipping poll because of no wakeup fd.
E0429 15:10:27.744857000 4681657856 ev_posix.cc:228]                   No event engine could be initialized from all
```

when running `lsof | wc -l` I start to get crazy numbers of open files after a while.

You can run this test code with clip-server running to expose leak as fast as possible:

```python
from clip_client import Client


c = Client("grpc://localhost:51000")

while True:
   c.encode([" "]) 
```

hanxiao:
A better visualization on diff

```python
import os
import subprocess
import sys
import webbrowser
from difflib import HtmlDiff

from jina import Flow, Client

pid = os.getpid()
print('pid', pid)
before = subprocess.getoutput(f'lsof -p {pid}')
with Flow() as f:
    c = Client(port=f.port)
    for _ in range(10):
        c.post('/')
    after = subprocess.getoutput(f'lsof -p {pid}')

with open('diff.html', 'w') as fp:
    fp.writelines(HtmlDiff().make_file(before.split('\n'), after.split('\n')))

webbrowser.open('file://' + os.path.realpath('diff.html'), new=2)


```


<img width="1616" alt="image" src="https://user-images.githubusercontent.com/2041322/166095488-a07ba663-9454-4145-9b36-fb377872a950.png">

hanxiao:
Related issue to help core team:

https://github.com/QuarkChain/pyquarkchain/issues/411#issuecomment-484720553
https://github.com/python/cpython/issues/77262#issuecomment-1093779482

MrBeeMovie:
@hanxiao Thank you so much for your efforts! Hope this gets fixed soon.

jacobowitz:
I think our `run_async` helper function is at fault here. I got an even simpler mwu to reproduce with this:

`import psutil

from jina.helper import run_async


async def do_something():
    result = []
    result.append('hello')

for i in range(1):
    run_async(do_something)

p = psutil.Process()
print(p.num_fds())`

jacobowitz:
[...]

PR #4734, opened 2022-05-02
fix: close loop from run_async

jacobowitz:
We are leaking `asyncio` loops in `run_ascynio`. We use `get_or_reuse_loop()` there to create a new `asyncio` loop in each call. This loop is then used in `run_until_complete`. This completes the coroutine, but does NOT close the loop.

What is strange is that we are trying to set the new event loop and later reuse it. That mechanism does not appear to be working as intended. I think we may want to raise a separate issue for this and use this as the quick fix?

I also add `psutil` as a new dependency for testing. I could also do its purpose manually, but it seems more convenient to use this library here and I dont think we need to be too restrictive for CI purposes?

Fixes: https://github.com/jina-ai/jina/issues/4725


jina/helper.py
```diff
@@ -1322,7 +1322,10 @@ def run(self):
                 'please report this issue here: https://github.com/jina-ai/jina'
             )
     else:
-        return get_or_reuse_loop().run_until_complete(func(*args, **kwargs))
+        loop = get_or_reuse_loop()
+        result = loop.run_until_complete(func(*args, **kwargs))
+        loop.close()
```
deepankarm, line 1327: We can only close a loop, if we created it. If it was already running, we shouldn't.

JoanFM: we should use some context manager here

JoanFM (changes requested):

jina/helper.py
jacobowitz: we could, its a bit tricky though. The underlying design issue is that the caller does not necessarily own the loop.

jina/helper.py
```diff
@@ -1322,9 +1333,10 @@ def run(self):
                 'please report this issue here: https://github.com/jina-ai/jina'
             )
     else:
-        loop = get_or_reuse_loop()
+        loop, is_new_loop = get_or_reuse_loop_with_info()
         result = loop.run_until_complete(func(*args, **kwargs))
```
JoanFM, line 1337: wrap in context manager or something to guarantee is closed if something wrong happens only if is owned

JoanFM (changes requested):

jina/helper.py
jacobowitz: I implemented now in the helper and used [here](https://github.com/jina-ai/jina/pull/4734/files#diff-91bf316ed025400aa871f71855a824d46198a819fe561abae015373e9cbb7f65R1346)

jina/helper.py
```diff
@@ -1000,7 +1000,17 @@ def _update_policy():
             )
 
 
-def get_or_reuse_loop_with_info():
+class ManagedLoop(object):
```
JoanFM: u will need to redefine `__getattr__` to forward it to `_loop` to make it transparent, or access the private loop to start it

jina/helper.py
```diff
@@ -1333,11 +1343,8 @@ def run(self):
                 'please report this issue here: https://github.com/jina-ai/jina'
             )
     else:
-        loop, is_new_loop = get_or_reuse_loop_with_info()
-        result = loop.run_until_complete(func(*args, **kwargs))
-        if is_new_loop:
-            loop.close()
-        return result
+        with ManagedLoop() as loop:
+            return loop.run_until_complete(func(*args, **kwargs))
```
JoanFM, line 1347: this can't work

jacobowitz:
After discussion with @deepankarm we came up with a better solution which fixes the underlying issue: Not reusing existing loops. Lets see if it shows any undesired side effects in the tests. If not, lets use this solution.

jacobowitz:
actually there are some cases where we need to create a loop. We should not try to retrieve a running loop though. Any non closed loop is fine.
Also rebased against master

jina/helper.py
```diff
@@ -1006,13 +1007,13 @@ def get_or_reuse_loop():
 
     :return: A new eventloop or reuse the current opened eventloop.
     """
+    _update_policy()
     try:
-        loop = asyncio.get_running_loop()
+        loop = asyncio.get_event_loop()
```
jacobowitz, line 1012: this is the relevant change. We should use any loop (unless closed as checked in the next line).

jacobowitz:
I reset the loop now on forking due to this [upstream issue](https://github.com/python/cpython/issues/66197)

PR #4737, opened 2022-05-02
feat: add a decoratator for monitoring method

samsja:
Follow up of https://github.com/jina-ai/clip-as-service/pull/674

Goal:

Having an decorator to monitor internal method of an Executor

```python
class DummyExecutor(Executor):
    @requests(on='/foo')
    def foo(self, docs, **kwargs):
        self._proces(docs)
        self.proces_2(docs)

    @monitor(name='metrics_name',  documentation='metrics description')
    def _proces(self, docs):
        ...

    @monitor()
    def proces_2(self, docs):
        ...
```

Screenshot from 2022-05-02 14-44-11

### DIFF
--- a/jina/serve/executors/__init__.py
+++ b/jina/serve/executors/__init__.py
@@ -132,8 +132,11 @@ class BaseExecutor(JAMLCompatible, metaclass=ExecutorType):
                 namespace='jina',
                 labelnames=('executor', 'endpoint'),
             )
+            self._metrics_buffer = {'process_request_seconds': self._summary_method}
+
         else:
             self._summary_method = None
+            self._metrics_buffer = None
 
     def _add_requests(self, _requests: Optional[Dict]):
         if not hasattr(self, 'requests'):
--- a/jina/serve/executors/__init__.py
+++ b/jina/serve/executors/__init__.py
@@ -502,22 +502,6 @@ class BaseExecutor(JAMLCompatible, metaclass=ExecutorType):
         else:
             return None
 
-    def log(
-        self,
-        value: Any,
-        name: Optional[str] = None,
-        documentation: Optional[str] = None,
-    ):
-        """
-        observe a value to a metric, if the metric does not exist yet, it will create it and store it in a buffer.
-        :param value: value to be observe by the metric,
-        :param name: the name of the metrics
-        :param documentation:  the description of the metrics
-        """
-        metric = self.get_metrics(name, documentation)
-        if metric:
-            metric.observe(value)
-
 
 class ReducerExecutor(BaseExecutor):
     """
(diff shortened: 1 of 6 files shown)


jina/serve/executors/__init__.py
```diff
@@ -15,8 +15,9 @@
 from jina.serve.executors.decorators import requests, store_init_kwargs, wrap_func
 
 if TYPE_CHECKING:
-    from jina import DocumentArray
+    from prometheus_client import Summary
 
+    from jina import DocumentArray
```
JoanFM, line 20:
```suggestion
    from docarray import DocumentArray
```

JoanFM (changes requested):

jina/serve/executors/__init__.py
JoanFM, line 20:
```suggestion
    from docarray import DocumentArray
```

jina/serve/executors/decorators.py
```diff
@@ -136,3 +136,44 @@ def __set_name__(self, owner, name):
...
+        )
+
+        @functools.wraps(func)
+        def _f(self, *args, **kwargs):
+            metric = self.get_metrics(name_, documentation_)
+
+            if metric:
+                with metric.time():
+                    return func(self, *args, **kwargs)
+            else:  # should raise a warning or log something about the fact the monitoring is disabled
```
JoanFM, line 174: what about this comment, why is not implemented?

JoanFM (changes requested):

hanxiao:
```
@monitor('metrics_name', 'metrics description')
```

1. what are the available matric names?
2. it is better to force the explicit kwargs, with `*`: https://stackoverflow.com/a/14298976
3. if metrics_name is basically an enum, then why the description is required, isn't the description best described by us?

samsja:
> 

'metric_name' is the name that will appear on prometheus. For example if you are monitoring the time that clip as a service spend processing some text you will call it `text_preprocessing`. **It is not an enum for the type of the metric**, therefore the description is used to give more context on what the function which is wrapped by monitoring is doing.
Btw the user can as well just call `@monitor` and it will produce a name and a description automatically based on the name of the function ( everything will be documented in the documentation pr)

samsja:
> ```
> @monitor('metrics_name', 'metrics description')
> ```
> 
>     1. what are the available matric names?
> 
>     2. it is better to force the explicit kwargs, with `*`: https://stackoverflow.com/a/14298976
> 
>     3. if metrics_name is basically an enum, then why the description is required, isn't the description best described by us?

for point 2. you are right I will update it

JoanFM (changes requested):
I am missing unit tests for the monitor decorator

JoanFM (approved):

PR #4738, opened 2022-05-02
feat: inherit monitoring arg from Flow

samsja:
Right  now ( jina 3.3.18) when doing 

```python
from jina import Flow
Flow(monitoring=True).add(uses='jinahub//SimpleIndexer')
```

the monitoring is not enable on the executor

to enable it we need to to :

```python
Flow(monitoring=True).add(uses='jinahub//SimpleIndexer',monitoring=True)
```

it should be the other way around, we monitoring is set to True on the Flow level all of the executor should inherit from it unless we precisely disable it by doing

```python
Flow(monitoring=True).add(uses='jinahub//SimpleIndexer',monitoring=False)
```

### DIFF
--- a/tests/integration/monitoring/test_monitoring.py
+++ b/tests/integration/monitoring/test_monitoring.py
@@ -21,8 +21,8 @@ def test_enable_monitoring_deployment(port_generator):
     port1 = port_generator()
     port2 = port_generator()
 
-    with Flow().add(uses=DummyExecutor, monitoring=True, port_monitoring=port1).add(
-        uses=DummyExecutor, monitoring=True, port_monitoring=port2
+    with Flow().add(uses=DummyExecutor, port_monitoring=port1, monitoring=True).add(
+        uses=DummyExecutor, port_monitoring=port2, monitoring=True
     ) as f:
         for port in [port1, port2]:
             resp = req.get(f'http://localhost:{port}/')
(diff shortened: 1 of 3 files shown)


tests/integration/monitoring/test_monitoring.py
```diff
@@ -115,3 +115,35 @@ def test_document_processed_total(port_generator):
             f'jina_document_processed_total{{endpoint="/foo",executor="DummyExecutor"}} 4.0'  # check that we nothing change on bar count
             in str(resp.content)
         )
+
+
+def test_disable_monitoring_on_pods(port_generator):
+    port0 = port_generator()
+    port1 = port_generator()
+
+    with Flow(monitoring=True, port_monitoring=port0).add(
```
JoanFM, line 124: this should be a unit tests and not an integration one. U can check when the Flow construction happens that the `args` are what u expect. I am not saying to erase this test but to also keep the simpler one

JoanFM (changes requested):

tests/integration/monitoring/test_monitoring.py
samsja: I added unit tests

JoanFM (approved):
Please remember to properly document this behavior

samsja:
> Please remember to properly document this behavior

Sure

PR #4752, opened 2022-05-05
feat: add runtime name as label for monitoring

samsja:
We need a specific name on each endpoint so that we can differentiate them. For example we need to differentiate the metrics coming from the executors from the one coming from the gateway.

Therefore we need to add a common label on every endpoint with the name of the pod.#

- [x] Added `runtime_name` to the labels on every metrics
- [x] include metrics from the decorator: https://github.com/jina-ai/jina/pull/4737


JoanFM (changes requested):
How will this feature affect in Kubernetes? In Kubernetes these names will not be different? Will we tell by replica?

JoanFM (approved):

PR #4770, opened 2022-05-09
feat(hubble): executor push/pull user integration

delgermurun:
Ref https://github.com/jina-ai/hubble/issues/355

### DIFF
--- a/tests/unit/hubble/test_hubio.py
+++ b/tests/unit/hubble/test_hubio.py
@@ -35,6 +35,21 @@ def clear_function_caches():
     _get_hub_config.cache_clear()
 
 
+@pytest.fixture(scope='function')
+def auth_token(tmpdir):
+    clear_function_caches()
+    os.environ['JINA_HUB_ROOT'] = str(tmpdir)
+
+    token = 'test-auth-token'
+    with open(tmpdir / 'config.json', 'w') as f:
+        json.dump({'auth_token': token}, f)
+
+    yield token
+
+    clear_function_caches()
+    del os.environ['JINA_HUB_ROOT']
+
+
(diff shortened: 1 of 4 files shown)


tests/unit/hubble/test_hubio.py
```diff
@@ -212,6 +223,41 @@ def _mock_post(url, data, headers=None, stream=True):
     )
 
 
+def test_push_with_authorization(mocker, monkeypatch, tmpdir):
+    clear_function_caches()
+    os.environ['JINA_HUB_ROOT'] = str(tmpdir)
```
JoanFM, line 228: u should use a fixture to make sure that this env variable is properly set and unset at the beginning and at the end of the test regardless of the success of failure of it

JoanFM (changes requested):

tests/unit/hubble/test_hubio.py
```diff
@@ -293,6 +339,34 @@ def _mock_post(url, json, headers=None):
     assert mock.call_count == 6  # mock must be called 3+3
 
 
+def test_fetch_with_authorization(mocker, monkeypatch, tmpdir):
+    clear_function_caches()
+    os.environ['JINA_HUB_ROOT'] = str(tmpdir)
```
JoanFM, line 344: same here

JoanFM (changes requested):

tests/unit/hubble/test_hubio.py
delgermurun: I created `auth_token` fixture. Please check it out.

mapleeit (approved):

JoanFM (approved):

PR #4779, opened 2022-05-11
feat: add grpc health checking

JoanFM:
This PR aims to change the way health check is done in the core for Executors but also for any runtime using grpc. We aim to use the standard grpc https://github.com/grpc/grpc/blob/master/doc/health-checking.md as suggested by grpc, and which is something that lots of Load Balancers, etc ... will use to make sure grpc-based microservices are healthy.

This also has the benefit that `ControlRequest` can be removed from the code.


jina/orchestrate/deployments/__init__.py
```diff
@@ -284,6 +285,17 @@ def is_sandbox(self) -> bool:
         is_sandbox = uses.startswith('jinahub+sandbox://')
         return is_sandbox
 
+    @property
+    def _is_docker(self) -> bool:
+        """
+        Check if this deployment is a sandbox.
```
samsja, line 291:
```suggestion
        Check if this deployment is to be run in docker.
```

jina/serve/runtimes/gateway/grpc/__init__.py
```diff
@@ -13,12 +15,22 @@
...
+        self,
+        args: argparse.Namespace,
+        **kwargs,
+    ):
+        """Initialize the runtime
+        :param args: args from CLI
+        :param kwargs: keyword args
+        """
+        self._health_servicer = health.HealthServicer(experimental_non_blocking=True)
+        super().__init__(args, **kwargs)
```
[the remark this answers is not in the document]
JoanFM: I think it was because not all of them have it

jina/__init__.py
```diff
@@ -68,7 +68,7 @@ def _warning_on_one_line(message, category, filename, lineno, *args, **kwargs):
 
 # do not change this line manually
 # this is managed by proto/build-proto.sh and updated on every execution
-__proto_version__ = '0.1.10'
+__proto_version__ = '0.1.12'
```
JoanFM, line 71:
```suggestion
__proto_version__ = '0.1.11'
```

jina/orchestrate/deployments/__init__.py
```diff
@@ -284,6 +285,17 @@ def is_sandbox(self) -> bool:
...
 
+    @property
+    def _is_docker(self) -> bool:
+        """
+        Check if this deployment is to be run in docker.
+
+        :return: True if this deployment is to be run in docker
+        """
+        uses = getattr(self.args, 'uses', '')
+        is_docker = uses.startswith('jinahub+docker://') or uses.startswith('docker://')
```
JohannesMessner, line 296: What about sandbox executors? I don't actually know, but I imagine they also run inside docker? Also more generally, can we know if external executors are dockerized? Or is it irrelevant in that case?

jina/orchestrate/deployments/__init__.py
```diff
@@ -459,52 +471,27 @@ def num_pods(self) -> int:
...
+        :param pod_is_container: boolean specifying if pod is to be run in container
+        :param head_is_container: boolean specifying if head pod is to be run in container
+        :return: host to pass in connection list of the head
         """
         # Check if the current pod and head are both containerized on the same host
         # If so __docker_host__ needs to be advertised as the worker's address to the head
         worker_host = (
             __docker_host__
-            if Deployment._is_container_to_container(pod, head_pod)
+            if (pod_is_container and (head_is_container or in_docker()))
```
JohannesMessner, line 489: The naming of `_is_docker` and `in_docker` is very similar. From what I understand the one checks a provided deployment, while the other checks the current process. Could this be clarified in the naming?

jina/serve/networking.py
```diff
@@ -920,22 +831,51 @@ def send_request_sync(
...
+                    tls=tls,
+                    root_certificates=root_certificates,
+                ) as channel:
+                    health_check_req = health_pb2.HealthCheckRequest()
+                    health_check_req.service = ''
+                    stub = health_pb2_grpc.HealthStub(channel)
+                    return stub.Check(health_check_req, timeout=timeout)
+            except grpc.RpcError as e:
+                if e.code() != grpc.StatusCode.UNAVAILABLE or i == 2:
+                    raise
```
JohannesMessner, line 877: Do we need error handling here a la `InternalNetworkError`, reporting back to the client which target failed, that sort of stuff? Or will this be handled differently?

JoanFM: here  u are sending the health check to a single entity, this check is not about sending to the complete Flow, that is in the other PR

docs/fundamentals/health-check.md
```diff
@@ -0,0 +1,238 @@
...
+```text
+{}%
+```
+
+## Readiness of a Flow exposed to the client
+
+A lot of times, from the client perspective, it is useful to know if a Flow, as a complete set of microservices, is ready to receive requests. This is why the gateway 
+exposes an endpoint for each of the supported protocols to know the health and readiness of a Flow. 
+
+Also Jina Flow and Client contains a convenient API to query these endpoints. You can call `flow.dry_run()` and `client.dry_run()`
```
JohannesMessner, line 84:
```suggestion
Jina `Flow` and `Client` offer a convenient API to query these readiness endpoints. You can call `flow.dry_run()` and `client.dry_run()`, which will return `True` if the Flow is healthy and ready, and `False` otherwise.
```

JohannesMessner (changes requested):
I would put the part about checking the entire Flow, `dry_run()` etc at the top. To me it seems like the most useful and commonly needed feature.

Also, what is the difference between the sections at the top about checking Executor/Gateway, and the sections at the bottom about 'from the outside'? I didn't leave comments on the latter because it seemed the same to me. Could you explain the difference?

docs/fundamentals/health-check.md
```diff
@@ -0,0 +1,238 @@
...
+By having a Flow with http or websocket as protocol:
+
+```python
+from jina import Flow
+
+f = Flow(protocol='http', port=12345).add()
+with f:
+    f.block()
+```
+
```
JohannesMessner, line 69:
```suggestion
Then, you can query the "empty" endpoint:
```

docs/fundamentals/health-check.md
```diff
@@ -0,0 +1,238 @@
+# Health and readiness check of Jina Services
+
```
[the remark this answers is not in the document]
JoanFM: this comes from the `document` of GRPC linked there

docs/fundamentals/flow/health-check.md
```diff
@@ -0,0 +1,275 @@
+# Health and readiness check of Jina Flow
```
JohannesMessner, line 1:
```suggestion
# Health and readiness check
```

Issue #4784, opened 2022-05-13
Grpc gateway not working properly with every cert pair


samsja:
We can't turn on TLS for grpc, it always says port can not be bind
```
Failed to bind to address 0.0.0.0:51000;
       set GRPC_VERBOSITY=debug environment variable to see
       detailed error message.
```

```yaml
jtype: Flow
version: '1'
with:
  port: 51000
  protocol: grpc
  uvicorn_kwargs:
    ssl_keyfile_password: abcd
  ssl_certfile: cert.perm
  ssl_keyfile: key.perm
```
The same yaml when changing protocol: http works

cc @hanxiao

PR #4785, opened 2022-05-13
fix: better tls support for grpc

samsja:
fixing https://github.com/jina-ai/jina/issues/4784


samsja:
Closing issue as the the certificates were not valid

Issue #4773, opened 2022-05-10
It shoud not be possible to pass argument to the Flow init function when loading from a yaml file


samsja:
It shoud not be possible to pass argument to the Flow init function when loading from a yaml file

```python
from jina import Flow
f = Flow(port=8080).load_config('config.yaml')
```

Here port=8080 could be override because load config will override everything.

load config should be a static method, i.e we should not be able to call it on an instance. 

How do we keep backward compatibility ? Maybe deprecated the class method and add a warning

PR #4787, opened 2022-05-16
fix: warn user when calling load config from flow instance

samsja:
fixing https://github.com/jina-ai/jina/issues/4773

Now when doing 

```python
from jina import Flow

Flow().load_config('config.yaml')
```

you got this warning 

```
UserWarning: Calling load_config from a Flow instance will override all of its initial parameters. You should probably use `Flow.load_config(...)` instead
```

### DIFF
--- a/jina/orchestrate/flow/base.py
+++ b/jina/orchestrate/flow/base.py
@@ -98,11 +98,7 @@ FALLBACK_PARSERS = [
 ]
 
 
-class _BaseClassFlow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
-    ...
-
-
-class Flow(_BaseClassFlow):
+class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
     """Flow is how Jina streamlines and distributes Executors."""
 
     # overload_inject_start_client_flow
@@ -2045,7 +2041,7 @@ class Flow(_BaseClassFlow):
         self._common_kwargs.update(kwargs)
 
     def __getattribute__(self, item):
-        obj = super(_BaseClassFlow, self).__getattribute__(item)
+        obj = super().__getattribute__(item)
 
         if (
             item == 'load_config' and inspect.ismethod(obj) and obj.__self__ is Flow
--- a/jina/orchestrate/flow/base.py
+++ b/jina/orchestrate/flow/base.py
@@ -2047,7 +2047,7 @@ class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
             item == 'load_config' and inspect.ismethod(obj) and obj.__self__ is Flow
         ):  # check if obj load config call from an instance and not the Class
             warnings.warn(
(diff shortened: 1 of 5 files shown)


jina/orchestrate/flow/base.py
```diff
@@ -96,7 +98,11 @@ class FlowType(type(ExitStack), type(JAMLCompatible)):
 ]
 
 
-class Flow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
+class _BaseClassFlow(PostMixin, JAMLCompatible, ExitStack, metaclass=FlowType):
```
JoanFM, line 101: why is this new class needed?

tests/integration/issues/github_4773/test_load_config.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pytest
+
+from jina import Flow
+
+
+def test_load_config_instance():
+    with pytest.warns(UserWarning):  # assert that a warning has been emited
+        f = Flow().load_config('config.yaml')
+    f.build()
```
JoanFM, line 11: no need to build

tests/integration/issues/github_4773/test_load_config.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_load_config_class():  # assert that no warnings were emited
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        f = Flow.load_config('config.yaml')
+
+    f.build()
```
JoanFM, line 19: no need to build

JoanFM (changes requested):

jina/orchestrate/flow/base.py
samsja: I need this class here : https://github.com/jina-ai/jina/blob/b70aa02900aa3957531719e3b88a00dc25b447de/jina/orchestrate/flow/base.py#L2048

Otherwise I can't have a default `__getattribute__` function which I overload to check the if the function called is `load_config`

JoanFM: u can not get it from object? just doing `super().__getattribute__(item)`? It would surprise me this is not possible

samsja: yes ofc

jina/orchestrate/flow/base.py
```diff
@@ -2037,3 +2039,15 @@ def update_network_interface(self, **kwargs):
...
         self._common_kwargs.update(kwargs)
+
+    def __getattribute__(self, item):
+        obj = super().__getattribute__(item)
+
+        if (
+            item == 'load_config' and inspect.ismethod(obj) and obj.__self__ is Flow
+        ):  # check if obj load config call from an instance and not the Class
+            warnings.warn(
+                'Calling load_config from a Flow instance will override all of its initial parameters. You should probably use `Flow.load_config(...)` instead'
```
JohannesMessner, line 2050:
```suggestion
                'Calling load_config from a Flow instance will override all of the instance's initial parameters. We recommend to use `Flow.load_config(...)` instead'
```

JoanFM (approved):

JoanFM (approved):

Issue #4731, opened 2022-05-02
Implement all the metrics that we want to expose post release


samsja:
Follow up of https://github.com/jina-ai/jina/issues/4589

- [ ] Number of fails: How many requests failed, exception type etc 
- [x] Request size: In bytes of the request object https://github.com/jina-ai/jina/pull/4789
- [ ]  Number of Clients connected: gateway only 
- [x] Number of pending requests: gateway only https://github.com/jina-ai/jina/pull/4789

PR #4789, opened 2022-05-16
feat: add additional metrics

samsja:
linked to https://github.com/jina-ai/jina/issues/4731

add new metrics

- [x] Request size: In bytes of the request object
- [x] Number of pending requests: gateway only

SKIP
- [x] (skip to do in another PR)  Number of fails: How many requests failed, exception type etc 
- [x]  (skip to do in another PR) Number of Clients connected: gateway only

### DIFF
--- a/tests/integration/monitoring/test_monitoring.py
+++ b/tests/integration/monitoring/test_monitoring.py
@@ -190,3 +190,63 @@ def test_requests_size(port_generator, executor):
         measured_request_bytes_sum = _get_request_bytes_size()
 
         assert measured_request_bytes_sum > measured_request_bytes_sum_init
+
+
+def test_pending_request(port_generator):
+    port0 = port_generator()
+    port1 = port_generator()
+
+    class SlowExecutor(Executor):
+        @requests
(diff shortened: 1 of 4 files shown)


JoanFM (changes requested):
Some tests should be added

samsja:
> Some tests should be added

I added integration tests for both of the current metrics

tests/integration/monitoring/test_monitoring.py
```diff
@@ -152,3 +153,94 @@ def test_disable_monitoring_on_gatway_only(port_generator, executor):
...
+        p_send.start()
+        time.sleep(1)
+        p_while.start()
+
+        for p in [p_before, p_send, p_while]:
+            p.join()
+
+        for p in [p_before, p_send, p_while]:
+            p.terminate()
+            assert not p.exitcode
```
JoanFM, line 244: collect the exit codes and assert after all of them have been terminated, to avoid timeouts

JoanFM (changes requested):

JoanFM (approved):

PR #4790, opened 2022-05-16
fix: add check that jaml parses to correct class

JohannesMessner:
Fixes silent error when invalid YAML is passed as an Executor

### DIFF
--- a/tests/unit/jaml/test_type_parse.py
+++ b/tests/unit/jaml/test_type_parse.py
@@ -2,9 +2,10 @@ import os
 
 import pytest
 
-from jina.serve.executors import BaseExecutor
+from jina import Flow, __default_executor__, requests
+from jina.excepts import BadConfigSource
 from jina.jaml import JAML, JAMLCompatible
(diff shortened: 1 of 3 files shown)


tests/unit/jaml/test_type_parse.py
```diff
@@ -179,3 +180,11 @@ def test_parsing_brackets_in_envvar():
         assert b['executors'][0]['env']['var2'] == 'a'
         assert b['executors'][0]['env']['var3'] == '{"1": "2"}-a'
         assert b['executors'][0]['env']['var4'] == '-{"1": "2"}'
+
+
+def test_exception_invalid_yaml():
+    with pytest.raises(BadConfigSource):
+        BaseExecutor.load_config('./invalid.yml')
```
JoanFM, line 187: do not use `./invalid.yml`, use some `cur_dir` to make sure test is runnable from anywhere

JoanFM (changes requested):

JoanFM (approved):

PR #4794, opened 2022-05-17
feat: smart replica retry

JohannesMessner:
Right now, when a replica fails (on local), the gateway will stubbornly retry to connect to that same replica, instead of trying its luck with one of the other replicas. This PR changes that, applying round robing scheduling also for retries, not just successful calls.

**ToDo:**

- [x] smart retry mechanism
- [x] test for retry mechanism
- [x] appropriate error message if all replicas fail
- [x] test for this failure case

### DIFF
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -310,7 +310,7 @@ class GrpcConnectionPool:
             entity_id: Optional[int] = None,
             increase_access_count: bool = True,
         ) -> ReplicaList:
-            # returns all replicas of a given deployment, using a "random" shard
+            # returns all replicas of a given deployment, using a given shard
             if deployment in self._deployments:
                 type_ = 'heads' if head else 'shards'
                 if entity_id is None and head:
@@ -730,10 +730,10 @@ class GrpcConnectionPool:
             tried_addresses = set()
             num_retries = max(3, len(connections.get_all_connections()))
             for i in range(num_retries):
-                connection = connections.get_next_connection()
-                tried_addresses.add(connection.address)
+                current_connection = connections.get_next_connection()
+                tried_addresses.add(current_connection.address)
                 try:
-                    return await connection.send_requests(
+                    return await current_connection.send_requests(
                         requests=requests,
                         metadata=metadata,
                         compression=self.compression,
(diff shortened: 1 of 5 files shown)


jina/serve/networking.py
```diff
@@ -310,6 +310,7 @@ def get_replicas(
             entity_id: Optional[int] = None,
             increase_access_count: bool = True,
         ) -> ReplicaList:
+            # returns all replicas of a given deployment, using a "random" shard
```
JoanFM, line 313: Not a `random` but a given one

JoanFM (changes requested):

jina/serve/networking.py
```diff
@@ -707,21 +712,26 @@ def _handle_aiorpcerror(
...
     ) -> asyncio.Task:
         # this wraps the awaitable object from grpc as a coroutine so it can be used as a task
         # the grpc call function is not a coroutine but some _AioCall
         async def task_wrapper():
             metadata = (('endpoint', endpoint),) if endpoint else None
-            for i in range(3):
+            tried_addresses = set()
+            num_retries = max(3, len(connections.get_all_connections()))
+            for i in range(num_retries):
+                connection = connections.get_next_connection()
```
samsja, line 733: I find it really confusing to have `connection` for the single entity and `connections` for the group.
Maybe `current_connexion` ?

tests/integration/runtimes/test_network_failures.py
```diff
@@ -411,3 +413,116 @@ async def test_runtimes_graphql(port_generator):
...
+        ready_or_shutdown_event=multiprocessing.Event(),
+    )
+
+    try:
+        # ----------- 1. ping Flow once to trigger endpoint discovery -----------
+        # we have to do this in a new process because otherwise grpc will be sad and everything will crash :(
+        p = multiprocessing.Process(target=_send_request, args=(gateway_port, 'grpc'))
+        p.start()
+        p.join()
+        assert p.exitcode == 0
```
JoanFM, line 453: assert these after u have finished every assert in the test, otherwise it may be left hanging

JoanFM (changes requested):

JoanFM (approved):

PR #4795, opened 2022-05-17
docs: add specification of executor and flow yaml

JohannesMessner:
TODO:

- [x] mention environment variables and internal links

### DIFF
--- a/docs/fundamentals/flow/yaml-spec.md
+++ b/docs/fundamentals/flow/yaml-spec.md
@@ -2,11 +2,14 @@
 # YAML specification
 
 This page outlines the specification for valid Flow YAML files.
-Such YAML configurations can be used to generate a Flow via `Flow.load_config('flow.yml')`.
+
+Such YAML configurations can be used to generate a Flow object via `Flow.load_config('flow.yml')`.
+
+To generate a YAML configuration from a `Flow` Python object, run `f.save_config()`.
 
 ## Example
 
-The following constitutes a typical Flow configuration:
+The following constitutes an example Flow configuration:
 
 `flow.yml`/`flow.yaml`:
 ```yaml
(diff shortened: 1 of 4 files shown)


docs/fundamentals/flow/yaml-spec.md
```diff
@@ -0,0 +1,66 @@
...
+
+`flow.yml`/`flow.yaml`:
+```yaml
+jtype: Flow
+version: '1'
+with:
+  protocol: http
+metas:
+  - py_modules:
+      - executor.py
```
JohannesMessner, line 22: This patterns does not seem to work. But also not sure if we even want to promote it.

docs/fundamentals/flow/yaml-spec.md
```diff
@@ -0,0 +1,66 @@
...
+# inline Executor YAML
+- name: firstexec
+  uses:
+    jtype: MyExec
+    py_modules:
+      - executor.py
+# reference to Executor YAML
+- name: secondexec
+  uses: indexer.yml
+  workspace: /home/johannes/.config/JetBrains/PyCharmCE2022.1/scratches/indexed
```
JoanFM, line 33: what about this workspace? why is not explained? I also would suggest for u just to pass some random value here

JoanFM (changes requested):

docs/fundamentals/flow/yaml-spec.md
JohannesMessner: Below it says that here we can put any argument that `add()` would take, and links to where to find that info. This is meant just as an example of where to actually put such an arg (how to indent etc)

docs/fundamentals/flow/yaml-spec.md
JoanFM: then remove if it does not seem to work

JoanFM (approved):

PR #4799, opened 2022-05-18
fix(qabot): hide the slot elements on the docs page

shazhou2015:
Goals:
Add a template to prevent the qabot slot elements from showing on the docs pages.


nomagick (approved):
LGTM

PR #4805, opened 2022-05-19
docs: round robin gpu scheduling 

samsja:
Add docs for the round robin cuda features

### DIFF
--- a/docs/fundamentals/flow/create-flow.md
+++ b/docs/fundamentals/flow/create-flow.md
@@ -676,10 +676,21 @@ The above Flow will create a topology with three Replicas of Executor `slow_enco
 request to exactly one of the three instances. Then the replica will send its result to `fast_indexer`.
 
 
-### Replicate on Multiple GPUs
+### Replicate on multiple GPUs
 
-If you have multiple GPU devices, you can leverage them via `CUDA_VISIBLE_DEVICES=RR`. For example, if you have 3 GPUs 
-and one of your Executor has 5 replicas then 
+You can be in a setup when you want to replicate your Executor so that each replicas can use a different GPU on your machine.
+To do so you need to tell the Flow to leverage multiple GPUs, by passing `CUDA_VISIBLE_DEVICES=RR` as an environment variable,
+the Flow will then assign each available GPU to each replicas in a round-robin fashion.
+
+```{caution} 
+Replicate on multiple GPUs by using `CUDA_VISIBLE_DEVICES=RR` should only be used locally.  
+```
+
+```{tip}
+When working in Kubernetes or with Docker Compose you shoud allocate GPU ressources to each replica directly in the configuration files.
+```
+
+For example, if you have 3 GPUs and one of your Executor has 5 replicas then 
 
 ````{tab} Python
  In a `flow.py` file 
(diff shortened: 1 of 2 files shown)


docs/fundamentals/flow/create-flow.md
```diff
@@ -675,6 +675,65 @@ Flow with 3 replicas of slow_encoder and 1 replica of fast_indexer
...
+  install_requirements: True
+  replicas: 5  
+```
+
+```shell
+CUDA_VISIBLE_DEVICES=RR jina flow --uses flow.yaml
+```
+````
+
+Will assign GPU devices to the following round-robin fashion:
```
JohannesMessner, line 715:
```suggestion
Will assign GPU devices in the following round-robin fashion:
```

JohannesMessner (changes requested):
I would first explain *what* we are trying to achieve, and then *how* we achieve it. Basically first explain round robing scheduling of the GPU's, then show the code. Otherwise the uses reads instructions for something he doesn't even really know what it is. Do you agree?

JoanFM (changes requested):
This has to clarify that this does not apply when using Kubernetes or docker compose

samsja:
> 

done

samsja:
> I would first explain _what_ we are trying to achieve, and then _how_ we achieve it. Basically first explain round robing scheduling of the GPU's, then show the code. Otherwise the uses reads instructions for something he doesn't even really know what it is. Do you agree?

I agree, done as well

JoanFM (approved):

docs/fundamentals/flow/create-flow.md
```diff
@@ -675,6 +675,76 @@ Flow with 3 replicas of slow_encoder and 1 replica of fast_indexer
...
+| GPU device | Replica ID |
+|------------|------------|
+| 0          | 0          |
+| 1          | 1          |
+| 2          | 2          |
+| 0          | 3          |
+| 1          | 4          |
+
+ 
+You can also restrict the visible devices in round-robin assignment by `CUDA_VISIBLE_DEVICES=RR0:2`, where `0:2` has the same meaning as Python slice. This will create the following assignment:
```
alaeddine-13, line 737: Although this is not documented in the clip as service documentation, I believe it's good to let users know that they might need to separate the allocation of devices to executors.
you can check how this separation is implemented in [dalle flow](https://github.com/jina-ai/dalle-flow/blob/main/flow.yml#L41).
In general, users should be aware that they can use this feature in order to allocate 1 device for exec 1, RR1:3 for a replicated executor 2, etc. This can help them avoid issues like many processes trying to access the same resource

samsja: it is documented in clip as as service : https://clip-as-service.jina.ai/user-guides/server/?utm_source=clip-as-service#serving-on-multiple-gpus.

You think that the current documentation on the round robin things is not enough ? What would you like to add ?

alaeddine-13: For me, restricting the gpus for the entire flow is not the same as restricting gpus for some executors. The feature can also serve as a protection mechanism against executors accessing the same resource (and htat's how we are trying to use it currently).
What I mean is giving an example where a user can do:
```yaml

  - name: Exec1
    ...
    env:
      CUDA_VISIBLE_DEVICES: 0

  - name: Exec2
    replicas: 2
    ...
    env:
      CUDA_VISIBLE_DEVICES: RR1:3

  - name: Exec3
    ...
    env:
      CUDA_VISIBLE_DEVICES: 3:5
```

samsja: Okay now I understand your point

JohannesMessner (approved):

PR #4809, opened 2022-05-19
fix: remove tag from save config yaml

samsja:
The output of `f.save_config()` should not include `!Flow` since it is deprecated syntax and not compatible with jcloud.

### DIFF
--- a/jina/jaml/__init__.py
+++ b/jina/jaml/__init__.py
@@ -554,9 +554,16 @@ class JAMLCompatible(metaclass=JAMLCompatibleType):
         """
         from jina.jaml.parsers import get_parser
 
-        tmp = {'jtype': cls.__name__}
-        tmp.update(get_parser(cls, version=data._version).dump(data))
-        return representer.represent_mapping('!' + cls.__name__, tmp)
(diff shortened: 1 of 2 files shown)


samsja:
Turns out it is not that easy ...

https://github.com/yaml/pyyaml/issues/408

JoanFM (changes requested):
A test would be needed

JoanFM (commented):
didn't we want to have a custom `Dumper`?

JohannesMessner:
> didn't we want to have a custom `Dumper`?

We did, but
1) I don't think it is necessary if we can fix the issue with 1 LOC, and
2) not sure if it would even work. Internally, pyyaml represents the yaml as a graph of `node`s, and every `node` needs to have a `tag`. This cannot be `''` or `None`. So I suspect that even with a custom Dumper we would end up with the same solution anyways. For example, dumping a normal Python `dict` internally also does the same thing (just even hackier, this magic string is the same as the tag that I use):
```python
def represent_dict(self, data):
    return self.represent_mapping('tag:yaml.org,2002:map', data)
```

JoanFM (approved):

PR #4851, opened 2022-05-31
uploaded missing image of containerized environment in documentation

vkvbit:
While reading the documentation I encountered that an image is missing.


vkvbit (commented):
containerized environment image is missing from documentation, image is attached
container-env

JoanFM:
Hey @vkvbit , I need to understand where do u consider this is missing

JoanFM:
THis documentation is not up to date, therefore I would rather rewrite it

PR #4858, opened 2022-06-02
docs: add default port client

samsja:
This pr add docs for the default port client


JoanFM (changes requested):
Does this also apply to the gateway?

samsja:
> Does this also apply to the gateway?

No only for the Client

JoanFM (approved):

PR #4883, opened 2022-06-07
feat: add argument to control number of grpc retries

JohannesMessner:
This allows the user to configure the number of retries that the gateway and head perform for each grpc call.
This includes the ability to disable retries altogether, which is useful when the user wants to user a service mesh level retry mechanism instead.

**ToDos:**

- [x] Add in Gateway
- [x] Add in Head
- [x] Test for gateway
- [x] Test for head
- [x] Add to docs

**Feature usage:**
Now the user can set the following:
```python
f = Flow()  # default policy: try every replica at least once, but perform at least 3 tries
f = Flow(retries=5)  # perform 5 retries, so 6 tries in total
f = Flow(retries=0)  # perform 0 retries, so 1 try in total
f = Flow(retries=-1)  # default policy again
```


jina/parsers/orchestrate/pod.py
```diff
@@ -103,3 +103,11 @@ def mixin_pod_parser(parser):
         dest='port_monitoring',
         help=f'The port on which the prometheus server is exposed, default port is {__default_port_monitoring__} ',
     )
+
+    gp.add_argument(
+        '--retries',
+        type=int,
+        default=None,  # default prometheus server port
```
JoanFM, line 110: it should be 3 right? plus wrong comment

JoanFM (changes requested):

jina/parsers/orchestrate/pod.py
JohannesMessner: Wrong comment, true, thanks for the catch.

But I would leave the default to `None`, which allows to keep the previous policy in place: Try at least three times, but if there are more than 3 replicas, try every replica at least once. That seems reasonable to me, because otherwise there might be a healthy replica but the call still fails. If the default is 3, we don't know if it was user set, and can't implement the default policy.

An alternative would be to always try every replica at least once, regardless of what the user puts in (`actual_num_retries = max(retries, num_replicas)`). But it feels a bit off to me to take that control away from the user.

jina/serve/networking.py
```diff
@@ -698,14 +710,18 @@ def _send_requests(
...
     ) -> asyncio.Task:
         # this wraps the awaitable object from grpc as a coroutine so it can be used as a task
         # the grpc call function is not a coroutine but some _AioCall
         async def task_wrapper():
             metadata = (('endpoint', endpoint),) if endpoint else None
             tried_addresses = set()
-            num_retries = max(3, len(connections.get_all_connections()))
-            for i in range(num_retries):
+            if retries is None or retries < 0:
+                total_num_tries = max(3, len(connections.get_all_connections())) + 1
```
JoanFM, line 721: may we have this 3 as a constant somewhere?

JoanFM (changes requested):

JoanFM (approved):

Issue #4876, opened 2022-06-07
Make the monitor decorator a context manager as well


samsja:
If you always force user to create new functions, how do you do this
```python
def foo():
   shared_var
   some_code_block_to_measure1
   more_shared_var
   some_code_block_to_measure2
   more_shared_var
   some_code_block_to_measure3
```
adding decorator to the three blocks means pulling them out with nontrivial refactoring, and as a developer, I really do not appreciate it.
Hence, it is better to use with context manager
```python
def foo():
   shared_var
   with monitor(...):
     some_code_block_to_measure1
   more_shared_var
   with monitor(...):
     some_code_block_to_measure2
   more_shared_var
   with monitor(...):
     some_code_block_to_measure3
```
As the developer of CAS, I would rather not have monitor feature at all, then seeing monitor feature breaks my project structure. The reason is simple, developers do not trust the pattern the @monitor imposed on the over their own pattern. If developers have to refactor their function structure very awkwardly to add monitoring, they'd rather not have monitoring.

PR #4892, opened 2022-06-07
feat: custom monitoring with context manager

samsja:
# Custom monitoring with context manager
## Context

having only the possibility to monitor custom part of Executor with a decorator is a limitation. We need to have a context manager as well. See full issue here : https://github.com/jina-ai/jina/issues/4876

## What this PR do

This pr allow developer of Executor to use custom metrics with a context manager

example:

```python
from jina import requests, Executor

def preprocessing(self, docs: DocumentArray):
    ...

def model_inference(self, tensor):
    ...

class MyExecutor(Executor):

    @requests
    def encode(self, docs: DocumentArray, **kwargs):
        with self.monitor('preprocessing_seconds', 'Time preprocessing the requests'):
            docs.tensors = preprocessing(docs)
        with self.monitor('model_inference_seconds', 'Time doing inference the requests'):
            docs.embedding = model_inference(docs.tensors)
```

### DIFF
--- a/tests/integration/monitoring/test_executor.py
+++ b/tests/integration/monitoring/test_executor.py
@@ -59,3 +59,41 @@ def test_decorator_interface(port_generator):
             f'jina_proces_2_seconds_count{{runtime_name="executor0/rep-0"}} 1.0'
             in str(resp.content)
         )
+
+
+def test_context_manager_interface(port_generator):
+    class DummyExecutor(Executor):
+        @requests(on='/foo')
+        def foo(self, docs, **kwargs):
+
+            with self.monitor(
+                name='proces_seconds', documentation='process time in seconds '
+            ):
+                self._proces(docs)
+
+            with self.monitor(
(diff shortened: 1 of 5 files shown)


tests/integration/monitoring/test_executor.py
```diff
@@ -59,3 +59,41 @@ def proces_2(self, docs):
...
+                name='proces_seconds', documentation='process time in seconds '
+            ):
+                self._proces(docs)
+
+            with self.monitor(
+                name='proces_2_seconds', documentation='process 2 time in seconds '
+            ):
+                self.proces_2(docs)
+
+        def _proces(self, docs):
```
JoanFM, line 79:
```suggestion
        def _process(self, docs):
```

JoanFM (approved):
Just a typo, if u want to change everywhere

hanxiao:
btw are we sure `monitor` is the right terminology?

JoanFM (approved):

PR #4913, opened 2022-06-10
docs: create api reference

JohannesMessner:
Goals:

Create a proper API reference page.

The changed commit bot leads to the following behaviour:
- For any **class/method that gets its arguments from a parser, docstrings are generated automatically** (this was already the case before)
- Now, **if there are `@overload` methods, all :param: information will automatically also be added to the final stub**, the one that holds the actual implementation
- If the parser changes, this information is updated automatically. This means that **`:param:` descriptions come from the parser and should be modified there**
- the bot is smart enough to distinguish the auto-generated part of the docstring (all the `:param:`s etc.), from the hand written top level description. **This means that class/method descriptions can be edited directly in the docstring.**
- If there is a `@overload` stub that does _not_ get its arguments from a parser, then the bot can not know about them automatically. In that case, the `:param:` info has to be added in the bot script, like [here](https://github.com/jina-ai/jina/pull/4913/files#diff-386893c98d26e69d4f2e67efee373ebdf8f35dc01621373ed4d76b4ed0a25dd8R292).

Tasks:

- [x] make it so that inherited properties and methods show up
- [x] create landing page that only shows classes/functions that are actually interesting to the user (our "official" API)
  - [x] fix alignment/whitespace in a non-hacky way (ideas welcome)
- [x] make it so that `__init__()` docstrings are shown on the class level
- [x] make it so that docstrings from overloaded methods/functions are all aggregated and displayed
- [ ] write better docstrings (optional, can be its own PR)


JoanFM (changes requested):
Please add, whatever script you used to generate this reference, and make it part of some CI/CD action to update it

samsja:
> Please add, whatever script you used to generate this reference, and make it part of some CI/CD action to update it

The beautifull part is that we did not used any additional script. Turns out sphinx allow to do that automatically. Finding the parameters to do it what an other challenge though

JohannesMessner:
> Please add, whatever script you used to generate this reference, and make it part of some CI/CD action to update it

There is no script, we hand generated the "landing page", following the example of sklearn: https://scikit-learn.org/stable/modules/classes.html
We think it makes sense this way, because it lets us actually choose what our official API is, and what should be hidden from the user; and it lets us put those important objects to the front instead of them being hidden deep in our nested package structure.

Also note that this is still highly WIP, we are having trouble to make these tables look not-horrible.

JohannesMessner:
We are currently **blocked** for the following reason:

For overloaded methods/functions (`@overload`), *only the docstring of the last signature* (the one that holds the actual implementation) *is shown* in the generated docs. The problem is that all the useful information is at the level of the overloaded stubs, since they show the arguments as the users sees them, whereas the implemenetation signature usually only holds `args` and `kwargs`.
An **[example](https://github.com/jina-ai/jina/blob/master/jina/orchestrate/flow/base.py#L105)** for this is the Flow class, where the [implementation signature](https://github.com/jina-ai/jina/blob/master/jina/orchestrate/flow/base.py#L321) has no docstring at all, but the overloaded stubs do.

As I see it, there are two ways of solving this:
1. Have sphinx automatically compile all the overloaded docstrings and display them all. I don't know if this is possible. I found [this](https://sphinx-toolbox.readthedocs.io/en/latest/extensions/more_autodoc/overloads.html) extension which might (?) do what we need it to, but it throws an exception when I try it.
2. Put the entire and complete docstring below the implementation (last) signature, for every overloaded method. Does anyone know if it would be possible to enforce this in linting, similar to how we currently already enforce docstrings for public methods?

JohannesMessner:
CI isn't passing because one commit message is too long. This is a bit of a pain to fix because it is so far back in history, but the problem should solve itself when squash & merge is performed.

jina/orchestrate/flow/base.py
```diff
@@ -916,126 +806,45 @@ def add(
...
-                  * an Executor YAML file (.yml, .yaml, .jaml)
-                  * a Jina Hub Executor (must start with `jinahub://` or `jinahub+docker://`)
-                  * a docker image (must start with `docker://`)
-                  * the string literal of a YAML config (must start with `!` or `jtype: `)
-                  * the string literal of a JSON config
-
-                  When use it under Python, one can use the following values additionally:
-                  - a Python dict that represents the config
-                  - a text file stream has `.read()` interface
-        :param uses_after: The executor attached after the Pods described by --uses, typically used for receiving from all shards, accepted type follows `--uses`. This argument only applies for sharded Deployments (shards > 1).
```
JoanFM: ans where all they go?

jina/orchestrate/flow/base.py
```diff
@@ -812,21 +931,134 @@ def add(
...
+                  * an Executor YAML file (.yml, .yaml, .jaml)
+                  * a Jina Hub Executor (must start with `jinahub://` or `jinahub+docker://`)
+                  * a docker image (must start with `docker://`)
+                  * the string literal of a YAML config (must start with `!` or `jtype: `)
+                  * the string literal of a JSON config
+
+                  When use it under Python, one can use the following values additionally:
+                  - a Python dict that represents the config
+                  - a text file stream has `.read()` interface
+        :param uses_after: The executor attached after the Pods described by --uses, typically used for receiving from all shards, accepted type follows `--uses`. This argument only applies for sharded Deployments (shards > 1).
```
JohannesMessner, line 1033: @JoanFM I think here is what you are looking for. It is present both under the `@overload` stub and under the implementation stub.

JohannesMessner:
Converting back to draft because I may have found the clean solution to create a nicely aligned landing page, including having a single page for each class/function, and automatically pulling descriptions from the docstring. I'll give it a try tomorrow, let's wait with the merge until then.

samsja:
For me the next step of this PR are :

- [ ] Try to show in the search response whereas it is in the docstring on in the User guide
- [ ] remove the classes that are not in the new Python API reference from the search result

JohannesMessner:
> For me the next step of this PR are :
> 
> * [ ]  Try to show in the search response whereas it is in the docstring on in the User guide
> * [ ]  remove the classes that are not in the new Python API reference from the search result

I think search should go in a separate PR, it is its own issue, in this here can provide value without improved search, so we should merge it as soon as we can.

jina/orchestrate/flow/base.py
```diff
@@ -823,9 +931,118 @@ def add(
...
         :param kwargs: other keyword-value arguments that the Deployment CLI supports
         :return: a (new) Flow object with modification
+        :return: a (new) Flow object with modification
+
+        .. # noqa: DAR102
+        .. # noqa: DAR202
+        .. # noqa: DAR101
+        .. # noqa: DAR003
         """
         needs = kwargs.get('needs', None)
```
JohannesMessner, line 1047:
```suggestion
        """
        ## implementation_stub_inject_end_add
        needs = kwargs.get('needs', None)
```

jina/orchestrate/flow/base.py
```diff
@@ -628,72 +739,70 @@ def needs_all(self, name: str = 'joiner', *args, **kwargs) -> 'Flow':
...
+        uses_before: Optional[Union[str, Type['BaseExecutor'], dict]] = None, 
+        uses_before_address: Optional[str] = None, 
+        uses_metas: Optional[dict] = None, 
+        uses_requests: Optional[dict] = None, 
+        uses_with: Optional[dict] = None, 
+        volumes: Optional[List[str]] = None, 
+        when: Optional[dict] = None, 
+        workspace: Optional[str] = None, 
+        **kwargs) -> Union['Flow', 'AsyncFlow']:
         """Add an Executor to the current Flow object.
```
JohannesMessner, line 790:
```suggestion
        **kwargs) -> Union['Flow', 'AsyncFlow']:
         implementation_stub_inject_start_add
         
        """Add an Executor to the current Flow object.

```

JohannesMessner, line 790:
```suggestion
        **kwargs) -> Union['Flow', 'AsyncFlow']:
        
        # implementation_stub_inject_start_add
        """Add an Executor to the current Flow object.

```

PR #4915, opened 2022-06-10
chore: raise warning if port_monitoring is default and already in use

alaeddine-13:
chore: raise warning if port_monitoring is default and already in use

When a user tries to setup monitoring locally and he does not change the default monitoring port for executors, raise a warning to give him a hint

### DIFF
--- a/jina/serve/runtimes/monitoring.py
+++ b/jina/serve/runtimes/monitoring.py
@@ -1,3 +1,6 @@
+from jina import __default_port_monitoring__
+
+
 class MonitoringMixin:
     """The Monitoring Mixin for pods"""
 
@@ -17,4 +20,19 @@ class MonitoringMixin:
 
             from prometheus_client import start_http_server
 
-            start_http_server(self.args.port_monitoring, registry=self.metrics_registry)
+            try:
+                start_http_server(
+                    self.args.port_monitoring, registry=self.metrics_registry
+                )
+            except OSError as e:
+                if (
+                    e.args[0] == 98
+                    and self.args.port_monitoring == __default_port_monitoring__
+                ):
+                    self.logger.warning(
+                        f'The port monitoring {self.args.port_monitoring} is already in use. If you are running Jina '
+                        f'Flow natively, you may have not set different ports for each Executor'
+                    )
+                    raise
+                else:
+                    raise
--- a/jina/serve/runtimes/monitoring.py
+++ b/jina/serve/runtimes/monitoring.py
@@ -33,7 +33,7 @@ class MonitoringMixin:
                 ):
                     self.logger.warning(
                         f'The port monitoring {self.args.port_monitoring} is already in use. If you are running Jina '
-                        f'Flow natively, you may have not set different ports for each Executor'
+                        f'Flow locally, you may have not set different port_monitoring for each Executor'
                     )
                     raise
                 else:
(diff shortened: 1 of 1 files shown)


jina/serve/runtimes/monitoring.py
```diff
@@ -17,4 +20,19 @@ def _setup_monitoring(self):
...
             from prometheus_client import start_http_server
 
-            start_http_server(self.args.port_monitoring, registry=self.metrics_registry)
+            try:
+                start_http_server(
+                    self.args.port_monitoring, registry=self.metrics_registry
+                )
+            except OSError as e:
+                if (
+                    e.args[0] == 98
```
JoanFM, line 29: what is this 98? lets create some constants or something

JoanFM (changes requested):

jina/serve/runtimes/monitoring.py
alaeddine-13: OSError: port already in use

alaeddine-13: updated

hanxiao (changes requested):
`port_monitoring` why is this so special to have a default? we have a lot of `port` using random_port generation, why this `port_monitoring` is so special. I'm guessing it is because it forgets to use `random_port`

samsja:
> `port_monitoring` why is this so special to have a default? we have a lot of `port` using random_port generation, why this `port_monitoring` is so special. I'm guessing it is because it forgets to use `random_port`

**We explicitly decided not to use random port for the monitoring port**. 
The reason is that to leverage the monitoring you need to pass a configuration file with all of the url of the Executor+Gateway to Prometheus so that he can monitor them. If these URLs( port in local case) change at every restart of the Flow this is intractable for Prometheus which will need to be restarted each time. 

The user **needs to provide** a port by himself to the different Executor for monitoring

jina/serve/runtimes/monitoring.py
```diff
@@ -17,4 +22,19 @@ def _setup_monitoring(self):
...
+                    self.args.port_monitoring, registry=self.metrics_registry
+                )
+            except OSError as e:
+                if (
+                    e.args[0] == errno.ENOSR
+                    and self.args.port_monitoring == __default_port_monitoring__
+                ):
+                    self.logger.warning(
+                        f'The port monitoring {self.args.port_monitoring} is already in use. If you are running Jina '
+                        f'Flow natively, you may have not set different ports for each Executor'
```
samsja, line 36:
```suggestion
                        f'Flow locally, you may have not set different port_monitoring for each Executor'
```

samsja (approved):

hanxiao:
> We explicitly decided not to use random port for the monitoring port.
The reason is that to leverage the monitoring you need to pass a configuration file with all of the url of the Executor+Gateway to Prometheus so that he can monitor them. If these URLs( port in local case) change at every restart of the Flow this is intractable for Prometheus which will need to be restarted each time.

This logic is not valid

hanxiao (changes requested):
inconvincing logic. Please use `random_default` strategy

hanxiao:
PR here: https://github.com/jina-ai/jina/pull/4927

hanxiao:
deprecated as #4927 merged

PR #4923, opened 2022-06-13
feat: allow having parameters defined per executor

JoanFM:
Proposal:

The goal is to allow `Parameters` to be sent to only specific Executor without being a breaking change.

This way the proposal is.

In the `parameters` look for the key (KEY COULD BE CHANGED (__jina_parameters_per_executor__)) And if it has the value `True` then we expect that every Executor will receive the parameters that are under its name:

Let's see an example:

```python
from jina import Flow

f.post(parameters= {'__jina_parameters_per_executor__': True, 'exec1': {'key1': val1}}' # like this Exec1 will receive {'key1': val1} in their params and exec2 will receive empty.

f.post(parameters= {'__jina_parameters_per_executor__': False,  'key1': val1}' # like this Exec1  and Exec2 will receive {'key1': val1}

 f.post(parameters= {'key1': val1}' # like this Exec1  and Exec2 will receive {'key1': val1} to keep backwards compatibility
```

TODO:

- [ ] Check optimizations for lazy request access. Check test (tests/integration/gateway_clients/test_clients_gateways.py::test_grpc_gateway_runtime_lazy_request_access)

### DIFF
--- a/jina/clients/mixin.py
+++ b/jina/clients/mixin.py
@@ -3,7 +3,7 @@ from functools import partialmethod, wraps
 from inspect import signature
 from typing import TYPE_CHECKING, AsyncGenerator, Dict, List, Optional, Union
 
-from jina.helper import get_or_reuse_loop, run_async
+from jina.helper import __key_results_parameter__, get_or_reuse_loop, run_async
 from jina.importer import ImportExtensions
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


samsja:
My suggestion:

Using the `key__executor_name` instead of `key` to precise that it should only be sent to the given Executor.

The `__` operator is use for this kind of things in the python word. We use it in [docarray](https://docarray.jina.ai/fundamentals/documentarray/access-attributes/?highlight=tags__#dunder-syntax-for-nested-attributes) : da[:,'tags__key'].

).

```python
from jina import Flow

```

What I don't like regarding the first solution is that it is not easy to pass syntax when we send to all of the Executor to the new one.

Moreover : regarding performance issue of deserialization on the gateway side I suggest to avoid doing any deserialization on the gateway side by sending the full parameters to the Worker and only passing the one relevant to the Executor

PR #4928, opened 2022-06-14
fix(doc): update the description and default questions for qabot

shazhou2015:
Goals:

The current title, description, and default questions are about DocArray in the qabot, correct them to jina.


nan-wang (approved):
LGTM👍

Issue #4919, opened 2022-06-13
Parameters per Executor


JoanFM:
you should be able to send parameters that is going through the flow, but only accepted by one Executor

numb3r3:
For example, we have the following flow consisting of several executors
```
class Filter(Executor):
    @requests
    def search(self, docs, parameters, **kwargs):
        traversal_paths = paramerters.get('traversal_paths', '@r')
        limit = parameters.get('traversal_paths', 100)
        flats = docs[traversal_paths]
        ....
        for d in flats:
            d.matches = d.matches[:limit]

class Indexer(Executor):
    @requests
    def search(self, docs, parameters, **kwargs):
        traversal_paths = paramerters.get('traversal_paths', '@r')
        limit = parameters.get('traversal_paths', 100)
        flats = docs[traversal_paths]
        ....
        for d in flats:
            d.matches = d.matches[:limit]

class Encoder(Executor):
    @requests
    def encode(self, docs, parameters, **kwargs):
        traversal_paths = paramerters.get('traversal_paths', '@r')
        
        flats = docs[traversal_paths]
        flats.embed(...)
        
f = Flow().add(uses=Encoder).add(uses=Filter).add(uses=Indexer)
```

For each request:
- The `Encoder` want to work on '@r', however `Filter` and `Indexer` want to work on `@c`
- The `Filter` want to use `limit=100`, `Indexer` want to use `limit=10`

To solve the above case, can we provide a way to send specific parameters for target executors:

```
client.post(on='/', inputs=da, parameters={
    'version': 1,
    {
        'Encoder': {'traversal_paths': '@r'},
        'Filter': {'traversal_paths': '@c', 'limit': 100},
        'Indexer': {'traversal_paths': '@c', 'limit': 10}
    }
})
```
**Note**: `version` is used here for illustration, maybe we have a better flag name.

PR #4934, opened 2022-06-16
feat: add specific executor param

samsja:
# Context

ATM when you pass a parameters to an Executor `client.index(docs, parameters = {...}` every Executor will receive all of them.
Meaning that if you want to pass different parameters to each Executor you will need to change the name.That's problematic because some Executor except the same parameters, like `traversal_path`.

For instance you can't embed on the chunk with one executor and embed on the root with an other one because you can only pass one `traversal_path`.

**We need to be able to send parameters to different Executor.**

original issue here : https://github.com/jina-ai/jina/issues/4919
this PR take over the old proposition which we decide not to go for : https://github.com/jina-ai/jina/pull/4923 

## Proposal for a Solution

Using the `key__executor_name` instead of `key` to precise that it should only be sent to the given Executor.

The `__` operator is use for this kind of things in the python world. We use it in [docarray](https://docarray.jina.ai/fundamentals/documentarray/access-attributes/?highlight=tags__#dunder-syntax-for-nested-attributes) : da[:,'tags__key'].

).

```python
from jina import Flow

```

Moreover : regarding performance issue of deserialization on the gateway side I suggest to avoid doing any deserialization on the gateway side by sending the full parameters to the Worker and only passing the one relevant to the Executor

## What this PR do:

- [x] Allow to use the `__` syntax to pass executor to one Executor
- [x] tests
- [ ] documentation

## What still need to be fix/done:

- [ ] find a better way to retrieve the original name of the Executor
- [ ] better pasring machanism, something like `__results__` or `__something__` should not be considerate while parsing

### DIFF
--- a/jina/serve/runtimes/request_handlers/data_request_handler.py
+++ b/jina/serve/runtimes/request_handlers/data_request_handler.py
@@ -112,6 +112,15 @@ class DataRequestHandler:
             raise
 
     @staticmethod
+    def _parse_params(parameters: Dict, executor_name: str):
+        parsed_params = parameters
+        specific_parameters = parameters.get(executor_name, None)
+        if specific_parameters:
+            parsed_params.update(**specific_parameters)
+
+        return parsed_params
+
+    @staticmethod
     def _parse_specific_params(parameters: Dict, executor_name: str):
         """Parse the parameters dictionary to filter executor specific parameters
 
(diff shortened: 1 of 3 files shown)


JoanFM (changes requested):
for the logic u mentioned it should be 'executorname__param` and not the other way around right?

PR #4939, opened 2022-06-20
feat: param per executor gateway

samsja:
# Context

fork of https://github.com/jina-ai/jina/pull/4934 to to filtering inside the gateway and not at the Executor level

ATM when you pass a parameters to an Executor `client.index(docs, parameters = {...}` every Executor will receive all of them.
Meaning that if you want to pass different parameters to each Executor you will need to change the name.That's problematic because some Executor except the same parameters, like `traversal_path`.

For instance you can't embed on the chunk with one executor and embed on the root with an other one because you can only pass one `traversal_path`.

**We need to be able to send parameters to different Executor.**

original issue here : https://github.com/jina-ai/jina/issues/4919
this PR take over the old proposition which we decide not to go for : https://github.com/jina-ai/jina/pull/4923 

## Proposal for a Solution

Using the `executor_name__key` instead of `key` to precise that it should only be sent to the given Executor.

The `__` operator is use for this kind of things in the python world. We use it in [docarray](https://docarray.jina.ai/fundamentals/documentarray/access-attributes/?highlight=tags__#dunder-syntax-for-nested-attributes) : da[:,'tags__key'].

).

```python
from jina import Flow

```

In this PR filtering of the key is done  is done on the gateway side to handle correctly the external executor

## What this PR do:

- [x] Allow to use the `__` syntax to pass executor to one Executor
- [x] tests
- [x] documentation

- [x] handle the case `__hello__` in the parsing

### DIFF
--- a/jina/serve/runtimes/gateway/graph/topology_graph.py
+++ b/jina/serve/runtimes/gateway/graph/topology_graph.py
@@ -10,6 +10,7 @@ import grpc.aio
 from jina import __default_endpoint__
 from jina.excepts import InternalNetworkError
 from jina.serve.networking import GrpcConnectionPool
+from jina.serve.runtimes.helper import _parse_specific_params
 from jina.serve.runtimes.request_handlers.data_request_handler import DataRequestHandler
 from jina.types.request.data import DataRequest
 
@@ -60,7 +61,17 @@ class TopologyGraph:
                 copy_req = copy.deepcopy(self.parts_to_send[i])
                 filtered_docs = copy_req.docs.find(self._filter_condition)
                 copy_req.data.docs = filtered_docs
+
                 self.parts_to_send[i] = copy_req
+
+        def _update_request_by_params(
+            self, original_parameters: Dict, deployment_name: str
+        ):
+            specific_paraneters = _parse_specific_params(
+                original_parameters, deployment_name
+            )
+            for i in range(len(self.parts_to_send)):
+                self.parts_to_send[i].parameters = specific_paraneters
 
         def _handle_internalnetworkerror(self, err):
             err_code = err.code()
@@ -105,6 +116,8 @@ class TopologyGraph:
                 # this is a specific needs
                 if len(self.parts_to_send) == self.number_of_parts:
                     self.start_time = datetime.utcnow()
+                    original_parameters = copy.deepcopy(request.parameters)
+                    self._update_request_by_params(original_parameters, self.name)
                     if self._filter_condition is not None:
                         self._update_requests()
                     if self._reduce and len(self.parts_to_send) > 1:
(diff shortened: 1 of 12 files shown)


samsja:
Two propositions to access the parameters without deserializing  the DocumentArray of the DataRequest:
* in the gRPC call dissociate the Parameters from the DocumentArray in order to have two Protos.
* Fake the deserialization of DocumentArray the same way we are doing in core with the lazy access

tests/integration/specific_executor_params/test_specific_params.py
```diff
@@ -0,0 +1,32 @@
...
+        Flow()
+        .add(uses=MyExec, name='exec1', uses_with={'params_awaited': {'key_1': True}})
+        .add(
+            uses=MyExec,
+            name='exec2',
+            uses_with={'params_awaited': {'key_1': True, 'key_2': False}},
+        )
+    )
+
+    with flow:
```
JoanFM, line 28: I prefer the test to return the parameters received in some docs tags and assert on the client side.

JoanFM (changes requested):

samsja:
performance benchmark :

```python
from jina import Executor, requests, Flow
import time


class DummyExecutor(Executor):
    @requests
    def foo(self, docs, **kwargs):
        pass


n_docs = 100000
with Flow().add(uses=DummyExecutor, name='exec') as f:

    initial_time = time.time()
    f.index(
        inputs=DocumentArray(
            [Document(id='1', text='hello') for _ in range(n_docs)],
            parameters={'exec__hello': 'world'},
        ),
    )

    final_time = time.time() - initial_time

    print(f'{n_docs} docs in {final_time} seconds')
    print(f'{n_docs / final_time} docs/sec')
    print(f'{n_docs / final_time / 60} docs/min')
```

current branch : 
```
100000 docs in 3.252817392349243 seconds
30742.580335190043 docs/sec
512.3763389198341 docs/min
```

master :

```
100000 docs in 3.252817392349243 seconds
30742.580335190043 docs/sec
512.3763389198341 docs/min
```

JohannesMessner (dismissed):
It seems like currently we don't explain parameters at all. Please add that as well, and make sure that it appears before this more specific section!

jina/serve/runtimes/gateway/graph/topology_graph.py
```diff
@@ -57,8 +57,9 @@ def __init__(
         def leaf(self):
             return len(self.outgoing_nodes) == 0
 
-        def _update_requests(self):
+        def _update_requests_with_filter_condition(self):
             for i in range(len(self.parts_to_send)):
+                self.parts_to_send[i].docs
```
JoanFM, line 62: why do u need this? is very weird. it is a code smell

tests/integration/conditions_feature/test_condition_behavior.py
```diff
@@ -88,7 +88,7 @@ def flow(request, temp_workspace):
     return f
 
 
-@pytest.mark.parametrize('flow', ['python', 'yaml'], indirect=True)
```
JoanFM, line 91: be careful with this, this test must work

JoanFM (changes requested):

tests/integration/conditions_feature/test_condition_behavior.py
samsja: oups forgot to uncomment

jina/serve/runtimes/gateway/graph/topology_graph.py
samsja: Fixed it know by overriding the deepcopy method of DataRequest

JoanFM (approved):

PR #4941, opened 2022-06-20
fix: fix reconnect issues

JohannesMessner:
Goals:

Currently, local Gateways struggle to re-connect to a previously failing Executor: It can take 1-5sec, despite the Executor being back up.
This PR makes reconnection faster and reliable.

**TODO**:
- [x] Fix the problem when very first request (-> endpoint discovery fails)
- [x] Fix the problem when very first request succeeds, and failure/reconnect happens later
- [x] Solve race condition when multiple request do remove-and-add of same connection
~~- [ ] implement fix in the head~~ not needed, already covered
- [x] remove `delete_connection()` / merge it with `reset_connection()`
- [x] max number of retries in `get_next_connection()`


jina/serve/networking.py
```diff
@@ -740,32 +810,41 @@ async def task_wrapper():
...
-        connection: ConnectionStubs,
+        connection_list: ReplicaList,
         timeout: Optional[float] = None,
     ) -> asyncio.Task:
         # this wraps the awaitable object from grpc as a coroutine so it can be used as a task
         # the grpc call function is not a coroutine but some _AioCall
         async def task_wrapper():
+            connection = None
+            if connection_list:
+                connection = await connection_list.get_next_connection()
```
JohannesMessner, line 835: @JoanFM is there a particular reason this was previously done outside the `async` scope? I had to pull it into here, I think it should not be a problem, but maybe you want to check.

jina/serve/networking.py
```diff
@@ -41,6 +42,52 @@ def __init__(self, summary):
         self._rr_counter = 0  # round robin counter
         self.summary = summary
 
+    async def reset_connection(self, address: str):
```
JoanFM, line 45: please add type annotation on return value

jina/serve/networking.py
```diff
@@ -41,6 +42,52 @@ def __init__(self, summary):
...
+
+        :param address: Target address of this connection
+        :returns: The reset connection or None if there was no connection for the given address
+        """
+        if (
+            address in self._address_to_connection_idx
+            and self._address_to_connection_idx[address] is not None
+        ):
+            # remove connection:
+            self._rr_counter = (
```
JoanFM, line 59: If this code is heavily shared with `add_` and `remove_connection` I would suggest you break those into smaller shared pieces

JoanFM (changes requested):

jina/serve/networking.py
```diff
@@ -41,6 +42,52 @@ def __init__(self, summary):
...
+            self._address_to_channel[address] = None
+            # we should handle graceful termination better, 0.5 is a rather random number here
+            await closing_channel.close(0.5)
+
+            # re-add connection:
+            try:
+                parsed_address = urlparse(address)
+                address = parsed_address.netloc if parsed_address.netloc else address
+                use_tls = parsed_address.scheme in TLS_PROTOCOL_SCHEMES
+            except:
```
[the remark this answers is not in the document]
JohannesMessner: True, but not of my making. I am just making it consistent with what happens in `add_connection`.

tests/integration/runtimes/test_network_failures.py
```diff
@@ -62,12 +63,8 @@ def _create_head(port, connection_list_dict, polling, retries=-1):
 def _send_request(gateway_port, protocol):
```
JoanFM, line 63: I think is better to return and leave the `asserts` to the actual `test` functions so is easier to spot

JoanFM (changes requested):

JoanFM (changes requested):

jina/serve/networking.py
```diff
@@ -89,15 +112,55 @@ async def remove_connection(self, address: str):
...
+                connection = self._connections[internal_rr_counter]
+                # connection is None if it is currently being reset. In that case, try different connection
+                if connection is not None:
+                    break
+            if connection is None and num_retries <= 0:
+                # if still None, then all connections are currently being reset
+                if num_retries <= 0:
+                    raise ConnectionError(
+                        f'All connections in {self._connections} are currently being reset and cannot be used.'
+                    )
```
JohannesMessner, line 157: Here it would probably be better to raise an Exception that inherits from `BaseJinaExceptions`, but there isn't one that fits (in particular, `InternalNetworkError` sounds right, but is reserved as a wrapper around `AioRpcError` and gets special treatment).
Creating a new Exception class just for this seemed excessive.

@JoanFM any suggestions?

tests/integration/runtimes/test_network_failures.py
JohannesMessner: I am not sure how this can be done, since this is often run in its own process (due to reasons of gRPC and pytest not liking each other)

jina/serve/networking.py
JoanFM: u can create a new one with no problem

jina/serve/networking.py
```diff
@@ -86,9 +86,16 @@ def add_connection(self, address: str):
             self._address_to_channel[address] = channel
             self._connections.append(stubs)
 
-    async def remove_connection(self, address: str) -> Union[grpc.aio.Channel, None]:
+    async def _remove_connection(self, address: str) -> Union[grpc.aio.Channel, None]:
```
JoanFM, line 89: This is not private, since it is called from a different class, we cannot just decide to make it private because we want so

JoanFM (changes requested):

JoanFM (approved):

PR #4944, opened 2022-06-21
docs: add wolf annotations document

tarrantro:
Goals:

- ...
- ...

### DIFF
--- a/docs/fundamentals/jcloud/advanced.md
+++ b/docs/fundamentals/jcloud/advanced.md
@@ -90,3 +90,26 @@ executors:
   - name: custom
     uses: jinahub+docker://CustomExecutor
 ```
+
+## Flow retention days
+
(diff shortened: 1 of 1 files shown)


docs/fundamentals/jcloud/advanced.md
```diff
@@ -90,3 +90,26 @@ executors:
   - name: custom
     uses: jinahub+docker://CustomExecutor
 ```
+
+## Flow retention days
+
+In Jcloud, we have default life-cycle(24hrs) for each flow. That means we will remove flows periodically. To avoid it, or manage it by yourself, you can setup `retention_days` args in `jcloud`. `-1` is never expired, `0` is follow default life-cycle, `X`(0<X<365), which means keep the flow utill X days. `0` is default value if you don't pass `retention_days` argument.
```
zac-li, line 96:
```suggestion
In Jcloud, we have default life-cycle(24hrs) for each flow and we will remove flows periodically if they are beyond the life-cycle. To change the default behavior and manage it by yourself, you can setup `retention_days` args in `jcloud`. `-1` is never expired, `0` is to use the default life-cycle, or `X`(0<X<365), which means keep the flow utill X days. `0` is the default value if you don't pass `retention_days` argument.
```

zac-li (commented):
looks good to me otherwise

zac-li (approved):

PR #4950, opened 2022-06-22
feat: allow to load parameters from jina data request proto wo loading the full docs

samsja:
# Context:

Right now there is no way to access the parameters from a `DataRequest` without deserializing the docs which is costly.

What we have now is the lazy loading of the proto which allow to pass the DataRequest from the Gateway to the Executor without any proto deserialization unless we directly call one of the field in which case the deserialization will be done lazily.

This is not enough and we need to be able to only deserialize the header and the parameters of the proto without paying the cost of the docs deserialization.

This will be useful for this [PR](https://github.com/jina-ai/jina/pull/4939) where we  want to modify the parameters on the fly without touching the docs.

## What this PR do:

- [ ] Add a new proto `DataRequestWoDocs` which will be used when we don't need the docs
- [ ] Parameters/Header access use the `DataRequestWoDocs` at lazy deserialization 
- [ ] tests
- [ ] Documentation

### DIFF
--- a/jina/types/request/data.py
+++ b/jina/types/request/data.py
@@ -98,6 +98,8 @@ class DataRequest(Request):
         request: Optional[RequestSourceType] = None,
     ):
         self.buffer = None
+        self._buffer_when_load_wo_data = None
+
         try:
             if isinstance(request, jina_pb2.DataRequestProto):
                 self._pb_body = request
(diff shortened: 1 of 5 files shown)


jina/types/request/data.py
```diff
@@ -98,6 +98,8 @@ def __init__(
         request: Optional[RequestSourceType] = None,
     ):
         self.buffer = None
+        self._buffer_when_load_wo_data = None
```
JoanFM, line 101: I think this naming is weird. I would like to see. buffer_with_docs, buffer_wo_docs

jina/types/request/data.py
```diff
@@ -185,7 +241,7 @@ def parameters(self) -> Dict:
         :return: a Python dict view of the parameters.
         """
         # if u get this u need to have it decompressed
-        return json_format.MessageToDict(self.proto.parameters)
+        return json_format.MessageToDict(self.proto_wo_data.parameters)
```
JoanFM, line 244: if I want to get the data, it is very weird that I access proto_wo_data

tests/unit/types/request/test_request.py
```diff
@@ -116,4 +118,26 @@ def test_status():
...
+    r = DataRequest()
+    da = r.docs
+    da.extend([Document(text='534534534er5yr5y645745675675675345')] * doc_count)
+    r.data.docs = da
+
+    parameters = {'a': 0}
+    r.parameters = parameters
+    byte_array = DataRequestProto.SerializeToString(r)
+
+    deserialized_request = DataRequestProto.FromString(byte_array)
```
JoanFM, line 136: This should be callig the regular DataRequest contstructor, this is a unit test independent of the Proto optimization we have in serializer

JoanFM (changes requested):

samsja:
@JoanFM I will polish the PR once most of the test are passing in the CI. By polishing I mean working on the renaming of the variable and certainly some re factorization.

I want to be sure that I am not breaking to much things with the current solution before spending time on code optimization

jina/types/request/data.py
samsja: Probably a bug in my solution then. I agree that it makes little sense

samsja:
`tests.integration.reduce.test_reduce::test_reduce_status` is failing. I think I know the reason. It might be that once we serialize the DataRequest with a DataRequestWoDocs under the hood for proto we don't send the docs. ( Which is definitely a big issue )

JoanFM (changes requested):
It looks very good to me. I know it must already exist indirectly somewhere, but I would like to have an integration test related directly to this feature. A Flow where both parameters and `docs` are used in an Executor that proves still works.

samsja:
> It looks very good to me. Congrats on this idea! I know it must already exist indirectly somewhere, but I would like to have an integration test related directly to this feature. A Flow where both parameters and `docs` are used in an Executor that proves still works.

You are right I will design a specific test !

JohannesMessner:
Looks quite clear to me, good job!
The only thing that i find a little bit confusing is the naming convention betwenn `wo_data` and `wo_docs`; on the proto level it is `wo_docs`, but on the `DataRequest` level all methods etc. are named with `wo_data`, but from my understanding they are actually both the same.
I think this should be unified, technically `wo_data` would be more correct I guess because the entire `DataContent` is removed.

JoanFM (approved):

PR #4956, opened 2022-06-24
fix: handle replicas in ui monitoring

samsja:
# Context:

In the UI the `port_monitoring` of the replicas are not shown. 

# What this pr do 

Fix the behavior and iterate over the pods to show the replicas `port_monitoring`
```python
from jina import Flow

with Flow(monitoring=True).add(name='hello', port_monitoring=9090, replicas=2) as f:
    f.block()
```

Before
Screenshot from 2022-06-24 12-17-00

After 
Screenshot from 2022-06-24 12-17-25


hanxiao:
image

PR #4961, opened 2022-06-24
feat: allow to pass a list of port monitoring to replicas 

samsja:
# Context:

Right now it is difficult to use the monitoring feature with replicas/shards because it is impossible to gave a list of port to the Deployment. The current behavior is to randomly assign the port when there is a replicas. It is actually the same behavior as for the normal port. But since the `port_monitoring` should be access by Prometheus that is external we can't only rely on this behavior.

It would be great if we could pass several port to a Deployment.

# What this PR do:

Allow to pass a list of port to the Deployment. The ports should be separated by a comma.

```python
from jina import Flow, Executor

with Flow(monitoring=True).add(uses=Executor, replicas=2, port_monitoring='9092,9091'):
   ...
```

- [x] Allow to pass a list
- [x] Add tests
- [x] Add documentation


JoanFM (changes requested):
I guess in k8s it does not apply right?

jina/orchestrate/deployments/__init__.py
```diff
@@ -691,6 +710,36 @@ def _set_pod_args(args: Namespace) -> Dict[int, List[Namespace]]:
                     # if there are no shards/replicas, we dont need to distribute ports randomly
                     # we should rather use the pre assigned one
                     args.port = args.port
+                    _args.port_monitoring = args.all_port_monitoring[0]
+                elif args.shards == 1 and args.replicas > 1:
+                    _args.port_monitoring = (
```
JoanFM, line 715: can't this be generalized as having one counter with a max_num_ports? and some custom iteration?

JoanFM (changes requested):

jina/orchestrate/deployments/__init__.py
samsja: Yes probably I will do some refactoring

samsja:
> I guess in k8s it does not apply right?

No need needed for k8s

docs/fundamentals/flow/monitoring-flow.md
```diff
@@ -91,6 +91,51 @@ On the other hand, If you want to only enable the monitoring on a given Executor
 Flow().add(...).add(uses=MyExecutor, monitoring=True)
 ```
 
+### Enable monitoring with replicas and shards
+
+When using replicas or shards (see this how-to scale-out) all of the duplicate Executor will expose their own metrics.
```
JohannesMessner, line 96: This doesn't look like a proper internal link

docs/fundamentals/flow/monitoring-flow.md
```diff
@@ -91,6 +91,51 @@ On the other hand, If you want to only enable the monitoring on a given Executor
...
+
+When deploying a Flow locally (i.e without Kubernetes or Docker Compose) all the duplicate will use a random `port_monitoring`
+expect one that will use the `port_monitoring` that you precise. You can as well pass a list of ports to be used by the different replicas 
+for the monitoring by passing a string of several ports separated by a comma.
```
JohannesMessner, line 100: Could we rephrase this?
I would first explain how to pass a list using the comma, then give the examples.
Only below that i would describe what happens if only one port is given, as that seems like a special case to me (and there i would first say that one replica gets the specified port, and the other ones are random, instead of the other way around. I think that would be clearer.)

jina/helper.py
```diff
@@ -1618,3 +1618,30 @@ def is_port_free(host, port):
             return False
         else:
             return True
+
+
+def _parse_port(port: str) -> Union[int, List]:
```
JohannesMessner, line 1623:
```suggestion
def _parse_ports(port: str) -> Union[int, List]:
```
i think this should be plural

JohannesMessner (changes requested):
Looks mostly good!

docs/fundamentals/flow/monitoring-flow.md
samsja: Oh that's because I forgot to add the link

docs/fundamentals/flow/monitoring-flow.md
samsja: Sure

jina/helper.py
samsja: will do

jina/orchestrate/deployments/__init__.py
```diff
@@ -101,6 +101,13 @@ def head_port(self):
         """
         return self.head_args.port if self.head_args else None
 
+    @property
+    def head_port_monitoring(self):
```
JoanFM, line 105: why is this property relevant?

JohannesMessner (dismissed):

jina/orchestrate/deployments/__init__.py
samsja: Well there is one for the port so it could be usefull to have one fort port_monitoring. It think that it is mainly used in test for now

JoanFM (approved):

PR #4971, opened 2022-06-27
build: add hub integration

floralatin:
# Context : 

In order to catch errors before the Jina core's release, we need to test Hubble integration on the core's master branch

# What the pr is doing
make some base api sanity check https://github.com/jina-ai/hub-integration

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -245,15 +245,6 @@ jobs:
           JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
           JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
 
-  hub-integration: 
-    needs: [commit-lint, lint-flake-8, code-injection]
-    uses: jina-ai/hub-integration/.github/workflows/main.yml@main
-    with:
-      actions: 'all'
-    secrets:
-      jina_dev_bot: ${{ secrets.JINA_DEV_BOT }}
-      jina_auth_token: ${{ secrets.JINA_AUTH_TOKEN }}
-
   k8s-test:
     needs: update-schema
     runs-on: ubuntu-latest
@@ -396,7 +387,7 @@ jobs:
   # just for blocking the merge until all parallel core-test are successful
   success-all-steps:
     runs-on: ubuntu-latest
-    needs: [core-test, import-test, hub-test, hub-integration, k8s-test, k8s-failures-test, docker-compose-test, docker-image-test, benchmark-pre-release, update-schema, update-docker]
+    needs: [core-test, import-test, hub-test, k8s-test, k8s-failures-test, docker-compose-test, docker-image-test, benchmark-pre-release, update-schema, update-docker]
     if: always()
     steps:
       - uses: technote-space/workflow-conclusion-action@v2
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -245,6 +245,15 @@ jobs:
           JINAHUB_USERNAME: ${{ secrets.JINAHUB_USERNAME }}
           JINAHUB_PASSWORD: ${{ secrets.JINAHUB_PASSWORD }}
 
+  hub-integration: 
+    needs: [commit-lint, lint-flake-8, code-injection]
+    uses: jina-ai/hub-integration/.github/workflows/main.yml@main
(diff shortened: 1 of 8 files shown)


JoanFM (commented):
Doesn't this generate a very dangerous cross-dependency @mapleeit @nomagick ?

mapleeit:
What do you mean by very dangerous?

JoanFM:
> What do you mean by very dangerous?

that there will be a deadlock between Hub and Jina Core PRs

mapleeit:
We should only add this before release as we discussed before. @floralatin 

Then I guess it's fine since it will be only triggered when releasing? @JoanFM

floralatin:
> We should only add this before release as we discussed before. @floralatin
> 
> Then I guess it's fine since it will be only triggered when releasing? @JoanFM

i see

mapleeit (changes requested):
Remove it in `ci.yml`. Only trigger it before releasing

JoanFM:
Can we then do it in `CD` after merge but not requiring it to have a successfull CD? So that Hub can get the alert but not pollute the CD status?

mapleeit:
> Can we then do it in CD after merge but not requiring it to have a successfull CD? So that Hub can get the alert but not pollute the CD status?

But then we'll need another patch version if things go wrong. The purpose is to make sure that **before releasing**, we know something is wrong and we need to fix something instead of after. 

The same logic comes for Wolf, Jina Now and other products as well.

floralatin:
> Remove it in `ci.yml`. Only trigger it before releasing

yes, i will improve it

JoanFM:
> > Can we then do it in CD after merge but not requiring it to have a successfull CD? So that Hub can get the alert but not pollute the CD status?
> 
> But then we'll need another patch version if things go wrong. The purpose is to make sure that **before releasing**, we know something is wrong and we need to fix something instead of after.
> 
> The same logic comes for Wolf, Jina Now and other products as well.

WOLF and NOW depend on Jina and they treat every Jina version update as a PR where they run their own tests. Having Jina NOW fail is not a reason for Core not to release. For Hub it could be different, but then CD is not the right place to be, and CI should be.

mapleeit:
Fair point. But the test have some limits. It can't be used in PR since we can't install merged code in Hubble in a unmerged PR context. (We can specify a branch or commit for Hubble to use, but for unmerged PR, it's hard)

So right now, the best place is CD. And in Hubble, we also use the same test before releasing.

https://github.com/jina-ai/hubble/blob/master/.github/workflows/deploy-on-tag.yml#L111

JoanFM:
> Fair point. But the test have some limits. It can't be used in PR since we can't install unmerged code in Hubble. (We can specify a branch or commit for Hubble to use, but it's hard to use merged content in a unmerged PR)
> 
> So right now, the best place is CD. And in Hubble, we also use the same test before releasing.
> 
> https://github.com/jina-ai/hubble/blob/master/.github/workflows/deploy-on-tag.yml#L111

but on CD, if we have already released, what is the point?

mapleeit:
By CD, I mean before official release. Should be "Manual Release" in Jina Core?

floralatin:
so now. we only add  this before jina core release? and  we make it have some optional ：blocking and on-blocking ？

floralatin:
so we only add this before jina release ?

mapleeit:
> so we only add this before jina release ?

Yes. I think it should be `force-release.yml`, but @floralatin please make sure it's right place. 

Let's improve whenever it goes not well.

floralatin:
> > so we only add this before jina release ?
> 
> Yes. Joan said it should be `force-release.yml`, but @floralatin please make sure it's right place.
> 
> Let's improve whenever it goes not well.

fine

floralatin:
> Remove it in `ci.yml`. Only trigger it before releasing
done

mapleeit (approved):
LGTM 

---

before merge. We need to add the `JINA_AUTH_TOKEN` to secrets.

floralatin:
> Can we add description on the PR before we merge ?
> 
> At minima
> 
> * the context of the pr : why we need the PR
> * What the pr is doing

the context of the pr :  In order to catch errors before the Jina core's release, we need to test Hubble integration on the core's master branch 

What the pr is doing:  make some  base api sanity check

JoanFM (approved):

PR #4974, opened 2022-06-27
fix: fix monitoring update of metrics on exception

JoanFM:
Goals:

When requests fail due to connection errors, the metrics related do not get updated


samsja (approved):
lgtm

PR #4985, opened 2022-06-30
feat: set prefetch default to 1000

JoanFM:
Goals:

Setting prefetch default value to 1000 to avoid common OOM in Gateway


jina/serve/stream/helper.py
```diff
@@ -1,18 +1,48 @@
-from typing import Iterator, AsyncIterator, Union
+import asyncio
+from typing import AsyncIterator, Iterator, Optional, Union
 
 from jina.helper import get_or_reuse_loop
 
 
+class RequestsCounter:
```
samsja, line 7: I am wondering if we should not name it `_RequestsCounter`. I remember that in the past we have been locked into not refactoring helper function (rename them or move them) because they were not private and thus users could rely on it. If it is a private Class it kind of gave the right to refactor as we want.

samsja (commented):
LGTM !

One small comments though on the name of the helper Class

jina/serve/stream/helper.py
JoanFM: I will do later in another PR related to this part of the code

samsja: sure

PR #4991, opened 2022-07-04
fix: fix compress method

samsja:
# Context

This pr is the direct follow up of https://github.com/jina-ai/jina/pull/4950.

The last pr allow us to access the parameters of the DataRequest wo loading the data( DocumentArray). 

We need sometime to set the parameters wo loading the data as well. Especially for https://github.com/jina-ai/jina/pull/4939

# What this pr do:
This PR allow to set the parameters wo loading the data. 

- [ ] implementation
- [ ] tests


jina/types/request/data.py
```diff
@@ -127,15 +128,25 @@ def __init__(
...
-        self._pb_body_wo_data = None
-
     @property
     def is_decompressed(self) -> bool:
+        """
+        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto` or
+        :class:`jina.proto.jina_pb2.DataRequestProtoWoData`
+           :return: True if the proto was deserialized before
+        """
+        return type(self._pb_body) in [
```
JoanFM, line 138: why was last logic not valid?

JoanFM (commented):
what was the error?

jina/types/request/data.py
```diff
@@ -127,15 +128,25 @@ def __init__(
                 f'fail to construct a {self.__class__} object from {request}'
             ) from ex
 
-        self._pb_body_wo_data = None
-
     @property
     def is_decompressed(self) -> bool:
+        """
+        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto` or
```
JohannesMessner, line 134: Maybe explicitly state in the docstring that this does not necessarily mean that the data was decompressed? Maybe there even is a name for this property that would immediately convey that.

jina/types/request/data.py
```diff
@@ -157,32 +168,33 @@ def proto_wo_data(
...
         """
-        if self.is_decompressed:
-            return self._pb_body
-        elif self.is_decompressed_wo_data:
-            return self._pb_body_wo_data
-        else:
+        if self._pb_body is None:
             self._decompress_wo_data()
-            return self._pb_body_wo_data
+        return self._pb_body
```
JohannesMessner, line 173: If `self._pb_body` is a proto *with* data (so not `None`), this will return a proto with data, right? That would be unexpected for me.

jina/types/request/data.py
```diff
@@ -157,32 +168,33 @@ def proto_wo_data(
...
         """
         Cast ``self`` to a :class:`jina_pb2.DataRequestProto`. Laziness will be broken and serialization will be recomputed when calling.
-        Under the hood it is calling :meth:`jina.types.request.data.DataRequest.proto_wo_data`. This method is keep for legacy.
         :meth:`SerializeToString`.
         :return: DataRequestProto protobuf instance
         """
-        return self.proto_data
+        if not self.is_decompressed:
+            self._decompress()
+        return self._pb_body
```
JohannesMessner, line 186: I think this could return either a proto with or without data, but the docstring doesn't match that behaviour (the type hint does, though).

JohannesMessner (changes requested):

jina/types/request/data.py
samsja: It rellied on the fact that buffer is None if _pb_body. This one is more precise and clear with the new change

jina/types/request/data.py
samsja: Even if you originally wanted to only access only the parameters, if the proto is already loaded with data then you can just use it because the expensive step of deserialize the data is already done.

JohannesMessner: I am guessing the user experience is exactly the same if a body with data is returned, but one without data is expected? If so, then it's ok

jina/types/request/data.py
```diff
@@ -127,25 +128,34 @@ def __init__(
...
+        """
+        return type(self._pb_body) in [
+            jina_pb2.DataRequestProto,
+            jina_pb2.DataRequestProtoWoData,
+        ]
+
+    @property
+    def is_decompressed_with_data(self) -> bool:
+        """
+        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto`. In this case the full proto is loaded including tha data ( docs )
```
JohannesMessner, line 146:
```suggestion
        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto`. In this case the full proto is decompressed, including the data (docs).
```

jina/types/request/data.py
```diff
@@ -127,25 +128,34 @@ def __init__(
...
             ) from ex
 
-        self._pb_body_wo_data = None
-
     @property
     def is_decompressed(self) -> bool:
         """
-        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto`
+        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto` or
+        :class:`jina.proto.jina_pb2.DataRequestProtoWoData` This does not necessarily mean that the docs are loaded.
```
JohannesMessner, line 135:
```suggestion
        :class:`jina.proto.jina_pb2.DataRequestProtoWoData`. This does not necessarily mean that the data (docs) inside the request is also decompressed.
```

jina/types/request/data.py
```diff
@@ -127,25 +128,34 @@ def __init__(
...
+        return type(self._pb_body) in [
+            jina_pb2.DataRequestProto,
+            jina_pb2.DataRequestProtoWoData,
+        ]
+
+    @property
+    def is_decompressed_with_data(self) -> bool:
+        """
+        Checks if the underlying proto object was already deserialized into a :class:`jina.proto.jina_pb2.DataRequestProto`. In this case the full proto is loaded including tha data ( docs )
            :return: True if the proto was deserialized before
```
JohannesMessner, line 147:
```suggestion
           :return: True if the proto was deserialized before, including the data (docs)
```

JohannesMessner (changes requested):

PR #4992, opened 2022-07-04
(GitHub CI) Bump actions/checkout and actions/setup-python versions in ci.yml

ArjunSharda:
This will bump versions for actions/checkout and actions/setup-python

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
   commit-lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


JoanFM:
Hello @ArjunSharda ,

Thanks for the contribution, however we prefer to keep the stability on our CI that we currently have with current versions

ArjunSharda:
Ah alright. I don't think updating them will do a breaking change though.

ArjunSharda:
And if you want I can also change more of the workflow files to update it.

PR #5001, opened 2022-07-07
fix(hubble): fix the way to generate executor name of the func hub pull

floralatin:
Fix the way to generate executor name of the func hub pull

Close: https://github.com/jina-ai/hubble/issues/475

### DIFF
--- a/jina/hubble/hubio.py
+++ b/jina/hubble/hubio.py
@@ -787,7 +787,7 @@ metas:
                     force=need_pull,
                 )
 
-                presented_id = getattr(executor, 'name') if getattr(executor, 'name') is not None else executor.uuid
+                presented_id = executor.name if executor.name else executor.uuid
                 executor_name = (
                     f'{presented_id}'
                     if executor.visibility == 'public' or not secret
--- a/jina/hubble/hubio.py
+++ b/jina/hubble/hubio.py
@@ -787,7 +787,7 @@ metas:
                     force=need_pull,
                 )
 
-                presented_id = getattr(executor, 'name', executor.uuid)
+                presented_id = getattr(executor, 'name') if getattr(executor, 'name') is not None else executor.uuid
                 executor_name = (
                     f'{presented_id}'
                     if executor.visibility == 'public' or not secret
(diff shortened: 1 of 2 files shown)


mapleeit (changes requested):
I think the change here is invalid, have you tested it?

floralatin:
> I think the change here is invalid, have you tested it?

https://files.slack.com/files-pri/T0189QW56UE-F03N6QCG2S3/image.png

mapleeit (approved):

jina/hubble/hubio.py
```diff
@@ -787,7 +787,7 @@ def pull(self) -> str:
                     force=need_pull,
                 )
 
-                presented_id = getattr(executor, 'name', executor.uuid)
+                presented_id = getattr(executor, 'name') if getattr(executor, 'name') is not None else executor.uuid
```
mapleeit, line 790: Does it make sense if we just use `executor.name` instead of `getattr(executor, 'name)`

floralatin: yes

mapleeit (approved):

JoanFM (changes requested):
Please add a test covering this

floralatin:
> Please add a test covering this

yes, thanks

mapleeit (changes requested):
The test doesn't make sense, it doesn't cover https://github.com/jina-ai/hubble/issues/475. We need to make sure assert the display name.

floralatin:
> The test doesn't make sense, it doesn't cover [jina-ai/hubble#475](https://github.com/jina-ai/hubble/issues/475). We need to make sure assert the display name.

 let me think and  fix it

tests/unit/hubble/test_hubio.py
```diff
@@ -375,8 +375,8 @@ def iter_content(self, buffer=32 * 1024):
     def status_code(self):
         return self.response_code
 
-
-def test_pull(test_envs, mocker, monkeypatch):
+@pytest.mark.parametrize('executer_name', ['alias_dummy',None])
```
mapleeit, line 378:
```suggestion
@pytest.mark.parametrize('executer_name', ['alias_dummy', None])
```

mapleeit (approved):

JoanFM (approved):

PR #5038, opened 2022-07-28
fix: fix specific params problem with branches

JoanFM:
Solves an issue with `executor specific parameters` per Executor.

Original problem is that `request` is concurrently altered and there are issues on how they are updated before they are sent.

This problem is now solved by deepcopying the request before sending, but we have to analyze if this is the right approach since it can imply performance degradation. To solve this I added some optimizations in the code to detect when it is actually needed

Here I just want to validate this functionality would work, specially if the `__results__` handling is correct

### DIFF
--- a/jina/serve/runtimes/gateway/graph/topology_graph.py
+++ b/jina/serve/runtimes/gateway/graph/topology_graph.py
@@ -101,7 +101,8 @@ class TopologyGraph:
             endpoint: Optional[str],
             executor_endpoint_mapping: Optional[Dict] = None,
             target_executor_pattern: Optional[str] = None,
-            request_input_parameters: Dict = {}
+            request_input_parameters: Dict = {},
+            copy_request_at_send: bool = False
         ):
             # Check my condition and send request with the condition
             metadata = {}
@@ -114,7 +115,10 @@ class TopologyGraph:
                 request.parameters = _parse_specific_params(
                     request.parameters, self.name
                 )
-                self.parts_to_send.append(copy.deepcopy(request))
+                if copy_request_at_send:
+                    self.parts_to_send.append(copy.deepcopy(request))
+                else:
+                    self.parts_to_send.append(request)
                 # this is a specific needs
                 if len(self.parts_to_send) == self.number_of_parts:
                     self.start_time = datetime.utcnow()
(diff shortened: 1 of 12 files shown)


jina/serve/runtimes/gateway/graph/topology_graph.py
```diff
@@ -44,33 +44,34 @@ def __init__(
...
+                req = self.parts_to_send[i] if not need_copy else copy.deepcopy(self.parts_to_send[i])
+                filtered_docs = req.docs.find(self._filter_condition)
+                req.data.docs = filtered_docs
 
-                self.parts_to_send[i] = copy_req
-
-        def _update_request_by_params(self, deployment_name: str):
+                self.parts_to_send[i] = req
+                # filtered_docs = self.parts_to_send[i].docs.find(self._filter_condition)
+                # self.parts_to_send[i].data.docs = filtered_docs
```
JohannesMessner, line 69: Commented code

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -130,6 +132,13 @@ def _handle_request(request: 'Request') -> 'Tuple[Future, Optional[Future]]':
             r.start_time.GetCurrentTime()
             # If the request is targeting a specific deployment, we can send directly to the deployment instead of
             # querying the graph
+            num_outgoing_nodes = len(request_graph.origin_nodes)
+            has_specific_params = False
+            request_input_parameters = request.parameters
+            for key in request_input_parameters:
+                if _is_param_for_specific_executor(key):
+                    has_specific_params = True
+
```
JohannesMessner, line 141: I think we can directly exit the loop after the `if` is true for the first time

tests/integration/specific_executor_params/test_specific_params.py
```diff
@@ -32,3 +33,65 @@ def process(self, docs, parameters, **kwargs):
...
+    class EmbeddingCombinerTestSpecific(Executor):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.model = lambda emb1, emb2: np.concatenate(
+                [emb1, emb2], axis=1
+            )  # initialize dummy model to combine embeddings
+
+        @requests(on='/encode')
+        def combine(self, docs, parameters, **kwargs):
+            text1_path = parameters.get('text1_access_path', None)
```
JohannesMessner, line 62: Can we have asserts directly in the Executors to make sure that the params arrive correctly?

jina/serve/runtimes/gateway/request_handling.py
JoanFM: yes

tests/integration/specific_executor_params/test_specific_params.py
JoanFM: if they Get None the Executor will raise exception so this is also checked.

JohannesMessner (approved):

PR #5039, opened 2022-07-28
docs: add walkthrough in docs

JoanFM:


docs/how-to/walkthrough.md
```diff
@@ -0,0 +1,314 @@
...
+ - How to deploy it in Kubernetes
+
+For that we want to build a Flow that looks like this:
+
+```{figure} simple_flow_walkthrough.svg
+:width: 70%
+
+```
+
+In this case, we are going to see how to build a Flow when we will use a public Executor from the `Hub` and a private Executor built and developed by us.
```
CatStark, line 18: I don't get this part, it's about creating a Flow to be used with public executors. And then, different topic, creating a private Executor?

PR #5041, opened 2022-08-03
fix: do not send target_executor to Executors

JoanFM:
Goals:

Fix problem trying to connect to `CaS` as External Executor with `target_executor` parameter.

Since it is a Gateway-Gateway communication, you are actually passing the target_executor also to the second gateway, and in the external Flow, the executor may not be called the same and then it does not match any and returns the Docs unchanged.
The fix is to remove target_executor from the request once it is extracte


samsja (approved):
lgtm

PR #5053, opened 2022-08-10
feat: add tests for disable external exec

samsja:
# context

fix some issue when saying `external=False` but still having `host` 

## What this pr do :

```python
Flow().add(host='grpc://localhost:12345', external=False, use=MyExec)
```

this will not use the external Executor anymore as intended


samsja:
Original issue has been closed not need to continue to work on this pr

PR #5054, opened 2022-08-10
fix: add type checksum conversion of the function fetch_meta

floralatin:
Goals:

there was a bug where pulling an Executor with build env vars would not work, because it tried to cache the keys of these env vars, but the DictKeys object can't be pickled. This PR converts these keys to a list, thus fixing the bug.

The diffs are large, but that's only because of the black.
The only relevant line is build_env=list(buildEnv.keys()) if buildEnv and isinstance(buildEnv, dict) else [], line 715 in hubio.py (and the added test).

### DIFF
--- a/jina/hubble/hubio.py
+++ b/jina/hubble/hubio.py
@@ -698,7 +698,7 @@ metas:
             image_name=image_name,
             archive_url=resp['package']['download'],
             md5sum=resp['package']['md5'],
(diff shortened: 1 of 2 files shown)


floralatin:
Have fixed  by this pull request   https://github.com/jina-ai/jina/pull/5061

PR #5061, opened 2022-08-15
fix: cache hub pull with build env

JohannesMessner:
there was a bug where pulling an Executor with build env vars would not work, because it tried to cache the keys of these env vars, but `DictKeys` object can't be pickled. This PR converts these keys to a list, thus fixing the bug.

The diffs are large, but that's only because of `black`.
The only relevant line is `build_env=list(buildEnv.keys()) if buildEnv else [],`, line 715 in `hubio.py` (and the added test).


floralatin:
There is already a pr to fix it https://github.com/jina-ai/jina/pull/5054

JohannesMessner:
> There is already a pr to fix it #5054

I see now. Would it be possible to always add a description to your PRs? Otherwise we won't know what your PR does, and in a past retro we also agreed not to review PRs without description. Thank you!

jina/hubble/hubio.py
```diff
@@ -347,43 +347,53 @@ def push(self) -> None:
...
+                        f'The --build-env parameter: `{env}` is wrong format. you can use: `--build-env {env}=YOUR_VALUE`.'
+                    )
                 if check_requirements_env_variable(env_list[0]) is False:
-                    raise Exception( f'The --build-env parameter key:`{env_list[0]}` can only consist of uppercase letter and number and underline.')
+                    raise Exception(
+                        f'The --build-env parameter key:`{env_list[0]}` can only consist of uppercase letter and number and underline.'
+                    )
                 build_env_dict[env_list[0]] = env_list[1]
-            build_env = build_env_dict if len(list(build_env_dict.keys()))>0 else None
+            build_env = build_env_dict if len(list(build_env_dict.keys())) > 0 else None
```
nan-wang, line 366:
```suggestion
            build_env = build_env_dict if build_env_dict else None
```

jina/hubble/hubio.py
```diff
@@ -347,43 +347,53 @@ def push(self) -> None:
...
-                if (len(env_list) != 2):
-                    raise Exception( f'The --build-env parameter: `{env}` is wrong format. you can use: `--build-env {env}=YOUR_VALUE`.')
+                if len(env_list) != 2:
+                    raise Exception(
+                        f'The --build-env parameter: `{env}` is wrong format. you can use: `--build-env {env}=YOUR_VALUE`.'
+                    )
                 if check_requirements_env_variable(env_list[0]) is False:
-                    raise Exception( f'The --build-env parameter key:`{env_list[0]}` can only consist of uppercase letter and number and underline.')
+                    raise Exception(
+                        f'The --build-env parameter key:`{env_list[0]}` can only consist of uppercase letter and number and underline.'
```
nan-wang, line 363:
```suggestion
                        f'The `--build-env` parameter key:`{env_list[0]}` can only consist of uppercase letter and number and underline.'
```

nan-wang (changes requested):
minor changes

floralatin:
I think the following code should be used

`build_env=list(buildEnv.keys()) if buildEnv and isinstance(buildEnv, dict) else []`

nan-wang (approved):
LGTM👍

floralatin (approved):

floralatin (changes requested):
should use`build_env = build_env_dict if len(list(build_env_dict.keys())) > 0 else None`
not ` build_env = build_env_dict if build_env_dict else None`. if use  ` build_env = build_env_dict if build_env_dict else None` it is always true

hanxiao (approved):

PR #5080, opened 2022-08-23
docs: remove kong/alb from jcloud document

tarrantro:
**Goals:**

- remove alb/kong from jcloud gateway tab to avoid confusion.
- correct some GPU notes

- [ ] check and update documentation. See [guide](https://github.com/jina-ai/jina/CONTRIBUTING.md#documentation-guidelines) and ask the team.

### DIFF
--- a/docs/fundamentals/jcloud/yaml-spec.md
+++ b/docs/fundamentals/jcloud/yaml-spec.md
@@ -60,7 +60,7 @@ executors:
 
 JCloud supports GPU workloads with two different usages: `shared` or `dedicated`. 
 
-If GPU is enabled, JCloud will provide NVIDIA A10G Tensor Core GPUs for workloads in both usage types.
+If GPU is enabled, JCloud will provide NVIDIA A10G Tensor Core GPUs with 24G memory for workloads in both usage types.
 
 ```{note}
 When using GPU resources, it may take few extra mins until all Executors ready to serve traffic.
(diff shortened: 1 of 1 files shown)


docs/fundamentals/jcloud/yaml-spec.md
```diff
@@ -81,12 +81,8 @@ executors:
         gpu: shared
 ```
 
-```{note}
-When using shared GPU resources, it will share the GPU memory across pods(24G memory total). If your application is memory consuming, we suggest using a dedicated GPU.
-```
-
 ```{caution}
-There are no special provisions in place to isolate replicas that run on the same underlying GPU. Each workload has access to the GPU memory and runs in the same fault-domain as of all the others. Therefore, if one workload crashes, they all do. 
+The tradeoffs with `shared` GPU are increased latency, jitter, and potential out-of-memory (OOM) conditions when many different applications are time-slicing on the GPU. This mechanism is what we focus on improving. If your application is memory consuming, we suggest using a dedicated GPU.
```
deepankarm, line 85: > This mechanism is what we focus on improving.

We can skip mentioning this, I guess.

deepankarm (changes requested):
Can you also fix the typo here? [Other deployment pptions](https://docs.jina.ai/fundamentals/jcloud/yaml-spec/#other-deployment-pptions)

docs/fundamentals/jcloud/yaml-spec.md
tarrantro: ok

deepankarm (approved):

PR #5082, opened 2022-08-23
feat: bump protobuf version

alaeddine-13:
**Goals:**
We aim to use latest protobuf version in python: 4.21
However, since some Executors might contain dependencies that only work with 3.20 or less, we want to be able to only use latest protobuf version when possible.

We solve this using the following method:

1. create new compiler v21 dockerfile and update build script accordingly
2. generate protobuf modules twice using compilers v20 and v21
3. load the appropriate modules dynamically depending on the installed protobuf version

To be merged in conjunction with https://github.com/jina-ai/docarray/pull/371

closes: https://github.com/jina-ai/jina/issues/5083


samsja (changes requested):
we need to keep the new proto files (`jina_pb2_grpc.py` and `jina_pb2.py`) into `jina/proto`)
otherwise it is a breaking change for ppl doing import like this
```python
from jina.proto.jina_pb2 import DataRequestProto
```

so I suggest the following files tree:

jina.proto:
       -jina_pb2_grpc.py 
       -jina_pb2.py
       -pb2:
                 -jina_pb2_grpc.py 
                 -jina_pb2.py

the pb2 folder will disappear at some point anyway when all the upstream dependency will move to proto 4

Issue #5085, opened 2022-08-24
feat: Grafana executor latency breakdown by endpoints


florian-hoenicke:
As you can see in this graphic, it looks like our executors got faster over time, but it is not the case.
<img width="641" alt="image" src="https://user-images.githubusercontent.com/11627845/186400078-b31478c8-ecf1-4a68-be46-8776d04abc46.png">
The reason why the processing time is increasing is that we first made indexing requests and then search requests. Indexing takes longer for the preprocessing executor because we need to process the videos. Searching is fast, because, we just forward the search text and no huge preprocessing is done.
So, we loose basically all information in this graphic. We don't know how long indexing take and don't know how long search take.

User acceptance criteria:
- [ ] each endpoint which got called will cause a new graphic in the dashboard

samsja:
Thanks @florian-hoenicke for reporting the issue.

We actually already cover this feature. You should use the `jina_receiving_request_seconds` which will have as a labels the `exec_endpoint`. I believe right now your dashboard is just using the wrong metrics `jina_process_request_seconds` More info in the [documentation](https://docs.jina.ai/fundamentals/flow/monitoring-flow/#executor-pods)

Morevover, the jina responsibility about metrics is to expose them. We recommend to use Grafana to visualize it but we do not maintain any official dashboard. Therefore it will be better for us and for future people opening the issue if next you could report the issue at the metrics levels and not at the dashboard levels, i.e. by saying "metrics `jina_abcd` does not support this".

samsja:
Thanks @florian-hoenicke for reporting the issue.

We actually already cover this feature. You should use the `jina_receiving_request_seconds` which will have as a labels the `exec_endpoint`. I believe right now your dashboard is just using the wrong metrics `jina_process_request_seconds` More info in the [documentation](https://docs.jina.ai/fundamentals/flow/monitoring-flow/#executor-pods)

Morevover, the jina responsibility about metrics is to expose them. We recommend to use Grafana to visualize it but we do not maintain any official dashboard. Therefore it will be better for us and for future people opening the issue if next you could report the issue at the metrics levels and not at the dashboard levels, i.e. by saying "metrics `jina_abcd` does not support this".

PR #5087, opened 2022-08-24
refactor: improve request handler monitoring code

samsja:
# Context

Atm default Executor metrics aggregate their value for all of the routes of the Executor making it impossible to monitor only one routes. This could be a problem when one want to detect bottleneck for instance. This behavior was raised in issue: https://github.com/jina-ai/jina/issues/5085 

# What this PR do ?:

- [x] refactor the code the default metrics in the `RequestHandler`. It notably fix a big when failed request time are measured whereas they should not
- [ ] Make it so that all of the metrics work on the routes level and not on the Executor level


samsja:
Close as the upstream issue is resolve

PR #5088, opened 2022-08-24
refactor: refactor data request handler monitoring

samsja:
# Context

The behavior of the `receiving_request_seconds` is not perfect atm since it record the time for failed request as well.

# What this pr do ?

- [x] fix the behavior of the metric `receiving_request_seconds`
- [x] refactor the RequestHandler

### DIFF
--- a/jina/serve/runtimes/gateway/request_handling.py
+++ b/jina/serve/runtimes/gateway/request_handling.py
@@ -48,7 +48,7 @@ class RequestHandler:
 
             self._receiving_request_metrics = Summary(
                 'receiving_request_seconds',
-                'Time spent processing request',
+                'Time spent processing successful request',
                 registry=metrics_registry,
                 namespace='jina',
                 labelnames=('runtime_name',),
--- a/jina/serve/runtimes/gateway/request_handling.py
+++ b/jina/serve/runtimes/gateway/request_handling.py
@@ -132,7 +132,7 @@ class RequestHandler(MonitoringRequestMixin):
         metrics_registry: Optional['CollectorRegistry'] = None,
         runtime_name: Optional[str] = None,
     ):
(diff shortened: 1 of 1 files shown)


jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -90,25 +89,53 @@ def _update_start_request_metrics(self, request: 'Request'):
...
+    :param metrics_registry: optional metrics registry for prometheus used if we need to expose metrics from the executor or from the data request handler
+    :param runtime_name: optional runtime_name that will be registered during monitoring
+    """
+
+    def __init__(
+        self,
+        metrics_registry: Optional['CollectorRegistry'] = None,
+        runtime_name: Optional[str] = None,
+    ):
+        super(RequestHandler, self).__init__(metrics_registry, runtime_name)
```
JohannesMessner, line 135: can't we just do `super().__init__(..)`?

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -22,9 +22,9 @@
     from jina.types.request import Request
 
 
-class RequestHandler:
+class MonitoringRequestMixin:
```
JohannesMessner, line 25: Should we have this in a separate file, together with other monitoring related stuff? To have it all in one place and not clutter the main business logic anywhere?

alaeddine-13 (dismissed):
LGTM

jina/serve/runtimes/gateway/request_handling.py
samsja: I don't think that it should go into other monitoring things are they are speadout everywhere in the codebase. Putting it out of the file is a good idea though

Issue #4958, opened 2022-06-24
Investigate packet loss on Dalle Flow


samsja:
Network package loss is now easy to reproduce. 

just open client notebook, send some request, and if it gets stuck there for 1 hour you reproduce it.

PR #5092, opened 2022-08-25
feat: expose grpc parameters and add production ready keepalive parameters

samsja:
# Context:

We have been noticing some case where our client was hanging forever, waiting for a requests to return which never happened. This only happen when a [deployed application](https://github.com/jina-ai/jina/issues/4958) was under high pressure ( very high QPS or DDOS  ) and started throttling.

Turns out that when one of the pod went OOM some of the connection would stop abruptly without the pod beeing able to close the gRPC connection properly like it would do when sending a SIGKILL. However this failed connection will not trigger anything on Client side which will just not be aware that it is a broken connection since nothing will be send.

gRPC has actually a mechanism to detect such zombie connections. However the default frequency of this check is 2 hours which in practice in too long and the client will hang for two hours thiking that his requests is just being processed but slowly.

More info in this [blog post](https://cs.mcgill.ca/~mxia3/2019/02/23/Using-gRPC-in-Production/) 

related issue : https://github.com/jina-ai/jina/issues/4958

# What this PR do ?

- [x] implement production ready default value for the `keepalive` feature of gRPC
- [x] expose grpc parameters
- [x] test the feature


jina/serve/runtimes/gateway/grpc/__init__.py
```diff
@@ -48,6 +48,19 @@ async def async_setup(self):
             options=[
                 ('grpc.max_send_message_length', -1),
                 ('grpc.max_receive_message_length', -1),
+                # for the following see this blog post for the choice of default value https://cs.mcgill.ca/~mxia3/2019/02/23/Using-gRPC-in-Production/
+                ('grpc.keepalive_time_ms', 10000),
+                # send keepalive ping every 10 second, default is 2 hours.
+                ('grpc.keepalive_timeout_ms', 5000),
```
JoanFM, line 54: may it be too short? This could be a problem with dalle flow perhaps with so slow processing time? should we maybe expose it as parameter?

samsja:
@JoanFM Let me clarify. 

The `keepalive` things will just check if the connection is healthy. With my current settings even though it is written 10 seconds frequency it does not mean after 10 seconds the connection will be drop. It means that after 10 seconds the client will send a small ping to the server to check than everything is okay. But the server can take as much time as he want to answers as long as it communicates to the client that he is still processing the requests. 

I will emphasize this behavior in the tests to be sure that it won't be a problem for dalle flow or other slow processing Executor

jina/serve/runtimes/gateway/grpc/__init__.py
samsja: I exposed it as parameters now as well

jina/helper.py
```diff
@@ -1684,3 +1684,34 @@ def _telemetry():
...
+    # allow grpc pings from client every 10 seconds
+    ('grpc.http2.min_ping_interval_without_data_ms', 5000),
+    # allow grpc pings from client without data every 5 seconds
+]
+
+
+def _get_grpc_server_options(option_from_args: Dict) -> List[Tuple[str, Any]]:
+
+    if option_from_args:
+        options_from_dict = []
```
JohannesMessner, line 1711: Shouldn't this also be filled with all options in `_DEFAULT_GRPC_OPTION` that are not given in `option_from_args`? Right now it looks like as soon as the user specifies one option in `option_from_args`, all of our defaults will be ignored.

samsja: Yes I had the same thinking.

The things is I don't like than the stuff happen under the hoods from the users. Only advanced users will use it anyway so I will assume that they know what they are doing and prefer this behavior.

What do you think @alaeddine-13 ?

alaeddine-13: Actually I think grouping grpc parameters into the same grpc options argument does not go well with the pattern we follow:
we do not group parameters for docker or executor,...
But in case we opt for grouping parameters in the same argument, I'd go with @JohannesMessner , I find it more intuitive

samsja: Yeah but I can't expose all of the possible grpc parameters in the Flow right ? What would you suggest ?
The things is that we don't usually let users do override our parameters

JohannesMessner: I also wouldn't expose all possible grpc arguments independently. I see this similarly to how you just pass an entire config to the storage backend in docarray, and what you don't precise gets filled in with the defaults. I would do the same here.

JohannesMessner (changes requested):
The blog you link claims that these same options should also be set on the client side. I think in our case that would include our actual client as well as the client side on the gateway <-> executor communication.
I might be missing it, but I don't see those args being passed there, is there a reason for that?

samsja:
> The blog you link claims that these same options should also be set on the client side. I think in our case that would include our actual client as well as the client side on the gateway <-> executor communication. I might be missing it, but I don't see those args being passed there, is there a reason for that?

I check myself and it does not matter if it done on client side or not.

If I have the options on server side and not on the client side it is working
If I have the options only on client side it is not working
(well If I have both it is working as well)

So I assume we don't need it on client side. Moreoever external client would not know that they need to pass this value so it is better that the options work as long as it is set on server side.

But I might be wrong, would you say we put it for safety ?

JohannesMessner:
> > The blog you link claims that these same options should also be set on the client side. I think in our case that would include our actual client as well as the client side on the gateway <-> executor communication. I might be missing it, but I don't see those args being passed there, is there a reason for that?
> 
> I check myself and it does not matter if it done on client side or not.
> 
> If I have the options on server side and not on the client side it is working If I have the options only on client side it is not working (well If I have both it is working as well)
> 
> So I assume we don't need it on client side. Moreoever external client would not know that they need to pass this value so it is better that the options work as long as it is set on server side.
> 
> But I might be wrong, would you say we put it for safety ?

If it works let's keep it like this. Your point about 3rd party clients is important.

jina/helper.py
```diff
@@ -1684,3 +1684,41 @@ def _telemetry():
...
+
+    option_from_args = (
+        {**_DEFAULT_GRPC_OPTION, **option_from_args}
+        if option_from_args
+        else _DEFAULT_GRPC_OPTION
+    )  # merge new and default args
+
+    options_from_dict = []
+    for key, value in option_from_args.items():
+        options_from_dict.append((key, value))
```
JohannesMessner, line 1722: can't we just do `return list(option_from_args.items())`?

samsja: done

JohannesMessner (dismissed):

tests/integration/gateway_clients/test_long_flow_keepalive.py
```diff
@@ -0,0 +1,29 @@
...
+
+from jina import Executor, Flow, requests
+
+
+@pytest.fixture()
+def slow_executor() -> Executor:
+    class MySlowExec(Executor):
+        @requests
+        def slow(self, docs, **kwargs):
+            time.sleep(300)
```
alaeddine-13, line 14: can we reduce this ? just to ease our CI

docs/fundamentals/flow/executor-args.md
```diff
@@ -16,6 +16,7 @@
 | `host_in` | The host address for binding to, by default it is 0.0.0.0 | `string` | `0.0.0.0` |
 | `native` | If set, only native Executors is allowed, and the Executor is always run inside WorkerRuntime. | `boolean` | `False` |
 | `output_array_type` | The type of array `tensor` and `embedding` will be serialized to.<br><br>Supports the same types as `docarray.to_protobuf(.., ndarray_type=...)`, which can be found <br>`here <https://docarray.jina.ai/fundamentals/document/serialization/#from-to-protobuf>`.<br>Defaults to retaining whatever type is returned by the Executor. | `string` | `None` |
+| `grpc_server_options` | Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server | `object` | `None` |
```
alaeddine-13, line 19: can we give an example ? it helps the user understand the syntax when writing options

docs/fundamentals/flow/gateway-args.md
```diff
@@ -16,6 +16,7 @@
 | `host_in` | The host address for binding to, by default it is 0.0.0.0 | `string` | `0.0.0.0` |
 | `native` | If set, only native Executors is allowed, and the Executor is always run inside WorkerRuntime. | `boolean` | `False` |
 | `output_array_type` | The type of array `tensor` and `embedding` will be serialized to.<br><br>Supports the same types as `docarray.to_protobuf(.., ndarray_type=...)`, which can be found <br>`here <https://docarray.jina.ai/fundamentals/document/serialization/#from-to-protobuf>`.<br>Defaults to retaining whatever type is returned by the Executor. | `string` | `None` |
+| `grpc_server_options` | Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server | `object` | `None` |
```
alaeddine-13, line 19: same as above

tests/integration/gateway_clients/test_long_flow_keepalive.py
samsja: good point, done

docs/fundamentals/flow/executor-args.md
```diff
@@ -16,6 +16,7 @@
 | `host_in` | The host address for binding to, by default it is 0.0.0.0 | `string` | `0.0.0.0` |
 | `native` | If set, only native Executors is allowed, and the Executor is always run inside WorkerRuntime. | `boolean` | `False` |
 | `output_array_type` | The type of array `tensor` and `embedding` will be serialized to.<br><br>Supports the same types as `docarray.to_protobuf(.., ndarray_type=...)`, which can be found <br>`here <https://docarray.jina.ai/fundamentals/document/serialization/#from-to-protobuf>`.<br>Defaults to retaining whatever type is returned by the Executor. | `string` | `None` |
+| `grpc_server_options` | Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {grpc.max_send_message_length': -1} | `object` | `None` |
```
alaeddine-13, line 19:
```suggestion
| `grpc_server_options` | Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1} | `object` | `None` |
```

docs/fundamentals/flow/gateway-args.md
```diff
@@ -16,6 +16,7 @@
 | `host_in` | The host address for binding to, by default it is 0.0.0.0 | `string` | `0.0.0.0` |
 | `native` | If set, only native Executors is allowed, and the Executor is always run inside WorkerRuntime. | `boolean` | `False` |
 | `output_array_type` | The type of array `tensor` and `embedding` will be serialized to.<br><br>Supports the same types as `docarray.to_protobuf(.., ndarray_type=...)`, which can be found <br>`here <https://docarray.jina.ai/fundamentals/document/serialization/#from-to-protobuf>`.<br>Defaults to retaining whatever type is returned by the Executor. | `string` | `None` |
+| `grpc_server_options` | Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {grpc.max_send_message_length': -1} | `object` | `None` |
```
alaeddine-13, line 19:
```suggestion
| `grpc_server_options` | Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1} | `object` | `None` |
```

docs/fundamentals/flow/executor-args.md
alaeddine-13: actually users must be aware they have to pass valid json, so they need to pass double quotes like this:
`--grpc-server-options "{\"grpc.max_send_message_length\": -1}"` but not sure where this should go

jina/orchestrate/flow/base.py
```diff
@@ -198,7 +198,7 @@ def __init__(
         :param floating: If set, the current Pod/Deployment can not be further chained, and the next `.add()` will chain after the last Pod/Deployment not this current one.
         :param graph_conditions: Dictionary stating which filtering conditions each Executor in the graph requires to receive Documents.
         :param graph_description: Routing graph for the gateway
-        :param grpc_server_kwargs: Dictionary of kwargs arguments that will be passed to the grpc server when starting the server # todo update
+        :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {grpc.max_send_message_length': -1}
```
alaeddine-13, line 201:
```suggestion
        :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}
```

jina/orchestrate/flow/base.py
```diff
@@ -372,7 +372,7 @@ def __init__(
...
```
alaeddine-13, line 375:
```suggestion
        :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}
```

jina/orchestrate/flow/base.py
```diff
@@ -849,6 +850,7 @@ def add(
               - To access specified gpus based on device id, use `--gpus device=[YOUR-GPU-DEVICE-ID]`
               - To access specified gpus based on multiple device id, use `--gpus device=[YOUR-GPU-DEVICE-ID1],device=[YOUR-GPU-DEVICE-ID2]`
               - To specify more parameters, use `--gpus device=[YOUR-GPU-DEVICE-ID],runtime=nvidia,capabilities=display
+        :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {grpc.max_send_message_length': -1}
```
alaeddine-13, line 853:
```suggestion
        :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}
```

jina/orchestrate/flow/base.py
```diff
@@ -995,6 +997,7 @@ def add(
...
```
alaeddine-13, line 1000:
```suggestion
        :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}
```

docs/fundamentals/flow/executor-args.md
samsja: Yes but this part will be copy paste to the python docstring. So we can't talk about the json stuff here

alaeddine-13 (approved):

jina/parsers/orchestrate/runtimes/worker.py
```diff
@@ -103,3 +103,12 @@ def mixin_worker_runtime_parser(parser):
 Defaults to retaining whatever type is returned by the Executor.
 ''',
     )
+
+    gp.add_argument(
+        '--grpc-server-options',
+        action=KVAppendAction,
+        metavar='KEY: VALUE',
+        nargs='*',
+        help="Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}",
```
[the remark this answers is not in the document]
samsja: to which documentation ? the grpc one ?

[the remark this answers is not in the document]
JoanFM: yes

PR #5098, opened 2022-08-29
fix: update to protobuf 4.21 new types

JohannesMessner:
CI is currently failing, it looks like a type changed in the protbuf library, so a type check we did fails.
This fixes the issue by just seeing if the needed attribute is there, regardless of the type, so this issue should not happen anymore in the future.


alaeddine-13 (approved):
I think we can merge this if CI is passing with pre-released docarray

alaeddine-13 (approved):

Issue #4940, opened 2022-06-20
Monitoring: add send/receive bytes as a metrics


samsja:
It could be usefull to track the number of bytes received and send betweem the client and the Gateway during a gRPC connection.

Especially since we need this metric here https://github.com/jina-ai/jina/blob/9dac3f693b6b7d89b9b02095b68e7e07cb358087/jina/proto/serializer.py#L26

samsja:
Putting this as priority since @florian-hoenicke might need such a metrics for defining alerting in jina-now

PR #5111, opened 2022-08-31
feat(monitoring): add monitoring of requests size in bytes at all level

samsja:
# Context

one requested metrics is to measure the size in bytes of the requests

issue : https://github.com/jina-ai/jina/issues/4940

```
┌─────────┐
│ Client  │
└─┬───────┘
  │    ▲
  │    │
  │    │
 A▼    │ F
┌──────┴──┐
│ Gateway │
└─┬───────┘
 B│     ▲  E
  │     │
  │     │
 C▼     │  D
┌───────┴──┐
│ Executor │
└──────────┘
```

# What this pr do
Monitor everywhere the the size of the requests everywhere

Gateway 
- [x] size of the received request (A) `jina_received_request_bytes`
- [x] size of the send request to a pod (B) `jina_sent_request_bytes`
- [x] size of the return request from an Executor/Head (E) `jina_received_response_bytes`
- [x] size of the return request to the client (F) `jina_sent_response_bytes`

Head :
- [x] size of the send request to a pod (B) `jina_sent_request_bytes`
- [x] size of the return request from an Executor (E)  `jina_received_response_bytes`

Executor:
- [x] size of the received request (C) `jina_received_request_bytes`
- [x] size of the send request (D) `jina_sent_response_bytes`

- [x] add  runtime label in the metrics of the Connection pool

documentation : https://feat-monitoring-bytes--jina-docs.netlify.app/fundamentals/flow/monitoring-flow/#available-metrics

I am not a big fan of the current naming. Unfortunately `jina_request_size_bytes` is here from before and is poorly named (my bad) so changing it will be a breaking change.

EDIT: not a breaking change since we introduced metric deprecation


hanxiao (dismissed):

JoanFM:
I think these names are confusing:

```
jina_request_size_bytes

jina_send_request_bytes
```

Why one has `size` and not the `other`?

JoanFM:
Accoridng to diagram, in `D` I would call `respond` and not `send`

docs/fundamentals/flow/monitoring-flow.md
```diff
@@ -156,40 +156,50 @@ Because not all Pods have the same role, they expose different kinds of metrics:
...
+| `jina_receiving_request_seconds`        | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)    | Measure the time elapsed between receiving a request from the gateway and sending back the response.          |
+| `jina_sending_request_seconds`          | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)    | Measure the time elapsed between sending a downstream request to an Executor and receiving the response back. |
+| `jina_sending_request_bytes`            | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)    | Measure the size of the downstream requests send to an Executor in Bytes                                      |
+| `jina_failed_requests_total`            | [Counter](https://prometheus.io/docs/concepts/metric_types/#counter)    | Counts the number of failed requests returned by the gateway                                                  |
+| `jina_send_request_bytes`               | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)    | Measures the size in Bytes of the request send by the Head to the Pod                                         |
```
JoanFM, line 183: `sent` is the right word and not `send`.

`jina_send_request_bytes` is from the `Head` to the `Executor`?

docs/fundamentals/flow/monitoring-flow.md
```diff
@@ -156,40 +156,50 @@ Because not all Pods have the same role, they expose different kinds of metrics:
...
+| `jina_return_request_bytes`             | [Summary](https://prometheus.io/docs/concepts/metric_types/#summary)    | Measures the size in Bytes of the request returned by the Pod                                                 |
```
JoanFM, line 184: returned by the Pod or the Executor?

JoanFM (changes requested):

JoanFM:
In the `diagram` you write in description u do not write about the `Head`. It can be interesting to see how u approach it there in terms of naming

samsja:
> Accoridng to diagram, in `D` I would call `respond` and not `send`

I see something like this `jina_respond_request_bytes` instead of `jina_send_request_bytes`

jina/serve/networking.py
```diff
@@ -326,18 +353,40 @@ async def send_requests(
                         compression=compression,
                         timeout=timeout,
                     )
-                    with self._summary_time:
+                    if self._metrics.send_requests_bytes_metrics:
+                        self._metrics.send_requests_bytes_metrics.observe(
+                            requests[0].nbytes
```
alaeddine-13, line 358: why just `requests[0]`

jina/serve/networking.py
```diff
@@ -326,18 +353,40 @@ async def send_requests(
...
+                        )
+
+                    with self._metrics.sending_requests_time_metrics.time() if self._metrics.sending_requests_time_metrics else contextlib.nullcontext():
```
alaeddine-13, line 361: add helper function `get_context_or_null` and use it

jina/serve/networking.py
```diff
@@ -360,14 +409,19 @@ async def send_requests(
                 raise ValueError(f'Unsupported request type {type(requests[0])}')
 
     class _ConnectionPoolMap:
-        def __init__(self, logger: Optional[JinaLogger], summary):
+        def __init__(
+            self,
+            logger: Optional[JinaLogger],
+            metrics: _NetworkingMetrics,
+            runtime_name: str,
```
alaeddine-13, line 416: let's try to have similar signature and follow the order: runtime_name, logger, metrics

jina/serve/networking.py
```diff
@@ -138,7 +156,10 @@ def _create_connection(self, address):
         use_tls = parsed_address.scheme in TLS_PROTOCOL_SCHEMES
 
         stubs, channel = GrpcConnectionPool.create_async_channel_stub(
-            address, tls=use_tls, summary=self.summary
+            address,
+            metrics=self._metrics,
+            runtime_name=self.runtime_name,
+            tls=use_tls,
```
alaeddine-13, line 162:
```suggestion
            address, self._metrics, self.runtime_name, tls=use_tls
```
unless they are not really required parameters and in this case let's make them keyword arguments in the signature

jina/serve/networking.py
```diff
@@ -252,10 +273,16 @@ class ConnectionStubs:
             'jina.JinaInfoRPC': jina_pb2_grpc.JinaInfoRPCStub,
         }
 
-        def __init__(self, address, channel, summary):
+        def __init__(
+            self,
+            address,
+            channel,
+            runtime_name: str,
```
alaeddine-13, line 280: looks like this was not used ?

alaeddine-13 (commented):
it may help future readers to add comments above each defined metric with some description

alaeddine-13:
my suggestion regarding naming, is to introduce a breaking change in `jina_request_size_bytes` and have more clear naming:
use `received`/`sent`, `request`/`response`, just bytes instead of size_bytes + we add runtime name as follows:
┌─────────┐
│ Client  │
└─┬───────┘
  │    ▲
  │    │
  │    │
 A▼    │ F
┌──────┴──┐
│ Gateway │
└─┬───────┘
 B│     ▲  E
  │     │
  │     │
 C▼     │  D
┌───────┴──┐
│ Executor │
└──────────┘

(A) jina_request_size_bytes -> gateway_received_request_bytes
(B) jina_send_request_bytes -> gateway_sent_request_bytes
(E) jina_return_request_bytes -> gateway_received_response_bytes
(F) jina_return_client_request_bytes -> gateway_sent_response_bytes

Executor:

(C) jina_request_size_bytes -> worker_received_request_bytes

(D) jina_send_request_bytes -> worker_sent_response_bytes
PS: the head metrics should be named same as head imo (just change runtime name)

samsja:
> 

gr

samsja:
> my suggestion regarding naming, is to introduce a breaking change in `jina_request_size_bytes` and have more clear naming: use `received`/`sent`, `request`/`response`, just bytes instead of size_bytes + we add runtime name as follows:
> 
> (A) jina_request_size_bytes -> gateway_received_request_bytes (B) jina_send_request_bytes -> gateway_sent_request_bytes (E) jina_return_request_bytes -> gateway_received_response_bytes (F) jina_return_client_request_bytes -> gateway_sent_response_bytes
> 
> Executor:
> 
> (C) jina_request_size_bytes -> worker_received_request_bytes
> 
> (D) jina_send_request_bytes -> worker_sent_response_bytes PS: the head metrics should be named same as head imo (just change runtime name)

metric should not include where they come from (worker/gateway) this is the job of label. Name should be unified

Mevertheless I like the `received`/`sent`, `request`/`response` I will rename to this

docs/fundamentals/flow/monitoring-flow.md
samsja: I change the documentation so that it is clear

jina/serve/networking.py
samsja: here len(requests) == 1 as you can see in the above line

jina/serve/networking.py
samsja: but having them as named argument is just more powerful than having them as keyword right ? And it does behave like keyword arugments as well. I am not sure to get why it would be better

jina/serve/networking.py
samsja: good catch

jina/serve/networking.py
samsja: good idea

alaeddine-13:
> > it may help future readers to add comments above each defined metric with some description
> 
> @alaeddine-13 you mean in the documentation ? How would you see it ?

I meant in the source code

tests/integration/monitoring/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    for line in raw_metrics.split('\n'):
+        if not line.startswith('#') and ' ' in line:
+            line_split = line.split(' ')
+
+            metric_name = ''.join(line_split[0:-1])
+
+            metric_value = float(line_split[-1])
+
+            metrics[metric_name] = metric_value
+
```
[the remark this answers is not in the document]
samsja: I personally does not agree that regular expression is more readable and robust that doing small python code as long as the feature is simple (okay maybe it is bias bc I don't like regex bc it is a brainf*ck). 

I would say not to over engineer here

tests/integration/monitoring/test_request_size.py
```diff
@@ -0,0 +1,173 @@
...
+    assert (
+        size_return_from_exec_at_gateway == size_send_by_executor
+    )  # both should have the same size since it is just the same request send from executor to gateway
+
+    assert (
+        _compare_relative_difference(
+            size_return_to_the_client, size_return_from_exec_at_gateway
+        )
+        < 0.1
+    )
```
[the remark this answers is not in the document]
alaeddine-13: yes also the choice of the coefficients may be unjustified. My concern is also about whether one day we will end up maintaining these tests and they become flaky.
my suggestion is:
size_received_at_executor < 10 * size_return_from_exec_at_gateway => let's just check if smaller
and I would remove _compare_relative_difference and just assert both values are >0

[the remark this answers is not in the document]
samsja: So basically I remove the two assert with the _compare_relative_difference. I agree with this 
> My concern is also about whether one day we will end up maintaining these tests and they become flaky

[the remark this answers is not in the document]
samsja: okay done

tests/integration/complex_tags/test_complex_tags.py
```diff
@@ -41,12 +41,12 @@ def validate(resp):
...
 
-    mock = mocker.Mock()
+    # mock = mocker.Mock()
     f = Flow().add(uses=executor_class)
     with f:
-        f.index(inputs=docs, on_done=mock)
+        f.index(inputs=docs)  # , on_done=mock)
 
-    validate_callback(mock, validate)
+    # validate_callback(mock, validate)
```
alaeddine-13, line 49: why commented ?

samsja: hmmm I might have missed that

tests/integration/monitoring/test_deprecated_metric.py
```diff
@@ -0,0 +1,93 @@
+import time
+
+import requests as req
+from prometheus_client import CollectorRegistry, start_http_server
+
+from jina.serve.monitoring import _SummaryDeprecated
+
+from . import get_metric_values
```
JoanFM, line 8: do not use this type of import, use absolute

JoanFM (commented):
Check alaeddine comment

tests/integration/monitoring/test_deprecated_metric.py
samsja: I am not a big fan on using absolute import in test

alaeddine-13 (approved):

JoanFM (approved):

Issue #5113, opened 2022-09-01
Fix test async flow http test


samsja:
# Context

`tests/unit/orchestrate/flow/flow-async/test_asyncflow.py::test_run_async_flow_other_task_concurrent` is failing on master and this is not flakyness. https://github.com/jina-ai/jina/blob/6b24bf698b907e2104ad4b4d7dc093bfab0b6518/tests/unit/orchestrate/flow/flow-async/test_asyncflow.py#L136
```
test_run_async_flow_other_task_concurrent[http] failed; it passed 0 out of the required 1 times.
	<class 'AssertionError'>
	assert 10.227989813000022 < 10
```

PR #5115, opened 2022-09-01
test: fix test  concurrent async flow

alaeddine-13:
This PR fixes a test that asserts async flow and an async task run in parallel and optimize timing
However, the test also measures the flow construction and deconstruction which might exceed the 10s threshhold.
This PR excludes the construction and deconstruction from the measurements and lowers the threshhold.

closes: https://github.com/jina-ai/jina/issues/5113


JoanFM (approved):

samsja (approved):
Yeah it makes way more sense now

PR #5126, opened 2022-09-05
feat(serve): add argument to specify exceptions that will exit the runtime

### LINKED ISSUE #5110: Add --exit-exceptions parameter to Executor
**Describe the feature**
Add --exit-exceptions parameter to the Executor, such that when Executor raises this Exception from a `bound` method, the Runtime sends `exit(1)` signal so that K8s reschedules the Pod.

Usage example:

```jina executor ... --exit-exceptions CUDA_MEMORY_ERROR,OUT_OF_MEMORY_ERROR```

### PR DESCRIPTION
**Pull request by girishc13**
Adds a new `--exit-on-exceptions` argument for the executor arguments list. The command accepts one or more exception class names and performing a system exit to terminate the process. This will enable cloud deployments to kill the pod and allocate a new pod which may lead to a recovery of the executor.

**Goals:**

- resolves #5110 
- the system exit command must be run on the executor process to correctly kill the executor pod.
- the argument supports a list of exception class names to handle different types of scenarios.

### CODE REVIEW

File: jina/parsers/orchestrate/base.py [hunk 1]
```diff
@@ -119,3 +119,11 @@ def mixin_base_ppr_parser(parser):
     
     ''',
     )
+
+    gp.add_argument(
+        '--exit-on-exceptions',
+        type=str,
+        default=[],
+        nargs='*',
+        help='List of exceptions that will cause the executor to shutdown.',
```
JohannesMessner, line 128:
```suggestion
        help='List of exceptions that will cause the Executor to shut down.',
```

girishc13:
Follow up question: Should I call the `self.async_teardown()` before calling the `sys.exit(1)` for better resource handling?

JoanFM:
> Follow up question: Should I call the `self.async_teardown()` before calling the `sys.exit(1)` for better resource handling?

Or u can raise the Exception that will make sure that the `async_teardown` happens

girishc13:
We can raise a `RuntimeInterrupted` exception to trigger the `__exit__` method which will also do the execute the `teardown` method.

tests/unit/serve/runtimes/worker/test_worker_runtime.py
```diff
@@ -426,3 +429,54 @@ def start_runtime(args, cancel_event):
...
+        ) as channel:
+            stub = jina_pb2_grpc.JinaSingleDataRequestRPCStub(channel)
+            response, call = stub.process_single_data.with_call(
+                _create_test_data_message()
+            )
+
+        assert response.header.status.code == jina_pb2.StatusProto.ERROR
+        assert 'is-error' in dict(call.trailing_metadata())
+    except:
+        pass
```
alaeddine-13, line 477: 1. can't we use the grpc client used by the gateway runtime ?
2. I believe asserts shouldn't be wrapped in a try-catch

tests/unit/serve/runtimes/worker/test_worker_runtime.py
```diff
@@ -231,7 +232,7 @@ def start_runtime(args, cancel_event):
             try:
                 s.connect((args.host, args.port))
                 connected = True
-            except ConnectionRefusedError:
+            except:
```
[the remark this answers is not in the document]
girishc13: On my Mac the error thrown is different than the single expected error. This was a blocker for running the tests in the `worker` module so I removed it because the exception itself was not used for asserts.

tests/unit/serve/runtimes/worker/test_worker_runtime.py
girishc13: 1. I was using the first example as my sample to create the new test. I checked the other tests and replaced the call with the available helper method.
2. I removed the try-catch since the Runtime gracefully handles the exception and returns a generic error request. I wasn't able to validate before because of I couldn't run the test locally. Anyway now I can run the test partially and verify which works the best.

tests/k8s/test_k8s.py
```diff
@@ -696,3 +697,65 @@ async def _create_external_deployment(api_client, app_client, docker_images, tmp
...
+    await asyncio.sleep(0.5)
+
+    pods = core_client.list_namespaced_pod(namespace=namespace).items
+    pod_restarts = [item.status.container_statuses[0].restart_count for item in pods]
+    assert any([count for count in pod_restarts if count > 0])
+
+    await asyncio.sleep(2)
+    pods = core_client.list_namespaced_pod(namespace=namespace).items
+    pod_phases = [item.status.phase for item in pods]
+    assert len(pod_phases) == 2
```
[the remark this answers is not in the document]
girishc13: I think this was more of a debug statement to check number of running pods. The tests doesn't check the number of pods after deployment so it's really helpful to check again after the test request. I think its sufficient to check for `Running` pods.

girishc13:
reworked in #5165

PR #5127, opened 2022-09-05
test: fix k8s tests return responses params

JoanFM:
Fixes CI problems with K8s. When the `return_responses` param was deprecated these tests were not properly adapted

### DIFF
--- a/tests/k8s/test_graceful_request_handling.py
+++ b/tests/k8s/test_graceful_request_handling.py
@@ -1,4 +1,4 @@
-import asyncio
+/mport asyncio
 import multiprocessing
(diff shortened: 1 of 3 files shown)


tests/k8s/test_graceful_request_handling.py
```diff
@@ -1,4 +1,4 @@
-import asyncio
+/mport asyncio
```
alaeddine-13, line 1:
```suggestion
import asyncio
```

alaeddine-13 (approved):

JohannesMessner (approved):

PR #5134, opened 2022-09-06
feat: pass internal flag to telemetry

JoanFM:
- Pass internal flag to determine if CI is from jina-ai org

### DIFF
--- a/tests/unit/orchestrate/flow/flow-orchestrate/test_flow_complex_topology.py
+++ b/tests/unit/orchestrate/flow/flow-orchestrate/test_flow_complex_topology.py
@@ -1,4 +1,5 @@
 import threading
+import multiprocessing
 import time
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/unit/orchestrate/flow/flow-orchestrate/test_flow_complex_topology.py
```diff
@@ -36,15 +37,14 @@ def test_flow_external_executor_with_gateway():
...
 
-    e = threading.Event()
-    t = threading.Thread(
+    e = multiprocessing.Event()
+    t = multiprocessing.Process(
         name='serve-exec',
         target=serve_exec,
-        kwargs={'port_expose': external_gateway_port, 'stop_event': e},
-        daemon=True,
+        kwargs={'port': external_gateway_port, 'stop_event': e},
```
alaeddine-13, line 44: daemon=True proved to prevent CI hanging when we do multiprocessing.
are we sure we want to remove it ?
@samsja fyi

JoanFM: if we do not remove, we cannot start Processes in it. And in multithreads, no signal handling is provided.

JohannesMessner:
Do we need to update the docs?

JoanFM:
> Do we need to update the docs?

In the docs, I point to the exact implementation so that people can look at the details

JohannesMessner (approved):

PR #5138, opened 2022-09-06
feat(telemetry): send telemetry data about exceptions

JoanFM:
Add some information about Exceptions raised by Executors and at the Client side


JoanFM:
https://github.com/jina-ai/internal-tasks/issues/421

PR #5141, opened 2022-09-07
fix: invalid input raise exception

JoanFM:
When the input iterator raises an Error, right now no Exception is provided by `client.post` method


delgermurun (approved):
LGTM!

Issue #5032, opened 2022-07-27
CustomGateway: Allow usage of Flow connectivity from a custom Gateway


JoanFM:
# Custom Gateway 

We want to allow users to define their own Gateway. 

## Context

The point is that atm many users put a fastAI in front of our gateway to translate the docarray syntax to their own specific needs.
This is an issue since for performance. 

## What could be done

Our gateway is basically two things: an API (grpc or http server) and the core logic to handle the Flow.We could have more flexibility and let the user create their own API to fit to their requirement and logic. 

- [x] refactor code so that the `TopolofyGrapg` the `ConnectionPool` and the `Streamer` live into the same object https://github.com/jina-ai/jina/pull/5031

Left to do:
- [x] Refactor and separate gateway logic from runtime https://github.com/jina-ai/jina/pull/5153
- [x] Rename any BFF to CustomGateway
- [x] Rename `GatewayBFF` into a new name and clarify arguments and API so that it can be useful by user.
- [x] Allow `ContainerPod` to work with a `Gateway`
- [x] investigate whether it is possible to start gateway with local object/module
- [x] Allow users to pass `uses` for the `gateway`. Here we have 2 options: Allow only to pass  a docker image or maybe we can also allow to pass something local (In this case we could assume is a `FastAPI` app or something (it is the most common usage observed so far). https://github.com/jina-ai/jina/pull/5189
- [x] Have a syntax on the Flow side,  maybe `Flow(uses=CustomGateway)`
- [x] Add expose gateway API in Flow API as a separate set of args:
`Flow().config_gateway()` (or add gateway)
- [x] Define the proper user journey to enable this feature both locally and with docker.
- [x] Define how this new docker container or object will receive the parameters that will allow the user to instantiate this `GatewayBFF` and use it.
- [x] Document this
- [ ] Blog post about the feature and what enables

alaeddine-13:
As Documentation was merged, I'm closing this issue

PR #5153, opened 2022-09-08
refactor: extract gateway app logic into custom gateway class

alaeddine-13:
Since we aim to implement Custom Gateway in https://github.com/jina-ai/jina/issues/5032, we start by refactoring all logic that creates gateway application (fast api app in http/websocket gateway runtime and grpc.aio.server in grpc gateway).
We try to extract this logic into the `BaseGateway` class and extend this class into 3 subclasses (grpc, http and websockets). These classes are to be exposed to the user in case he wants to customize his gateway.

When implementing this PR we have the following options:
1. let user define everything (app + server) and start/stop server without expecting him to return uvicorn server
* he doesn't expose a server object
* he must implement shutdown and graceful shutdown functionalities (basically everything in gateway runtime). This means we're basicly telling the user to implement gateway runtime.
* user can define any wsgi/asgi server (no limitation on the framework)
* User must implement extra app functionality to preserve compatibility with API (dryrun/health check,...)
2. let user define everything (app + server), start, stop server and he also exposes uvicorn server to us
* we can implement functionalities in gateway runtime for the user (user only cares about defining the app and the server)
* we depend on the server interface (probably only asgi uvicorn server => user can only implement asgi apps. [List of asgi frameworks supported by uvicorn](https://www.uvicorn.org/#asgi-frameworks)
* if there is no flexibility over the server to be exposed, why don't we make the user just implment the asgi app (we allow him to configure the uvicorn server)
* User must implement extra app functionality to preserve compatibility with API (dryrun/health check,...)
3. User only defines the app:
* we expect certain objects for the user per runtime (asgi app to be run on uvicorn server for http/websocket, grpc server for grpc)
* We handle running the server
* User only implements asgi apps
* We relatively can implement extra app functionalities (dryrun/grpc) but but only if we force a framework for the user in http and websockets

In this PR, I decide to adopt option 3 and implement health check/dryrun/reflection features by default only for grpc because we know the app object returned by the user (grpc ai server)

**PS: not implementing health checks might break compatibility with kubernetes deployment**

**Goals:**

- extract gateway app logic into custom gateway class


jina/serve/gateway.py
```diff
@@ -0,0 +1,45 @@
...
+    gateway logic
+
+    :class:`jina.Gateway` as an alias for this class.
+    """
+
+    def __init__(
+        self,
+        streamer: Optional[GatewayStreamer] = None,
+        args: 'argparse.Namespace' = None,
+        logger: 'JinaLogger' = None,
```
JoanFM, line 24: same for logger if it JAML Compatible, it should only have args

jina/serve/runtimes/gateway/grpc/__init__.py
```diff
@@ -20,18 +19,6 @@
...
-    def __init__(
-        self,
-        args: argparse.Namespace,
-        **kwargs,
-    ):
-        """Initialize the runtime
-        :param args: args from CLI
-        :param kwargs: keyword args
-        """
-        self._health_servicer = health.HealthServicer(experimental_non_blocking=True)
```
JoanFM, line 32: why do u need to change this?

JoanFM:
We should only `expose` "start" and "stop" which they need to implement, and our Runtime ensures that they are called properly

jina/serve/gateway.py
alaeddine-13: it should be jaml compatible now

jina/serve/runtimes/gateway/grpc/__init__.py
alaeddine-13: no more changed

jina/serve/gateway.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        :param streamer: configured gateway streamer
+        :param args: args
+        :param logger: jina logger object
+        :param kwargs: additional extra keyword arguments to avoid failing when extra params ara passed that are not expected
+        """
+        self.streamer = streamer
+        self.args = args
+        self.logger = logger
+        self.app = self.get_app()
```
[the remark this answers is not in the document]
alaeddine-13: now BaseGateway requires server implementation (but not as simple as start and stop, rather, it requires setup, start, teardown and stop)

jina/serve/runtimes/gateway/http/gateway.py
```diff
@@ -0,0 +1,21 @@
+from jina.helper import extend_rest_interface
+
+from ....gateway import BaseGateway
+from . import get_fastapi_app
+
+
+class HTTPGateway(BaseGateway):
+    """HTTP Gateway implementation"""
+
+    def get_app(self):
```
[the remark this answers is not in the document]
alaeddine-13: it's no more get_app

jina/serve/gateway.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    The base class of all custom Gateways, can be used to build a custom interface to a Jina Flow that supports
+    gateway logic
+
+    :class:`jina.Gateway` as an alias for this class.
+    """
+
+    def __init__(
+        self,
+        streamer: Optional[GatewayStreamer] = None,
```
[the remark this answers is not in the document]
alaeddine-13: it is no exposed rn

jina/serve/gateway.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        :param name: Gateway pod name
+        :param kwargs: additional extra keyword arguments to avoid failing when extra params ara passed that are not expected
+        """
+        self.streamer = None
+        self.name = name
+        # TODO: original implementation also passes args, maybe move this to a setter/initializer func
+        self.logger = JinaLogger(self.name)
+
+    def _set_streamer(
```
JoanFM, line 37: why is `_set_streamer` not called in the `Init`?

JoanFM (commented):
I like it in general, I just want to make sure `_set_streamer` is called in `__init__` (RAII pattern). 

I am not fan of the `should_exit` property but let's keep it for now

JoanFM (approved):

PR #5157, opened 2022-09-09
feat: add duration info in events

JoanFM:
Goals:
Add information about duration of Flows and Runtimes in telemetry


jina/serve/runtimes/asyncio.py
```diff
@@ -65,7 +65,13 @@ def __init__(
...
 
         self._setup_monitoring()
-        send_telemetry_event(event='start', obj=self)
+        flow_id = self._flow_id
+        telemetry_kwargs = {}
+
+        if flow_id is not None:
+            telemetry_kwargs = {'flow_id': flow_id}
+        send_telemetry_event(event='start', obj=self, **telemetry_kwargs)
+        self._start_time = time.time()
```
alaeddine-13, line 74: I think it's better to have function `_get_telemetry_extra_kwargs` and override it in gateway runtime to flow id.
I don't like much how we set a `None` flow id for all pods.
I'm also okay with renaming to `pod_id` and then each runtime can implement it on his own

jina/serve/runtimes/asyncio.py
```diff
@@ -193,3 +206,7 @@ def _log_data_request(self, request: DataRequest):
         self.logger.debug(
             f'recv DataRequest at {request.header.exec_endpoint} with id: {request.header.request_id}'
         )
+
+    @property
+    def _flow_id(self):
+        return None
```
alaeddine-13, line 212: I think it's better to either rename to `pod_id` or have another function like _get_telemetry_extra_kwargs which is applicable to all runtimes in general

jina/orchestrate/flow/base.py
```diff
@@ -2401,3 +2410,14 @@ def __getattribute__(self, item):
...
+
+    @property
+    def _flow_id(self) -> str:
+        import hashlib
+
+        if self._build_level.value < FlowBuildLevel.GRAPH.value:
+            self.build(copy_flow=False)
+
+        return hashlib.sha256(
+            json.dumps(self._get_graph_representation()).encode('utf-8')
```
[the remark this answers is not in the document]
JoanFM: but is the most representing thing. Other args I consider them less important. Which args would u add?

jina/serve/runtimes/asyncio.py
JoanFM: We do not sent a `None` flow_id, we just do not send it

jina/serve/runtimes/asyncio.py
JoanFM: I want it to be a `Flow` representing ID. Not generally `pod_id`. `Flow_id` can be interpreted by us, while `pod_id` not

jina/orchestrate/flow/base.py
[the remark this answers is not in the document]
alaeddine-13: yes but it might not be enough. I'd say it would be great if we manage to get executor names, otherwise this can generate false metrics.
check code snippet for reference
```python
import hashlib
import json

from jina import Flow

flow1 = Flow().add(uses='jinahub://DummyHubExecutor').add(needs='gateway')
flow2 = Flow().add().add(needs='gateway')


with flow1, flow2:
    print(hashlib.sha256(
        json.dumps(flow1._get_graph_representation()).encode('utf-8')).hexdigest())
    print(hashlib.sha256(
        json.dumps(flow2._get_graph_representation()).encode('utf-8')).hexdigest())

```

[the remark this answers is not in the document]
JoanFM: Executor names are not available at the `gateway` level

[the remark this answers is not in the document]
JoanFM: But you are right is too poor. I will make it a draft

jina/orchestrate/flow/base.py
```diff
@@ -1479,7 +1487,8 @@ def start(self):
 
         self._build_level = FlowBuildLevel.RUNNING
 
-        send_telemetry_event(event='start', obj=self)
+        self._start_time = time.time()
+        send_telemetry_event(event='start', obj=self, entity_id=self._entity_id)
```
alaeddine-13, line 1491: I don't understand why do we need to define _entity_id and send start event in Flow ? isn't this already handled in asyncio runtime when it's a gateway ?

alaeddine-13 (approved):

PR #5175, opened 2022-09-15
feat(instrumentation): add OpenTelemetry tracing and metrics with basic configurations

### LINKED ISSUE #5155: Add OpenTelemetry to increase observability
**Describe the feature**

Improve observability using [OpenTelemetry](https://opentelemetry.io) and the available sdk implementations to

- enable tracing of requests
- standardize the already available prometheus metrics collection
- API is need to enable measurement and SDK is required to collect and export metrics

Standard API's are available for monitoring different operations with varying granularity which reduces the effort of using different metrics collection and aggregation vendors. The feature is optional and can be independently configured by users depending on the telemetry collector implementation.

**Your proposal**

Add [OpenTelemetry](https://opentelemetry.io) API's and standard SDK's for:

- tracing network requests within the flow ecosystem
- convert the existing prometheus collection and export to the new open telemetry compatible metrics collector
- allow users to better tracks the operation in the executors

Available packages:

- https://pypi.org/project/opentelemetry-api/
- https://pypi.org/project/opentelemetry-sdk/
- https://pypi.org/project/opentelemetry-instrumentation-grpc/

---

**Environment**

**Screenshots**

girishc13:
**Goals:**

- resolves #5155 
- [x] Integrate OpenTelemetry API and SDK.
- ~[ ] Provide environment variable configurations to enable tracking when required. Use console exporter for now.~
- [x] Trace gRPC requests within the Flow.
- [x] Add helpers for creating traces on request methods with default span attributes.
- ~[ ] Convert `send_health_check_sync` or `is_ready` method to async to prevent the grpc aio interceptor from throwing and capturing an exception.~
- [x] Extract tracing context from the server and make it available for the Executor methods in the `kwargs` list or arguments. 
- [ ] check and update documentation. See [guide](https://github.com/jina-ai/jina/CONTRIBUTING.md#documentation-guidelines) and ask the team.

### Flow
```yaml
jtype: Flow
version: '1'
with:
  protocol: grpc
  port: 54321
  tracing: true
  traces_exporter_host: '0.0.0.0'
  traces_exporter_port: 4317
executors:
  - uses: executor1/config.yml
    name: toyExecutor
  - uses: executor2/config.yml
    name: toyExecutor2
```

### Executor
```python
import functools

from jina import DocumentArray, Executor, requests
from opentelemetry.context.context import Context
from opentelemetry.semconv.trace import SpanAttributes
from opentelemetry.trace import Status, StatusCode


class MyExecutor(Executor):

    @requests
    def foo(self, docs: DocumentArray, tracing_context: Context, **kwargs):
        with self.tracer.start_span("foo", context=tracing_context) as span:
            try:
                span.set_attribute("len_added_docs", len(docs))
                span.set_attribute(SpanAttributes.RPC_METHOD, functools.__name__)

                docs[0].text = 'hello, world!'
                docs[1].text = 'goodbye, world!'
            except Exception as ex:
                span.set_status(Status(StatusCode.ERROR))
                span.record_exception(ex)
            finally:
                span.set_status(Status(StatusCode.OK))
```

### Client
```python
from jina import Client, DocumentArray
import time

if __name__ == '__main__':
    c = Client(
        host='grpc://0.0.0.0:54321',
        tracing=True,
        traces_exporter_host='0.0.0.0',
        traces_exporter_port=4317,
    )

    da = c.post('/', DocumentArray.empty(4))
    print(da.texts)

    time.sleep(3)

```

#### Collecting Data
Please check the `docker-compose.yml` and `otel-collector-config.yml` under the folder `tests/integration/instrumentation` for running the OpenTelemetry collector and Jaeger UI locally.


girishc13:
The python OpenTelemetry sdk currently doesn't have support for the `grpc.aio.Server`. There is an open [pull request](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1245) which will make the implementation easy for us.

jina/serve/instrumentation/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+                metric_readers=[metric_reader], resource=resource
+            )
+            metrics.set_meter_provider(meter_provider)
+            self.meter = metrics.get_meter(name)
+
+    def aio_tracing_server_interceptor(self):
+        '''Create a gRPC aio server interceptor.
+        :returns: A service-side aio interceptor object.
+        '''
+        from . import _aio_server
```
JoanFM, line 54: please use `absolute` modules syntax if possible

jina/serve/instrumentation/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+        return _aio_server.OpenTelemetryAioServerInterceptor(self.tracer)
+
+    @staticmethod
+    def aio_tracing_client_interceptors():
+        '''Create a gRPC client aio channel interceptor.
+        :returns: An invocation-side list of aio interceptor objects.
+        '''
+        tracer = trace.get_tracer(__name__)
```
JoanFM, line 63: where does `name` come from?

jina/serve/instrumentation/_aio_client.py
```diff
@@ -0,0 +1,319 @@
...
+                        description=details,
+                    )
+                )
+        finally:
+            span.end()
+
+    return callback
+
+
+class _BaseAioClientInterceptor(OpenTelemetryClientInterceptor):
```
JoanFM, line 51: why do we need to copy this code here?

jina/serve/instrumentation/__init__.py
girishc13: This will be the package name.

jina/serve/instrumentation/_aio_client.py
girishc13: The opentelemty-python package doesn't yet support `grpc.aio.*` implementations. The interfaces are same or similar to the normal grpc implementation but the `grpc.aio.*` expects async methods which has been provided in this code. This can be removed once the official package provides support.

jina/serve/gateway.py
```diff
@@ -22,28 +23,30 @@ class BaseGateway(JAMLCompatible):
 
     def __init__(
         self,
+        args: 'argparse.Namespace',
```
alaeddine-13, line 26: we can't add this here and I actually don't see why it needs to be added, it's only needed in set_streamer and is already passed as parameter

jina/serve/gateway.py
```diff
@@ -52,21 +55,21 @@ def set_streamer(
...
         from jina.serve.streamer import GatewayStreamer
 
-        graph_description = json.loads(args.graph_description)
-        graph_conditions = json.loads(args.graph_conditions)
-        deployments_addresses = json.loads(args.deployments_addresses)
-        deployments_disable_reduce = json.loads(args.deployments_disable_reduce)
+        graph_description = json.loads(self.args.graph_description)
+        graph_conditions = json.loads(self.args.graph_conditions)
+        deployments_addresses = json.loads(self.args.deployments_addresses)
+        deployments_disable_reduce = json.loads(self.args.deployments_disable_reduce)
```
alaeddine-13, line 61: to be reverted (same as other usage below/above)

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -18,26 +19,30 @@ class GRPCGateway(BaseGateway):
 
     def __init__(
         self,
+        args: 'argparse.Namespace',
```
alaeddine-13, line 22: we can't add this here

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -18,26 +19,30 @@ class GRPCGateway(BaseGateway):
...
         """Initialize the gateway
+        :param args: runtime args
         :param port: The port of the Gateway, which the client should connect to.
         :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}
         :param ssl_keyfile: the path to the key file
         :param ssl_certfile: the path to the certificate file
         :param kwargs: keyword args
         """
-        super().__init__(**kwargs)
+        super().__init__(args=args, **kwargs)
```
alaeddine-13, line 37: same as above

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -18,26 +19,30 @@ class GRPCGateway(BaseGateway):
...
+        self._setup_instrumentation()
         self.port = port
         self.grpc_server_options = grpc_server_options
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
         self.server = grpc.aio.server(
-            options=_get_grpc_server_options(self.grpc_server_options)
+            options=_get_grpc_server_options(self.grpc_server_options),
+            interceptors=[self.aio_tracing_server_interceptor()],
```
alaeddine-13, line 45: Since I'm asking to remove args from constructor, if this needs a specific parameter,that parameter should be added to the constructor and then passed in this method call

jina/serve/runtimes/gateway/http/__init__.py
```diff
@@ -35,7 +36,6 @@ async def async_setup(self):
         )
 
         self.gateway.set_streamer(
-            args=self.args,
```
alaeddine-13, line 38: why was this removed ? this can result in failing tests

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -15,6 +17,7 @@
 
 
 def get_fastapi_app(
+    args: 'argparse.Namespace',
```
alaeddine-13, line 20: why is this needed ? if any specific parameter is really needed, it should be added to this func signature and to gateway constructor

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -18,26 +19,30 @@ class GRPCGateway(BaseGateway):
...
+        :param args: runtime args
         :param port: The port of the Gateway, which the client should connect to.
         :param grpc_server_options: Dictionary of kwargs arguments that will be passed to the grpc server as options when starting the server, example : {'grpc.max_send_message_length': -1}
         :param ssl_keyfile: the path to the key file
         :param ssl_certfile: the path to the certificate file
         :param kwargs: keyword args
         """
-        super().__init__(**kwargs)
+        super().__init__(args=args, **kwargs)
+        self._setup_instrumentation()
```
alaeddine-13, line 38: this one expects an args in self. However I don't think it's possible. What should be done instead is one of these options:
* extract all needed parameters from args and accept them in signature (and possible turn the mixin into a set of helper functions)
* modify the usage of _setup_instrumentation in order to follow monitoring: it is to be applied on runtimes not on gateway class and the runtime receives instrumentation objects and injects them into the gateway class (we already do this dependency injection logic in set_streamer method)

jina/serve/runtimes/gateway/http/gateway.py
```diff
@@ -56,6 +59,7 @@ def __init__(
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
         self.uvicorn_kwargs = uvicorn_kwargs
+        self._setup_instrumentation()
```
alaeddine-13, line 62: same as in grpc gateway

jina/serve/runtimes/gateway/websocket/__init__.py
```diff
@@ -20,6 +20,7 @@ async def async_setup(self):
 
         self.gateway = WebSocketGateway(
             name=self.name,
+            args=self.args,
```
alaeddine-13, line 23: same as in grpc gateway runtime

jina/serve/runtimes/gateway/websocket/__init__.py
```diff
@@ -28,7 +29,6 @@ async def async_setup(self):
         )
 
         self.gateway.set_streamer(
-            args=self.args,
```
alaeddine-13, line 31: same as in grpc gateway runtime

jina/serve/runtimes/gateway/websocket/app.py
```diff
@@ -22,12 +24,14 @@ def _fits_ws_close_msg(msg: str):
 
 
 def get_fastapi_app(
+    args: 'argparse.Namespace',
```
alaeddine-13, line 27: same as in http gateway

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -59,6 +63,9 @@ def get_fastapi_app(
         version=__version__,
     )
 
+    if args.opentelemetry_tracing:
+        FastAPIInstrumentor.instrument_app(app)
```
alaeddine-13, line 67: add `opentelemetry_tracing` as parameter and populate it in gateway constructor

jina/serve/runtimes/gateway/websocket/app.py
```diff
@@ -109,6 +113,9 @@ async def send(
 
     app = FastAPI()
 
+    if args.opentelemetry_tracing:
+        FastAPIInstrumentor.instrument_app(app)
```
alaeddine-13, line 117: same as in http gateway

jina/serve/runtimes/gateway/websocket/gateway.py
```diff
@@ -14,24 +15,27 @@ class WebSocketGateway(BaseGateway):
 
     def __init__(
         self,
+        args: 'argparse.Namespace',
```
alaeddine-13, line 18: same as in grpc gateway

jina/serve/runtimes/gateway/websocket/gateway.py
```diff
@@ -14,24 +15,27 @@ class WebSocketGateway(BaseGateway):
...
         """Initialize the gateway
+        :param args: runtime args
         :param port: The port of the Gateway, which the client should connect to.
         :param ssl_keyfile: the path to the key file
         :param ssl_certfile: the path to the certificate file
         :param uvicorn_kwargs: Dictionary of kwargs arguments that will be passed to Uvicorn server when starting the server
         :param kwargs: keyword args
         """
-        super().__init__(**kwargs)
+        super().__init__(args=args, **kwargs)
```
alaeddine-13, line 33: same as in grpc gateway

jina/serve/runtimes/gateway/websocket/gateway.py
```diff
@@ -14,24 +15,27 @@ class WebSocketGateway(BaseGateway):
...
         self.port = port
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
         self.uvicorn_kwargs = uvicorn_kwargs
+        self._setup_instrumentation()
```
alaeddine-13, line 38: same as in grpc gateway

jina/serve/runtimes/gateway/http/gateway.py
```diff
@@ -65,6 +69,7 @@ async def setup_server(self):
 
         self.app = extend_rest_interface(
             get_fastapi_app(
+                args=self.args,
```
alaeddine-13, line 72: only pass the parameters needed, not all args

jina/serve/runtimes/gateway/websocket/gateway.py
```diff
@@ -41,6 +45,7 @@ async def setup_server(self):
 
         self.app = extend_rest_interface(
             get_fastapi_app(
+                args=self.args,
```
alaeddine-13, line 48: same as in http gateway

alaeddine-13 (commented):
In general, changes requested for Gateway runtime logic:
* runtime can receive args
* gateway class (and all methods it uses) only receives JAMLCompatible parameters and cannot receive all args.

girishc13:
All these usage of 'args' to create or initialise global parameters is not suitable. I will work on removing the instrumentation mixin and move the initialisation to module level. This will work fine when deployed to k8s but there might be some slight unknowns or side effects when running the flow locally.

deepankarm (commented):
I think, tracing should be an optional feature. We should remove related requirements from `core` tag & remove top-level imports.

extra-requirements.txt
```diff
@@ -35,8 +35,17 @@ packaging>=20.0:            core
...
+opentelemetry-instrumentation-grpc>=0.33b0:  core 
 uvloop:                     perf,standard,devel
 prometheus_client:          perf,standard,devel
+opentelemetry-sdk>=1.12.0:   perf,standard,devel
+opentelemetry-exporter-otlp>=1.12.0:  perf,standard,devel
+opentelemetry-exporter-prometheus>=1.12.0rc1:  perf,standard,devel
+opentelemetry-semantic-conventions>=0.33b0:    perf,standard,devel
+opentelemetry-instrumentation-aiohttp-client>=0.33b0:    perf,standard,devel
+opentelemetry-instrumentation-fastapi>=0.33b0: perf,standard,devel
+opentelemetry-exporter-jaeger>=1.12.0: perf,standrad,devel
```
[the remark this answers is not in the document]
JoanFM: it should be treated as `prometheus`

jina/parsers/client.py
```diff
@@ -30,3 +30,33 @@ def mixin_client_features_parser(parser):
         default=False,
         help='If set, then the input and output of this Client work in an asynchronous manner. ',
     )
+
+    parser.add_argument(
+        '--opentelemetry-tracing',
```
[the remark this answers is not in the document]
JoanFM: I agree here, I am not sure people are interested in the OpenTelemetry detail?

JoanFM (changes requested):
Agree with @deepankarm , I would simply use the name `tracing` and `monitoring` without the `opentelemetry` prefix or suffix

jina/serve/runtimes/gateway/http/__init__.py
```diff
@@ -1,71 +1 @@
...
-        await self.gateway.teardown()
-
-    async def async_cancel(self):
-        """Stop the server."""
-        await self.gateway.stop_server()
-
-    async def async_run_forever(self):
-        """Running method of the server."""
-        await self.gateway.run_server()
+from .gateway import HTTPGateway
```
JoanFM, line 1: please use absolute import

JoanFM (changes requested):

jina/clients/base/helper.py
```diff
@@ -1,9 +1,13 @@
+import argparse
 import asyncio
 import random
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 from aiohttp import WSMsgType
+from opentelemetry import trace
```
JoanFM, line 8: can we hide these `imports` when they are needed, so that the dependency is not needed if trace is not activated?

JoanFM (changes requested):
Please rename to `tracing` the argument

jina/clients/base/helper.py
girishc13: We cannot hide the `opentelmetry` api module imports. Otherwise we cannot use any typing. Hiding also doesn't make sense to me because the dependency injection happens in the class constructor for most implementations. These imports are just api's and don't execute code unless some methods are executed.

The sdk imports are already hidden in the `instrumentation` module.

jina/serve/instrumentation/__init__.py
```diff
@@ -1,6 +1,7 @@
 from typing import Optional
```
JoanFM, line 2: remember to hide the `opentelemetry` imports if possible as `prometheus-client` was doing

jina/clients/base/helper.py
```diff
@@ -30,7 +26,7 @@ def __init__(
         initial_backoff: float = 0.5,
         max_backoff: float = 0.1,
         backoff_multiplier: float = 1.5,
-        tracer_provider: Optional[trace.TracerProvider] = None,
+        tracer_provider: Optional[Any] = None,
```
JoanFM, line 29: why Any? it should not be any

JoanFM (changes requested):

jina/clients/base/helper.py
girishc13: I'm using Any for now to check if hiding imports will help in any way. I understand module level imports should also be hidden but also in some cases this import here is purely for typing info and nothing else.

JoanFM: u can do the typing by hiding in 'trace.TraceProvider', just the string

tests/conftest.py
```diff
@@ -88,3 +88,8 @@ def event_loop(request):
     loop = asyncio.get_event_loop_policy().new_event_loop()
     yield loop
     loop.close()
+
+
+@pytest.fixture(autouse=True)
+def set_start_method():
```
JoanFM, line 94: later we can take this out and try

jina/serve/runtimes/worker/__init__.py
```diff
@@ -6,6 +6,7 @@
 import grpc
 from grpc_health.v1 import health, health_pb2, health_pb2_grpc
 from grpc_reflection.v1alpha import reflection
+from opentelemetry.propagate import Context, extract
```
JoanFM, line 9: try to hide import

jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -169,6 +190,7 @@ async def handle(self, requests: List['DataRequest']) -> DataRequest:
                 requests,
                 field='docs',
             ),
+            otel_context=otel_context,
```
JoanFM, line 193: we have to rethink the name @JohannesMessner @samsja @alaeddine-13 any suggestion?

jina/serve/streamer.py
```diff
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union
```
JoanFM, line 1: remove Any

tests/integration/instrumentation/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+    return (server_spans, client_spans, internal_spans)
+
+
+class ExecutorTestWithTracing(Executor):
+    def __init__(
+        self,
+        metas: Optional[Dict] = None,
+        requests: Optional[Dict] = None,
+        runtime_args: Optional[Dict] = None,
```
JoanFM, line 57: do not use these, just use `**kwargs`, there is no need to expose this as this is handled by runtime

JoanFM (changes requested):
I have some comments: 

Besides some leftovers of imports, I have 3 main concerns.

- I think we do not need to use `NoOp` operations, we can use `None` so that user can use Jina without `opentelemetry` dependency
- We need to rethink the `otel_naming` and working
- We need to make sure that the basic `span` works for the `Executor` and `gateway` without any code from developer

jina/serve/runtimes/request_handlers/data_request_handler.py
alaeddine-13: `tracing_context` ?
we use `monitoring` as reference to prometheus, we can come up with a term like `tracing` to reference open telemetry as well.

jina/serve/runtimes/gateway/http/app.py
```diff
@@ -39,6 +43,8 @@ def get_fastapi_app(
     :param expose_graphql_endpoint: If set, /graphql endpoint is added to HTTP interface.
     :param cors: If set, a CORS middleware is added to FastAPI frontend to allow cross-origin access.
     :param logger: Jina logger.
+    :param tracing: Enables tracing is set to True.
```
JohannesMessner, line 46:
```suggestion
    :param tracing: Enables tracing if set to True.
```

jina/serve/instrumentation/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class InstrumentationMixin:
+    '''Instrumentation mixin for OpenTelemetery Tracing and Metrics handling'''
+
+    def _setup_instrumentation(
+        self,
+        name: str,
+        tracing: Optional[bool] = False,
+        span_exporter_host: Optional[str] = '0.0.0.0',
```
JohannesMessner, line 11: I think for monitoring we have this naming the other way around, so `port_monitoring` and not `monitoring_port`. I think we should do the same and change this to `port_span_exporter` and `host_span_exporter`.

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -36,15 +36,17 @@ def __init__(
...
-        self.server = grpc.aio.server(
-            options=_get_grpc_server_options(self.grpc_server_options)
-        )
         self.health_servicer = health.HealthServicer(experimental_non_blocking=True)
 
     async def setup_server(self):
         """
         setup GRPC server
         """
+        self.server = grpc.aio.server(
```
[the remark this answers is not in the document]
girishc13: Can't create the server in `init` because the opentelmetry dependencies are not available at the `init` level. These are set by the `inject_dependencies` method.

tests/integration/instrumentation/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+        workspace: Optional[str] = None,
+        **kwargs,
+    ):
+        super().__init__(metas, requests, runtime_args, workspace, **kwargs)
+        self.docs_counter = self.meter.create_counter(name='docs_counter')
+
+    @requests(on='/index')
```
JohannesMessner, line 64: So the `requests` decorator does not do any tracing in the background?

jina/serve/runtimes/gateway/__init__.py
```diff
@@ -83,11 +86,16 @@ async def async_setup(self):
             extra_search_paths=self.args.extra_search_paths,
         )
 
-        self.gateway.set_streamer(
+        self.gateway.inject_dependencies(
```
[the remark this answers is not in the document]
girishc13: Why not? There are 4 dependencies for tracing.

tests/integration/instrumentation/__init__.py
JoanFM: We will have to do it, in the `requests` or right in the `DataRequestHandler`

jina/serve/instrumentation/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+            from opentelemetry.sdk.metrics import MeterProvider
+            from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+            from opentelemetry.sdk.resources import SERVICE_NAME, Resource
+
+            resource = Resource(attributes={SERVICE_NAME: name})
+
+            metric_reader = PeriodicExportingMetricReader(
+                OTLPMetricExporter(
+                    endpoint=f'{metrics_exporter_host}:{metrics_exporter_port}',
+                    insecure=True,
```
alaeddine-13, line 60: can't we just have args: metric-exporter-kwargs ? (like in docker-kwargs parameter)
there are quite some interesting parameters left here

jina/serve/runtimes/gateway/http/__init__.py
girishc13: This is not my change. Using absolute import won't work. Maybe @alaeddine-13 can explain.

jina/serve/networking.py
```diff
@@ -987,13 +1012,55 @@ async def task_wrapper():
...
+
     @staticmethod
     def get_grpc_channel(
         address: str,
         options: Optional[list] = None,
         asyncio: bool = False,
         tls: bool = False,
         root_certificates: Optional[str] = None,
+        aio_tracing_client_interceptors: Optional[Sequence[Any]] = None,
+        tracing_client_interceptor: Optional[Any] = None,
```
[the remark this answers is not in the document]
girishc13: Both sync and async methods are used at runtime so we cannot decide this during dependency injection.

jina/serve/instrumentation/__init__.py
girishc13: We can but this will have to be provided in the next iteration because there are still quite a few details that we need to decide upon. This is currently not production ready.

docs/fundamentals/flow/executor-args.md
```diff
@@ -35,6 +35,12 @@
 | `port_monitoring` | The port on which the prometheus server is exposed, default is a random port between [49152, 65535] | `string` | `random in [49152, 65535]` |
 | `retries` | Number of retries per gRPC call. If <0 it defaults to max(3, num_replicas) | `number` | `-1` |
 | `floating` | If set, the current Pod/Deployment can not be further chained, and the next `.add()` will chain after the last Pod/Deployment not this current one. | `boolean` | `False` |
+| `tracing` | If set, the sdk implementation of the OpenTelemetry tracer will be available and will be enabled for automatic tracing of requests and customer span creation. Otherwise a no-op implementation will be provided. | `boolean` | `False` |
+| `span_exporter_host` | If tracing is enabled, this hostname will be used to configure the trace exporter agent. | `string` | `None` |
+| `span_exporter_port` | If tracing is enabled, this port will be used to configure the trace exporter agent. | `number` | `None` |
+| `metrics` | If set, the sdk implementation of the OpenTelemetry metrics will be available for default monitoring and custom measurements. Otherwise a no-op implementation will be provided. | `boolean` | `False` |
```
samsja, line 41: I don't understand the sentence. Isn't it going to overlap with the `monitoring` ?

samsja (changes requested):
I think that we should merge the new `metrics` argument with the already existing `monitoring` one.

Great PR I am looking forward to it

docs/fundamentals/flow/executor-args.md
```diff
@@ -35,6 +35,12 @@
 | `port_monitoring` | The port on which the prometheus server is exposed, default is a random port between [49152, 65535] | `string` | `random in [49152, 65535]` |
 | `retries` | Number of retries per gRPC call. If <0 it defaults to max(3, num_replicas) | `number` | `-1` |
 | `floating` | If set, the current Pod/Deployment can not be further chained, and the next `.add()` will chain after the last Pod/Deployment not this current one. | `boolean` | `False` |
+| `tracing` | If set, the sdk implementation of the OpenTelemetry tracer will be available and will be enabled for automatic tracing of requests and customer span creation. Otherwise a no-op implementation will be provided. | `boolean` | `False` |
+| `span_exporter_host` | If tracing is enabled, this hostname will be used to configure the trace exporter agent. | `string` | `None` |
+| `span_exporter_port` | If tracing is enabled, this port will be used to configure the trace exporter agent. | `number` | `None` |
```
JohannesMessner, line 40: I know this is a small thing that I mentioned already, so sorry to be a PITA about this, but I really think we should switch these around to ''host/port_span_exporter" to align them with the nomenclature of the prometheus feature. It's the small things that make a good user experience imo

samsja: @JohannesMessner what name would u suggest ?

jina/serve/executors/__init__.py
```diff
@@ -330,11 +340,18 @@ async def __acall_endpoint__(self, req_endpoint, **kwargs):
             else contextlib.nullcontext()
         )
 
-        with _summary:
-            if iscoroutinefunction(func):
-                return await func(self, **kwargs)
-            else:
-                return func(self, **kwargs)
+        if self.tracer:
+            with self.tracer.start_span(req_endpoint, context=tracing_context) as _:
```
JoanFM, line 344: but does it check if it is successfull?

JoanFM (changes requested):

jina/serve/executors/__init__.py
girishc13: The span uses a context manager so it's handled automatically. We can add a try-except block in case we want to add custom attributes, record exceptions or set specific error status codes.

docs/fundamentals/flow/executor-args.md
girishc13: The port_monitoring won't exist in the near future and there will be only the `span_exporter` attributes. I'm generally used to seeing and using `_host` as the suffix rather than as a prefix.

samsja: but then we might introduce a breaking change right ? We need to be careful

docs/fundamentals/flow/executor-args.md
girishc13: Yes, my intention is to use the same terms as OpenTelemetry. If people read the OpenTelemetry documentation then the terms are aligned.

docs/fundamentals/flow/executor-args.md
samsja: we can deprecated an argument if needed but this should be thinked ahead.

@girishc13 could you show here what would be the relevant arguments on this near future where `port_monitoring` does not exist ?

girishc13: Also you can think of the naming in terms on the yaml configuration for the OpenTelemetry collector. The hierarchy that I'm implicitly used to is: `dependency -> service -> host, port, ...`. So this naturally follows the convention of `service.host` and `service.port`.

```
version: "3"
services:
  # Jaeger
  jaeger:
    image: jaegertracing/all-in-one:latest
    ports:
      - "16686:16686"
      - "14250"

  otel-collector:
    image: otel/opentelemetry-collector:0.61.0
    command: [ "--config=/etc/otel-collector-config.yml" ]
    volumes:
      - ${PWD}/otel-collector-config.yml:/etc/otel-collector-config.yml
    ports:
      - "1888:1888" # pprof extension
      - "8888:8888" # Prometheus metrics exposed by the collector
      - "8889:8889" # Prometheus exporter metrics
      - "13133:13133" # health_check extension
      - "55679:55679" # zpages extension
      - "4317:4317" # OTLP gRPC receiver
      - "4318:4318" # OTLP http receiver
    depends_on:
      - jaeger
```

jina/serve/executors/__init__.py
JoanFM: Please add a test, I do not understand what u say, I can see from this `span` if it succeeded or failed?

tests/integration/instrumentation/test_flow_instrumentation.py
```diff
@@ -61,3 +63,39 @@ def test_gateway_instrumentation(
...
+        tracing=True,
+        span_exporter_host='localhost',
+        span_exporter_port=4317,
+    ).add(uses=ExecutorFailureWithTracing)
+
+    with f:
+        from jina import DocumentArray
+
+        try:
+            f.post(
```
JoanFM, line 79: no need to do this, just pass `continue_on_error=True`

JoanFM (changes requested):

docs/fundamentals/flow/executor-args.md
JohannesMessner: I'm not a big fan of deprecating our current `port_monitoring` so quickly after it being introduced, but if it leads to a nicer and more unified experience moving forward then we'll have to do it.

But apart from the argument naming, am I understanding correctly that, according to this plan, the user won't be able to use prometheus to collect metrics anymore? Or will the setup on the user side remain the same, and we only change the way we expose these metrics from our internals?
Because on the otel collector site I still see some prometheus logos but some of them are not connected to the system, so I am a bit lost.

If this is the case, then I don't think we should remove the current way users set up their metrics pipeline. This would be a huge breaking change.

girishc13: > But apart from the argument naming, am I understanding correctly that, according to this plan, the user won't be able to use prometheus to collect metrics anymore? Or will the setup on the user side remain the same, and we only change the way we expose these metrics from our internals?

The main concern from my understanding is introducing a breaking change for the metrics data which requires new setup. Do we have data on how many users are using the Prometheus client for monitoring except for JCloud users? Also the lack of interior between OpenTelemetry monitoring and Prometheus monitoring makes it a bit hard to just remove the current monitoring setup.

I can think of the following ways to tackle this:

1. We can also choose to release only the tracing instrumentation and work on the metrics later if we get feedback from the users. I also believe that the OpenTelemetry metrics does not provide rich features when compared to Prometheus but it's still the direction to go early to avoid the users from investing too much into the Prometheus only solution.
2. We deprecate Prometheus monitoring and continue supporting OpenTelemetry tracing and monitoring for users that want to work with OpenTelemetry. The decision is up to the user and we might have some more work to maintain both.

JoanFM: I would declared the old metric system as deprecated (TO BE REMOVED in a couple of minors) and go with full OpenTelemetry approach

girishc13: The official Prometheus library already [supports](https://github.com/prometheus/prometheus/pull/9724) OpenTelemetry api's and sdk's. The OpenTelemetry Collector also supports [scraping](https://opentelemetry.io/docs/collector/configuration/#receivers) data from the existing Prometheus client. We might need some elaborate configuration for metrics and OpenTelemetry Collector to support the existing mechanism but OpenTelemetry is the way to go.

jina/serve/runtimes/gateway/__init__.py
```diff
@@ -83,11 +86,16 @@ async def async_setup(self):
...
             args=self.args,
             timeout_send=self.timeout_send,
             metrics_registry=self.metrics_registry,
             runtime_name=self.args.name,
+            tracing=self.tracing,
+            tracer_provider=self.tracer_provider,
+            grpc_tracing_server_interceptors=self.aio_tracing_server_interceptor(),
```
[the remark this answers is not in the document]
girishc13: Yes, I probably missed the method name when I did the refactoring to hide the imports.

jina/serve/runtimes/gateway/http/__init__.py
JoanFM, line 1: please use the `absolute` import

JoanFM (changes requested):

jina/serve/runtimes/gateway/http/__init__.py
JoanFM: ok will do in separate PR

docs/fundamentals/flow/executor-args.md
girishc13: Will be renamed to `traces_exporter_host`?

samsja (approved):
Lgtm

JoanFM (approved):

JohannesMessner (approved):

PR #5176, opened 2022-09-15
feat: add multiple attempts options to client.post API

JoanFM:
Goal of this PR is to enable `retry_mechanism` in the client.

- [x] Expose `max_attempts`, `initial_backoff`, `max_backoff` and `backoff_multiplier` in the `client.post` API
- [x] Clarify usage of `continue_on_error` with any kind of Errors
- [ ] Test features
- [x] Document

**Changes**
- Allow to pass `max_attempts`, `initial_backoff`, `max_backoff` and `backoff_multiplier` to client.post API
   - In gRPC, these are added as configuration to the grpcChannel
   - In HTTP and websocket, we do a manual retry following the formula proposed by grpc
```The initial retry attempt will occur at random(0, initialBackoff). In general, the n-th attempt will occur at random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff)).```

- Document `continue_on_error` behavior.

### CODE REVIEW

File: docs/fundamentals/client/client.md [hunk 1]
```diff
@@ -668,6 +663,20 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
...
+````{admonition} What errors can be handled by the callback
+:class: caution
+Callback can handle errors that are caused inside the Flow, it can be from an Executor raising an Exception or from the Gateway having connectivity errors with the Executors.
+It will not receive errors of connectivity between the Client and the Gateway.
+````
+
+### Transient fault handling with retries
+
+`client.post()` accepts `num_retries` and `retry_delay` parameters to control the capacity to retry requests when a transient connectivity error occurs.
+This can help to overcome transient network connectivity issues.
```
alaeddine-13, line 675: it feels like we're hiding the retry mechanism on users. If it's meant to be the same for all protocols I'd decorate all `send_request` methods with the same decorator, otherwise say that in grpc it relies on grpc retry mechanism

docs/fundamentals/client/client.md
```diff
@@ -668,6 +663,20 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
...
 
+````{admonition} What errors can be handled by the callback
+:class: caution
+Callback can handle errors that are caused inside the Flow, it can be from an Executor raising an Exception or from the Gateway having connectivity errors with the Executors.
+It will not receive errors of connectivity between the Client and the Gateway.
+````
+
+### Transient fault handling with retries
+
+`client.post()` accepts `num_retries` and `retry_delay` parameters to control the capacity to retry requests when a transient connectivity error occurs.
```
samsja, line 674: Can we expecitly explain what these two parameters control ?

jina/clients/base/grpc.py
```diff
@@ -84,58 +109,65 @@ async def _get_results(
...
-                            on_done=on_done,
-                            on_always=on_always,
-                            continue_on_error=self.continue_on_error,
-                            logger=self.logger,
-                        )
-                        if self.show_progress:
-                            p_bar.update()
-                        yield resp
+                    try:
+                        async for resp in stub.Call(
```
florian-hoenicke, line 113: Should we make the retry here? As I understand this code,it will still fail in case of a network error. Right?
So what about the case where the wifi is down for a few seconds? Will it retry?

jina/clients/base/helper.py
```diff
@@ -17,11 +17,13 @@
 class AioHttpClientlet(ABC):
     """aiohttp session manager"""
 
-    def __init__(self, url: str, logger: 'JinaLogger', **kwargs) -> None:
+    def __init__(self, url: str, logger: 'JinaLogger',  num_retries: int = 1, retry_delay: float = 0.5,  **kwargs) -> None:
```
samsja, line 20: isn't 0.5 a bit too long ?

jina/clients/base/grpc.py
JoanFM: the `grpc` channel retry options should be the one handling this.

docs/fundamentals/client/client.md
JoanFM: I will say

samsja (commented):
minor comments, it looks good to me

docs/fundamentals/client/client.md
JoanFM: Okey, I will explain better

jina/clients/base/helper.py
JoanFM: it won't affect in the default

jina/clients/base/helper.py
```diff
@@ -161,10 +188,21 @@ async def send_message(self, request: 'Request'):
...
+            except ConnectionResetError:
+                if retry == self.max_attempts - 1:
+                    self.logger.critical(f'server connection closed already!')
+                    raise
+                else:
+                    if retry == 0:
+                        wait_time = random.uniform(0, self.initial_backoff)
+                    else:
+                        wait_time = random.uniform(0, min(self.initial_backoff*self.backoff_multiplier**(retry-1), self.max_backoff))
+                    await asyncio.sleep(wait_time)
```
alaeddine-13, line 205:
```suggestion
        for retry in range(1, self.max_attempts + 1):
            try:
                return await self.websocket.send_bytes(request.SerializeToString())
            except ConnectionResetError:
                if retry == self.max_attempts:
                    self.logger.critical(f'server connection closed already!')
                    raise
                else:
                    wait_time = random.uniform(0, min(self.initial_backoff*self.backoff_multiplier**(retry-1), self.max_backoff))
                    await asyncio.sleep(wait_time)
```

docs/fundamentals/client/client.md
```diff
@@ -668,7 +663,26 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
...
+`client.post()` accepts `continue_on_error` parameter. When set to True, `Client` will keep trying to send the following requests. `continue_on_error` parameter will only apply
+to Exceptions caused by Executor exceptions, but in case of network connectivity issues, an Exception will be raised.
+
+### Transient fault handling with retries
+
+`client.post()` accepts `max_attempts`, `initial_backoff`, `max_backoff` and `backoff_multiplier` parameters to control the capacity to retry requests when a transient connectivity error occurs.
+This can help to overcome transient network connectivity issues. 
+
+The `max_attempts` determines the number of sending attempts, including the original request.
+The `initial_backoff`, `max_backoff`, and `backoff_multiplier` parameters determine the randomized delay before retry attempts.
```
alaeddine-13, line 683:
```suggestion
The `initial_backoff`, `max_backoff`, and `backoff_multiplier` parameters determine the randomized delay in seconds before retry attempts.
```

jina/clients/base/helper.py
```diff
@@ -103,7 +118,19 @@ async def send_message(self, request: 'Request'):
...
+                return res
+            except:
+                if retry == self.max_attempts - 1:
+                    raise
+                else:
+                    if retry == 0:
+                        wait_time = random.uniform(0, self.initial_backoff)
+                    else:
+                        wait_time = random.uniform(0, min(self.initial_backoff*self.backoff_multiplier**(retry-1), self.max_backoff))
+                    await asyncio.sleep(wait_time)
```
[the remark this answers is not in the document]
JoanFM: I do not think it is worth the pain to try to track the Exceptions, just retry and eventually it will fail again

jina/clients/base/grpc.py
```diff
@@ -62,6 +63,10 @@
         on_error: Optional['CallbackFnType'] = None,
         on_always: Optional['CallbackFnType'] = None,
         compression: Optional[str] = None,
+        max_attempts: bool = 1,
```
samsja, line 66:
```suggestion
max_attempts: int = 1
```

jina/clients/base/helper.py
```diff
@@ -103,7 +118,15 @@
...
             req_dict['target_executor'] = req_dict['header']['target_executor']
-        return await self.session.post(url=self.url, json=req_dict).__aenter__()
+        for retry in range(1, self.max_attempts + 1):
+            try:
+                return await self.session.post(url=self.url, json=req_dict).__aenter__()
+            except:
+                if retry == self.max_attempts:
+                    raise
+                else:
+                    wait_time = random.uniform(0, min(self.initial_backoff*self.backoff_multiplier**(retry-1), self.max_backoff))
```
samsja, line 128: I might have missed the design discussion on this but why is the waiting time random here ?

docs/fundamentals/client/client.md
```diff
@@ -668,7 +663,26 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
...
+`client.post()` accepts `continue_on_error` parameter. When set to True, `Client` will keep trying to send the following requests. `continue_on_error` parameter will only apply
+to Exceptions caused by Executor exceptions, but in case of network connectivity issues, an Exception will be raised.
+
+### Transient fault handling with retries
+
+`client.post()` accepts `max_attempts`, `initial_backoff`, `max_backoff` and `backoff_multiplier` parameters to control the capacity to retry requests, using exponential backoff strategy, when a transient connectivity error occurs.
+This can help to overcome transient network connectivity issues. 
+
+The `max_attempts` determines the number of sending attempts, including the original request.
+The `initial_backoff`, `max_backoff`, and `backoff_multiplier` parameters determine the randomized delay in seconds before retry attempts.
```
samsja, line 683: IMO this is still unclear in the documentation what this parameters actually do.

Maybe we should include a small example and explain when the client will retry to send the requests with respect to the given parameters

samsja (dismissed):

docs/fundamentals/client/client.md
JoanFM: I added the formula

jina/clients/base/helper.py
JoanFM: it is the definition of what is done in grpc

docs/fundamentals/client/client.md
```diff
@@ -668,7 +663,26 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
     )
 ```
 
+````{admonition} What errors can be handled by the callback
+:class: caution
+Callback can handle errors that are caused by Executors raising an Exception.
+It will not receive exceptions or from the Gateway having connectivity errors with the Executors or between the Client and the Gateway.
```
JohannesMessner, line 669: I don't understand this sentence :/

docs/fundamentals/client/client.md
```diff
@@ -668,7 +663,26 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
...
+````
 
+### Continue streaming when an error occurs
+
+`client.post()` accepts `continue_on_error` parameter. When set to True, `Client` will keep trying to send the following requests. `continue_on_error` parameter will only apply
+to Exceptions caused by Executor exceptions, but in case of network connectivity issues, an Exception will be raised.
```
JohannesMessner, line 675:
```suggestion
`client.post()` accepts a `continue_on_error` parameter. When set to `True`, the Client will keep trying to send the remaining requests. The `continue_on_error` parameter will only apply
to Exceptions caused by an Executor, but in case of network connectivity issues, an Exception will be raised.
```

docs/fundamentals/client/client.md
```diff
@@ -668,7 +663,26 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
...
+
+### Transient fault handling with retries
+
+`client.post()` accepts `max_attempts`, `initial_backoff`, `max_backoff` and `backoff_multiplier` parameters to control the capacity to retry requests, using exponential backoff strategy, when a transient connectivity error occurs.
```
JohannesMessner, line 679:
```suggestion
`client.post()` accepts `max_attempts`, `initial_backoff`, `max_backoff` and `backoff_multiplier` parameters to control the capacity to retry requests, when a transient connectivity error occurs, using an exponential backoff strategy, 
```

JohannesMessner (commented):
This seems like a slightly more sophisticated retry mechanism compared to the one that we currently have between gateway and executors.

There (I believe) we don't use the built-in gRPC mechanism, but we do it manually (I think because we want to be able to switch to a different replica when one of them doesn't respond).
Should we implement this exponential backoff strategy there as well?

docs/fundamentals/client/client.md
JoanFM: that "networking" exceptions are not passed to the `on_error` callback, how could I better phrase this?

docs/fundamentals/client/client.md
```diff
@@ -668,7 +663,26 @@ with Flow().add() as f, open('output.txt', 'w') as fp:
     )
 ```
 
+````{admonition} What errors can be handled by the callback
```
alexcg1, line 666:
```suggestion
````{admonition} What errors can be handled by the callback?
```

docs/fundamentals/client/client.md
alexcg1, line 669:
```suggestion
Callbacks can handle errors that are caused by Executors raising an Exception.

A callback will not receive exceptions:
- from the Gateway having connectivity errors with the Executors.
- between the Client and the Gateway.
```

I'm not clear on the "or from" part here. Is this a typo? Regardless, I've changed it as best I can. Does the admonition box support bullets?

alexcg1 (changes requested):
Some clarification needed

alexcg1 (approved):
LGTM :+1:

alaeddine-13 (approved):

Issue #5192, opened 2022-09-22
Document grpc client threading limitation


alaeddine-13:
Currently it's not clear to users that the GRPC Client cannot be used in multi-threaded environment like http servers or so.
We should mention that the client can only be used in a single thread and users either have to use 1 thread (use multi processing for servers instead) or use another protocol like http.

PR #5193, opened 2022-09-22
docs: document grpc client limitation

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/5192

Currently it's not clear to users that the GRPC Client cannot be used in multi-threaded environment like http servers or so.
This pr  mentions that the client can only be used in a single thread and users either have to use 1 thread (use multi processing for servers instead) or use another protocol like http.

### DIFF
--- a/docs/fundamentals/client/client.md
+++ b/docs/fundamentals/client/client.md
@@ -148,8 +148,8 @@ Client(host='https://my.awesome.flow:1234', port=4321)
 ```
 ````
 
-````{admonition} Important
-:class: important
+````{admonition} Caution
+:class: caution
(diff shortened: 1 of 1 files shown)


JoanFM (dismissed):

samsja (changes requested):
please stick to Hint Caution or see also for the admonition

docs/fundamentals/client/client.md
```diff
@@ -148,6 +148,15 @@ Client(host='https://my.awesome.flow:1234', port=4321)
 ```
 ````
 
+````{admonition} Important
+:class: important
```
samsja, line 152:
```suggestion
:class: caution
```

JoanFM (approved):

PR #5202, opened 2022-09-23
docs: clarify exec endpoint usage in http

JoanFM:
**Goals**

Clarify confusion between HTTP endpoints and Executor endpoints

### DIFF
--- a/docs/fundamentals/gateway/index.md
+++ b/docs/fundamentals/gateway/index.md
@@ -218,7 +218,9 @@ curl --request POST \
 The above cURL command is equivalent to passing the `on` parameter to `client.post` as follows:
 
 ```python
-from jina import Client, DocumentArray, Document
+from docarray import DocumentArray, Document
+from jina import Client
+
 client = Client(port=12345, protocol='http')
 client.post(on='/foo', inputs=DocumentArray([Document(text='hello world')]))
 ```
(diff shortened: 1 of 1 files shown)


docs/fundamentals/gateway/index.md
```diff
@@ -205,6 +205,24 @@ with:
...
+However, if you want to send requests to a different Executor endpoint, you can still do it without exposing it in the HTTP endpoint, by sending an HTTP request to the `/post` HTTP endpoint while setting  
+`execEndpoint` in the request.
+
+```text
+curl --request POST \
+'http://localhost:12345/post' \
+--header 'Content-Type: application/json' -d '{"data": [{"text": "hello world"}], "execEndpoint": "/foo"}'
+```
+
+this would be equivalent to passing the `on` parameter to `client.post` like:
```
alexcg1, line 218:
```suggestion
The above cURL command is equivalent to passing the `on` parameter to `client.post` as follows:
```

docs/fundamentals/gateway/index.md
```diff
@@ -205,6 +205,24 @@ with:
...
+
+```python
+from jina import Client, DocumentArray, Document
```
alexcg1, line 221: Should we be importing everything from `jina` or some from `docarray?`

alexcg1 (changes requested):
Looking good. Just a small change and question

docs/fundamentals/gateway/index.md
JoanFM: I like `docarray` good point

alexcg1 (approved):
LGTM :+1:

PR #5205, opened 2022-09-26
fix(hubio.fetch_meta): move significant params to sequential

### LINKED ISSUE #5204: buggy hubio.fetch_image causes `docker.errors.NullResource: Resource ID was not provided`
**Describe the bug**

People sometimes get `docker.errors.NullResource: Resource ID was not provided` when pulling hub images.

To reproduce:

```bash
jina hub pull jinahub://DummyHubExecutor --force
``` 
```bash
jina hub pull jinahub+docker://DummyHubExecutor
``` 

**Describe how you solve it**

```bash
jina hub pull jinahub+docker://DummyHubExecutor --force
``` 
---

This is because of the buggy cache behaviour of `jina.hubio.fetch_image`

nomagick:
**Goals:**

- resolves #5204

So `hubio.fetch_meta` is decorated by `disk_cache_offline` which implements the cache.
https://github.com/jina-ai/jina/blob/1399e36cea5d29621c0d180726685c1aa8a14657/jina/hubble/hubio.py#L958-L981

If you look into the docstring more carefully, it actually warns about the difference between `*args` and `**kwargs`.

This is because inside the implementation of `disk_cache_offline`, the `call_hash `, which is used for the cache key, only considers the sequential parameters.
https://github.com/jina-ai/jina/blob/1399e36cea5d29621c0d180726685c1aa8a14657/jina/hubble/helper.py#L372-L379

For `hubio.fetch_meta`'s case, the `image_required` parameter controls if the API would return docker image names. And `rebuild_image` controls if the returned docker image would likely be different as before.
These parameters are as significant as the `name` and `tag` so they need to be made into sequential parameters.


nomagick:
These should verify the fixed behaviour:
https://github.com/jina-ai/jina/blob/71ca97a51af7054db32c48bd258146777c36d5fa/tests/unit/hubble/test_hubio.py#L1155-L1175

Where the `pull jinahub+sandbox://...` and `pull jinahub+docker://...` doesn't share the same cache.

Actually, we should compare `pull jinahub://...`, but it has too many file system operations. So I used the `jinahub+sandbox` instead.

mapleeit (approved):
LGTM

Issue #5135, opened 2022-09-06
GRPC keep alive need to be configured on client side as well


samsja:
# Context

According to this https://github.com/grpc/grpc/blob/master/doc/keepalive.md we need to setup the client in the same way as the server for the keepalive parameters to work

PR #5211, opened 2022-09-26
feat: use default grpc parameters for grpc connection pool connection

samsja:
# Context

According to this https://github.com/grpc/grpc/blob/master/doc/keepalive.md we need to setup the client in the same way as the server for the keepalive parameters to work

fixing : https://github.com/jina-ai/jina/issues/5135

# What this pr do:

- [x] add the default grpc param to the connection pool


alaeddine-13 (commented):
what about options that are specific to server only, they have no effect on client right ?

samsja:
> what about options that are specific to server only, they have no effect on client right ?

Yes you are right I should separate the concern

jina_cli/autocomplete.py
```diff
@@ -203,6 +204,7 @@
             '--version',
             '--loglevel',
```
JoanFM, line 205: why are these changes related?

JoanFM (changes requested):

jina_cli/autocomplete.py
samsja: let me check this come from our best friend jina bot I believe

alaeddine-13: I think with upstream dependencies and more args added in jcloud, jina bot infers it and adds it to our autocomplete

alaeddine-13: here they added logout https://github.com/jina-ai/jcloud/pull/84

samsja: good catch

alaeddine-13 (approved):

JoanFM (approved):

PR #5217, opened 2022-09-27
feat: distributed replicas across different hosts

### LINKED ISSUE #5103: How can I serve replicas of an executor in different hosts?
First, I thank Jina for an easy deployment/scaling framework. In my scenario, I have independent physical instance GPU resources, so is there any way to host replicas of an executor type in a different place? For example, is a `external replica` exist? Or suppose I have to do that manually. In that case, any tool in the Jina ecosystem or third party can help me to solve this problem?

### LINKED ISSUE #5231: Bug: For external Executors, passing entire address to `host` does not work
The following syntax, which is supported for the gateway, does not work for external executors:

```python
f.add(host='grpc://localhost:1234', external=True)
```

Instead, host and port have to be precised separately:

```python
f.add(host='localhost', port=1234, external=True)
```

Not that this bug also applies to replicated external Executors:

```python
f.add(host='grpc://localhost:1234,grpc://localhost:1235', external=True)
```

JohannesMessner:
## Goal

Allow replicas of the same Executor to run in a distributed manner (on different machines).
Closes #5103
Closes #5231 

## Usage

```python
f.add(host='localhost,91.198.174.192', port='12345,12346', external=True)
```
Alternatively:
```python
f.add(host='localhost:12345,91.198.174.192:12346', external=True)
```
will connect to `grpc://localhost:12345` and `grpc://91.198.174.192:12346` as two replicas of the same Executor.

## ToDo
- [x] allow different hosts for replicas of the same Executor
- [x] expose this option to the user
- [x] make it work in combination with passing the entire address to `host`
- [ ] what about shards?
- [x] testing
  - [x] unit tests for new helpers
  - [x] simple case
    - [x] passing `host` an `port` separately
    - [x] parsing whole schema from `host`~~ separate bug
  - [x] using docker
- [x] Documentation
  - [x] User guide, including distinction between external and shared executor
  - [x] docstring


docs/fundamentals/flow/executor-args.md
```diff
@@ -12,7 +12,7 @@
 | `uses_metas` | Dictionary of keyword arguments that will override the `metas` configuration in `uses` | `object` | `None` |
 | `uses_requests` | Dictionary of keyword arguments that will override the `requests` configuration in `uses` | `object` | `None` |
 | `py_modules` | The customized python modules need to be imported before loading the executor<br><br>Note that the recommended way is to only import a single module - a simple python file, if your<br>executor can be defined in a single file, or an ``__init__.py`` file if you have multiple files,<br>which should be structured as a python package. For more details, please see the<br>`Executor cookbook <https://docs.jina.ai/fundamentals/executor/executor-files/>`__ | `array` | `None` |
-| `port` | The port for input data to bind to, default is a random port between [49152, 65535] | `number` | `random in [49152, 65535]` |
+| `port` | The port for input data to bind to, default is a random port between [49152, 65535] | `string` | `random in [49152, 65535]` |
```
JoanFM, line 15: are we sure? isn't this quite breaking?

jina/orchestrate/deployments/__init__.py
```diff
@@ -251,6 +251,8 @@ def __init__(
             needs or set()
         )  #: used in the :class:`jina.flow.Flow` to build the graph
 
+        self.ext_repl_ports = getattr(args, 'external_replica_ports', None)
```
JoanFM, line 254: why would args have these values?

jina/orchestrate/flow/base.py
```diff
@@ -1181,6 +1195,13 @@ def add(
                 'External Executors with multiple needs have to do auto reduce.'
             )
 
+        args.external_replica_ports = make_iterable(_parse_ports(str(args.port)))
```
JoanFM, line 1198: do not pass this as `args` to `Deployment`. `Deployment` itself should do this parsing

JoanFM (changes requested):

docs/fundamentals/flow/executor-args.md
JohannesMessner: You can still pass a string and it will just get converted, so no breaking change from the user perspective.
This is necessary to be able to pass multiple ports for multiple replicas, and this follows the same syntax that was introduced by monitoring (where you pass multiple monitoring ports).

jina/orchestrate/flow/base.py
JohannesMessner: Other args like workspace and random port assignment get also handled and changed here, then passed to the deployment as args, so I wanted to follow that pattern.
But I can of course change it if you want.

JoanFM: but this seems like `new` arguments, that do not match `CLI` params

JohannesMessner: fair enough, will change it

jina/orchestrate/deployments/__init__.py
```diff
@@ -282,6 +306,13 @@ def update_pod_args(self):
         else:
             self.pod_args = self._parse_args(self.args)
 
+        if self.external:
```
JoanFM, line 309: What does this do?

jina/orchestrate/flow/base.py
```diff
@@ -1181,6 +1195,20 @@ def add(
                 'External Executors with multiple needs have to do auto reduce.'
             )
 
+        # args.external_replica_ports = make_iterable(_parse_ports(str(args.port)))
+        # args.external_replica_hosts = make_iterable(_parse_hosts(str(args.host)))
```
JoanFM, line 1199: remember to remve if needed

tests/integration/external_deployment/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM jinaai/jina:master
```
JoanFM, line 1: `test-pip`

JoanFM, line 1:
```suggestion
FROM jinaai/jina:test-pip
```

tests/integration/external_deployment/test_external_deployment.py
```diff
@@ -339,5 +381,46 @@ def foo(self, docs, **kwargs):
         with f:
             response = f.post(on='/', inputs=d, target_executor='external_executor')
 
-
     assert response[0].text == 'external'
+
+
+@pytest.mark.parametrize('hosts', ['localhost,localhost', 'localhost'])
```
JoanFM, line 387: External deployment and external replicas should be considered 2 different features and in my opinion tested differently in different test folders

JoanFM (changes requested):

tests/integration/external_deployment/test_external_deployment.py
JohannesMessner: ok will refactor

jina/orchestrate/deployments/__init__.py
JohannesMessner: When the deployment is not external then we do not expect the user to pass ports and hosts. Therefore we should not re-assign the port and host already associated with a given pod

docs/fundamentals/flow/topologies.md
```diff
@@ -160,6 +160,28 @@ You can also restrict the visible devices in round-robin assignment by assigning
...
+:class: seealso
+For more information about starting Executors outside of a Flow, see our {ref}`how-to on external Executors <external-executor>`.
+````
+
+Then, you can add them by specifying their hosts, ports, and `external=True`:
+
+```python
+from jina import Flow
+
+Flow().add(host='localhost,91.198.174.192', port='12345,12346', external=True)
```
JoanFM, line 179: is there a chance to enable the syntax:

```python
Flow().add(host='localhost:12345,91.198.174.192:12345')
```

JoanFM (changes requested):

JoanFM (dismissed):
If tests, pass GO

docs/fundamentals/executor/executor-serve.md
```diff
@@ -9,6 +9,24 @@ There are different options for deploying and running a standalone Executor:
 * Run the static {meth}`~jina.serve.executors.BaseExecutor.to_kubernetes_yaml()` method to generate K8s deployment configuration files
 * Run the static {meth}`~jina.serve.executors.BaseExecutor.to_docker_compose_yaml()` method to generate a Docker Compose service file
 
+````{admonition} Served vs. shared Executor
+:class: hint
+
+In Jina there are two ways of running standalone Executors: *Served Executors* and *shared Executors*.
+
+- A **served Executor** is launched by one of the methods described here (`.serve()`, `to_kubernetes_yaml()`, or `to_docker_compose_yaml()`).
```
alaeddine-13, line 17: are you trying to reference a link ?

docs/fundamentals/executor/executor-serve.md
```diff
@@ -9,6 +9,24 @@ There are different options for deploying and running a standalone Executor:
...
+It resides behind a {ref}`Gateway <architecture-overview>` and can thus be directly accessed by a {ref}`Client <client>`.
```
alaeddine-13, line 18:
```suggestion
It resides behind a {ref}`Gateway <architecture-overview>` and thus, can be directly accessed by a {ref}`Client <client>`.
```

docs/fundamentals/executor/executor-serve.md
```diff
@@ -9,6 +9,24 @@ There are different options for deploying and running a standalone Executor:
...
+It can also be used as part of a Flow.
+
+- A **shared Executor** is launched using the [Jina CLI](../cli/index.rst) and does *not* sit behind a Gateway.
+It is intended to be used in one or more Flows.
+Because a shared Executor does not reside behid a Gataway, it can not be directly accessed by a Client, but it requires
```
alaeddine-13, line 23:
```suggestion
Because a shared Executor does not reside behind a Gataway, it cannot be directly accessed by a Client, but it requires
```

docs/fundamentals/flow/add-executors.md
```diff
@@ -94,17 +94,29 @@ f = Flow(extra_search_paths=['../executor']).add(uses='config1.yml').add(uses='c
...
 
 ```python
 from jina import Flow
 
 Flow().add(host='123.45.67.89', port=12345, external=True)
 ```
 
-This is adding an external Executor to the Flow. The Flow will not start or stop this Executor and assumes that is externally managed and available at `123.45.67.89:12345`
+This is adding an external Executor to the Flow.
+The Flow will not start or stop this Executor and assumes that is externally managed and available at `123.45.67.89:12345`.
```
alaeddine-13, line 119:
```suggestion
The Flow will not start or stop this Executor and assumes that it is externally managed and available at `123.45.67.89:12345`.
```

jina/clients/__init__.py
```diff
@@ -36,7 +36,7 @@ def Client(
     """Create a Client. Client is how user interact with Flow
 
     :param asyncio: If set, then the input and output of this Client work in an asynchronous manner.
-    :param host: The host address of the runtime, by default it is 0.0.0.0.
+    :param host: The host address of the runtime, by default it is 0.0.0.0. In the case of an external Executor (`--external` or `external=True`) this can be a list of hosts, separated by commas. Then, every resulting address will be considered as one replica of the Executor.
```
alaeddine-13, line 39: is the new description relevant to client ? or is it just auto-generated ?

jina/orchestrate/deployments/__init__.py
```diff
@@ -265,6 +283,48 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
+
+    def _parse_external_replica_hosts_and_ports(self):
+        # splits user provided lists of hosts and ports into a host and port for every distributed replica
+        ext_repl_ports = make_iterable(_parse_ports(str(self.args.port)))
+        ext_repl_hosts = make_iterable(_parse_hosts(str(self.args.host)))
+        if len(ext_repl_hosts) < len(ext_repl_ports):
+            if (
+                len(ext_repl_hosts) == 1
+            ):  # only one host given, assume replicas are on the same host
+                ext_repl_hosts = ext_repl_hosts * len(ext_repl_ports)
```
alaeddine-13, line 312: else raise error ?

jina/orchestrate/deployments/__init__.py
```diff
@@ -265,6 +283,48 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
+        elif len(ext_repl_hosts) > len(ext_repl_ports):
+            if (
+                len(ext_repl_ports) == 1
+            ):  # only one port given, assume replicas are on the same port
+                ext_repl_ports = ext_repl_ports * len(ext_repl_hosts)
```
alaeddine-13, line 317: same , raise error in else

jina/parsers/orchestrate/runtimes/remote.py
```diff
@@ -145,7 +145,9 @@ def _add_host(arg_group):
         '--host',
         type=str,
         default=__default_host__,
-        help=f'The host address of the runtime, by default it is {__default_host__}.',
+        help=f'The host address of the runtime, by default it is {__default_host__}.'
+        ' In the case of an external Executor (`--external` or `external=True`) this can be a list of hosts, separated by commas.'
+        ' Then, every resulting address will be considered as one replica of the Executor.',
```
alaeddine-13, line 150: I find it weird that other components like the client share the exact same arg here. The extra description doesn't really apply to the client. I think we can either follow pattern [here](https://github.com/jina-ai/jina/blob/master/jina/parsers/__init__.py#L91) or make clone this parameter in client args with different description.
PS: we can handle this in separate PR especially if we have many more similar args

jina/clients/__init__.py
JohannesMessner: It is auto generated, due to our parser structure I don't think that there is a way to disentangle these things.

jina/parsers/orchestrate/runtimes/remote.py
JohannesMessner: I think this is an issue in many places. I am in favour of tackling this separately

docs/fundamentals/executor/executor-serve.md
JohannesMessner: no

alaeddine-13: okay
```suggestion
- A **served Executor** is launched by one of the following methods: `.serve()`, `to_kubernetes_yaml()`, or `to_docker_compose_yaml()`.
```

JoanFM (approved):

PR #5224, opened 2022-09-28
refactor: separate gateway and asyncio runtime readiness checks

alaeddine-13:
Currently, we only define readiness check in AsyncNewLoopRuntime which just applies a grpc health check test.
However, the GatewayRuntime which can also be an http or websocket server, does not override this behavior.
We override this and refactor code in AsyncNewLoopRuntime in order to correctly re-use this function

### DIFF
--- a/jina/serve/runtimes/gateway/__init__.py
+++ b/jina/serve/runtimes/gateway/__init__.py
@@ -1,5 +1,7 @@
 import argparse
+import urllib
 from abc import ABC
+from http import HTTPStatus
 from typing import TYPE_CHECKING, Optional, Union
 
 from jina.serve.runtimes.asyncio import AsyncNewLoopRuntime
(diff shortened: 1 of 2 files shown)


jina/serve/runtimes/gateway/__init__.py
```diff
@@ -28,3 +30,54 @@ def __init__(
         if self.timeout_send:
             self.timeout_send /= 1e3  # convert ms to seconds
         super().__init__(args, cancel_event, **kwargs)
+
+    @classmethod
```
JoanFM, line 34: staticmethods cannot be inherited?

alaeddine-13: I changed this, but wait_for_ready_or_shutdown has to stay classmethod

jina/serve/runtimes/gateway/__init__.py
```diff
@@ -28,3 +30,52 @@ def __init__(
...
+
+        :param timeout: The time to wait before readiness or failure is determined
+        :param ctrl_address: the address where the control message needs to be sent
+        :param ready_or_shutdown_event: the multiprocessing event to detect if the process failed or is ready
+        :param protocol: protocol of the gateway runtime
+        :param kwargs: extra keyword arguments
+
+        :return: True if is ready or it needs to be shutdown
+        """
+        return super().wait_for_ready_or_shutdown(
```
JoanFM, line 76: I do not understand, why do u need to override?

JoanFM (changes requested):

JoanFM (approved):

Issue #2527, opened 2021-06-01
Flow is hanging when using multi-threading


Yongxuanzhang:
**Describe the bug**

Flow is hanging when adding RuntimeBackendType.THREAD into one executor in the flow.
![image](https://user-images.githubusercontent.com/44033547/120252881-d1a11100-c253-11eb-8e42-bfa002044b9a.png)

**To reproduce it:**

```
from jina import Flow
from jina.enums import RuntimeBackendType

f = Flow().add(name='p1').add(name='p2').add(name='p3',needs='p2').add(name='p4', runtime_backend=RuntimeBackendType.THREAD).add(name='p5').add(name='p6',needs='p2').add(name='p7',needs=['p6','p5'])
f.plot()
with f:
    pass
```
---

**Environment**

jina==2.0.0rc2.dev57

**Screenshots**

Yongxuanzhang:
To address this, use it at Flow level instead:

```
from jina import Flow
from jina.enums import RuntimeBackendType

f = Flow(runtime_backend=RuntimeBackendType.THREAD).add(name='p1').add(name='p2').add(name='p3',needs='p2').add(name='p4').add(name='p5').add(name='p6',needs='p2').add(name='p7',needs=['p6','p5'])
f.plot()
with f:
    pass
```

JoanFM:
It is curious, if u remove the `last` pod it seems to work

jacobowitz:
Code to reproduce the issue without Jina:

```
import multiprocessing
import threading


def process_target():
    import slow_module
    print('process done')


def thread_target():
    import slow_module
    print(f'thread done')


p = multiprocessing.Process(target=process_target)
t = threading.Thread(target=thread_target)
t.start()
p.start()
t.join()
p.join()

print(f'done')
```

slow_module being something like this:
```
from time import sleep

sleep(1.0)
```

FionnD:
Closing this issue because it the feature will be added in #2527

PR #5246, opened 2022-10-06
feat(orchestrate): use spawn start method by default

girishc13:
**Goals:**

- resolves #ISSUE-NUMBER
- Spawn processes by default to avoid hanging flows as in #2527.
- ...

- ...
- ...


JoanFM (changes requested):
I would not do this, I would only add a `fixture` with general scope in `conftests` where we set `JINA_MP_START_METHOD=spawn` so that it is set by default

PR #5295, opened 2022-10-20
chore: remove unused class

JoanFM:


jina/jaml/parsers/__init__.py
```diff
@@ -22,8 +22,6 @@ def _get_all_parser(cls: Type['JAMLCompatible']):
         return _get_exec_parser()
     elif issubclass(cls, BaseGateway):
         return _get_gateway_parser()
-    else:
```
alaeddine-13, line 25: else, raise error.
Ideally, each JAML compatible class should select its parser rather than having to add code to select it (but not so important for now)

PR #5314, opened 2022-10-26
refactor: rename request handler objects

Jackmin801:
**Goals:**
- [x] Address https://github.com/jina-ai/jina/issues/5307
- [x] check and update documentation? I believe the Request Handlers are never mentioned in the docs.


jina/serve/runtimes/request_handlers/data_request_handler.py
```diff
@@ -18,7 +18,7 @@
     from jina.logging.logger import JinaLogger
 
 
-class DataRequestHandler:
+class ExecutorRequestHandler:
```
JohannesMessner, line 21: I think `WorkerRequestHandler` would be even better, since internally we have the `WorkerRuntime` that handles Executors, and also uses this handler.

JohannesMessner (commented):
Should we also rename the files?

JoanFM (approved):

PR #5325, opened 2022-10-28
fix: nicer error message for rpc errors

### LINKED ISSUE #5136: GRPC keep alive need a nice error message
# Context

When the keep alive mechanism is triggered we need to catch the exception and display a nice error message

Jackmin801:
**Goals:**

- resolves #5136

### DIFF
--- a/jina/clients/base/grpc.py
+++ b/jina/clients/base/grpc.py
@@ -3,6 +3,7 @@ import json
 from typing import TYPE_CHECKING, Optional
 
 import grpc
+from grpc import RpcError
 
 from jina.clients.base import BaseClient
 from jina.clients.helper import callback_exec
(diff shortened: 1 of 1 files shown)


Jackmin801:
Main issue is that all `RpcError`s have `.details()` except for https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_channel.py#L361-L453

I have no idea when that one is triggered though. We can catch it in the line above the `RpcError` catch to explicitly deal with it but I wouldn't know what information would be useful in the error message.

JoanFM (approved):

Jackmin801:
Oh and the way I reproduce the error is to run a gateway on a different computer in my network then try to ping it with the CLI:

`jina ping flow grpc://192.168.100.56:12345 --attempts 10`

Local pings to local executors aren't able to produce the `"keepalive watchdog timeout"` error because the error is thrown when a request is mid-flight then suddenly stops replying without closing the connection. If the ping and executor live in the same network namespace, this is not possible as the connection will always be known as closed by the pinger.

samsja (approved):
lgtm, sometime the best solution only requires a couple of line of code !

Issue #5033, opened 2022-07-27
Allow dynamic batching in Executor


JoanFM:
**Describe the feature**
We have seen that the simple `Request-Response` model with a static `request_size` is limiting in some cases. For instance some `Encoders` would rather accumulate `requests` until a big `enough` batch is accumulated and then all encoded at once. This could be used to improve `throughput` at a potential cost of `latency`.

How can I imagine the Feature:

In `executor.py`:

```python
from jina import Executor, requests, dynamic_batch

class DynamicBatchExecutor(Executor):

    @requests
    @dynamic_batch(batch_size=...) # potentially other arguments
     def foo(self, docs, **kwargs): 
           # decorator makes sure X docs arrive here (X being batch size)
           pass
```

Should the `Executor` do it or allowed to be passed dynamically in the `Runtime`? To be decided as it is more of a UI issue

This would affect the `Req-Response` model in the gateway. How I would solve it without affecting the `Request-Response` model from the client?

- [ ] In our proto, have some field in the `Response` stating that Response is yet to be served.
- [ ] The gateway keeps a map with the `request_id` and the `doc_ids` present there. 
- [ ] The gateway waits for the next Response that is served from that node, and then fires the continuation of all the waiting `Requests` with the appropiate `Documents` attached.

This way, we achieve.

- [ ] Dynamic batching
- [ ] The rest of the Flow and Executors do not realize about this. 
- [ ] The interface with the Client is kept. Single `request-response` model is kept from user perspective

Small hidden implementation details:
- How to deal with the last hanging docs.
- Imagine I have dynamic_batch_size of 50 and I just send 102 Documents. How to handle the 2 remaining Docs? I just have a timeout or something?

mister-fahrenheit:
@JoanFM @JohannesMessner hey!
I am very interested in this feature. Can you tell, what is the progress here?

Also, can you add `timeout` parameter for `@dynamic_batch` which tells Executer to process all pending requests if some of them is older than `timeout` and their amount lower than `batch_size`?

JohannesMessner:
I am currently finishing off some other work and will start with the implementation of this feature this week.
And yes, we will most likely have a mechanism that allows non-complete batches to be processed, a `timeout` seems reasonable.

@mister-fahrenheit do you have any other requirements or wishes for this feature? Anything about your use case that you would want to share? We are always happy to get inspiration from our users for features like this!

JohannesMessner:
[...]

PR #5340, opened 2022-11-01
feat: dynamic batching

JohannesMessner:
allow dynamic batching for executors. Feature described in detail [here](https://github.com/jina-ai/jina/issues/5033).

## ToDo

- [x] create the decorator
- [ ] make overridable in YAML
- [ ] implement queue and batching in the worker
- [ ] implement waiting logic in the gateway
- [ ] documentation
- [ ] tests
- [ ] **Question:** what about mis-matching tensor shapes? let the executors handle it? let the decorator define an expected shape? Do smart batching based on the shape? (I like the last one)

### DIFF
--- a/jina/serve/runtimes/request_handlers/worker_request_handler.py
+++ b/jina/serve/runtimes/request_handlers/worker_request_handler.py
@@ -168,7 +168,7 @@ class WorkerRequestHandler:
 
     @staticmethod
     def _parse_params(parameters: Dict, executor_name: str):
-        parsed_params = parameters
+        parsed_params = parameters # TODO: Copy? This still points to the same object
         specific_parameters = parameters.get(executor_name, None)
         if specific_parameters:
             parsed_params.update(**specific_parameters)
@@ -266,6 +266,10 @@ class WorkerRequestHandler:
         :param tracing_context: Optional OpenTelemetry tracing context from the originating request.
         :returns: the processed message
         """
+        # JOAN: This true? Why list?
+        # TODO: Just leave this then run all the tests
+        assert len(requests) == 1, "There are places where the handler is called with more than one request __apparently__"
+
         # skip executor if endpoints mismatch
         if (
             requests[0].header.exec_endpoint not in self._executor.requests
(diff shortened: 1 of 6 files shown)


JoanFM:
> allow dynamic batching for executors. Feature described in detail [here](https://github.com/jina-ai/jina/issues/5033).
> 
> ## ToDo
> * [x]  create the decorator
> * [ ]  make overridable in YAML
> * [ ]  implement queue and batching in the worker
> * [ ]  implement waiting logic in the gateway
> * [ ]  documentation
> * [ ]  tests
> * [ ]  **Question:** what about mis-matching tensor shapes? let the executors handle it? let the decorator define an expected shape? Do smart batching based on the shape? (I like the last one)

As per the question, in our case batching is done at DocArray level so decotator does nothing with respect to the tensor for now.

A question from me: Have we researched the Triton behavior?

JoanFM:
Make overridable in YAML means that there will be `runtime_args` that will make it overridable? I would suggest that we make it overridable even without the `decorator` so that we do not need to modify Hub Executors to use this behavior

jina/serve/runtimes/request_handlers/worker_request_handler.py
```diff
@@ -266,6 +266,10 @@ async def handle(
         :param tracing_context: Optional OpenTelemetry tracing context from the originating request.
         :returns: the processed message
         """
+        # JOAN: This true? Why list?
+        # TODO: Just leave this then run all the tests
+        assert len(requests) == 1, "There are places where the handler is called with more than one request __apparently__"
```
JoanFM, line 271: Sometimes if u have a Flow like this u will see:

```python
from jina import Flow, DocumentArray
with Flow().add(name='1').add(name='2', needs=['gateway']).add(name='3', needs=['1', '2'], no_reduce=True) as f:
   f.post(on='/', inputs=DocumentArray.empty(3))
```

JohannesMessner:
> Make overridable in YAML means that there will be `runtime_args` that will make it overridable? I would suggest that we make it overridable even without the `decorator` so that we do not need to modify Hub Executors to use this behavior

definitely, it should work even if there is no decorator.

PR #5341, opened 2022-11-01
fix: new timer should keep labels

Jackmin801:
**Goals:**

- [x] Add tests that would have caught the bug
- [x] Implement fix

The current decorator behaviour for MetricsTimer does not correctly save the labels assigned to the histogram.
This PR fixes this.
An example of the problem is in the tests.
The previous implementation would fail on the line:
```python
assert labels == histogram_metric['data']['data_points'][0]['attributes']
```


girishc13 (dismissed):

JoanFM:
@Jackmin801 do we need to add some tests?

JoanFM (approved):

girishc13 (approved):

Issue #5215, opened 2022-09-27
Custom Gateway: expose gateway api in flow api as a separate set of args


alaeddine-13:
As part of the Custom Gateway feature, we want to avoid confusion between gateway args and flow args.
We need to separate the concerns between gateway API and Flow API where the common parameter `uses` can be specified for both independently.
However, we do not want to introduce a breaking change and aim to keep existing gateway parameters at flow API and merge between those common gateway args and flow args in the following order:
1. gateway api args
2. flow api args
3 .default gateway api args
4. default flow api args
to be also added in flow yaml like this:
```jtype: Flow
... flow args
gateway:
    uses:
    ... # gateway args
executors:
    ...
```

PR #5342, opened 2022-11-02
feat: add gateway API

alaeddine-13:
As part of the Custom Gateway feature, we want to avoid confusion between gateway args and flow args.
We need to separate the concerns between gateway API and Flow API where the common parameter `uses` can be specified for both independently.
However, we do not want to introduce a breaking change and aim to keep existing gateway parameters at flow API and merge between those common gateway args and flow args.

The gateway API is added using the config_gateway method:
```python
f = Flow().config_gateway(uses='MyCustomGateway', port=12345)
```
and at yaml level:
```yaml
jtype: !Flow
gateway:
    uses: gateway.yml
    port: 12345
...
```
Since we don't want to introduce a breaking change, we aim to keep existing gateway parameters at flow API.
This means the gateway parameters inherit some valid Flow parameters and overrides them if specified (in the Gateway API):
```python
# ignores gateway params and uses gateway params instead
f = Flow(port=12344, protocol='http'').config_gateway(port=12345, protocol='grpc')
```

closes: https://github.com/jina-ai/jina/issues/5215

todos:

- [x] Flow.config_gateway
- [x] add gateway to json schema
- [x] update json schema tests
- [x] test custom gateway in Flow API (python class + yaml config + yaml and nested gateway yaml + docker ) in tests/unit/orchestrate/flow
- [ ] ~~fix the cli overload script and overload minimal args for each interface (flow) + add cli overload for the gateway method~~ document the inheritence of args from flow to executors behavior (tbd in https://github.com/jina-ai/jina/issues/5353)
- [x] Add `config_gateway` to cli overload script
- [x] yaml config tests in tests/unit/test_yaml_parser.py (yaml + nested gateway yaml)
- [x] Flow.save_config and Flow.load_config
- [x] Adapt k8s tests to new API
- [x] Adapt docker-compose tests to new API
- [x] check whether Flow API doesn't use Flow(uses=custom_gateway) anymore
- [x] pass CI
- [x] TODOs in PR


JoanFM:
Please add a good description

jina/parsers/__init__.py
```diff
@@ -3,6 +3,8 @@
 from jina.parsers.orchestrate.runtimes.container import mixin_container_runtime_parser
 from jina.parsers.orchestrate.runtimes.head import mixin_head_parser
 
+GATEWAY_NAME = 'gateway'
```
JoanFM, line 6: add this to some `helper` and use it everywhere in the code

JoanFM (changes requested):

jina/jaml/parsers/flow/v1.py
```diff
@@ -121,13 +122,21 @@ def dump(self, data: 'Flow') -> Dict:
             for t in _get_taboo(parser):
                 if t in kwargs:
                     kwargs.pop(t)
-            if k == 'gateway':
-                if 'JINA_FULL_CLI' in os.environ:
-                    r['with'].update(kwargs)
-                else:
-                    continue
-            else:
+            if k != 'gateway':
```
JoanFM, line 125: let's have a constant variable to hold `gateway` so that we can use everyehere

JoanFM (approved):
Just a minor comment

jina/schemas/__init__.py
```diff
@@ -1,16 +1,20 @@
+from jina.schemas.gateway import schema_gateway
```
numb3r3, line 1: is there any special reason to import the schema_gateway at the begining, rather than inside the `get_full_schema()`?

alaeddine-13: no good reason, thanks for catching this

jina/parsers/__init__.py
alaeddine-13: done

tests/unit/yaml/test-flow-custom-gateway-nested-config.yml
```diff
@@ -1,12 +1,7 @@
 !Flow
-with:
-  protocol: grpc
-  port: 12345
 gateway:
   uses: test-custom-gateway.yml
```
JoanFM, line 3: does it have `with` inside?

scripts/inject-cli-as-overload.py
```diff
@@ -278,6 +278,16 @@ def fill_implementation_stub(
         overload_fn='Client',
         class_method=False,
     ),
+    dict(
+        cli_entrypoint='gateway',
+        doc_str_title='Configure the Gateway inside a Flow. The Gateway exposes your Flow logic as a service to the internet according to the protocol and configuration you choose.',
+        doc_str_return='the new Flow object',
```
JoanFM, line 284: are we sure about this?

alaeddine-13: why not ? to me it makes sense because we do it for `Flow.add` and `Flow.__init__` : 
* args come from cli
* need to be maintained with a bot

JoanFM: doc_str_return I mean

tests/unit/yaml/test-flow-custom-gateway-nested-config.yml
JoanFM: then let's remove the `uses-with` right?

scripts/inject-cli-as-overload.py
alaeddine-13: Yes sure about this because we want to be able to chain methods

tests/unit/yaml/test-flow-custom-gateway-nested-config.yml
alaeddine-13: removed

JoanFM (approved):

PR #5344, opened 2022-11-03
feat: add deployment label in grpc stub metrics

### LINKED ISSUE #5154: Metrics on the gateway should always contained a label based on the Executor target
# Context

At the moment the gateway average all of its metrics and does not allow users to make distinction between different Executor.

To give an example:

Lets say we have a flow with two executors, one takes a lot of time to process the other one take few times.

At the moment the `jina_sending_request_seconds` metrics for instance will just average both of them

Actually this was not an issue initially since you can actually do the separation at the at the Executor level but the issue is with external Executor where you will have metrics coming from all the different flow.

# Proposition

Therefore we should rather measure the metrics on gateway level depending on which Executor the request is send.

Note: telemetry and tracing will also fix this problem

Jackmin801:
**Goals:**

- Resolves #5154
- [x] Handle connection reset case
- [x] Add regression tests

### DIFF
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -112,7 +112,7 @@ class ReplicaList:
         handling of race condition if multiple callers reset a connection at the same time.
 
         :param address: Target address of this connection
-        :param address: Target deployment of this connection
+        :param deployment_name: Target deployment of this connection
         :returns: The reset connection or None if there was no connection for the given address
         """
         self._logger.debug(f'resetting connection to {address}')
@@ -146,7 +146,7 @@ class ReplicaList:
         """
         Add connection with address to the connection list
         :param address: Target address of this connection
-        :param address: Target deployment of this connection
+        :param deployment_name: Target deployment of this connection
         """
         if address not in self._address_to_connection_idx:
             self._address_to_connection_idx[address] = len(self._connections)
@@ -980,6 +980,7 @@ class GrpcConnectionPool:
         },  # same deployment can have multiple addresses (replicas)
         total_num_tries: int = 1,  # number of retries + 1
         current_address: str = '',  # the specific address that was contacted during this attempt
+        current_deployment: str = '',  # the specific deployment that was contacted during this attempt
         connection_list: Optional[ReplicaList] = None,
     ):
         # connection failures, cancelled requests, and timed out requests should be retried
@@ -1005,7 +1006,7 @@ class GrpcConnectionPool:
             # after connection failure the gRPC `channel` gets stuck in a failure state for a few seconds
             # removing and re-adding the connection (stub) is faster & more reliable than just waiting
             if connection_list:
-                await connection_list.reset_connection(current_address)
+                await connection_list.reset_connection(current_address, current_deployment)
 
             raise InternalNetworkError(
                 og_exception=error,
@@ -1067,6 +1068,7 @@ class GrpcConnectionPool:
                         tried_addresses=tried_addresses,
                         total_num_tries=total_num_tries,
                         current_address=current_connection.address,
+                        current_deployment=current_connection.deployment_name,
                         connection_list=connections,
                     )
 
@@ -1108,6 +1110,7 @@ class GrpcConnectionPool:
                         retry_i=i,
                         tried_addresses=tried_addresses,
                         current_address=connection.address,
+                        current_deployment=connection.deployment_name,
                         connection_list=connection_list,
                         total_num_tries=total_num_tries,
                     )
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -64,29 +64,28 @@ class _NetworkingHistograms:
     send_requests_bytes_metrics: Optional['Histogram'] = None
     histogram_metric_labels: Dict[str, str] = None
 
-    def record_sending_requests_time_metrics(self, value: int, additional_labels: Optional[Dict[str, str]] = None):
+    def get_labels(self, additional_labels: Optional[Dict[str, str]] = None):
+        # TODO: This is quite ugly. Surely there's a better way
+        if self.histogram_metric_labels is None:
+            return None
         if additional_labels is None:
-            labels = self.histogram_metric_labels
-        else:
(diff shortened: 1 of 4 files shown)


jina/serve/networking.py
```diff
@@ -315,15 +318,21 @@ def __init__(
...
             self.address = address
             self.channel = channel
+            self.deployment_name = deployment_name
             self._metrics = metrics
             self._histograms = histograms
             self._initialized = False
 
+            if self._histograms is not None:
+                if self._histograms.histogram_metric_labels is not None:
+                    self._histograms.histogram_metric_labels.update(deployment_name=deployment_name)
```
Jackmin801, line 334: Doing update on the histogram_metric_labels might be wrong because it is passed by reference. Make sure to check this

jina/serve/networking.py
```diff
@@ -393,73 +442,40 @@ async def send_requests(
...
-                            self._histograms.record_received_response_bytes(
-                                response.nbytes
-                            )
-
+                            self._record_received_bytes_metric(response.nbytes)
                             return response, None
+
             if request_type == DataRequest and len(requests) > 1:
+                # TODO: This section of the code has no observability
+                # TODO: Does this section not need to deal with streaming stubs?
```
JoanFM, line 470: I think this should be handled as well, perhaps open another ticket and see if it needs handling.

I believe this code is reached when there is a Flow like this:

```python
f = Flow().add(name='name1').add(name='name2', needs=['gateway']).add(name='name3', needs=['name1', 'name2'], disable_reduce=True)
```

then Gateway needs to send 2 requests to `name3` without reducing before.

JoanFM (changes requested):

jina/serve/networking.py
```diff
@@ -315,15 +330,23 @@ def __init__(
...
             histograms: _NetworkingHistograms,
         ):
             self.address = address
             self.channel = channel
+            self.deployment_name = deployment_name
             self._metrics = metrics
             self._histograms = histograms
             self._initialized = False
 
+            if self._histograms:
```
girishc13, line 344: No need for an `else` statement?

girishc13: I guess the `self._histograms` is always checked before accessing the `self.stub_specific_labels`.

girishc13 (commented):
Good work. We don't mention the available labels in the docs so I'm not sure if we need to update the docs.

jina/serve/networking.py
Jackmin801: Yeap. I think that's the better pattern rather than having a default None.

Jackmin801: That way it's obvious that self.stub_specific_labels is coupled to the existence of self._histograms.

jina/serve/networking.py
Jackmin801: Ticket opened: https://github.com/jina-ai/jina/issues/5359

girishc13 (dismissed):

JoanFM:
> Good work. We don't mention the available labels in the docs so I'm not sure if we need to update the docs.

@samsja, may u assist on whether docs need to be changed?

jina/serve/networking.py
```diff
@@ -64,19 +64,30 @@ class _NetworkingHistograms:
     send_requests_bytes_metrics: Optional['Histogram'] = None
     histogram_metric_labels: Dict[str, str] = None
 
-    def record_sending_requests_time_metrics(self, value: int):
+    def _get_labels(self, additional_labels: Optional[Dict[str, str]] = None):
```
samsja, line 67:
```suggestion
    def _get_labels(self, additional_labels: Optional[Dict[str, str]] = None) -> Optional[Dcjt[str, str]]:

```

jina/serve/networking.py
```diff
@@ -393,73 +442,40 @@ async def send_requests(
...
-                                )
-                            self._histograms.record_received_response_bytes(
-                                response.nbytes
-                            )
-
+                            self._record_received_bytes_metric(response.nbytes)
                             return response, None
+
             if request_type == DataRequest and len(requests) > 1:
+                # TODO: This section of the code has no observability
```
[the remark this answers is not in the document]
samsja: This is a weird part of the code. I am not sure where it is called tbh. In 99% of the case the len(requests) == 1 and I cannot find the case where it is more than 1

tests/unit/serve/instrumentation/test_gateway_metric_labels.py
```diff
@@ -0,0 +1,115 @@
...
+        tracing=False,
+        metrics=True,
+        metrics_exporter_host='localhost',
+        metrics_exporter_port=4317,
+        port=12345,
+    ).add(name='first_exec', uses=FirstExec).add(
+        name="second_exec", uses=SecondExec
+    ) as f:
+        f.post('/')
+        f.plot('flow.png')
```
samsja, line 27: we can remove this line I believe

samsja (changes requested):
Really good work. I added some comments

jina/serve/networking.py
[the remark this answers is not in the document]
samsja: @Jackmin801 what do u think is missing ?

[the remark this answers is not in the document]
JoanFM: an issue has been opened

tests/unit/serve/instrumentation/test_gateway_metric_labels.py
JoanFM: yes please remove this

Jackmin801: done

jina/serve/networking.py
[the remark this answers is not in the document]
Jackmin801: https://github.com/jina-ai/jina/issues/5359

samsja:
> u assist on whether docs need to be changed?

No need to add them

JoanFM (approved):

PR #5348, opened 2022-11-03
feat: add Secrets concept in Jina

JoanFM:
**Goals:**
Add `Secrets` concept to Jina


jina/helper.py
```diff
@@ -919,6 +919,41 @@ def is_valid_local_config_source(path: str) -> bool:
         return False
 
 
+secrets_regex_str = r'\${{\s?SECRETS\.[a-zA-Z0-9_]*\s?}}|\${{\s?secrets\.[a-zA-Z0-9_]*\s?}}'
+secret_name_regex_str = r'\.(.*?)(\s|\})'
+secrets_regex = re.compile(
+    secrets_regex_str
+)  # matches expressions of form '${{ SECRETS.var }}' or '${{ secrets.var }}'
```
JohannesMessner, line 926: If I am not wrong it also matches `'${{SECRETS.var}}'` or `'${{secrets.var}}'` (because of `\s?`). I think in the JAML parser we enforce the space, should we do the same here?

JohannesMessner, line 926: If I am not wrong it also matches `'${{SECRETS.var}}'` or `'${{secrets.var}}'` (because of `\s?`). I think in the JAML parser we enforce the space, should we do the same here?

jina/parsers/helper.py
```diff
@@ -55,6 +55,28 @@ def __call__(self, parser, args, values, option_string=None):
         setattr(args, self.dest, d)
 
 
+class ListofDictsAction(argparse.Action):
+    """argparse action to handle input and convert to List of Dictionaries
```
JohannesMessner, line 59: what does the input need to look like for this to work?

jina/helper.py
JoanFM: I do not really see why

jina/parsers/helper.py
JoanFM: I will work on Documentation later

PR #5351, opened 2022-11-03
chore: move setup requirements from setup.py to pyproject.toml

### LINKED ISSUE #5346: chore: DEPRECATION WARNING when installing Jina
```shell
DEPRECATION: jina is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```

Jackmin801:
**Goals:**

- resolves #5346

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=18.0", "wheel"]
+build-backend = "setuptools.build_meta"
--- a/setup.py
+++ b/setup.py
@@ -154,7 +154,6 @@ setup(
     long_description=_long_description,
     long_description_content_type='text/markdown',
     zip_safe=False,
-    setup_requires=['setuptools>=18.0', 'wheel'],
     install_requires=list(final_deps),
     extras_require=all_deps,
     entry_points={


Jackmin801:
I don't actually know if it works because I can't reproduce the deprecation warning. What I can see is that the first few lines in pip install have changed

Old:
image

New:
image

Jackmin801:
The first commit -- [fix: move the easy stuff to setup.cfg](https://github.com/jina-ai/jina/pull/5351/commits/e9844d3ddb0d2f424129b9ac0e8ed68fb24354d2) might also be unnecessary

setup.cfg
```diff
@@ -0,0 +1,49 @@
+[metadata]
+name = jina
+version = 3.11.1
```
JoanFM, line 3: we need to adapt our CI/CD to update this `version` then

JoanFM (changes requested):

setup.cfg
Jackmin801: yup

Jackmin801:
Hrmm the change broke some weird things. Let me try a smaller change. I think this might be enough to resolve the warning already. The first few lines of the pip install already use pyproject.toml

setup.cfg
Jackmin801: we can do this later. I think the current change of just moving the build configs is enough

Jackmin801:
Can confirm this fix works. Inside a python environment without `wheel` installed, the deprecation warning appears without the commit in this PR. With the commit in this PR, the deprecation warning does not appear.

JoanFM (approved):

PR #5367, opened 2022-11-09
fix: bootstrap multiprocessing inside main check to support spawn mp method

alaeddine-13:
It looks like we don't correctly follow [the python guidelines](https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods) when spawning new process for pods when the spawn method is `spawn`.

This means sometimes we still have issues even with `JINA_MP_START_METHOD=spawn` and we will receive:
```text
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...
```
For instance, one can reproduce the bug by running the following flow:
```
from jina import Flow

f = Flow().add()

with f:
    f.block()
```
then run
`JINA_MP_START_METHOD=spawn python flow.py`

currently we only support `JINA_MP_START_METHOD=spawn` if the flow is wrapped in
```python
if __name__ == '__main__':
    with f:
        f.block()
```

### DIFF
--- a/jina/orchestrate/pods/__init__.py
+++ b/jina/orchestrate/pods/__init__.py
@@ -4,6 +4,7 @@ import os
 import threading
 import time
 from abc import ABC, abstractmethod
+from multiprocessing import freeze_support
 from typing import Dict, Optional, Type, Union
 
 from jina import __ready_msg__, __stop_msg__, __windows__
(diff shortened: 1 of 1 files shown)


jina/orchestrate/pods/__init__.py
```diff
@@ -314,28 +315,31 @@ class Pod(BasePod):
...
-                'is_started': self.is_started,
-                'is_shutdown': self.is_shutdown,
-                'is_ready': self.is_ready,
-                'runtime_cls': self.runtime_cls,
-                'jaml_classes': JAML.registered_classes(),
-            },
-            name=self.name,
-            daemon=False,
-        )
+        if __name__ == '__main__':
```
JoanFM, line 318: add an else with an Exception then

alaeddine-13: not sure it should be an exception.
I think the principle is, if this is not the parent process, don't try to create the child process again from the child process itself. We should do nothing instead. We can surely improve the implementation I just try to confirm it

JoanFM: if we do not create `self.worker` everything will crash, so for sure we need to do Exception although I believe this is nothing that will happen, otherwise what u can do is to simply do:

```python
if __name__ == '__main__':
    freeze_support()
#rest of things
```

this way u just do the freeze when it is main and act normally elsewhere

jina/orchestrate/pods/__init__.py
```diff
@@ -314,28 +321,31 @@ class Pod(BasePod):
...
+                daemon=False,
+            )
 
     def start(self):
         """Start the Pod.
         This method calls :meth:`start` in :class:`threading.Thread` or :class:`multiprocesssing.Process`.
         .. #noqa: DAR201
         """
-        self.worker.start()
+        if _is_main_and_spawn():
```
JoanFM, line 347: leave as it is

jina/orchestrate/pods/__init__.py
```diff
@@ -350,15 +360,17 @@ def join(self, *args, **kwargs):
         :param kwargs: extra keyword arguments to pass to join
         """
         self.logger.debug(f'joining the process')
-        self.worker.join(*args, **kwargs)
+        if _is_main_and_spawn():
+            self.worker.join(*args, **kwargs)
```
JoanFM, line 364: do not change

jina/orchestrate/pods/__init__.py
```diff
@@ -350,15 +360,17 @@ def join(self, *args, **kwargs):
...
         self.logger.debug(f'successfully joined the process')
 
     def _terminate(self):
         """Terminate the Pod.
         This method calls :meth:`terminate` in :class:`threading.Thread` or :class:`multiprocesssing.Process`.
         """
         self.logger.debug(f'terminating the runtime process')
-        self.worker.terminate()
+        if _is_main_and_spawn():
+            self.worker.terminate()
```
JoanFM, line 373: do not change

JoanFM (changes requested):

Issue #5272, opened 2022-10-11
Align on OpenTelemetry service names and cloud semantic attributes.


girishc13:
The current tracers are using the default runtime name or module name to create spans. These are currently very generic and don't differentiate between different flows. Without unique service names and other cloud deployment attributes ([k8s](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/k8s/), [cloud provider attributes](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#cloud-provider-specific-attributes)), viewing traces will become un-filterable and very messy. 

# Proposal

- Allow users to provide/override the deployment name using environment variables which is then set in the instrumentation library.
- Jina AI Cloud specific attributes need to be manually supported and documented.
- Self hosted cloud specific attributes need to be set by the user using environment variables as defined in the OpenTelemetry [documentation](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#cloud-provider-specific-attributes).

PR #5372, opened 2022-11-10
feat: k8s resource attributes in instrumentation

Jackmin801:
**Goals:**

- resolves partially #5272
- [x] implement logic to pick up envs in instrumentation setup
- [x] declare envs in templates
- [x] add k8s regression tests
- [x] Update prometheus exporter config about resource labels https://github.com/jina-ai/jina/pull/5409


jina/serve/instrumentation/__init__.py
```diff
@@ -11,6 +11,32 @@
     from prometheus_client import Summary
 
 
+def _get_resource_attributes(service_name: str):
+    from opentelemetry.semconv.resource import ResourceAttributes
+    import os
+
+    attributes = {ResourceAttributes.SERVICE_NAME: service_name}
+    if "K8S_NAMESPACE_NAME" in os.environ:
+        attributes[ResourceAttributes.K8S_NAMESPACE_NAME] = os.environ[
```
JoanFM, line 20: please use single quote

jina/serve/instrumentation/__init__.py
```diff
@@ -11,6 +11,32 @@
     from prometheus_client import Summary
 
 
+def _get_resource_attributes(service_name: str):
+    from opentelemetry.semconv.resource import ResourceAttributes
+    import os
+
+    attributes = {ResourceAttributes.SERVICE_NAME: service_name}
+    if "K8S_NAMESPACE_NAME" in os.environ:
```
JoanFM, line 19: you could handle this in a for loop with some small reflection?

JoanFM (changes requested):
small comments

JoanFM (changes requested):
Please use single quotes and add a test in k8s test

tests/k8s_otel/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+import os
+from pathlib import Path
+from pytest_kind import KindCluster, cluster
+
+from jina.logging.logger import JinaLogger
+from .kind_wrapper import KindClusterWrapperV2
+
+# The default version broke cni at some point. That's why we need to specify the version here.
+# This can and probably should be put in env variable actually.
+cluster.KIND_VERSION = 'v0.11.1'
```
girishc13, line 11: You can probably run the subprocess command to get the KIND version.

tests/k8s_otel/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    return KindClusterWrapperV2(kind_cluster, JinaLogger('kubernetes-cluster-logger'))
+
+
+@pytest.fixture(scope='session')
+def otel_test_namespace(k8s_cluster_v2: KindClusterWrapperV2) -> str:
+    """
+    Returns a namespace with the otlp-collector, jaeger and prometheus services deployed.
+
+    Service endpoints:
+        - otlp-collector:4317 - OTLP GRPC metrics endpoint
```
girishc13, line 36: You can get random ports from `jina.helper.random_port()` and export them as environment variables so the hard coded ones don't accidentally collide.

tests/k8s_otel/kind_wrapper.py
```diff
@@ -0,0 +1,215 @@
...
+from docker import DockerClient
+from pytest_kind import KindCluster
+from subprocess import CalledProcessError
+from contextlib import contextmanager
+from typing import Generator
+
+from jina.logging.logger import JinaLogger
+
+
+class KindClusterWrapperV2:
```
girishc13, line 16: I think this is overkill. For tests it might be easier to just run kubectl commands for deploying the otel components. The Python code adds overhead in terms of maintenance. You created a separate step anyway so it's isolated from the other k8s test and kind clustes.

tests/k8s_otel/conftest.py
Jackmin801: Ah, the port forward chooses a random port on the host so shouldn't be an issue. The hard coded ports here live on the kubernetes services in which case a collision is fine; they are distinguished by service name.

tests/k8s_otel/conftest.py
Jackmin801: Oh this isn't to read the kind version, it is to get it to download the right kind version. I took it from the other k8s tests.

I saw in the source code for pytest_kind that it actually reads it from env variable so we can set it in the CI instead of in the code. I think that might be better.

tests/k8s_otel/kind_wrapper.py
Jackmin801: I agree with you actually. The python code here does just use kubectl in a subprocess in the end so it's not too different. New kubernetes API versions would not require a change in the python code since it just uses kubectl which should be updated instead. 

The old version of this wrapper in the other k8s test uses the python kubernetes client which could have this overhead of requiring an update separate from kubectl.

Oh and we can't just deploy otel components in a bash script outside of python because the lifecycle of the cluster is tied to python. Also exception handling and logging will ultimately need to be in pytest. So the python code does need to be the orchestrator in this case.

Jackmin801: I did try to do it without the wrapper. But it makes you duplicate a lot of common routines like loading images and apply then wait deployment. So the wrapper is probably a necessary evil. But I think this wrapper is less evil than the old one. The old one even pulls in linkerd though it doesn't seem like that is necessary and just creates an extra maintenance overhead of having to deal with linkerd versions

jina/serve/instrumentation/__init__.py
```diff
@@ -10,6 +10,26 @@
...
+
+
+def _get_resource_attributes(service_name: str):
+    from opentelemetry.semconv.resource import ResourceAttributes
+    import os
+
+    attributes = {ResourceAttributes.SERVICE_NAME: service_name}
+    for attribute in ENV_RESOURCE_ATTRIBUTES:
+        if attribute in os.environ:
+            attributes[ResourceAttributes.__dict__[attribute]] = os.environ[attribute]
```
Jackmin801, line 30: This line should be covered in the new test. Is codecov not happy because it is covered in the containers and not in the local jina execution?

tests/k8s_otel/__init__.py
```diff
@@ -0,0 +1 @@
+from .kind_wrapper import KindClusterWrapperV2
```
JoanFM, line 1: please use absolute imports

tests/k8s_otel/conftest.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+import os
+from pathlib import Path
+from pytest_kind import KindCluster, cluster
+
+from jina.logging.logger import JinaLogger
+from .kind_wrapper import KindClusterWrapperV2
```
JoanFM, line 7: please use absolute imports

tests/k8s_otel/kind_wrapper.py
JoanFM, line 16: Why is this needed? what is it doing that we were not doing before in other tests?

tests/k8s_otel/test_k8s_instrumentation.py
```diff
@@ -0,0 +1,102 @@
...
+
+    flow.to_kubernetes_yaml(dump_path, k8s_namespace=NAMESPACE)
+
+    # Deploy flow
+    k8s_cluster_v2.deploy_from_dir(dir=dump_path, namespace=NAMESPACE)
+
+    # Make client requests
+    with k8s_cluster_v2.port_forward('svc/gateway', NAMESPACE, svc_port=8080) as gateway_port:
+        from jina import Client
+        [i async for i in Client(port=gateway_port, asyncio=True).post("/")]
```
JoanFM, line 43: do not use `i`, these are `docs`

JoanFM (changes requested):

tests/k8s_otel/kind_wrapper.py
Jackmin801: Hrmm. It does things we were doing in other tests in a more consolidated and simpler fashion. Most notably the following:

**Deploying artifacts from a directory**
Other tests: https://github.com/jina-ai/jina/blob/master/tests/k8s/test_k8s.py#L26-L121
This test: https://github.com/jina-ai/jina/pull/5372/files#diff-22c1d4e56c2c85e03d9fca28239f7aa37559fd8d449f8629c73135afcbb654c4R57-R82

1. The old method requires you to know the exact names of your deployments and supply them in `deployment_replicas_expected`. The new one infers them and thus does not need this argument.
2. The old method uses `api_client`, `app_client` and `core_client` from python kubernetes as arguments. This creates an unnecessary dependency on the python implemention of kubernetes client when these methods can simply be performed with kubectl. Despite having all the clients, the old method still needs a kubernetes utils inside itself.
3. If the deployment were to fail, the old method provides no observability into the state of the cluster.

**Linkerd**
Linkerd is installed unnecessarily in the old tests.
1. This create issues with installing it, as mentioned here: https://github.com/jina-ai/jina/blob/master/tests/k8s/conftest.py#L69-L70.
2. Installing Linkerd forces all tests and development of the tests to need to work inside its service mesh framework which adds about a few minutes to debug operations that could have been done in 5 seconds because more time was spent on the overhead of figuring out why the service mesh isn't happy with your cert than actually looking at your containers.  

If there is some linkerd integration that needs to be tested, they should not be tested with the other k8s test. Though as far as I am aware, everything would work just fine without it. The only place other than installing itself in which it is referred is https://github.com/jina-ai/jina/blob/master/tests/unit/orchestrate/deployments/config/test_k8s_deployment_config.py#L339. Which imo, is more a test that linkerd works than a test that jina works with it.

JoanFM: then why do not u just change all the tests into using this? 

I do not want this much duplication, let's choose 1 way and only 1 to do these kind of tests

Jackmin801: Yep agreed. So I port the old ones to use this in another PR?

PR #5374, opened 2022-11-10
fix: cover missing metrics in HeadRuntime

### LINKED ISSUE #5184: Catch up with the missing metrics in the Head
# Context

Some metrics are only available in the gateway and not in the head this should not be the case.

List of metrics impacted:

- [ ] jina_number_of_pending_requests
- [ ] jina_successful_requests_total
- [ ] jina_failed_requests_total

See the docs for the available metrics : https://docs.jina.ai/fundamentals/flow/monitoring-flow/

JoanFM:
**Goals:**
Some metrics should be in the Head as they are in the Gateway

- resolves #5184 

- [x] Refactor HeaderRequestHandler inheriting MonitoringMixin
- [x] Do the proper private calls
- [x] Add tests
- [x] Update docs


jina/serve/runtimes/monitoring.py
```diff
@@ -14,9 +29,208 @@ def _setup_monitoring(self):
...
+    def _update_start_request_metrics(self, request: 'Request'):
+        if self._request_size_metrics:
+            self._request_size_metrics.observe(request.nbytes)
+        if self._request_size_histogram:
+            self._request_size_histogram.record(
+                request.nbytes, attributes=self._metric_labels
+            )
+
+        if self._receiving_request_metrics:
+            self._request_init_time[request.request_id] = time.time()
```
girishc13, line 171: Not related to this PR but I want to mention it. This map is not always safely cleared after every request ends. Especially in the head `_gather_worker_tasks` is not catching all exceptions. Right now it's focused on the `AioRpcError` and `InternalNetworkError` for retrying. In case if something else happens then the the `_update_end_successful_requests_metrics` method will not correctly remove entries from the dict.

jina/serve/runtimes/monitoring.py
```diff
@@ -14,9 +29,208 @@ def _setup_monitoring(self):
...
+        if self._failed_requests_metrics:
+            self._failed_requests_metrics.inc()
+        if self._failed_requests_counter:
+            self._failed_requests_counter.add(1, attributes=self._metric_labels)
+
+    def _update_end_request_metrics(self, result: 'Request'):
+        if result.status.code != jina_pb2.StatusProto.ERROR:
+            self._update_end_successful_requests_metrics(result)
+        else:
+            self._update_end_failed_requests_metrics()
```
girishc13, line 236: Should we also pop the Request entry in this method?

jina/serve/runtimes/head/request_handling.py
```diff
@@ -0,0 +1,180 @@
...
+
+    from jina.logging.logger import JinaLogger
+    from jina.types.request.data import DataRequest
+
+
+class HeaderRequestHandler(MonitoringRequestMixin):
+    """
+    Class that handles the requests arriving to the head and the results extracted from the requests future.
+
+    :param metrics_registry: optional metrics registry for prometheus used if we need to expose metrics from the executor or from the data request handler
```
JohannesMessner, line 19:
```suggestion
    :param metrics_registry: optional metrics registry for prometheus. Used if we need to expose metrics from the executor or from the data request handler
```

JohannesMessner (approved):
LGTM, I like the new file structure / refactoring aspect

jina/serve/runtimes/monitoring.py
JoanFM: I am just moving this method around

jina/serve/runtimes/monitoring.py
JoanFM: let's open a ticket for this, I am not sure there can be other errors,it should be handled in another PR

Issue #5358, opened 2022-11-07
Move the old Jina/Docarray relation docs from DocArray documenation to the jina one


samsja:
# Context

DocArray is moving into its own organization and reference to the jina project are slowly beeing removed.

Therefore we will loose this documentation at some point in docarray :  https://docarray.jina.ai/fundamentals/jina-support/ 

We need to move to content to the Jina documentation as it is still relevant.

It is more than copy pasting, work need to be done in the working because it is not jina info in docarray but docarray info in jina. The content is mostly the same though

PR #5382, opened 2022-11-11
docs: add section on coming from docarray

samsja:
# Context

DocArray is moving into its own organization and reference to the jina project are slowly beeing removed.

Therefore we will loose this documentation at some point in docarray :  https://docarray.jina.ai/fundamentals/jina-support/ 

We need to move to content to the Jina documentation as it is still relevant.

original issue : https://github.com/jina-ai/jina/issues/5358

# what this PR do

It add a section on coming from docarray


docs/index.md
```diff
@@ -143,8 +143,10 @@ fundamentals/flow/index
 fundamentals/gateway/index
 fundamentals/client/client
 fundamentals/k8s
+fundamentals/docarray-support/index
```
JoanFM, line 146: I would not call `docarray-support` but `dependency`?

samsja: this is just the folder. I renamed it to `docarray-dependency` since it is just for docarray.

The title of the section is `coming from docarray` because it is using the title on the md files rather than the fodler name

JoanFM (approved):

PR #5383, opened 2022-11-11
fix: decrement requests count in case of exception inside HeadRuntime

JoanFM:
In case of some Exceptions, the count of pending requests and some stats related were not properly decremented


girishc13 (approved):
LGTM
Please update the title and description.

PR #5391, opened 2022-11-14
fix: use start_as_current_span to maintain tracing operation chain

### LINKED ISSUE #5386: bug: worker runtime request handler span appears out of order
**Describe the bug**

The worker runtime request handler starts the span to record the executor request using `trace.start_as_span` which doesn't sequence the span correctly if the executor requests method creates another span using the `tracing_context`.

**Describe how you solve it**

To correctly sequence the span order, the worker runtime request handler must start it's span using `trace.start_as_current_span`
---

**Environment**

**Screenshots**

girishc13:
**Goals:**

- resolves #5386 
- the python SDK for whatever reason doesn't consider the span started from the origin request context as the direct child span. This causes inconsistency when the trace is viewed.

Before:
before-start-as-current-span

After:

after-start-as-span


jina/serve/executors/__init__.py
```diff
@@ -379,7 +382,9 @@ async def exec_func(
         }
 
         if self.tracer:
-            with self.tracer.start_span(req_endpoint, context=tracing_context) as _:
+            with self.tracer.start_as_current_span(
+                req_endpoint, context=tracing_context
+            ) as _:
```
JoanFM, line 387:
```suggestion
            ):
```

JoanFM (changes requested):

tests/integration/instrumentation/test_worker_instrumentation.py
```diff
@@ -0,0 +1,44 @@
+import time
+
+from jina import Flow
+from tests.integration.instrumentation import ExecutorTestWithTracing, get_traces
+
+
+def test_span_order(jaeger_port, otlp_collector, otlp_receiver_port):
+    # cannot use the pythonic f = Flow(...) because an actual gateway process is not started
```
JoanFM, line 8: I do not understand this

JoanFM (changes requested):

tests/integration/instrumentation/test_worker_instrumentation.py
girishc13: This might be a bug or something that is part of the orchestrate layer but if I start as flow using:
```python
f = Flow(tracing=True, traces_exporter_host='localhost', traces_exporter_port=4317)

with f:
  f.post(...)
```

This doesn't produce the proper chain (v.s. flow.yml) of requests to the executor.

girishc13: I tested this again on my local machine and I cannot reproduce this issue. Maybe something was broken in the test that caused me to use teh `flow.yml`.

JoanFM: can u get back to the Pythonic way then?

girishc13: Yes, the code has already been replaced with the pythonic code.

JoanFM (approved):

PR #5392, opened 2022-11-14
fix: health check exception with opentelemetry tracing interceptors

girishc13:
The `async` tracing interceptors expect the RPC stubs to be a `coroutine` function otherwise an exception is thrown for the RPC. When tracing was enabled, the health checks with docker or k8s containers failed silently and the Flow timed out eventually. The underlying issue was the usage of the default `grpc_health.HealthServicer` implementation with an `async` grpc server.

Goals:
- use `aio.HealthServicer` and `await` keyword when calling the health servicer.
- add an instrumentation test with a docker container to address the fix.


jina/serve/instrumentation/_aio_server.py
```diff
@@ -60,6 +60,8 @@ def telemetry_wrapper(behavior, request_streaming, response_streaming):
         return _wrap_rpc_behavior(next_handler, telemetry_wrapper)
 
     def _intercept_aio_server_unary(self, behavior, handler_call_details):
+        from jina.helper import iscoroutinefunction
```
JoanFM, line 63: why is this needed?

JoanFM (changes requested):

jina/serve/instrumentation/_aio_server.py
girishc13: Not needed if we decide to use the `aio.HealthServicer` approach. I was just experimenting if the interceptors can be easily modified to accept any method.

jina/serve/runtimes/asyncio.py
```diff
@@ -182,6 +182,7 @@ def is_ready(ctrl_address: str, **kwargs) -> bool:
         :return: True if status is ready else False.
         """
 
+        # print('--->address', ctrl_address)
```
JoanFM, line 185: remember to remove

JoanFM (changes requested):

JoanFM (approved):

PR #5404, opened 2022-11-17
feat: add option to return in order in client and streamer

JoanFM:
**Goals:**
Try returning in the order of requests from Streamer

### DIFF
--- a/docs/fundamentals/client/client.md
+++ b/docs/fundamentals/client/client.md
@@ -760,8 +760,8 @@ None
 
 ````
 
-A Client is used to connect to a Flow who processes Documents in an asynchronous and very distributed way. This means that the order in which the Requests are processed may not be the exact same order as they are sent by the Client.
-However, you can force the order of the results to be deterministic and the same as they enter the Flow by passing `results_in_order` parameter to {meth}`~jina.clients.mixin.PostMixin.post`.
+A Client connects to a Flow that processes Documents in an asynchronous and very distributed way. This means that the order of the Flow processing the requests may not be the same order as the Client sending the requests.
+However, you can force the order of the results to be deterministic and the same as when they enter the Flow by passing `results_in_order` parameter to {meth}`~jina.clients.mixin.PostMixin.post`.
 
 ```python
 import random
--- a/docs/fundamentals/client/client.md
+++ b/docs/fundamentals/client/client.md
@@ -769,14 +769,14 @@ import time
 from jina import Flow, Executor, requests, Client, DocumentArray, Document
 
 
-class ExecutorRandomSleepExecutor(Executor):
+class RandomSleepExecutor(Executor):
 
     @requests
     def foo(self, *args, **kwargs):
         rand_sleep = random.uniform(0.1, 1.3)
         time.sleep(rand_sleep)
 
-f = Flow().add(uses=ExecutorRandomSleepExecutor, replicas=3)
+f = Flow().add(uses=RandomSleepExecutor, replicas=3)
 input_text = [f'ordinal-{i}' for i in range(180)]
 input_da = DocumentArray([Document(text=t) for t in input_text])
 with f:
--- a/docs/fundamentals/client/client.md
+++ b/docs/fundamentals/client/client.md
@@ -760,6 +760,34 @@ None
 
 ````
 
+A Client is used to connect to a Flow who processes Documents in an asynchronous and very distributed way. This means that the order in which the Requests are processed may not be the exact same order as they are sent by the Client.
+However, you can force the order of the results to be deterministic and the same as they enter the Flow by passing `results_in_order` parameter to {meth}`~jina.clients.mixin.PostMixin.post`.
+
+```python
+import random
+import time
+from jina import Flow, Executor, requests, Client, DocumentArray, Document
+
+
+class ExecutorRandomSleepExecutor(Executor):
+
+    @requests
+    def foo(self, *args, **kwargs):
+        rand_sleep = random.uniform(0.1, 1.3)
+        time.sleep(rand_sleep)
+
+f = Flow().add(uses=ExecutorRandomSleepExecutor, replicas=3)
+input_text = [f'ordinal-{i}' for i in range(180)]
+input_da = DocumentArray([Document(text=t) for t in input_text])
+with f:
+    c = Client(port=f.port, protocol=f.protocol)
+    output_da = c.post('/', inputs=input_da, request_size=10, results_in_order=True)
+    for input, output in zip(input_da, output_da):
+        assert input.text == output.text
+    
+```
+
+
 (client-compress)=
 ## Enable compression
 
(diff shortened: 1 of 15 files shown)


docs/fundamentals/client/client.md
```diff
@@ -760,6 +760,34 @@ None
 
 ````
 
+A Client is used to connect to a Flow who processes Documents in an asynchronous and very distributed way. This means that the order in which the Requests are processed may not be the exact same order as they are sent by the Client.
```
alexcg1, line 763:
```suggestion
A Client connects to a Flow that processes Documents in an asynchronous and very distributed way. This means that the order of the Flow processing the requests may not be the same order as the Client sending the requests.
```
I think a short example would be useful here. Parsing that text was not simple.

Wait, when we say "who processes Documents in an asynchronous and very distributed way" does that refer to client or flow?

Maybe just rewrite this bit. There's ambiguity

docs/fundamentals/client/client.md
```diff
@@ -760,6 +760,34 @@ None
...
+However, you can force the order of the results to be deterministic and the same as they enter the Flow by passing `results_in_order` parameter to {meth}`~jina.clients.mixin.PostMixin.post`.
```
alexcg1, line 764:
```suggestion
However, you can force the order of the results to be deterministic and the same as when they enter the Flow by passing `results_in_order` parameter to {meth}`~jina.clients.mixin.PostMixin.post`.
```

alexcg1 (changes requested):

docs/fundamentals/client/client.md
JoanFM: the Flow processes it in an asynchronous and distributed way

JoanFM: There is a test already

alexcg1 (approved):
LGTM 👍

jina/serve/stream/__init__.py
```diff
@@ -206,15 +219,43 @@ def iterating_task_done(task):
...
+                # if return_in_order is needed, we have to check if the response is to be returned
+                response_request_id = await DataRequest.extract_id(response)
+                if response_request_id == request_ids[0]:
+                    # is the response's turn to be yielded
+                    yield response
+                    request_ids.pop(0)
+                else:
+                    # is not your turn, put to the end
+                    responses_ids[response_request_id] = len(responses_list)
+                    responses_list.append(response)
```
alaeddine-13, line 247: I don't see a need for responses_list here. I think we can directly put the response in the responses_ids map (actually let'ts map from ids to response objects and rename the variable)

jina/serve/stream/__init__.py
```diff
@@ -206,15 +219,43 @@ def iterating_task_done(task):
...
+
+                stop_yielding = False
+                while not stop_yielding and len(request_ids) > 0:
+                    next_request_id = request_ids[0]
+                    if next_request_id in responses_ids:
+                        response_index = responses_ids[next_request_id]
+                        del responses_ids[next_request_id]
+                        request_ids.pop(0)
+                        yield responses_list[response_index]
```
alaeddine-13, line 256: assuming we no more need responses_list, we can change the code like so:
```suggestion
                        request_ids.pop(0)
                        yield responses_ids.pop(next_request_id)
```

jina/serve/stream/__init__.py
```diff
@@ -206,15 +219,43 @@ def iterating_task_done(task):
...
+                    request_ids.pop(0)
+                else:
+                    # is not your turn, put to the end
+                    responses_ids[response_request_id] = len(responses_list)
+                    responses_list.append(response)
+
+                stop_yielding = False
+                while not stop_yielding and len(request_ids) > 0:
+                    next_request_id = request_ids[0]
+                    if next_request_id in responses_ids:
```
alaeddine-13, line 252: what happens if the next_request id is not received ? we don't yield anymore ?

tests/integration/test_return_order/test_return_order.py
```diff
@@ -0,0 +1,29 @@
+from jina import Flow, Executor, requests, Document, DocumentArray, Client
+import random
+import time
+
+import pytest
+
+
+@pytest.mark.parametrize('protocol', ['grpc'])
+def test_return_order_in_client(protocol):
+    class ExecutorRandomSleepExecutor(Executor):
```
alaeddine-13, line 10:
```suggestion
    class RandomSleepExecutor(Executor):
```

tests/integration/test_return_order/test_return_order.py
```diff
@@ -0,0 +1,29 @@
...
+
+        @requests
+        def foo(self, *args, **kwargs):
+            rand_sleep = random.uniform(0.1, 1.3)
+            time.sleep(rand_sleep)
+
+    f = Flow(protocol=protocol).add(uses=ExecutorRandomSleepExecutor, replicas=2)
```
alaeddine-13, line 17:
```suggestion
    f = Flow(protocol=protocol).add(uses=RandomSleepExecutor, replicas=2)
```

jina/serve/streamer.py
```diff
@@ -18,6 +18,25 @@
...
+    return TopologyGraph(
+        graph_representation=graph_description,
+        graph_conditions=graph_conditions,
+        deployments_metadata=deployments_metadata,
+        deployments_no_reduce=deployments_no_reduce,
+        timeout_send=timeout_send,
+        retries=retries,
+    )
+
+
```
JohannesMessner, line 39: I don't understand the purpose of this, it seems like it is the same as just calling the constructor directly?

jina/serve/stream/__init__.py
```diff
@@ -88,14 +94,19 @@ async def stream(
...
     ) -> AsyncIterator:
         """Implements request and response handling without prefetching
         :param request_iterator: requests iterator from Client
+        :param results_in_order: return the results in the same order as the request_iterator
         :yield: responses
         """
         result_queue = asyncio.Queue()
+        request_ids = []
+        responses_list = []
+        responses_ids = {}  # map from id to index in responses_list
```
JohannesMessner, line 109: I would rename to `response_id_to_index` or something like that

docs/fundamentals/client/client.md
```diff
@@ -760,6 +760,34 @@ None
 
 ````
 
+A Client connects to a Flow that processes Documents in an asynchronous and very distributed way. This means that the order of the Flow processing the requests may not be the same order as the Client sending the requests.
```
JohannesMessner, line 763: Is there a way to pass this parameter from a 3rd party client?

JoanFM: No, this is gonna be ready in the `streamer` for the CustomGateway. We can add it later I would say

JoanFM: Mostly external clients do not have this `streaming` capability so easily implemented, so I think is not so so important

alexcg1 (approved):
LGTM :+1:

Issue #5402, opened 2022-11-17
Bind to `host` instead of `default_host`


JoanFM:
**Describe the bug**
Flow accepts `host` parameter because it inherits from client and gateway but is confusing as shown in #5401

alaeddine-13:
In principle, Flow exposes parameters common to executors and the gateway and are inherited from the Flow. `host` is an important parameter for external executors, therefore, it is exposed by Flow.
We have an issue to document the behavior of parameter inheritance: https://github.com/jina-ai/jina/issues/5353
We can either just document this or make host an exception.

wqh17101:
All 0 monitoring is equivalent to binding all IPs of the machine, that is, it will cause the whole network to monitor, destroying the original three-sided isolation principle of the system, resulting in increased exposure or increasing the risk of lateral attack expansion.

This is a danger setting.
For example, i only want to expose this api internally instead of all the way to this machine.

JoanFM:
Hey @wqh17101, 

I am not sure to understand it.

What Jina is doing when you run Flow locally is to run a server listening to the `port` provided (or a random port) in your local machine.

We do not touch any rules on where your machine is exposed.

wqh17101:
@JoanFM 
As a web service, it can expose your port(api) on any ips on your machine.
So if you set host as 0.0.0.0,every ips will be available to the api. So everyone can reach your machine  by ping can attack your service like DDos.
if you set host as 127.0.0.1,only the calling on the server machine can reach.
You can get some info from https://stackoverflow.com/questions/20778771/what-is-the-difference-between-0-0-0-0-127-0-0-1-and-localhost

Sometimes i want a local-only interface not all network interface to be exposed for the security problem.

wqh17101:
According to the web framework thirdparty software such as `flask` or `pyramid`. It provides the parameter `host` for user to choose which ip to bind to. That is flexible that the users can adjust their host on their own, and the framework will not need to make some limitations.

PR #5378, opened 2022-11-11
feat: allow multiple port and protocols for gateway

alaeddine-13:
This PR adds support for multiple ports and protocols for the gateway runtime.

- [x] Gateway parser can receive list or single value
- [x] Flow `__init__` and `config_gateway` can receive multiple or single value. Flow.port and Flow.protocol return single or multiple values
- [x] GatewayRuntime expects list of values
- [x] BaseGateway can receive runtime args
- [ ] (optional) move injected dependencies to runtime_args, TBD in https://github.com/jina-ai/jina/issues/5396
- [x] implemented Gateway classes uses the port and protocol from runtime args
- [x] support in JAML
- [x] k8s gateway service
- [x] docker compose gateway service
- [x] health checks
* NetworkChecker
* docker and k8s checks
* Flow pod checks
- [ ] ~~apply the same principle for port_monitoring~~ TBD in a separate PR: https://github.com/jina-ai/jina/issues/5352
- [x] fix gateway output: https://github.com/jina-ai/jina/issues/5345
- [ ] pass CI
- [x] Add tests
follow up: https://github.com/jina-ai/jina/issues/5393
closes: https://github.com/jina-ai/jina/issues/5345, https://github.com/jina-ai/jina/issues/5343

PR #5405, opened 2022-11-17
fix: bind servers to `host` argument isntead of __default_host__

alaeddine-13:
Although we have a `host` parameter in pods, we only use `__default_host__` (either 0.0.0.0 or 127.0.0.1) to bind servers (grpc, http, websocket) in all pod types (head, worker, gateway).
This PR makes the runtime respect the user-provided host parameter and use it when binding servers.
Depends on https://github.com/jina-ai/jina/pull/5378 so that we can pass `host` as runtime argument in gateway classes
closes: https://github.com/jina-ai/jina/issues/5402

### DIFF
--- a/jina/orchestrate/deployments/__init__.py
+++ b/jina/orchestrate/deployments/__init__.py
@@ -792,9 +792,7 @@ class Deployment(BaseDeployment):
             selected_devices = []
             if device_str[2:]:
 
-                for device in Deployment._parse_devices(
-                    device_str[2:], num_devices
-                ):
+                for device in Deployment._parse_devices(device_str[2:], num_devices):
                     selected_devices.append(device)
             else:
                 selected_devices = range(num_devices)
@@ -880,7 +878,8 @@ class Deployment(BaseDeployment):
 
         _args = copy.deepcopy(args)
         _args.pod_role = PodRoleType.WORKER
-        _args.host = __default_host__
+        if not _args.host:
+            _args.host = __default_host__
         _args.port = helper.random_port()
 
         if _args.name:
--- a/jina/serve/runtimes/gateway/grpc/gateway.py
+++ b/jina/serve/runtimes/gateway/grpc/gateway.py
@@ -18,6 +18,7 @@ class GRPCGateway(BaseGateway):
     def __init__(
         self,
         port: Optional[int] = None,
+        host: Optional[str] = __default_host__,
         grpc_server_options: Optional[dict] = None,
         ssl_keyfile: Optional[str] = None,
         ssl_certfile: Optional[str] = None,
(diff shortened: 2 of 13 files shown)


jina/orchestrate/deployments/__init__.py
```diff
@@ -880,7 +878,8 @@ def _set_uses_before_after_args(args: Namespace, entity_type: str) -> Namespace:
 
         _args = copy.deepcopy(args)
         _args.pod_role = PodRoleType.WORKER
-        _args.host = __default_host__
+        if not _args.host:
```
JoanFM, line 881:
```suggestion
_args.host = _args.host or __default_host__
```

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -18,20 +18,23 @@ class GRPCGateway(BaseGateway):
     def __init__(
         self,
         port: Optional[int] = None,
+        host: Optional[str] = __default_host__,
```
JoanFM, line 21: then is not Optional

jina/serve/runtimes/gateway/http/gateway.py
```diff
@@ -14,6 +14,7 @@ class HTTPGateway(BaseGateway):
     def __init__(
         self,
         port: Optional[int] = None,
+        host: Optional[str] = __default_host__,
```
JoanFM, line 17: not Optional

JoanFM (changes requested):
Host should be a `runtime_arg`. Also test what happens when a stupid host is passed

tests/docker_compose/custom-gateway/dummy_gateway.py
```diff
@@ -21,16 +21,11 @@ class ProcessedResponseModel(BaseModel):
...
-        host: str = __default_host__,
-        arg1: str = None,
-        arg2: str = None,
-        arg3: str = 'default-arg3',
-        **kwargs
+        self, arg1: str = None, arg2: str = None, arg3: str = 'default-arg3', **kwargs
     ):
         super().__init__(**kwargs)
         self.port = self.runtime_args.port[0]
+        self.host = self.runtime_args.host
```
JoanFM, line 28: for the `gateway` we should provide these properties:

```python
@property
def port():
    return self.runtime_args.port[0]

@property
def ports():
    return self.runtime_args.port

@property
def host():
    return self.runtime_args.host
```

JoanFM (changes requested):

tests/docker_compose/custom-gateway/dummy_gateway.py
JoanFM: This we can do in a separate PR

alaeddine-13: okay makes sense to me

alaeddine-13: > Host should be a `runtime_arg`. Also test what happens when a stupid host is passed

added

JoanFM (approved):

PR #5410, opened 2022-11-18
feat: dynamic batching

Jackmin801:
allow dynamic batching for executors. Feature described in detail [here](https://github.com/jina-ai/jina/issues/5033).

## Dynamic Batching Constraints
- len(input) == len(output)
- merge must reduce docs (no docs_matrix)
- no dictionary return allowed

## To Do

- [x] create the decorator
- [x] make overridable in YAML
- [x] implement queue and batching in the worker
- [x] implement unit tests for batch queue
- [x] test that failures are propagated
- [x] make sure default endpoint is covered
- [x] allow endpoint specific configs
- [x] add support for parameters
- [x] ~~make max_batch_size work~~ removed afterwards
- [x] test endpoint specific configs
- [x] remove `max_batch_size`; this could be a useful addition for future iterations OR do it properly, splitting up individual requests that are longer than the max
- [x] verify user experience with `Executor.serve()`
- [x] handle sigterms
- [x] add more tests: 
* no dynamic_batching decorator, only specified in the yaml
* dynamic_batching decorator + yaml
* serve.reload + dynamic batching
- [x] what happens when request is received and the task is being flushed ? fixed in beec7cc08e18c3cd8bbe68bfa485d691e8abf27b
- [x] documentation

Do we still need this?
- implement waiting logic in the gateway
- fix tracing in executor call
  - It seems like the tracing works just fine even when we dont pass the context into the executor call.
  - Just the nesting levels are different.
  - Do we need to link spans here? @girishc13 
- make sure teardown flushes
  - this is not helpful because the grpc requests are all cancelled before teardown is called
  - this is true even without dynamic batching -- any executors executing halfway have their operations cancelled

**Example Usage**
As decorator:
```python
class MyExecutor(Executor):
    @requests(on=['/foo'])
    @dynamic_batching(preferred_batch_size=4, timeout=1000)
    def foo_fun(self, docs, **kwargs):
        for doc in docs:
            doc.text += FOO_SUCCESS_MSG
```
You can override this when adding to the flow:
```python
f = Flow(port=12345).add(
        uses=MyExecutor,
        name='exec0',
        uses_dynamic_batching={'/foo': {'preferred_batch_size': 2, 'timeout': 4000}},
    )
```
Or in the YAML:
```yaml
! MyExecutor
py_modules:
  - my_executor.py
dynamic_batching:
  /foo:
    preferred_batch_size: 2
    timeout: 4000
  /bar:
    preferred_batch_size: 4
    timeout: 2000
```


jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,187 @@
...
+    def _reset(self) -> None:
+        """Set all events and reset the batch queue."""
+        if getattr(self, '_events', None) is not None:
+            [i.set() for i in self._events]
+        self._events: List[Event] = []
+        self._requests: List[DataRequest] = []
+        self._request_lens: List[int] = []
+        self._big_doc: DocumentArray = DocumentArray.empty()
+
+        self._flush_trigger: Event = Event()
```
JohannesMessner, line 44: can we initialize these (also) directly in the `__init__()`? I like it when all the state is explicitly in the init, makes it easier to know what fields exist on this object.

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,187 @@
...
+
+        self._big_doc.extend(docs)
+        self._requests.append(request)
+        self._request_lens.append(len(docs))
+        self._events.append(event)
+
+        if len(self._big_doc) > self._preferred_batch_size:
+            self._flush_trigger.set()
+    
+    async def flush(self, trigger_event: Event) -> None:
```
JohannesMessner, line 81: this doesn't flush immediatelly, but rathe it awaits the event and then flushes. So maybe call it `await_and_flush()` or something like that?

JohannesMessner (changes requested):
We're moving in a good direction with this one, like it!

jina/serve/runtimes/worker/batch_queue.py
Jackmin801: Yea maybe we should just duplicate some of the code so that concern 1 moves into `__init__`

jina_cli/autocomplete.py
```diff
@@ -213,6 +213,7 @@
         'cloud login': ['--help'],
         'cloud logout': ['--help'],
         'cloud deploy': ['--help'],
+        'cloud normalize': ['--help'],
```
Jackmin801, line 216: What is this about?

Jackmin801:
Sample YAML spec:
```yaml
dynamic_batching:
  endpoint1:
    preferred_batch_size: 4
    timeout: 10000
  endpoint2:
    preferred_batch_size: 8
    timeout: 4000
```

- Each endpoint gets its own batch queue; even if the endpoints share functions

Reasoning
- User thinks about endpoints not method. This also allows the same method to have endpoints with different dynamic batching configurations.
- However users might want to batch everything into one queue for all endpoints of a function. This will not work by default. They will have to make one endpoint that has dynamic batching and have the other endpoints redirect to the one endpoint.

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,132 @@
+import asyncio
+from asyncio import Event, Task
+from typing import List, Dict, Any
+
+from jina import DocumentArray
+from jina.serve.executors import BaseExecutor
+from jina.types.request.data import DataRequest
+
+# TODO: Make sure we dont deserialize the request twice
+# TODO: SIGINT/SIGTEM: Flush
```
JoanFM, line 10: SIGINT SIGTERM is captured by the `WorkerRuntime` and calls `teardown`. So make sure the `teardown` method calls this properly, but do not add specific language about `signal handling` here

jina/serve/runtimes/worker/request_handling.py
```diff
@@ -58,6 +59,26 @@ def __init__(
...
+                for endpoint in func_endpoints[func_name]:
+                    self._batchqueue[endpoint] = BatchQueue(
+                        executor=self._executor,
+                        exec_endpoint=endpoint,
+                        args=self.args,
+                        preferred_batch_size=dbatch_config['preferred_batch_size'],
+                        timeout=dbatch_config['timeout'],
+                    )
+
+            self.logger.debug(f"Dynamic Batching configs: {self._executor.dynamic_batching}")
```
JoanFM, line 80: Single quotes

jina/serve/runtimes/worker/request_handling.py
```diff
@@ -267,42 +289,49 @@ async def handle(
...
         self._record_request_size_monitoring(requests)
 
-        params = self._parse_params(requests[0].parameters, self._executor.metas.name)
-        docs = WorkerRequestHandler.get_docs_from_request(
-            requests,
-            field='docs',
-        )
+        # TODO: This does not account for requests to "/"
+        if exec_endpoint in self._batchqueue:
+            assert len(requests) == 1, "dynamic batching does not support no_reduce"
```
JoanFM, line 307: single quote

jina_cli/autocomplete.py
JoanFM: this is something about `Jina Dev Bot` pushing some commits behind the back

tests/unit/orchestrate/deployments/test_batch_queue.py
```diff
@@ -0,0 +1,155 @@
+import asyncio
```
JoanFM, line 1: this unit test do not belong in `orchestrate`. They belong inside `serve`.

JoanFM (changes requested):

jina/serve/runtimes/worker/batch_queue.py
Jackmin801: grpc requests are cancelled before teardown.
even if the queue is flushed, the response wont return so it doesnt work; even when there is no dynamic batching. :(
not going to implement this for now.

tests/unit/serve/executors/test_bad_executor_constructor.py
```diff
@@ -13,7 +13,7 @@ def foo(self, **kwargs):
 
 
 class GoodExecutor2(Executor):
-    def __init__(self, metas, requests, runtime_args):
+    def __init__(self, metas, requests, runtime_args, dynamic_batching):
```
JoanFM, line 16: what are these `dynamic_batching`? why not part of `runtime_args`? @alaeddine-13 @JohannesMessner

tests/unit/serve/executors/test_decorators.py
```diff
@@ -2,9 +2,10 @@
 
 import pytest
 
+from jina import Executor
```
JoanFM, line 5: this test should not import `Executor` since this is unit test only for the decorator

tests/unit/serve/executors/test_decorators.py
```diff
@@ -63,3 +64,67 @@ async def fn_3(*args, **kwargs):
     assert not iscoroutinefunction(getattr(fn_2, 'fn'))
     assert hasattr(fn_3, 'fn')
     assert iscoroutinefunction(getattr(fn_3, 'fn'))
+
+
+@pytest.mark.parametrize(
```
JoanFM, line 69: make sure to be able to test the decorator without the Executor and move these tests to unit tests of Executor

tests/unit/serve/runtimes/worker/test_batch_queue.py
```diff
@@ -0,0 +1,253 @@
...
+    assert not single_task.done()
+    assert data_requests[0].data.docs[0].text == 'Not Done'
+    
+    await asyncio.sleep(0.5)
+    assert single_task.done()
+    assert data_requests[0].data.docs[0].text == 'Done'
+
+
+@pytest.mark.asyncio
+async def test_batch_queue_max_batch_size():
```
JoanFM, line 105: I think this test belongs to a different folder

JoanFM: here the runtime is not even used

JoanFM (changes requested):

jina/serve/runtimes/worker/__init__.py
```diff
@@ -224,6 +229,11 @@ async def endpoint_discovery(self, empty, context) -> jina_pb2.EndpointsProto:
         :param context: grpc context
         :returns: the response request
         """
+        if self._reject_incoming_requests:
```
JoanFM, line 232: This is not to be handled by us. This is the health check job to do. And thus is why we set it as graceful shutdown

jina/serve/runtimes/worker/__init__.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         :param args: args from CLI
         :param kwargs: keyword args
         """
+        self._reject_incoming_requests = False
```
JoanFM, line 37: This flag should not exist

JoanFM (changes requested):

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,172 @@
...
+        self._big_doc.extend(docs)
+        self._requests.append(request)
+        self._request_lens.append(len(docs))
+
+        if len(self._big_doc) >= self._preferred_batch_size:
+            self._flush_trigger.set()
+
+        return self._flush_task
+
+    async def await_then_flush(self, trigger_event: Event) -> None:
```
alaeddine-13, line 113: since the trigger_event is always self._flush_trigger, I think just always using it and removing the parameter improves readability

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,172 @@
...
+    def _reset(self) -> None:
+        """Set all events and reset the batch queue."""
+        self._requests: List[DataRequest] = []
+        self._request_lens: List[int] = []
+        self._big_doc: DocumentArray = DocumentArray.empty()
+
+        self._flush_trigger: Event = Event()
+        self._flush_task: Task = asyncio.create_task(
+            self.await_then_flush(self._flush_trigger)
+        )
```
alaeddine-13, line 81: very tiny perf improvement but can also improve code:
I think the flush_task should only be created when there was a push to the batch queue. So it is reset to None and only started in `push`.
flush_task == None => no timer stared + no data in queue
flush_task != None => timer has already been started and there's data in queue.
We would be able to remove _timer_started

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,172 @@
...
+        self._timer_started: bool = False
+
+    def _start_timer(self):
+        asyncio.create_task(sleep_then_set(self._timeout / 1000, self._flush_trigger))
```
alaeddine-13, line 85: I suggest simplifying code here by making sleep_then_set a private method of BatchQueue that directly uses timeout and flush_trigger (no need to receive parameters)

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,172 @@
...
+
+        :return: The task that will be set once the request has been processed.
+        """
+        if self._max_batch_size:
+            if len(self._big_doc) + len(request.docs) > self._max_batch_size:
+                self._flush_trigger.set()
+                await self._flush_task
+
+        if not self._timer_started:
+            self._start_timer()
```
alaeddine-13, line 101: according to what was suggested above, we can do:
```suggestion
        if not self. _flush_task:
            self._flush_task = asyncio.create_task(
                self.await_then_flush() # also taking in consideration another suggestion to remove parameter
            )
            asyncio.create_task(self.sleep_then_set()) # also taking in consideration another suggestion to remove parameter
```

jina/serve/runtimes/worker/batch_queue.py
alaeddine-13: maybe do this just for self._timeout, because self._flush_trigger is renewed on every flush and we don't want old timers to set triggers for new flushes

jina/serve/runtimes/worker/request_handling.py
```diff
@@ -299,43 +355,63 @@ async def handle(
...
-            docs=docs,
-            parameters=params,
-            docs_matrix=docs_matrix,
-            docs_map=docs_map,
-            tracing_context=tracing_context,
-        )
+        if exec_endpoint in self._batchqueue_config:
+            assert len(requests) == 1, 'dynamic batching does not support no_reduce'
+
+            param_key = str(params)
```
alaeddine-13, line 388: I don't think this is a reliable way to generate a key for parameters, try this for instance:
```python
d = {'a': 1, 'b': 'c'}
print(str(d)) # "{'a': 1, 'b': 'c'}"
d = {'b': 'c', 'a': 1}
print(str(d)) #  "{'b': 'c', 'a': 1}"
```
You can refer to this page for ways to hash a dict: https://stackoverflow.com/questions/5884066/hashing-a-dictionary
Ideally, `hash(frozenset(my_dict.items()))` would be great, if we can't use it, we'll have to use `json.dumps(d, sort_keys=True)`

jina/serve/runtimes/worker/request_handling.py
```diff
@@ -299,43 +355,63 @@ async def handle(
...
+            if param_key not in self._batchqueue_instances[exec_endpoint]:
+                self._batchqueue_instances[exec_endpoint][param_key] = BatchQueue(
+                    executor=self._executor,
+                    exec_endpoint=exec_endpoint,
+                    args=self.args,
+                    params=params,
+                    **self._batchqueue_config[exec_endpoint],
```
alaeddine-13, line 395: as mentioned above, I believe here we should only pass a function to BatchQueue (partial(executor.__acall__, exec_endpoint))
This should help decouple both classes and make BatchQueue easier to test (no more need for a mock executor)

jina/serve/runtimes/worker/__init__.py
JohannesMessner: My thought was that there is a brief moment where the worker is shutting down and processing pending requests, so the grpc server is still up and reachable, but should probably not accept new requests anymore.

jina/serve/runtimes/worker/__init__.py
JoanFM: Please remove this flag

JoanFM (changes requested):

jina/serve/runtimes/worker/request_handling.py
JoanFM: agree, this is not reliable

jina/serve/runtimes/worker/request_handling.py
JoanFM: Love it

tests/unit/serve/dynamic_batching/test_batch_queue.py
```diff
@@ -0,0 +1,254 @@
...
+import time
+
+import pytest
+
+from jina import Document, DocumentArray, Executor, requests
+from jina.serve.runtimes.worker.batch_queue import BatchQueue, sleep_then_set
+from jina.types.request.data import DataRequest
+
+
+@pytest.mark.skip(reason="Sleep then set doesnt count normal sleep here")
```
JoanFM, line 11: there is no need to test such a small function

tests/unit/serve/dynamic_batching/test_dynamic_batching_config.py
```diff
@@ -0,0 +1,31 @@
...
+        return DocumentArray([Document(text='cat')])
+
+    @requests()
+    @dynamic_batching(preferred_batch_size=10, timeout=10000)
+    def default_fun(self, docs, **kwargs):
+        return DocumentArray([Document(text='bar')])
+
+
+def test_dynamic_batching_config(tmpdir):
+    TMPDIR: Path = Path(tmpdir)
```
JoanFM, line 19: this is not a unit test, this is an integration test, and it should be reacher. Does it work when no `dynamic_batching` decorator exist?

tests/unit/serve/dynamic_batching/test_dynamic_batching_config.py
```diff
@@ -0,0 +1,31 @@
...
+    f = Flow(port=12345).add(
+        uses=MyExecutor,
+        name='exec0',
+        uses_dynamic_batching={'/foo': {'preferred_batch_size': 2, 'timeout': 4000}},
+    )
+    f.save_config(str(TMPDIR / 'flow0.yml'))
```
JoanFM, line 25: why do we need to `save` and load the config?

JoanFM (changes requested):

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,179 @@
...
+            else:
+                request.data.set_docs_convert_arrays(
+                    self._big_doc[left:right], ndarray_type=self.args.output_array_type
+                )
+            consumed_count += request_len
+
+    async def close(self):
+        """Closes the batch queue by flushing pending requests."""
+        if not self._is_closed:
+            self._flush_trigger.set()
```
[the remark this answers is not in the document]
JoanFM: after we close this we have to make sure WorkerRuntime returns the requests computed

alaeddine-13 (commented):
TODO: make sure this is compatible with executor reload

docs/concepts/executor/dynamic_batching.md
```diff
@@ -0,0 +1,123 @@
+(executor-dynamic-batching)=
+# Dynamic Batching
+Dynamic batching is a feature of Jina that allows requests to be accumulated and batched together before being sent to 
+an {class}`~jina.Executor`. The batch is created dynamically depending on the per-endpoint dynamic batching configuration.
+
+This feature is relevant especially for inference tasks where model inference is more optimized when done in batches in 
+order to efficiently use GPU resources.
+
+## Overview
+Enabling dynamic batching on Executor endpoints that perform inference typically results in increased throughput.
```
JoanFM, line 10: This is not true, it only happens in some circumstances. I think we should mention better hardware usage

docs/concepts/executor/dynamic_batching.md
```diff
@@ -0,0 +1,123 @@
...
+## Overview
+Enabling dynamic batching on Executor endpoints that perform inference typically results in increased throughput. 
+
+When Dynamic Batching is enabled requests incoming to an Executor endpoints with the same {ref}`request parameters<client-executor-parameters>`
+will be queued together. The Executor endpoint will be executed on the queue requests when the number of documents 
+accumulated exceeds the {ref}`preferred_batch_size<executor-dynamic-batching-parameters>` parameter or when the {ref}`timeout<executor-dynamic-batching-parameters` parameter is exceeded.
+
+Although this feature can work on {ref}`parametrized requests<client-executor-parameters>`, it is best used for endpoints
+that do not receive different parameters often.
+Creating a batch of requests typically results in increased throughput.
```
JoanFM, line 18: see above

jina/serve/runtimes/worker/__init__.py
```diff
@@ -194,6 +194,8 @@ async def async_cancel(self):
         self.logger.debug('cancel WorkerRuntime')
         if self._hot_reload_task is not None:
             self._hot_reload_task.cancel()
+        await self._request_handler.close()  # allow pending requests to be processed
```
JoanFM, line 197: no, u should enter the health servicer into graceful shutdown before closing the request so that u announce u are shutting down to the outside world

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -0,0 +1,161 @@
...
+                        f'The return type must be DocumentArray / `None` when using dynamic batching, '
+                        f'but getting {return_data!r}'
+                    )
+
+                # We need to re-slice the big doc array into the original requests
+                self._fan_out_return_data(return_data)
+            finally:
+                self._reset()
+
+    def _fan_out_return_data(self, return_data: Optional[DocumentArray]):
```
JoanFM, line 138: change name to a better understandable name

jina/serve/runtimes/worker/request_handling.py
```diff
@@ -208,7 +270,7 @@ def _refresh_executor(self, changed_files):
 
     @staticmethod
     def _parse_params(parameters: Dict, executor_name: str):
-        parsed_params = parameters
+        parsed_params = parameters  # TODO: Copy? This still points to the same object
```
JoanFM, line 273: remove TODO

tests/unit/serve/dynamic_batching/test_dynamic_batching_config.py
JoanFM: This is not a unit test move to integration

tests/unit/serve/dynamic_batching/test_dynamic_batching_config.py
JoanFM: no need to save

JoanFM (changes requested):

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -75,58 +76,64 @@ async def push(self, request: DataRequest) -> Task:
...
         docs = request.docs
-        # TODO: is this protected/synchronous ?
-        self._big_doc.extend(docs)
-        self._requests.append(request)
-        self._request_lens.append(len(docs))
 
-        if len(self._big_doc) >= self._preferred_batch_size:
-            self._flush_trigger.set()
+        # writes to shared data between tasks need to be mutually exclusive
+        async with self._data_lock:
```
JoanFM, line 82: does it need to protect also the task creation? and the start timer?

jina/serve/runtimes/worker/batch_queue.py
```diff
@@ -75,58 +76,64 @@ async def push(self, request: DataRequest) -> Task:
...
+            if not self._flush_task:
+                self._flush_task = asyncio.create_task(self._await_then_flush())
+            if not self._timer_task:
+                self._start_timer()
+
+            self._big_doc.extend(docs)
+            self._requests.append(request)
+            self._request_lens.append(len(docs))
+
+            if len(self._big_doc) >= self._preferred_batch_size:
```
JoanFM, line 92: also, does this need to be protected?

JoanFM (changes requested):

jina/serve/runtimes/worker/__init__.py
alaeddine-13: this might mean orchestrators kill the running pod right ?

docs/concepts/executor/dynamic_batching.md
```diff
@@ -0,0 +1,123 @@
...
+    def embed(self, docs: DocumentArray):
+        docs.embeddings = self.model(docs.tensors)
+        
+flow = Flow().add(uses=MyExecutor, uses_dynamic_batching={'/bar': {'preferred_batch_size': 10, 'timeout': 200}})
+```
+````
+
+````{tab} Using YAML configuration
+If you want to use YAML to enable dynamic batching on an Executor, you can use the `dynamic_batching` section in the 
+Executor section. Suppose the Executor is implemented like this:
```
alexcg1, line 82:
```suggestion
If you use YAML to enable dynamic batching on an Executor, you can use the `dynamic_batching` section in the 
Executor section. Suppose the Executor is implemented like this:
```

jina/serve/runtimes/worker/__init__.py
JoanFM: what do u mean?

JoanFM: yes, it may mean this

docs/concepts/executor/dynamic_batching.md
```diff
@@ -0,0 +1,123 @@
...
+    timeout: 200
+```
+````
+
+(executor-dynamic-batching-parameters)=
+## Dynamic Batching Parameters
+The following parameters allow you to configure the dynamic batching behavior on each Executor endpoint:
+* `preferred_batch_size`: Target number of Documents in a batch. The batcher will collect requests until 
+`preferred_batch_size` is reached, or until `timeout` is reached. Therefore, the actual batch size can be smaller or 
+larger than `preferred_batch_size`.
```
alexcg1, line 120:
```suggestion
* `preferred_batch_size`: Target number of Documents in a batch. The batcher collects requests until 
`preferred_batch_size` is reached, or until `timeout` is reached. Therefore, the actual batch size could be smaller or 
larger than `preferred_batch_size`.
```

docs/concepts/executor/dynamic_batching.md
```diff
@@ -0,0 +1,123 @@
...
+* `timeout`:  maximum time in milliseconds to wait for a request to be assigned to a batch.
+If the oldest request in the queue reaches a waiting time of `timeout`, the batch will be passed to the Executor, even 
+if it contains fewer than `preferred_batch_size` Documents. Default is 10_000ms (10 seconds).
```
alexcg1, line 123:
```suggestion
* `timeout`:  Maximum time in milliseconds to wait for a request to be assigned to a batch.
If the oldest request in the queue reaches a waiting time of `timeout`, the batch is passed to the Executor, even 
if it contains fewer than `preferred_batch_size` Documents. Default is 10,000ms (10 seconds).
```

jina/serve/runtimes/worker/batch_queue.py
alaeddine-13: in my opinion yes, the other block calls reset which also sets the task to None

jina/serve/runtimes/worker/batch_queue.py
JoanFM: okey good

JoanFM (approved):

yuanjie-ai:
donot work

JoanFM:
@yuanjie-ai,

can you open a ticket showing your set up ans explaining why it does not work?

Issue #5422, opened 2022-11-22
`metadata.uid` field doesn't exist in generated k8s YAMLs


zac-li:
**Describe the bug**

In the [generated YAML file](https://github.com/jina-ai/jina/blob/master/jina/resources/k8s/template/deployment-executor.yml), it has a reference to `metadata.uid`: `fieldPath: metadata.uid`. But looks like in `metadata` there's no `uid`:

```
metadata:
  name: {name}
  namespace: {namespace}
```

This is sometimes causing problem for our Operator use case; we had to manually remove the `POD_UID` block generated by `jina`. Are we going to address this?

**Describe how you solve it**

---

**Environment**

**Screenshots**

PR #5424, opened 2022-11-22
fix: remove pod uid block from k8s yamls

zac-li:
**Goals:**

This PR is raised to address issue https://github.com/jina-ai/jina/issues/5422.
Discussed with @deepankarm and @JoanFM, loos like the path forward is to remove these unused blocks.

### DIFF
--- a/jina/resources/k8s/template/deployment-executor.yml
+++ b/jina/resources/k8s/template/deployment-executor.yml
@@ -36,10 +36,6 @@ spec:
           - configMapRef:
               name: {name}-configmap
           env:
-          - name: POD_UID
-            valueFrom:
-              fieldRef:
-                fieldPath: metadata.uid
           - name: JINA_DEPLOYMENT_NAME
             value: {jina_deployment_name}
           - name: K8S_DEPLOYMENT_NAME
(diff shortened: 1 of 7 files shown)


girishc13:
Removing the code will just fix this temporary issues. Pod UID's are assigned by the k8s api. If it's not being set then there must be an issue with the k8s cluster setup or it has been intentionally not setup in the pod metadata.

girishc13:
Ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata

> Metadata
> 
> Every object kind MUST have the following metadata in a nested object field called "metadata":
> 
>     namespace: a namespace is a DNS compatible label that objects are subdivided into. The default namespace is 'default'. See [the namespace docs](https://kubernetes.io/docs/user-guide/namespaces/) for more.
>     name: a string that uniquely identifies this object within the current namespace (see [the identifiers docs](https://kubernetes.io/docs/user-guide/identifiers/)). This value is used in the path when retrieving an individual object.
>     uid: a unique in time and space value (typically an RFC 4122 generated identifier, see [the identifiers docs](https://kubernetes.io/docs/user-guide/identifiers/)) used to distinguish between objects with the same name that have been deleted and recreated

PR #5430, opened 2022-11-22
fix: dump env vars as str to k8s

JoanFM:
**Goals:**
Dump env vars as str so that Kubernetes does not complain


samsja (approved):
lgtm. I believe we might need to avoid converting "1234" to int from the flow.yml. But this one do the job as well

PR #5433, opened 2022-11-23
fix: unblock event loop to allow health service 

JoanFM:
**Goals:**
Runs the Executor methods in a protected thread so that it is async and unlocks the event lock so that health service checks can pass


tests/k8s/test_k8s.py
```diff
@@ -1140,9 +1140,55 @@ async def test_flow_with_stateful_executor(
...
         )
+    except Exception as exc:
+        logger.error(f' Exception raised {exc}')
+        raise exc
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    'docker_images', [['slow-process-executor', 'jinaai/jina']], indirect=True
+)
+async def test_slow_executor_readiness_probe_works(docker_images, tmpdir, logger):
```
alaeddine-13, line 1151: I don't understand this specific test
why don't we just send requests in a separate process, do port forwarding and execute a manual health check (WorkerRuntime.is_ready).
In another PR that adds liveness probe we can modify this test

JoanFM: what u say is done in the `test_runtimes`. This test adds little value for now. It proves that really slow works. We can remove later, or it will be more important when livenessProbe is there.

Let's keep for now

PR #5436, opened 2022-11-23
fix: test stateful executor tests

Jackmin801:
**Goals:**

- The stateful executor tests assert dont seem to be correct given the namespace deletion succeeds

### DIFF
--- a/tests/k8s/test_k8s.py
+++ b/tests/k8s/test_k8s.py
@@ -1120,6 +1120,7 @@ async def test_flow_with_stateful_executor(
         )
 
         core_client.delete_namespace(namespace)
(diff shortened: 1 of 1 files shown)


tests/k8s/test_k8s.py
```diff
@@ -1120,6 +1120,7 @@ async def test_flow_with_stateful_executor(
         )
 
         core_client.delete_namespace(namespace)
```
JoanFM, line 1122: The error is calling `delete namespace`, what we want is to restart the Pod, so we should delete the Pods

JoanFM (changes requested):

PR #5439, opened 2022-11-23
chore: add install ecosystem latest version test to ci

samsja:
Signed-off-by: Sami Jaghouar <sami.jaghouar@hotmail.fr>

# Context

It might happen that jina docarray hubble-sdk and jcloud are in conflict on the dependency. This is not intended and we should check before it happen

# What this pr do

Add a test in CI to install the current commit with all of the latest version

- [x] add test in ci
- [x] add nightly runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -418,9 +418,6 @@ jobs:
     needs: [lint-flake-8, code-injection ]
     strategy:
       fail-fast: false
-      matrix:
-        core: ['', 'true']
-        perf: ['', 'true']
     steps:
       - uses: actions/checkout@v2.5.0
       - name: Set up Python 3.7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -397,7 +397,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-    steps:
+        core: ['', 'true']
+        perf: ['', 'true']
+        exclude:
+          - core: 'true'
+            perf: 'true'
+      steps:
       - uses: actions/checkout@v2.5.0
       - name: Set up Python 3.7
         uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -412,7 +431,13 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install wheel
-          pip install --no-cache-dir .
+        
+          HUBBLE_VERSION=$(curl -L -s "https://pypi.org/pypi/jina-hubble-sdk/json" |  jq  -r '.releases | keys | .[]| select(startswith("0."))'  | sort -V | tail -1)
```
JoanFM, line 435: what is this doing?

samsja: getting the latest version on pypi. I copy paste from the executor CI

.github/workflows/ci.yml
```diff
@@ -392,16 +392,35 @@ jobs:
...
+
+  install-jina-ecosystem-test:
     runs-on: ubuntu-latest
     needs: [lint-flake-8, code-injection ]
     strategy:
       fail-fast: false
       matrix:
         core: ['', 'true']
         perf: ['', 'true']
-        exclude:
```
JoanFM, line 402: why this removal?

samsja: my mistake I believe

.github/workflows/ci.yml
```diff
@@ -421,6 +421,36 @@ jobs:
...
+          JINA_PIP_INSTALL_PERF: ${{ matrix.perf }}
+
+      - name: Test basic import
+        run: python -c 'from jina import Executor,requests'
+      - name: Test import all
+        run: python -c 'from jina import *'
+
 
   # just for blocking the merge until all parallel core-test are successful
   success-all-test:
```
JoanFM, line 456: why not added to all success?

JoanFM (changes requested):

.github/workflows/ci.yml
samsja: It some case we might want to introduce breaking change / causing a dependency conflicts between (Ex annlite and docarray). This test is more to be aware that this we cause a problem rather than blocking you

PR #5445, opened 2022-11-24
feat: add k8s preStop hook to the containers 

JoanFM:
**Goals:**

- adds preStop hook to the containers to allow more time to complete the in-flight requests and complete the grpc server shutdown
- reduce the load in `tests/test_k8s_failures.py` test because `kind` is not suitable for long running and high rps load setup. The load test aspects will be moved to a dedicated repo which can be deployed to an actual k8s cluster.


JoanFM:
> @JoanFM With the introduction of dynamic batching, I think we should increase the server stop timeout to 5 or 10 seconds. I think cpu and gpu intensive tasks will just get killed with the current 1 second timeout. If the requests finish sooner then the server will stop without waiting for too long.

Lets do in subsequent PR. Hopefully this is caught in a test there

PR #5450, opened 2022-11-25
docs: change executor uri to "jinaai"

delgermurun:
**Goals:**

- Adapt new executor URI on docs
- ...
- ...

- ...
- ...


docs/fundamentals/executor/hub/use-hub-executor.md
```diff
@@ -37,10 +37,10 @@ Use prebuilt images from Hub in your Python code:
 from jina import Flow
 
 # SECRET must be provided for private Executor
```
alaeddine-13, line 39: let's also handle a case for private executors. Maybe let's say now we have to login to jina hub ?

docs/fundamentals/executor/hub/use-hub-executor.md
```diff
@@ -37,10 +37,10 @@ Use prebuilt images from Hub in your Python code:
...
-f = Flow().add(uses='jinahub+docker://<UUID>[:<SECRET>][/<TAG>]')
+f = Flow().add(uses='jinaai+docker://<USERNAME>/<NAME>[:<TAG>]')
 ```
 
-If you do not provide a `/<TAG>`, it defaults to `/latest`.
+If you do not provide a `:<TAG>`, it defaults to `/latest`.
 
 ````{important}
 To use a private Executor, you must provide the `SECRET` which is generated after `jina hub push`.
```
alaeddine-13, line 46: let's also update this

JoanFM (changes requested):
In the `Hub` section we should add a section about `Resources` and a clear explanation of this pattern. Also explaining that the old behavior will still be suppoorted for backwards compatibility

JoanFM (changes requested):
This requires a lot of changes

JoanFM (approved):

Issue #5469, opened 2022-12-01
feat: improve grpc error message


girishc13:
**Describe the feature**

The default string method of the gRPC error and status code hides or combines the typical http errors into obscure status codes which are sometimes not actionable. The underlying http error or error message contained in the trailing metadata is more useful to debug the root cause.

**Your proposal**

Provide a custom formatter that displays the error class, gRPC status code and the context trailing metadata. 

---

**Environment**

**Screenshots**

PR #5475, opened 2022-12-01
refactor: add more debug info to prints

JoanFM:
**Goals:**
Try to add more information to ease debugging

### DIFF
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -112,6 +112,7 @@ class ReplicaList:
         runtime_name: str,
         aio_tracing_client_interceptors: Optional[Sequence['ClientInterceptor']] = None,
         tracing_client_interceptor: Optional['OpenTelemetryClientInterceptor'] = None,
+        deployment_name: str = ''
     ):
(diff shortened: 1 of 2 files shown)


samsja (dismissed):

jina/serve/networking.py
```diff
@@ -966,7 +969,7 @@ async def _handle_aiorpcerror(
         # requests usually gets cancelled when the server shuts down
         # retries for cancelled requests will hit another replica in K8s
         self._logger.debug(
-            f'GRPC call errored, getting error {error} for the {retry_i + 1}th time.'
+            f'GRPC call to {current_deployment} errored, getting error {error} for the {retry_i + 1}th time.'
```
girishc13, line 972: Printing the `error.trailing_metadata()` has the underlying http error. I created #5469  for better formatting the grpc error.

JoanFM: anyway, this will not harm right? I believe we can do this before

girishc13: It's not harmful.

PR #5490, opened 2022-12-05
ci: build python 3.10 and 3.11 as default jina docker images and make 3.8 the default jina docker image

alaeddine-13:
build python 3.10 and 3.11 as default jina docker images and make 3.8 the default jina docker image


JoanFM:
Why do we make 3.8 the default?

PR #5488, opened 2022-12-05
feat: add restart from YAML option to Flow and Executors

JoanFM:
**Goals:**
Community has been asking in Slack that to complement the reload option this feature to watch the Flow and Executor YAMLS would be interesting

PR #5494, opened 2022-12-06
feat: expand executor serve parameters

alaeddine-13:
feat: expand executor serve parameters


JoanFM:
Wait for #5488 since it touches a little bit

jina/serve/executors/__init__.py
```diff
@@ -526,6 +709,7 @@ def serve(
             uses_with=uses_with,
             uses_metas=uses_metas,
             uses_requests=uses_requests,
+            reload=reload,
         )
```
JoanFM, line 713: are u sure all these `kwargs` are passed to the Executor itself?

JoanFM (changes requested):

jina/serve/executors/__init__.py
alaeddine-13: I tried out the reload locally
Putting the executor definition and executor.serve() in the same file is not supported so I added a warning each time we attempt to reload the main module. In other cases it works

alaeddine-13: kwargs are passed to the Flow. That way, the user of serve will have full power over everything (gateway + executor).
I confirm that kwargs are passed correctly to Flow and the Flow in return, passes them to gateway and executor (beisdes the ones explicitly passed to the executors)

JoanFM (approved):

PR #5496, opened 2022-12-06
chore: send telemetry event before starting gateway object

alaeddine-13:
This PR makes the GatewayRuntime send a special telemetry event for gateway.
An event with name `start_gateway` is sent along with the gateway class and the protocol used.

**Note:**
A more efficient way would be to send information from gateway runtime only once but this means moving send_telemetry_event from init to async_setup


alaeddine-13:
it's confirmed that gateway telemetry works fine and results show up correctly

JoanFM (approved):

Issue #5502, opened 2022-12-07
Bug: Logger cuts off lines


joschkabraun:
**Describe the bug**
The logger cuts of lines. Similar to: #5241.

Code to reproduce:
```python
from jina import Executor,requests, Flow
from docarray import Document


class Exec(Executor):


    @requests()
    def foo(self, **kwargs):

        self.logger.info(','.join([str(i) for i in range(1000)]))


with Flow().add(uses=Exec) as f:
    f.post(on='/foo', inputs= Document(text='hello'))
```
Result:
![Screenshot 2022-12-07 at 17 17 05](https://user-images.githubusercontent.com/47435119/206232628-9e4cfce6-b764-4da1-a5ff-5a6028fa76fd.png)

**Environment**
```bash
- jina 3.12.0
- docarray 0.19.1
- jcloud 0.1.4
- jina-hubble-sdk 0.26.6
- jina-proto 0.1.13
- protobuf 3.20.3
- proto-backend cpp
- grpcio 1.47.0
- pyyaml 6.0
- python 3.10.8
- platform Linux
- platform-release 5.15.0-53-generic
- platform-version #59~20.04.1-Ubuntu SMP Thu Oct 20 15:10:22 UTC 2022
- architecture x86_64
- processor x86_64
- uid 185410946822994
- session-id 9e2bd57e-764a-11ed-8846-a8a1599f7352
- uptime 2022-12-07T17:17:17.021710
- ci-vendor (unset)
- internal False
* JINA_DEFAULT_HOST (unset)
* JINA_DEFAULT_TIMEOUT_CTRL (unset)
* JINA_DEPLOYMENT_NAME (unset)
* JINA_DISABLE_UVLOOP (unset)
* JINA_EARLY_STOP (unset)
* JINA_FULL_CLI (unset)
* JINA_GATEWAY_IMAGE (unset)
* JINA_GRPC_RECV_BYTES (unset)
* JINA_GRPC_SEND_BYTES (unset)
* JINA_HUB_NO_IMAGE_REBUILD (unset)
* JINA_LOG_CONFIG (unset)
* JINA_LOG_LEVEL (unset)
* JINA_LOG_NO_COLOR (unset)
* JINA_MP_START_METHOD (unset)
* JINA_OPTOUT_TELEMETRY (unset)
* JINA_RANDOM_PORT_MAX (unset)
* JINA_RANDOM_PORT_MIN (unset)
* JINA_LOCKS_ROOT (unset)
[...]

PR #5503, opened 2022-12-07
fix: fix the problen with log not unfolding

samsja:
Signed-off-by: Sami Jaghouar <sami.jaghouar@hotmail.fr>

# Context

fix https://github.com/jina-ai/jina/issues/5502

# What this pr do

use the normal rich logger and remove limit on length


samsja:
we closed since we prefer the old behavior

PR #5507, opened 2022-12-09
feat: implement unary_unary stub for Gateway Runtime

### LINKED ISSUE #5499: feat: implement unary_unary stub for Gateway Runtime
**Describe the feature**

Gateway currently only implements the stream-stream stub to handle all calls. As per the [documentation](https://grpc.io/docs/guides/performance/#general), the stream-stream rpc is not performant in python. Another observation is that the stream-stream stub uses at most a single channel to stream requests which can lead to more failures when deployed in Kubernetes since the http connection might be reset.

**Your proposal**

Implement a unary-unary stub in the Gateway and provide an argument in the Client to choose the unary-unary or stream-stream version.

girishc13:
**Goals:**

- resolves #5499
- add unary_unary stub to Gateway Runtime 
  -  the implementation currently returns responses in order.
  - unordered results will be implemented in #5510.
- add new `stream` boolean argument to the client. 
  - The argument only affects the gRPC protocol. By default the argument is set to True, which uses the available stream_stream interface to the Gateway. If set to False, the unary_unary interface to the Gateway is used.
  - `http` and `websocket` clients use the HTTP POST request to send the request to the Gateway. The connections are handled by the `aiohttp.ClientSession` implementation.
- unit tests
- read below for some profiling analysis 


girishc13:
# Profile or Performance Analysis

This feature was initiated because of this section in the gRPC [performance best practices](https://grpc.io/docs/guides/performance/#general) page. The special section for [Python](https://grpc.io/docs/guides/performance/#python) states the following:
```
- Streaming RPCs create extra threads for receiving and possibly sending the messages, which makes streaming RPCs much slower than unary RPCs in gRPC Python, unlike the other languages supported by gRPC.
- Using [asyncio](https://grpc.github.io/grpc/python/grpc_asyncio.html) could improve performance.
- Using the future API in the sync stack results in the creation of an extra thread. Avoid the future API if possible.
- (Experimental) An experimental single-threaded unary-stream implementation is available via the [SingleThreadedUnaryStream channel option](https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/experimental/__init__.py#L38), which can save up to 7% latency per message.
```

I ran the `tests/k8s/test_k8s_failures.py` kubernetes tests locally to do some profiling and performance analysis.
Using the default option in the test client, 

- `htop` reports that 5 threads are created by the gateway process
- `netstat` port monitoring lists the various tcp connection to local port 8080. The executor connections have ip `10.244.0.0/16` addresses. 

image

Using the `stream=False` option in the test client,

- `htop` reports 45 threads are created by the gateway process. Using asyncio gRPC server invalidates the remark about streaming rpc's being slow in Python. Unary call and asyncio server handling might be more expensive due to the spawning of multiple threads. This needs more investigation or looking up github issues related to asyncio gRPC performance.
- As in the previous example the `netstat` report is more of less the same because Python grpc channel controls the number of connections required to send the rpc's. This means that using stream or unary interface in the single `client.post` method doens't affect the number of tcp connections. 
image

# Summary
From the `htop` and `netstat` results, its not clear that the unary interface is performing better than the stream interface. For doing further load testing analysis and debugging its good to keep this implementation for internal use cases.

JoanFM:
Shouldn't stream=False default to the unary unary

girishc13:
> Shouldn't stream=False default to the unary unary

Yes, my mistake. I fixed the pr description.

girishc13:
On the other hand the executors are using gateway -> executor requests are made using unary rpc so the htop looks like this.

Screenshot from 2022-12-09 12-38-28

girishc13:
I think that testing with multiple clients and from multiple sources will give us a better view of the performance. Unfortunately the docs are not complete so its hard to configure.

jina/clients/base/grpc.py
```diff
@@ -123,31 +188,41 @@ async def _get_results(
...
-                                on_done=on_done,
-                                on_always=on_always,
-                                continue_on_error=continue_on_error,
-                                logger=self.logger,
-                            )
-                            if self.show_progress:
-                                p_bar.update()
-                            yield resp
+                        if stream:
+                            stub = jina_pb2_grpc.JinaRPCStub(channel)
```
alaeddine-13, line 198: can't we generate the stub inside _stream_rpc ? looks like _stream_rpc will only accept JinaRPCStub anyways

jina/clients/base/grpc.py
```diff
@@ -123,31 +188,41 @@ async def _get_results(
...
+                                on_error,
+                                on_done,
+                                on_always,
+                                continue_on_error,
+                                p_bar,
+                                **kwargs,
+                            ):
+                                yield resp
+                        else:
+                            stub = jina_pb2_grpc.JinaSingleDataRequestRPCStub(channel)
```
alaeddine-13, line 212: same, would be good to generate the stub inside  _unary_rpc

jina/clients/request/asyncio.py
```diff
@@ -46,15 +46,17 @@ async def request_generator(
...
-            async for batch in aiostream.stream.chunks(data, request_size):
-                yield _new_data_request_from_batch(
-                    _kwargs=kwargs,
-                    batch=batch,
-                    data_type=data_type,
-                    endpoint=exec_endpoint,
-                    target=target_executor,
-                    parameters=parameters,
-                )
+            # breakpoint()
```
alaeddine-13:
```suggestion
```

tests/unit/serve/runtimes/gateway/grpc/test_grpc_gateway_runtime.py
```diff
@@ -158,6 +158,20 @@ async def task_wrapper():
...
+    req = request_generator('/', DocumentArray([Document(text='client0-Request')]))
+    async for resp in streamer.stream(request_iterator=req):
+        responses.append(resp)
+    for req in request_generator(
+        '/',
+        DocumentArray([Document(text='client0-Request')]),
+    ):
+        unary_response = await streamer.process_single_data(request=req)
+        responses.append(unary_response)
+    return responses
```
alaeddine-13, line 172: isn't it better to add option stream as a parameter to this function and as a parameter to the test ? that way we'll be testing each feature on its own

jina/clients/request/asyncio.py
```diff
@@ -46,15 +46,17 @@ async def request_generator(
...
+            async with aiter(aiostream.stream.chunks(data, request_size)) as streamer:
```
girishc13, line 50: @alaeddine-13 The `aiter` function is available only from python version 3.10.

alaeddine-13: then we only do aiter and async with if python version is >3.10

tests/unit/serve/runtimes/gateway/grpc/test_grpc_gateway_runtime.py
girishc13: It can be done. The tests will be parametrized and will run both anyways. If it's about isolating tests then I'll refactor.

jina/clients/base/grpc.py
```diff
@@ -71,6 +137,7 @@ async def _get_results(
         max_backoff: float = 0.1,
         backoff_multiplier: float = 1.5,
         results_in_order: bool = False,
+        stream: bool = True,
```
JoanFM, line 140: shouldn't it be False by default?

jina/schemas/helper.py
```diff
@@ -10,7 +10,7 @@ def _python_type_to_schema_type(p):
     elif p == 'dict':
         dtype = 'object'
     else:
-        dtype = None
+        dtype = 'null'
```
JoanFM, line 13: what is this?

JoanFM (changes requested):
Shouldn't it be False by default?

JoanFM (changes requested):
It needs some mentioning in Documentation

jina/clients/base/grpc.py
girishc13: I wanted to get common understanding and consensus before making the unary call default. If there are no objections or doubt I will make the unary call default and update the documentation.

jina/schemas/helper.py
girishc13: The `tests/unit/schemas/test_schema.py` test was failing because None is not a valid json data type. It might have been enforced in the latest version.

docs/concepts/client/send-receive-data.md
```diff
@@ -196,6 +196,17 @@ with f:  # Using it as a Context Manager will start the Flow
 This will send the request to all Executors whose names start with 'bar', such as 'barExecutor'.
 In the simplest case, you can specify a precise Executor name, and the request will be sent only to that single Executor.
 
+## Use Unary or Streaming gRPC
+
+The Flow with **gRPC** protocol implements the unary and the streaming RPC lifecycle for communicating with the clients. When sending more than one request using the batching or the iterator mechanism, the RPC lifecycle for the {meth}`~jina.clients.mixin.PostMixin.post` method can be controlled using the `stream` boolean method argument. By default the stream option is set to `False` which uses the unary RPC to send the data to the Flow. If the stream option is set to `True` the streaming RPC is used to send the data to the Flow.
```
JoanFM, line 201: I would mention that behavior should not change

girishc13: I provided the link to the grpc documentation in the `hint` section. Should I be more explicit?

JoanFM: good then

jina/clients/base/grpc.py
JoanFM: put this default to False as the public API calling?

JoanFM (changes requested):

JoanFM (approved):

JoanFM (approved):

PR #5508, opened 2022-12-12
feat: dev install requirements

JoanFM:
**Goals:**
Add install_requirements capabilities to local Executors


jina/parsers/orchestrate/pod.py
```diff
@@ -114,6 +114,12 @@ def mixin_pod_parser(parser, pod_type: str = 'worker'):
             default=False,
             help='If set, the Gateway will restart while serving if the YAML configuration source is changed.'
         )
+    gp.add_argument(
+        '--install-requirements',
+        action='store_true',
+        default=False,
+        help='If set, try to install `requirements.txt` from the local Executor if exists in the Executor folder. If using Hub, install `requirements.txt` in the Hub Executor bundle to local.'
+    )
```
alaeddine-13, line 122: should not be added to the gateway in my opinion

alaeddine-13: I don't think this is much different from mixin_hub_pull_options_parser which comes from hubble.
I suggest just editing the hubble sdk rather than re-adding the argument

jina/parsers/orchestrate/pod.py
```diff
@@ -227,3 +233,18 @@ def mixin_pod_runtime_args_parser(arg_group, pod_type='worker'):
...
+    """
+
+    gp = add_arg_group(parser, title='Pull')
+    gp.add_argument(
+        '--force-update',
+        '--force',
+        action='store_true',
+        default=False,
+        help='If set, always pull the latest Hub Executor bundle even it exists on local',
+    )
```
alaeddine-13, line 250: I think we should just use the mixin mixin_hub_pull_options_parser from hubble, shouldn't take time to have hubble sdk release new version

samsja (approved):
lgtm

jina/orchestrate/pods/helper.py
```diff
@@ -88,3 +129,176 @@ def update_runtime_cls(args, copy=False) -> 'Namespace':
...
+def _get_package_path_from_uses(uses: str) -> Optional['Path']:
+    try:
+        from jina.jaml.helper import parse_config_source
+        from pathlib import Path
+        return Path(os.path.dirname(os.path.abspath(uses)))
+    except Exception as e:
+        # it may be a class or module
+        from jina.logging.predefined import default_logger
+        default_logger.warning(f'Error getting the directory name from {uses}: Exception {e}. `--install-requirements` option is only valid when `uses` is a configuration file.')
+        return None
```
alaeddine-13, line 304:
```suggestion
    if isinstance(uses, str) and os.path.exists(uses):
        from pathlib import Path
        return Path(os.path.dirname(os.path.abspath(uses)))
    else:
        from jina.logging.predefined import default_logger
        default_logger.warning(f'Error getting the directory name from {uses}: Exception {e}. `--install-requirements` option is only valid when `uses` is a configuration file.')
        return None
```

jina/orchestrate/pods/helper.py
```diff
@@ -88,3 +129,176 @@ def update_runtime_cls(args, copy=False) -> 'Namespace':
         _args.runtime_cls = 'HeadRuntime'
 
     return _args
+
+
+def _parse_fragment(fragment_string: str) -> Dict[str, str]:
+    """Takes a fragment string nd returns a dict of the components
```
alaeddine-13, line 135:
```suggestion
    """Takes a fragment string and returns a dict of the components
```

jina/orchestrate/pods/helper.py
```diff
@@ -11,11 +14,49 @@
 
 if TYPE_CHECKING:  # pragma: no cover
     from argparse import Namespace
+    from pathlib import Path
+
+VCS_SCHEMES = [
```
alaeddine-13, line 19: can't we just import all what is after this from hubble sdk ?

JoanFM: We cannot guarantee that hubble-sdk will mantain this

jina/parsers/orchestrate/pod.py
JoanFM: I do not think Hubble-sdk should redefine this on their end

jina/orchestrate/pods/helper.py
JoanFM: this is not hubble-sdk core and agreed with hubble that they should not care about installing from local

jina/parsers/orchestrate/pod.py
JoanFM: I do not want to force hubble sdk to change because of this

PR #5513, opened 2022-12-13
feat: using unary rpc from client respects results_in_order

JoanFM:
**Goals:**
Make unary_request really asynchronous and have it feel as streaming (Fast requests can be returned before slow ones). We need to add the concept of prefetch to this one in order to avoid overloading.


girishc13:
LGTM

jina/clients/base/grpc.py
```diff
@@ -122,24 +129,36 @@ async def _unary_rpc(
...
+            on_done: 'CallbackFnType',
+            on_error: Optional['CallbackFnType'] = None,
+            on_always: Optional['CallbackFnType'] = None,
+            compression: Optional[str] = None,
+            max_attempts: int = 1,
+            initial_backoff: float = 0.5,
+            max_backoff: float = 0.1,
+            backoff_multiplier: float = 1.5,
+            results_in_order: bool = False,
+            stream: bool = False,
```
girishc13, line 160: Small issue that we can fix later. This default value is different than the default value in the `.post` method. I forgot to update the default here last time.

girishc13 (approved):

PR #5517, opened 2022-12-14
chore: test on python 3.8 in CI/CD and use python version matrix for import test

alaeddine-13:
chore: test on python 3.8 in CI/CD and use python version matrix for import test


.github/workflows/cd.yml
```diff
@@ -167,15 +167,16 @@ jobs:
       matrix:
         core: ['', 'true']
         perf: ['', 'true']
+        python-env: [3.7, 3.8, 3.9, 3.10, 3.11]
```
samsja, line 170:
```suggestion
        python-env: ['3.7', '3.8', '3.9', '3.10', '3.11']
```

this is needed otherwise it try to install python 3.1 instead of 3.10

samsja (approved):

.github/workflows/ci.yml
```diff
@@ -399,15 +399,16 @@ jobs:
       matrix:
         core: ['', 'true']
         perf: ['', 'true']
+        python-env: [3.7, 3.8, 3.9, 3.10, 3.11]
```
alaeddine-13, line 402:
```suggestion
        python-env: ['3.7', '3.8', '3.9', '3.10', '3.11']
```

setup.py
```diff
@@ -149,6 +149,15 @@ def get_extra_requires(path, add_all=True):
...
     final_deps = perf_deps
 
+if sys.version_info.major == 3 and sys.version_info.minor >= 11:
+    grpcio_deps = []
+    for dep in final_deps:
+        if dep.startswith('grpcio'):
+            grpcio_deps.append(dep)
+    for dep in grpcio_deps:
+        final_deps.remove(dep)
+    final_deps.add('grpcio>=1.49.0')
```
JoanFM, line 159: u need some more grpcio related right?

JoanFM (changes requested):

setup.py
Jackmin801: ah yeap

Jackmin801:
BaseException needs to be used instead of Exception for asyncio exception for python >3.8
https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError
image

JoanFM (dismissed):

samsja (approved):

PR #5526, opened 2022-12-15
refactor: make server link to gateway so that custom gateway can inherit

JoanFM:
**Goals:**
Make the `GRPCGateway` be the one implementing the methods, so that we can have the CustomGateway developer inherit from it

### DIFF
--- a/jina/serve/gateway.py
+++ b/jina/serve/gateway.py
@@ -1,22 +1,12 @@
 import abc
-import argparse
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Dict, Optional, Sequence
+from typing import Dict, Optional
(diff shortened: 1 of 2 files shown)


jina/serve/streamer.py
```diff
@@ -189,8 +189,6 @@ async def close(self):
         await self._streamer.wait_floating_requests_end()
         await self._connection_pool.close()
 
-    Call = stream
```
alaeddine-13, line 192: I think for (internal) compatibility there's no harm in keeping the streamer bindable to the grpc server

JoanFM: yes, no harm at all u are right

PR #5530, opened 2022-12-15
test: add more tests to stream False

JoanFM:
**Goals:**
Try to show problems to improve CI


alaeddine-13 (commented):
since we change the way we import objects in the code base, I suggest adding more items to the `import-test` job in both ci and cd.
Something more exaustive: `python -c 'from jina import Executor,requests,Client,Flow,dynamic_batching'`

alaeddine-13 (approved):

PR #5532, opened 2022-12-16
fix: support composite gateway for k8s export

alaeddine-13:
fix: support composite gateway for k8s export


JoanFM (approved):

JoanFM (changes requested):
Can u add an example to the `k8s_tests`?

JoanFM (approved):

JoanFM (approved):

PR #5557, opened 2022-12-26
feat: add secrets concept to Jina(k8s)

### LINKED ISSUE #5238: Proposal feature: Add Secrets object concept to Jina
**Problem**
Right now we have a problem with Jina that users need to pass in code or YAML sensitive information as tokens, passwords etc in `Jina YAMLs` or `Jina Python scripts`.

This exposes our users security.

**Proposal**
As an MLOps framework, we can create our own Secret concept  analoguous to the Kubernetes Secret concept. The main difference is that in the Jina core we will not store or manage these secrets, only expose the logic on how these secrets have to be passed and read from the `Executors (microservices)`.

I suggest to create `secret` objects in Jina that can be attached to Flow (Gateway) or Executors both in YAML and in Python. Something similar to this (THESE DETAILS COULD CHANGE):

```yaml
jtype: Flow
secrets:
  - name: secret_gateway
    type: env
    key: JINA_SECRET_GATEWAY
executors:
  - name: preprocessor
    secrets:
      - name: secret_executor
        type: env
        key: JINA_SECRET1
    uses_with:
      confidential: ${{ SECRET.JINA_SECRET_EXECUTOR }}
```

```python
from jina import secrets

f = Flow(secrets=[{'name': 'secret_gateway', 'type': 'env','key': 'JINA_SECRET_GATEWAY'}]).add(secrets=[{'name': 'secret_executor', 'type': 'env','key': 'JINA_SECRET1'}], uses=PreProcessor, uses_with={'admin_emails': 'secrets.secret_executor'})
```

**User journey**
I believe the user journey can make sense accross all the ways Jina can be used. The same way `Kubernetes` separates the concerns of `creating` SECRETS and reading or accessing them in Pods we can do the same in Jina.

- Reading and accessing SECRETS in code is handled by these objects and the YAML python representation/interface

- Creating the SECRETS would remain in the user responsability in the following ways:
     - When running Flows locally, simply set environment variables or store information in file system
     - When running with Kubernetes, make sure to create K8s screts, `f.to_kubernetes_yaml` will do best effort to map `secrets` object to K8s secrets objects
     - JCloud would eventually offer a `secrets` API to manage these objects before a Flow is created in JCloud

**Example**

With this feature, this PR #5221 could work like this:

```YAML

jtype: Flow
secrets:
  - name: secret_external_deployment
    type: env
    key: JINA_SECRET_CAS
executors:
  - name: clip-as-service
    external: True
    grpc_metadata: 
         - authentication_token: ${{SECRETS.JINA_SECRET_CAS}}
```
This would allow the user to store this `Flow.yml` without storing confidential or security problematic information

**Extra helpers**
Maybe we can provide extra helpers in Executors to read automatically from secrets in the future

**Implementation**

- [ ] Create Secrets objects and define their key-value structure
- [ ] Make them a Deployment argument in parser and make sure they are handled by JAMLCompatible
- [ ] Make sure in the local Flow YAML, the envs are properly passed to the new processed and that the ${{SECRETS.XXX}} or any agreed syntax is properly implemented to access the `secrets` INSIDE the `Executor` or `Gateway` code
- [ ] Handle in `to_kubernetes_yaml` with best effort the transformation to the `readFromEnv` and `readFromValue` for `Deployments`
- [ ] JCloud offer an API to expose that easily in the K8s cluster

AnneYang720:
**Goals:**

- resolves #5238 
- [x] add `env_from_secret` to parser
- [x] support `env_from_secret` in k8s yaml generation
- [x] add related tests


tests/k8s/test_k8s.py
```diff
@@ -590,6 +603,86 @@ async def test_flow_with_configmap(k8s_flow_configmap, docker_images, tmpdir, lo
...
+            namespace,
+            command=['sh', '-c', 'echo $SECRET_PASSWORD'],
+            stderr=True,
+            stdin=False,
+            stdout=True,
+            tty=False,
+        ).strip("\n")
+
+        assert username == 'jina'
+        assert password == '123456'
```
alaeddine-13, line 679: maybe also one more assert where we send a request to the executor and it returns the value of those variables ? just add an extra endpoint to test-executor

docs/cloud-nativeness/k8s.md
```diff
@@ -58,7 +58,7 @@ Here are possible configuration options you may need to add or change
 - Add `requests` and `limits` for the resources of the different Pods 
 - Set up persistent volume storage to save your data on disk
 - Pass custom configuration to your Executor with `ConfigMap` 
-- Manage credentials of your Executor with secrets
+- Manage credentials of your Executor with secrets, you can use `f.add(..., env_from_secret={'SECRET_PASSWORD': {'name': 'mysecret', 'key': 'password'}})` to map them to Pod environment variables
```
alaeddine-13, line 61: Let's also say that this (for now) will only work for kubernetes secrets

alaeddine-13 (commented):
I believe we need more docs in https://docs.jina.ai/cloud-nativeness/kubernetes/. We can have a separate section where we explain how to define secrets and create/use them with jina and k8s

docs/cloud-nativeness/kubernetes.md
```diff
@@ -208,6 +208,39 @@ If you already have the simple Flow from the first example running on your clust
...
+            'SECRET_USERNAME': {'name': 'mysecret', 'key': 'username'},
+            'SECRET_PASSWORD': {'name': 'mysecret', 'key': 'password'},
+        },
+    )
+)
+
+f.to_kubernetes_yaml('./k8s_flow', k8s_namespace='custom-namespace')
+```
+
+After creating the namespace, you need to create the secrets metioned above:
```
JoanFM, line 235: Put a link to `secrets` in `kubernetes` so people know about

docs/concepts/flow/executor-args.md
```diff
@@ -12,6 +12,7 @@
 | `native` | If set, only native Executors is allowed, and the Executor is always run inside WorkerRuntime. | `boolean` | `False` |
 | `uses` | The config of the executor, it could be one of the followings:<br>        * the string literal of an Executor class name<br>        * an Executor YAML file (.yml, .yaml, .jaml)<br>        * a Jina Hub Executor (must start with `jinahub://` or `jinahub+docker://`)<br>        * a docker image (must start with `docker://`)<br>        * the string literal of a YAML config (must start with `!` or `jtype: `)<br>        * the string literal of a JSON config<br><br>        When use it under Python, one can use the following values additionally:<br>        - a Python dict that represents the config<br>        - a text file stream has `.read()` interface | `string` | `BaseExecutor` |
 | `uses_with` | Dictionary of keyword arguments that will override the `with` configuration in `uses` | `object` | `None` |
+| `env_from_secret` | Dictionary of keyword arguments that will override the `env_from_secret` configuration | `object` | `None` |
```
JoanFM, line 15: Are we sure is the good description?

jina/orchestrate/deployments/config/helper.py
```diff
@@ -112,6 +114,8 @@ def construct_runtime_container_args(cargs, uses_metas, uses_with, pod_type):
         container_args.extend(['--uses-metas', json.dumps(uses_metas)])
     if uses_with is not None:
         container_args.extend(['--uses-with', json.dumps(uses_with)])
+    if env_from_secret is not None:
+        container_args.extend(['--env-from-secret', json.dumps(env_from_secret)])
```
JoanFM, line 118: I do not think `env_from_secret` should be added to the `container`. This is an `orchestration` argument and should not be seen by the `runtime`

jina/orchestrate/deployments/config/k8s.py
```diff
@@ -125,10 +125,11 @@ def _get_image_name(self, uses: Optional[str]):
         def _get_container_args(self, cargs, pod_type):
             uses_metas = cargs.uses_metas or {}
             uses_with = self.deployment_args.uses_with
+            env_from_secret = self.deployment_args.env_from_secret
```
JoanFM, line 128: no need of this as well

jina/parsers/orchestrate/runtimes/worker.py
```diff
@@ -39,6 +39,15 @@ def mixin_worker_runtime_parser(parser):
     Dictionary of keyword arguments that will override the `with` configuration in `uses`
     ''',
     )
+    gp.add_argument(
+        '--env-from-secret',
+        action=KVAppendAction,
+        metavar='KEY: VALUE',
+        nargs='*',
+        help='''
+    Dictionary of keyword arguments that will override the `env_from_secret` configuration
```
JoanFM, line 48: improve description here. Also, the `parser` is not the correct, this should be in the `deployment` mixin

tests/k8s/test_k8s.py
```diff
@@ -590,6 +603,69 @@ async def test_flow_with_configmap(k8s_flow_configmap, docker_images, tmpdir, lo
...
+
+    try:
+        dump_path = os.path.join(str(tmpdir), 'test-flow-with-env-from-secret')
+        k8s_flow_env_from_secret.to_kubernetes_yaml(dump_path, k8s_namespace=namespace)
+
+        # create namespace
+        core_client.create_namespace(client.V1Namespace(metadata=client.V1ObjectMeta(name=namespace)))
+
+        # create secret
+        secret = client.V1Secret(
```
JoanFM, line 629: put this in a `fixture` I would say makes it more clear and clean

JoanFM (changes requested):

docs/cloud-nativeness/kubernetes.md
```diff
@@ -208,6 +208,39 @@ If you already have the simple Flow from the first example running on your clust
...
```
alaeddine-13, line 235:
```suggestion
After creating the namespace, you need to create the secrets mentioned above:
```

docs/cloud-nativeness/kubernetes.md
```diff
@@ -208,6 +208,39 @@ If you already have the simple Flow from the first example running on your clust
...
+        env={'k1': 'v1', 'k2': 'v2'},
+        env_from_secret={
+            'SECRET_USERNAME': {'name': 'mysecret', 'key': 'username'},
+            'SECRET_PASSWORD': {'name': 'mysecret', 'key': 'password'},
+        },
+    )
+)
+
+f.to_kubernetes_yaml('./k8s_flow', k8s_namespace='custom-namespace')
+```
```
alaeddine-13, line 233: can we create 2 examples ?
first one is in python (the current example) and the second one is with flow YAML and using command `jina export ...` command. Use a tab to separate between both examples

JoanFM (approved):

PR #5558, opened 2022-12-27
feat: Support communicating with individual Executors in CustomGateway

### LINKED ISSUE #5538: Support communicating with individual Executors with GatewayStreamer
When implementing a [custom gateway class](https://docs.jina.ai/concepts/gateway/customization), users need a [GatewayStreamer](https://docs.jina.ai/concepts/gateway/customization/#calling-executors-with-gatewaystreamer) to interact with Executors inside the gateway runtime.
The GatewayStreamer currently respects the Flow structure and only allows sending requests to the whole flow.

We want to make it possible to send requests to individual Executors using the GatewayStreamer.
This means we need to:
1. define how to select an executor in GatewayStreamer:
```python
# option 1, method to select an executor
connection = self.streamer.get_executor_connection('exec_name')

# option 2
connection = self.streamer['exec_name']

# option 3
self.executors['exec_name']
# feel free to come up with more options
```

2. When an executor is selected in GatewayStreamer, we need to wrap the connection pool with a class `ExecutorConnection` (feel free to come up with a better naming) that reuses the same connection pool and exposes an API to send requests using the connection pool.
The API of this class should contain methods that:
* receive documents
* convert documents to requests
* send requests to the right connection pool using a user provided endpoint
* convert responses back to documents and return them
Ideally, the API should be either similar to streamer's stream_docs or similar to Jina Client.

Jackmin801:
**Goals:**

- resolves #5538 
- [x] Add tests
- [x] Implement intended API


Jackmin801:
Gateway code:
```python

from jina.serve.runtimes.gateway.http.fastapi import FastAPIBaseGateway
from jina import Flow, Executor, requests

class MyGateway(FastAPIBaseGateway):
    @property
    def app(self):
        from fastapi import FastAPI

        app = FastAPI(title='Custom FastAPI Gateway')

        @app.get("/endpoint")
        async def get(text: str):
            docs = await self.executor['executor1'](exec_endpoint='/', docs=DocumentArray([Document(text=text)]), parameters={'k': 'v'})
            return {'result': docs[0].text}

        return app


class FirstExec(Executor):
    @requests
    def func(self, docs, **kwargs):
        for doc in docs:
            doc.text += " First"

class SecondExec(Executor):
    @requests
    def func(self, docs, **kwargs):
        print(kwargs)
        for doc in docs:
            doc.text += " Second"

with Flow(port=12345).config_gateway(uses=MyGateway, protocol='http', port=50975).add(
    uses=FirstExec, name='executor0'
).add(
    uses=SecondExec, name='executor1'
) as flow:
    flow.block()

```

Client code

jina/serve/streamer.py
```diff
@@ -224,31 +224,33 @@ def _set_env_streamer_args(**kwargs):
...
-                exec_endpoint=exec_endpoint,
-                docs=docs,
-                parameters=parameters,
-                request_size=request_size,
-                return_results=return_results,
-                results_in_order=results_in_order,
-                target_executor=executor_name
-            ):
-                result_docs.extend(_docs)
+            from jina.clients.request import request_generator
```
JoanFM, line 238: avoid this `request_generator`. Create a Request using the `Request` object. This `generator I would love to remove` since it is too complex

JoanFM (changes requested):

jina/serve/streamer.py
Jackmin801: ah ok then

docs/concepts/gateway/customization.md
```diff
@@ -259,6 +259,44 @@ Jina provides the Gateway with a list of ports and protocols to expose. Therefor
 on multiple ports using different protocols.
 ```
 
+(executor-streamer)=
+## Calling an Individual Executor with {class}`~jina.serve.streamer.ExecutorStreamer`
```
JoanFM, line 263: I would hide the class itself from the title, just ## Calling an Individual Executor

docs/concepts/gateway/customization.md
```diff
@@ -259,6 +259,44 @@ Jina provides the Gateway with a list of ports and protocols to expose. Therefor
...
+{class}`~jina.serve.streamer.ExecutorStreamer` allows you to interface with individual Executors from the Gateway. An instance of 
+this class knows where each Executor lives and allows you to call them individually. 
+An `executor` object (instance of {class}`~jina.serve.streamer.ExecutorStreamer`) is injected by Jina to your gateway class.
+
+After transforming requests that arrive to the gateway server into Documents, you can call the Executor in your Python code 
+using `self.executor['executor_name'](args)`.
```
JoanFM, line 269: the property should be `plural`

docs/concepts/gateway/customization.md
```diff
@@ -259,6 +259,44 @@ Jina provides the Gateway with a list of ports and protocols to expose. Therefor
...
+        @app.get("/endpoint")
+        async def get(text: str):
+            docs = await self.executor['executor1'](exec_endpoint='/', docs=DocumentArray([Document(text=text)]), parameters={'k': 'v'})
+            return {'result': [doc.text for doc in docs]}
+
+        return app
+```
+
+Keep in mind that calling Executors individually bypasses the Flow topology.
+You should only use this method for testing or if your use case is simple and does not require a pipeline of executors.
```
JoanFM, line 297: Avoid this, you can use it whenever u feel you need it

docs/concepts/gateway/customization.md
```diff
@@ -259,6 +259,44 @@ Jina provides the Gateway with a list of ports and protocols to expose. Therefor
...
+Otherwise, it is recommended to construct a Flow and use the {class}`~jina.serve.streamer.GatewayStreamer` to send Documents through the entire Flow.
```
JoanFM, line 298: Avoid this for now

JoanFM (changes requested):

jina/serve/streamer.py
```diff
@@ -221,3 +221,34 @@ def get_streamer():
     @staticmethod
     def _set_env_streamer_args(**kwargs):
         os.environ['JINA_STREAMER_ARGS'] = json.dumps(kwargs)
+
+
+class ExecutorStreamer:
```
JoanFM, line 226: This is not a good name, unless the `request_size` is something you consider

jina/serve/streamer.py
```diff
@@ -221,3 +221,34 @@ def get_streamer():
...
+    def __init__(self, connection_pool: GrpcConnectionPool) -> None:
+        self._connection_pool: GrpcConnectionPool = connection_pool
+
+    def __getitem__(self, executor_name: str):
+        async def _acall_executor(
+            docs: DocumentArray,
+            request_size: int = 100,
```
JoanFM, line 233: the request size is not respected, here you are not sending multiple `requests`

tests/unit/serve/gateway/test_gateway.py
```diff
@@ -157,3 +157,50 @@ def test_custom_gateway_no_executors(uses, uses_with, expected):
...
+        def func(self, docs, **kwargs):
+            for doc in docs:
+                doc.text += ' THIS SHOULD NOT HAVE HAPPENED!'
+    class SecondExec(Executor):
+        @requests
+        def func(self, docs, parameters, **kwargs):
+            for doc in docs:
+                doc.text += f' Second(parameters={str(parameters)})'
+
+    with Flow().config_gateway(uses=MyGateway, protocol=protocol).add(uses=FirstExec, name='executor0').add(uses=SecondExec, name='executor1') as flow:
```
JoanFM, line 203: how is this test supposed to work with `grpc` as protocol?

JoanFM (changes requested):

jina/serve/streamer.py
```diff
@@ -221,3 +221,34 @@ def get_streamer():
     @staticmethod
     def _set_env_streamer_args(**kwargs):
         os.environ['JINA_STREAMER_ARGS'] = json.dumps(kwargs)
+
+
+class ExecutorStreamer:
+    def __init__(self, connection_pool: GrpcConnectionPool) -> None:
+        self._connection_pool: GrpcConnectionPool = connection_pool
+
+    def __getitem__(self, executor_name: str):
```
JoanFM, line 230: also we wanted to have something `called` `post` to unify with the interface

JoanFM (changes requested):

docs/concepts/gateway/customization.md
Jackmin801: Oh but it is only one `ExecutorStreamer` though. The one class knows about all the executors

jina/serve/streamer.py
Jackmin801: Should we batch the requests then? Or keep this one simple and not have `request_size`?

jina/serve/streamer.py
Jackmin801: So the API is like this?:
```python
da = await self.executor['executor1'].post(exec_endpoint='/', docs=DocumentArray([Document(text=text)]), parameters={'k': 'v'})
```

tests/unit/serve/gateway/test_gateway.py
Jackmin801: To the client, it doesnt seem to make any difference. I was hoping you or @alaeddine-13 would know something

docs/concepts/gateway/customization.md
```diff
@@ -259,6 +259,44 @@ Jina provides the Gateway with a list of ports and protocols to expose. Therefor
...
+
+
+class MyGateway(FastAPIBaseGateway):
+    @property
+    def app(self):
+        app = FastAPI()
+
+        @app.get("/endpoint")
+        async def get(text: str):
+            docs = await self.executor['executor1'](exec_endpoint='/', docs=DocumentArray([Document(text=text)]), parameters={'k': 'v'})
```
alaeddine-13, line 290: I think this API is not as flexible as it should be. If we want to add a different behavior (for instance async/sync method, batching,...) we can't just add a new method.

jina/serve/streamer.py
JoanFM: this we should discuss now

tests/unit/serve/gateway/test_gateway.py
alaeddine-13: it looks like we don't make health checks to the gateway for local flows, probably that's why the test passes

alaeddine-13 (commented):
I suggest 2 options:
* keep introducing ExecutorStreamer, I think in this case it makes sense to rename GatewayStreamer to FlowStreamer. Since we are making a breaking change, it makes sense to unify the interface as well. This means having:
- self.flow_streamer.post(...)  # or self.flow.post(...)
- self.executor_streamer[exec].post(...) # or self.executors[exec].post(...)

* unify flow streamer and executor streamer under the same self.streamer object. I think this makes sense because otherwise, since the same pool is used, destroying one object corrupts the other

docs/concepts/gateway/customization.md
Jackmin801: Removed references to the class as suggested so this is not longer relevant I think

jina/serve/streamer.py
Jackmin801: decided to respect request_size

jina/serve/streamer.py
Jackmin801: decided to respect request_size

jina/serve/streamer.py
Jackmin801: implemented this API

docs/concepts/gateway/customization.md
Jackmin801: Ah ok. The implementation has been changed to conform to this

tests/unit/serve/gateway/test_gateway.py
JoanFM: We should be making health check at the start when we wait for it to be ready.

JoanFM: So please clarify this

JoanFM (changes requested):

docs/concepts/gateway/customization.md
```diff
@@ -224,6 +224,37 @@ class MyGateway(FastAPIBaseGateway):
...
 ```
 
+(executor-streamer)=
+## Calling an Individual Executor
+An `executor` object is injected by Jina to your gateway class which allows you to interface with individual Executors from the Gateway.
+
+After transforming requests that arrive to the gateway server into Documents, you can call the Executor in your Python code using `self.executor['executor_name'].post(args)`.
+This method expects a DocumentArray object and an endpoint exposed by the Executor (similar to Jina Client). 
+It returns a 'coroutine' which returns a DocumentArray.
+
```
alaeddine-13, line 234:
```suggestion
Check the method documentation for more information: {meth}`~ jina.serve.streamer._ExecutorStreamer.post()`

```

docs/concepts/gateway/customization.md
```diff
@@ -224,6 +224,37 @@ class MyGateway(FastAPIBaseGateway):
...
+Here is an example of calling an individual Executor named 'executor1':
```
alaeddine-13, line 235: I think we should give a more general example, for instance, adding a bunch of Executors to the Flow and calling one of them explicitly:
Flow.add(name='executor1', uses=...).add(name='executor2',...)...
// then show how to call one of them
Maybe we should also showcase how to call 2 executors in parallel using asyncio.

jina/serve/streamer.py
```diff
@@ -221,3 +221,36 @@ def get_streamer():
...
+        reqs = []
+        for docs_batch in inputs.batch(batch_size=request_size, shuffle=False):
+            req = DataRequest()
+            req.header.exec_endpoint = on
+            req.header.target_executor = self.executor_name
+            req.parameters = parameters
+            req.data.docs = docs_batch
+            reqs.append(req)
+
+        resp, _ = await self._connection_pool.send_requests_once(
```
JoanFM, line 248: do tests where u send more than one request, I believe this interface is not intended to be used like this.

Every `req` should lead to a `send_request` call and u should collect docs afterwards.

JoanFM (changes requested):

jina/serve/streamer.py
```diff
@@ -221,3 +221,36 @@ def get_streamer():
     @staticmethod
     def _set_env_streamer_args(**kwargs):
         os.environ['JINA_STREAMER_ARGS'] = json.dumps(kwargs)
+
+
+class _ExecutorStreamer:
```
alaeddine-13, line 226: @JoanFM  is it worth it to have an abstract client class common to both the post mixin and this class ?

docs/concepts/gateway/customization.md
Jackmin801: Yea sure. Done

tests/unit/serve/gateway/test_gateway.py
```diff
@@ -157,3 +157,43 @@ def test_custom_gateway_no_executors(uses, uses_with, expected):
 
     assert gateway_process.exitcode == 0
     assert worker_process.exitcode == 0
+
+
+def test_stream_individual_executor():
```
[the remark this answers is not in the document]
Jackmin801: done

jina/serve/streamer.py
Jackmin801: ah ok done.

Jackmin801:
zz bot

tests/unit/serve/gateway/test_gateway.py
```diff
@@ -227,7 +227,8 @@ def app(self):
             @app.get("/endpoint")
             async def get(text: str):
                 docs = await self.executor['executor1'].post(on='/', inputs=DocumentArray([Document(text=f"{text} {i}") for i in range(20)]), parameters=PARAMETERS, request_size=5)
```
JoanFM, line 229: make sure to send more documents, otherwise u may not hit all replicas?

Jackmin801: ah good idea

JoanFM (dismissed):
Technically approve, wait for @alexcg1 for the docs

docs/concepts/gateway/customization.md
```diff
@@ -224,6 +224,113 @@ class MyGateway(FastAPIBaseGateway):
...
+(executor-streamer)=
+## Calling an Individual Executor
+An `executor` object is injected by Jina to your gateway class which allows you to call individual Executors from the Gateway.
+
+After transforming requests that arrive to the gateway server into Documents, you can call the Executor in your Python code using `self.executor['executor_name'].post(args)`.
+This method expects a DocumentArray object and an endpoint exposed by the Executor (similar to Jina Client). 
+It returns a 'coroutine' which returns a DocumentArray.
+Check the method documentation for more information: {meth}`~ jina.serve.streamer._ExecutorStreamer.post()`
+
+In this example, we have a Flow with two executors ('executor1' and 'executor2'). We can call them individually using `self.executor['executor_name'].post`:
```
alexcg1, line 236:
```suggestion
In this example, we have a Flow with two Executors ('executor1' and 'executor2'). We can call them individually using `self.executor['executor_name'].post`:
```
Ensure "Executor" has capital when used outside code blocks

docs/concepts/gateway/customization.md
```diff
@@ -224,6 +224,113 @@ class MyGateway(FastAPIBaseGateway):
         return app
 ```
 
+(executor-streamer)=
+## Calling an Individual Executor
+An `executor` object is injected by Jina to your gateway class which allows you to call individual Executors from the Gateway.
+
+After transforming requests that arrive to the gateway server into Documents, you can call the Executor in your Python code using `self.executor['executor_name'].post(args)`.
+This method expects a DocumentArray object and an endpoint exposed by the Executor (similar to Jina Client).
```
alexcg1, line 232: Add link to Jina Client

docs/concepts/gateway/customization.md
```diff
@@ -224,6 +224,113 @@ class MyGateway(FastAPIBaseGateway):
         return app
 ```
 
+(executor-streamer)=
+## Calling an Individual Executor
```
alexcg1, line 228:
```suggestion
## Calling an individual Executor
```
Sentence case for headings (except level 1 headings)

alexcg1 (changes requested):

PR #5560, opened 2022-12-28
fix: respect timeout_ready when generating startup probe

alaeddine-13:
fix: respect timeout_ready when generating startup probe configurations


jina/orchestrate/deployments/config/k8slib/kubernetes_deployment.py
```diff
@@ -77,6 +84,12 @@ def get_template_yamls(
     else:
         protocols = protocol
 
+    if timeout_ready == -1:
+        raise ValueError(
```
JoanFM, line 88: I would perhaps maybe set some default value on our end?

JoanFM:
if timeout ready is -1 lets rrmove startupProbe

JoanFM (changes requested):
please add a test

JoanFM (approved):

JoanFM (approved):

PR #5564, opened 2023-01-02
feat: add beta support to scale Stateful Executors with consensus using RAFT

JoanFM:
**Goals:**
Provide a PoC of how StatefulExecutors could handle with consensus

Implementation:

- [x] Handle raftadmin add voters from jina orchestrate
- [x] Handle terminate signals on the raft node
- [x] SIGTERM Shutdown handling
- [x] Logger in raft naming editing (if they merge and release)
- [x] Check the replica trying for indexing #5601
- [x] Make it work with shards (workspace handling)
- [x] Expose RAFT options in parser
- [ ] Panic if FSM Apply fails https://github.com/hashicorp/raft/issues/307 (We have some failure reasons like grpc connectivity that should not be there)
- [x] Implement proper Snapshot and Restore
- [x] properly create workspace folders and log/data files, without the need for explicitly setting `raft_bootstrap` parameter
- [x] Improve AddVoter calls (we do 10 for now until it works)
- [x] Improve readiness check, (node ready before restore is done), RAFT node should expose a health check service which is a proxy to executor health check service. It should be used by Pod.wait_for_ready. 
- [x] Implement bulding wheels for different platforms (Look at ANNLite for wheels building)
- [x] HANDLE LICENSE (since we do not change the code, no need to do anything)
- [x] Handle ContainerPod
- [x] Clean logs
- [x] RPC should implement streaming endpoint
- [x] currently, ports are being assigned in an underterministic way. The snapshots store information about each node in the cluster. When a cluster needs to be restored, we have to make sure it is spawned with the same ports attributed to replicas.
- [ ] Fix raft library compilation for emulated python on mac (will build to `arm` architecture and therefore can't be linked to `x86_64` emulated python)
- [x] Handle better reduction in Shards (Head)
- [ ] (OPTIONAL) Implement ApplyBatch in the FSM that will help on reusing client and connection
- [ ] (OPTIONAL) Make jina.pb.go independent of docarray proto. (Maybe assume DocArray is just binary?)
- [ ] (OPTIONAL) Study if we can call WorkerRequestHandler directly from Golang

Testing:

- [ ] Unit test golang module
- [x] Integration test
- [ ] Try killing nodes, see leadership changes

Further understanding:
- [ ] How to handle timeout leader in snapshotting (Right now FAIL, can we give up leadership?, run snapshot in new process (copy-on-write)?
- [x] Understand restore behavior
- [x] Understand vote, bootstrap, etc ...


docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,183 @@ executors:
 Flow with three replicas of `slow_encoder` and one replica of `fast_indexer`
 ```
 
+(scale-consensus)=
+## Replicate stateful Executors with consensus using RAFT (Beta)
+
+````{admonition} docarray 0.30
```
alexcg1, line 61:
```suggestion
````{admonition} DocArray 0.30
```
Please fix throughout

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,183 @@ executors:
...
+:class: note
+
+Since docarray version > 0.30, docarray changed its interface and implementation drastically. We intend to support these new docarray versions
+in the near future, but not every feature is yet available {ref}`Check here <docarray-v2>`. This feature has been added with these new docarray versions support.
```
alexcg1, line 65:
```suggestion
Starting from DocArray version 0.30, DocArray changed its interface and implementation drastically. We intend to support these new versions in the near future, but not every feature is yet available Check {ref}`here <docarray-v2>` for more information. This feature has been added with the new DocArray support.
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,183 @@ executors:
...
+
+This feature is only available when using `grpc` as the protocol for the Deployment or when the Deployment is part of a Flow
```
alexcg1, line 67:
```suggestion
This feature is only available when using `grpc` as the protocol for the Deployment or when the Deployment is part of a Flow.
```

Please ensure punctuation throughout

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,183 @@ executors:
...
+
+````{admonition} gRPC protocol
+:class: note
+
+This feature is only available when using gRPC as the protocol for the Deployment or when the Deployment is part of a Flow
+````
+
+Replication is used to scale-out Executors by creating copies of them that can handle requests in parallel providing better RPS.
+However, when an Executor keeps some sort of state, then it is not simple to guarantee that each copy of the Executor keeps the same state,
+which can lead to undesired behavior, since each replica can provide different results, depending on the specific state they hold.
```
alexcg1, line 78:
```suggestion
Replication is used to scale out Executors by creating copies of them that can handle requests in parallel, providing better RPS.
However, when an Executor maintains some sort of state, then it is not simple to guarantee that each copy of the Executor maintains the *same* state,
which can lead to undesired behavior, since each replica can provide different results depending on the specific state they hold.
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,183 @@ executors:
...
+
+In Jina, you can also have replication while guaranteeing the consensus between Executors. For this, we rely on (RAFT)[https://raft.github.io/] which is
+an algorithm that guarantees eventual consistency between replicas.
```
alexcg1, line 81:
```suggestion
In Jina, you can also have replication while guaranteeing the consensus between Executors. For this, we rely on (RAFT)[https://raft.github.io/], which is
an algorithm that guarantees eventual consistency between replicas. 
```

"Eventual" makes it sound like this could take a long time

alexcg1 (changes requested):
Browser is slow, so I'll pause this review now, then jump in and finish it later

docs/concepts/orchestration/scale-out.md
JoanFM: it is the proper definition of what it does

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
 Flow with three replicas of `slow_encoder` and one replica of `fast_indexer`
 ```
 
+(scale-consensus)=
+## Replicate stateful Executors with consensus using RAFT (Beta)
+
+````{admonition} DocArray 0.30
+:class: note
+
+Starting from DocArray version 0.30, DocArray changed its interface and implementation drastically. We intend to support these new versions in the near future, but not every feature is yet available Check {ref}`here <docarray-v2>` for more information. This feature has been added with the new DocArray support.
```
alexcg1, line 64:
```suggestion
Starting from DocArray version 0.30, DocArray changed its interface and implementation drastically. We intend to support these new versions in the near future, but not every feature is yet available. Check {ref}`here <docarray-v2>` for more information. This feature has been added with the new DocArray support.
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
...
+
+This feature is only available when using `grpc` as the protocol for the Deployment or when the Deployment is part of a Flow.
```
alexcg1, line 66:
```suggestion
```
This is duplicated below

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
...
+```
+
+But at search time, the consensus module does not affect, and only one replica serves the queries.
+```text
+INFO   indexer/rep-0@902 Searching Document in index with 100000 documents indexed                                                                                                              [04/28/23 16:59:21]
+INFO   indexer/rep-1@910 Searching Document in index with 100000 documents indexed                                                                                                              [04/28/23 16:59:21]
+INFO   indexer/rep-2@923 Searching Document in index with 100000 documents indexed 
+```
+
+If you run the same example by setting replicas to 1 without the consensus module, you can see the benefits it has in the QPS at search time,
```
alexcg1, line 239:
```suggestion
If you run the same example by setting `replicas` to `1` without the consensus module, you can see the benefits it has in the QPS at search time,
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
...
+while there is a little cost on the time used for indexing.
+
+```python
+d = Deployment(name='indexer',
+               port=5555,
+               uses=ExactNNSearch,
+               replicas=1)
+```
+
+With 1 replica:
```
alexcg1, line 249:
```suggestion
With one replica:
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
...
+
+```text
+Indexing 100000 Documents took 18.93274688720703s
+Searching 1000 Queries took 385.96641397476196s
+```
+
+With 3 replicas and consensus:
```
alexcg1, line 256:
```suggestion
With three replicas and consensus:
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
...
+```text
+Indexing 100000 Documents took 35.066415548324585s
+Searching 1000 Queries took 202.07950615882874s
+```
+
+This allows to go from 2.5 to 5 QPS.
```
alexcg1, line 262:
```suggestion
This increases QPS from 2.5 to 5.
```

alexcg1 (changes requested):

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,212 @@ executors:
...
+an algorithm that guarantees eventual consistency between replicas. 
+
+Consensus-based replication using RAFT is a distributed algorithm designed to provide fault tolerance and consistency in a distributed system. In a distributed system, the nodes may fail, and messages may be lost or delayed, which can lead to inconsistencies in the system.
+The problem with traditional replication methods is that they can't guarantee consistency in a distributed system in the presence of failures. This is where consensus-based replication using RAFT comes in.
+In this approach, each Executor can be considered as a Finite State Machine, which means that it has a set of states that it can be in, and a set of transitions that it can make between those states. Each request that is sent to the Executor can be considered as a log entry that needs to be replicated across the cluster.
+
+In order to enable this kind of replication, we need to consider:
+
+- Specify which methods of the Executor {ref}` could be updating its internal state <stateful-executor>`.
+- Tell the deployment to use the RAFT consensus algorithm by setting the `--stateful` argument.
```
JoanFM, line 88:
```suggestion
- Tell the Deployment to use the RAFT consensus algorithm by setting the `--stateful` argument.
```

docs/concepts/orchestration/scale-out.md
```diff
@@ -55,6 +55,183 @@ executors:
...
+               replicas=3, 
+               stateful=True, 
+               peer_ports=[12345, 12346, 12347])
+with d:
+    d.block()
+```
+
+This capacity allows us not only to have replicas work with robustness and availability, it also can help us to achieve higher throughput in some cases.
+
+Let's imagine we write an Executor that is used to index and query documents from a vector index.
```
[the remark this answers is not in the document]
JoanFM: as u want, let's do this yes

docs/concepts/serving/executor/create.md
```diff
@@ -11,9 +11,9 @@ An {class}`~jina.Executor` is a self-contained microservice exposed using a gRPC
 It contains functions (decorated with `@requests`) that process `DocumentArray`s. Executors follow three principles:
 
 1. An Executor should subclass directly from the `jina.Executor` class.
-2. An Executor class is a bag of functions with shared state or configuration (via `self`); it can contain an arbitrary number of
-functions with arbitrary names.
+2. An Executor is a Python class; it can contain any number of functions.
 3. Functions decorated by {class}`~jina.requests` are exposed as services according to their `on=` endpoint. These functions can be coroutines (`async def`) or regular functions. This will be explained later in {ref}`Add Endpoints Section<exec-endpoint>`
+4. (Beta) Functions decorated by {class}`~jina.serve.executors.decorators.write` on top of their {class}`~jina.requests` decoration, are considered to update the internal state of the Executor. The `__init__` and `close` methods are exceptions. The reasons this is useful is explained in {ref}`Stateful-executor<stateful-executor>`.
```
alexcg1, line 16:
```suggestion
4. (Beta) Functions decorated by {class}`~jina.serve.executors.decorators.write` above their {class}`~jina.requests` decoration, are considered to update the internal state of the Executor. The `__init__` and `close` methods are exceptions. The reasons this is useful is explained in {ref}`Stateful-executor<stateful-executor>`.
```

alexcg1 (changes requested):

jina/serve/consensus/README.md
```diff
@@ -0,0 +1,124 @@
+# Raft implementation for Jina Docarray
```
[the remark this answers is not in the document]
JoanFM: this is infomrmative for us

PR #5566, opened 2023-01-02
refactor: removing seemingly useless line

Jackmin801:
**Goals:**

- resolves #ISSUE-NUMBER
- ...
- ...

- ...
- ...


JoanFM (approved):

Jackmin801:
@JoanFM Do u know when that branch is ever entered? I took out that line because it seems like its useless even if the condition is met but is that condition `if isinstance(content, tuple) and len(content) == 2` ever met?

JoanFM:
> @JoanFM Do u know when that branch is entered? I took out that line because it seems like its useless even if the condition is met but the condition `if isinstance(content, tuple) and len(content) == 2` seems oddly specific. Does it serve a particular use case?

It may be leftovers from when we used to support groundtruths

PR #5568, opened 2023-01-03
refactor: remove dead code, improve test coverage and fix convoluted client code

Jackmin801:
**Goals:**

- [x] Take out groundtruth code in client
- [x] Improve client code coverage
- [x] Take out kwargs logic in data to doc parsing because it creates silent failures

### DIFF
--- a/jina/clients/request/helper.py
+++ b/jina/clients/request/helper.py
@@ -31,10 +31,10 @@ def _new_data_request(endpoint: str, target: Optional[str], parameters: Optional
 
 
 def _new_doc_from_data(
-    data, data_type: DataInputType, **kwargs
+    data, data_type: DataInputType
 ) -> Tuple['Document', 'DataInputType']:
     def _build_doc_from_content():
-        return Document(content=data, **kwargs), DataInputType.CONTENT
+        return Document(content=data), DataInputType.CONTENT
 
     if data_type == DataInputType.DICT:
         return Document.from_dict(data), DataInputType.DICT
@@ -45,8 +45,8 @@ def _new_doc_from_data(
         elif isinstance(data, dict):
             return Document.from_dict(data), DataInputType.DICT
         try:
-            d = Document(data, **kwargs)
-            return d, DataInputType.DOCUMENT
+            d = Document(data)
+            return d, DataInputType.DOCUMENT # NOT HIT
         except ValueError:
             # AUTO has a fallback, now reconsider it as content
             if data_type == DataInputType.AUTO:
--- a/jina/clients/request/helper.py
+++ b/jina/clients/request/helper.py
@@ -7,12 +7,12 @@ from jina.types.request.data import DataRequest
 
 
 def _new_data_request_from_batch(
-    _kwargs, batch, data_type: DataInputType, endpoint: str, target: Optional[str], parameters: Optional[dict]
+    batch, data_type: DataInputType, endpoint: str, target: Optional[str], parameters: Optional[dict]
 ):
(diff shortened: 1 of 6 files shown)


jina/clients/request/helper.py
```diff
@@ -59,13 +58,9 @@ def _build_doc_from_content():
...
+def _add_docs(req: DataRequest, batch, data_type: DataInputType, _kwargs):
     da = DocumentArray()
     for content in batch:
-        if isinstance(content, tuple) and len(content) == 2:
-            d, data_type = _new_doc_from_data(content[0], data_type, **_kwargs)
-            da.append(d)
-        else:
-            d, data_type = _new_doc_from_data(content, data_type, **_kwargs)
-            da.append(d)
+        d, data_type = _new_doc_from_data(content, data_type, **_kwargs)
```
JoanFM, line 64: u are removing these `_kwargs` I think it is not needed to spend time on this

JoanFM (changes requested):

jina/clients/request/helper.py
Jackmin801: I will try remove them and see if they still pass tests. But it is possible to access the _kwargs here.

The _kwargs I removed are in another function where it is created but never used

Jackmin801: It technically might be used here. But it seems to never need it in the tests.
https://github.com/jina-ai/jina/blob/master/jina/clients/request/helper.py#L37-L38

Jackmin801:
The groundtruth test somehow fails when kwargs are removed; not when the condition that handles kwargs was removed

Jackmin801:
Mystery Solved
---
With kwargs on, when _new_doc_from_data is entered it has these parameters:
```python
data = (<Document ('id',) at...9d321ea9c>, <Document ('id',) at...c1263d4c1>)
data_type = <DataInputType.DOCUMENT: 1> | <DataInputType.AUTO: 0> # The first pair hits the auto the subsequent pairs have DOCUMENT because of the first pair setting the data_type
kwargs = {'proxy': False, 'host': '0.0.0.0', 'port': 12345, 'tls': False, 'asyncio': False, 'tracing': False, 'traces_exporter_host': None, 'traces_exporter_port': None, 'metrics': False, 'metrics_exporter_host': None, 'metrics_exporter_port': None, 'protocol': <GatewayProtocolType.GRPC: 0>, 'prefetch': 1000}
```

In this case, the kwargs are interpreted as tags and the Document constructor ignores that the data doesnt make sense
```python
Document(data, **kwargs) = <Document ('id', 'tags') at 10ab62829cef221b0712a30dfe174ad3>
```

Without the kwargs, it realizes the data makes no sense
```python
ValueError("Failed to initialize docarray.document.Document from obj=(<Document ('id',) at 0da4c26c7cc2909ee7f7c589d321ea9c>, <Document ('id',) at 61558bc5fe3d78093abf840c1263d4c1>), kwargs={}")
```

JoanFM:
what is the test with groundtruth? this should be rmeoved

Jackmin801:
https://github.com/jina-ai/jina/blob/master/tests/integration/v2_api/test_func_routing.py#L23

Yup. I've converted it to not use the groundtruth tuple.

jina/clients/request/helper.py
Jackmin801: I think it's a good idea to remove it. As described in https://github.com/jina-ai/jina/pull/5568#issuecomment-1369785922, it creates weird silent failures.

JoanFM (approved):

PR #5573, opened 2023-01-05
feat: run dry_run requests as warmup with 5 min max timeout

### LINKED ISSUE #5467: feat: run warmup requests on runtime startup to ensure that service is ready to accept connections
**Describe the feature**

Warm up requests are executed before the service is reported ready so that new incoming requests can be readily served without having to create new network connections to database dependency or sidecar container, etc or load the required modules. Warm up requests can be dummy or canary or health check type requests that can trigger the hot path at least once. Warm up requests don't need to be successful but need to be executed before reporting readiness.

**Your proposal**

Gateway Runtime: execute service discovery request to all executors.
Executor Runtime: execute empty post request on the default endpoint.

---

**Environment**

**Screenshots**

girishc13:
**Goals:**

- resolves #5467
- trigger warmup requests using the `client.is_flow_ready()` method from the gateway. 
- this runs the `dry_run` rpc to each exectuor thereby establishing the first connection and gRPC channel. 
- the warmup sequence runs for a max of 5 minutes and exits if all dry_run responses are successful. On failure, the warmup is retried after 0.2 seconds. 
- when debug logging is enabled the gateway will log connections errors until the Flow is ready or the warmup reaches the max timeout of 5 minutes.

### DIFF
--- a/jina/serve/gateway.py
+++ b/jina/serve/gateway.py
@@ -1,4 +1,5 @@
 import abc
+import time
 from types import SimpleNamespace
 from typing import Dict, Optional
 
@@ -175,6 +176,31 @@ class BaseGateway(JAMLCompatible, metaclass=GatewayType):
     async def shutdown(self):
         """Shutdown the server and free other allocated resources, e.g, streamer object, health check service, ..."""
         ...
+
+    async def warmup(self):
+        '''Run client._is_flow_ready() request to trigger the dry_run endpoint on each executor.
+        This forces the gateway to establish connection and open a gRPC channel to each executor so that the first
+        request doesn't need to experience the penalty of eastablishing a brand new gRPC channel.
+        '''
+        self.logger.debug('Running warmup')
+        from jina import Client
+
+        timeout = time.time() + 60 * 5  # 5 minutes from now
+
+        while True:
+            dry_run_responses = []
+            for port, protocol in zip(self.ports, self.protocols):
+                client = Client(
+                    host=f'{protocol}://{self.runtime_args.host}:{port}',
+                    asyncio=True,
+                    continue_on_error=True,
+                )
+                dry_run_responses.append(await client.is_flow_ready())
+
+            if time.time() > timeout or all(dry_run_responses):
+                break
+
+            time.sleep(0.2)
 
     def __enter__(self):
         return self
--- a/jina/serve/gateway.py
+++ b/jina/serve/gateway.py
@@ -115,6 +115,8 @@ class BaseGateway(JAMLCompatible, metaclass=GatewayType):
             prefetch=self.runtime_args.prefetch,
         )
 
+        self._warmup_task = None
+
     def _add_runtime_args(self, _runtime_args: Optional[Dict]):
         from jina.parsers import set_gateway_runtime_args_parser
 
@@ -172,17 +174,17 @@ class BaseGateway(JAMLCompatible, metaclass=GatewayType):
         """Run server forever"""
         ...
 
-    @abc.abstractmethod
     async def shutdown(self):
         """Shutdown the server and free other allocated resources, e.g, streamer object, health check service, ..."""
-        ...
+        if self._warmup_task and not self._warmup_task.done():
+            self._warmup_task.cancel()
 
     async def warmup(self):
         '''Run client._is_flow_ready() request to trigger the dry_run endpoint on each executor.
         This forces the gateway to establish connection and open a gRPC channel to each executor so that the first
         request doesn't need to experience the penalty of eastablishing a brand new gRPC channel.
         '''
-        self.logger.debug('Running warmup')
+        self.logger.debug(f'Running warmup for {self.__class__}')
         from jina import Client
 
         timeout = time.time() + 60 * 5  # 5 minutes from now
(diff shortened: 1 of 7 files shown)


jina/serve/gateway.py
```diff
@@ -176,6 +177,31 @@ async def shutdown(self):
...
+        '''
+        self.logger.debug('Running warmup')
+        from jina import Client
+
+        timeout = time.time() + 60 * 5  # 5 minutes from now
+
+        while True:
+            dry_run_responses = []
+            for port, protocol in zip(self.ports, self.protocols):
+                client = Client(
```
JoanFM, line 193: we should not be using `Client`, we should be using the `streamer`. Maybe you can add a `warmup` to the streamer

jina/serve/runtimes/gateway/grpc/gateway.py
```diff
@@ -107,6 +104,7 @@ async def shutdown(self):
 
     async def run_server(self):
         """Run GRPC server forever"""
+        await self.warmup()
```
JoanFM, line 107: Can u have a task that wraps the 2 of them and await both at the same time? I am scared about the capacity to shutdown the gateway while warmup

JoanFM (changes requested):
Can we also have a measure of the benefits?

jina/serve/runtimes/gateway/grpc/gateway.py
girishc13: This is indeed an issue since the max timeout is 5 mins so until then the co-routine keeps running without cancellation. I wanted to check if there is another entry point to trigger the warmup. Currently the warmup is just triggering requests while the executors are still starting. This adds some unnecessary debug logs if there are slow executors.

JoanFM: maybe u can `create` a task without awaiting it

JoanFM: so that is scheduled in the event loop and in the `teardown` u make sure to cancel the task or whatever

jina/serve/gateway.py
```diff
@@ -172,17 +174,17 @@ async def run_server(self):
         """Run server forever"""
         ...
 
-    @abc.abstractmethod
     async def shutdown(self):
         """Shutdown the server and free other allocated resources, e.g, streamer object, health check service, ..."""
-        ...
+        if self._warmup_task and not self._warmup_task.done():
```
JoanFM, line 179: try to do this in the `runtime` object and not the gateway, so that it is transparent to the developer implementing the `Gateway`

JoanFM (changes requested):

girishc13:
I would like to do a different change to the warm up setup but here is the benefit of doing a warmup. I'm using the following executors.
```python
from jina import DocumentArray, Executor, requests
from opentelemetry.trace import NoOpTracer

# fast executor
class MyExecutor(Executor):
        if not self.tracer:
            self.tracer = NoOpTracer()

    @requests
    def foo(self, docs: DocumentArray, tracing_context, **kwargs):
        with self.tracer.start_as_current_span('process_docs', context=tracing_context):
            docs[0].text = 'hello, world!'

# slowExecutor
class MyExecutor(Executor):
        if not self.tracer:
            self.tracer = NoOpTracer()

        time.sleep(5)

    @requests
    def foo(self, docs: DocumentArray, tracing_context, **kwargs):
        with self.tracer.start_as_current_span('process_docs', context=tracing_context):
            docs[0].text = 'hello, world!'
```

With flow.yml:
```
jtype: Flow
version: '1'
with:
  port: 54321
  env:
    JINA_LOG_LEVEL: DEBUG
executors:
- uses: executor1/config.yml
  name: toyExecutor
  env:
    JINA_LOG_LEVEL: DEBUG
- uses: slow-executor/config.yml
  name: slowExecutor
  env:
    JINA_LOG_LEVEL: DEBUG
```

With sample client:
```python
import time

from jina import Client, Document, DocumentArray

if __name__ == '__main__':
    c = Client(host='grpc://0.0.0.0:54321', tracing=True, traces_exporter_host='http://localhost', traces_exporter_port=4317)

    now = time.time_ns()
    da = c.post('/', DocumentArray.empty(2))
    print(f'Time take for request: {(time.time_ns() - now)/1000000}ms.')

    time.sleep(3)
```

Time taken for the first request:
```shell
# running flow with master branch
Time take for request: 141.843666ms.
# with warmup
Time take for request: 50.209816ms.
# after re-running the flow with warmup
Time take for request: 50.725921ms.
# re-running flow with master branch
Time take for request: 138.22857ms.
```

The results might vary on local machines but it's clear that without warmup the first request includes the penalty for establishing the gRPC channel from the gateway to the executor/s. There might be additional penalties if modules are being dynamically loaded on the `requests` method. 

The current warmup implementation only handles the first connection issue when the flow is first created and if the gateway scales up which gives a smoother client response instead of spikes to the client response. This doesn't handle the issue with executor scale up because the `requests` method is being triggered on an actual client request. There should be an additional warmup method in the Executor with option to override the default warmup method to allow the hot request path to be ready before the first client request is received by the executor.

girishc13:
I have a slightly different variation in branch `feat-serve-5467-runtime-warmup` which pushes the warmup to the runtime and adds additional warmup to the executor.

PR #5574, opened 2023-01-05
feat: offer a client as part of Deployment

### LINKED ISSUE #5551: Offer a client as part of Deployment when serving an Executor
When using a Deployment with the context manager, offer a client as well.
Basically, Deployment should extend PostMixin and by default, expose a gateway (like in Flow).
(This might mean adding more gateway args to deployment.)
In Flow, all usages of Deployment should not add a gateway

AnneYang720:
**Goals:**

- resolves #5551 
- [x] add client property and extend PostMixin in Deployment
- [x] tests
- [x] check and update documentation


docs/concepts/executor/serve.md
```diff
@@ -109,17 +108,6 @@ with Deployment(uses='docker://my-executor-image', port=12345, replicas=2) as de
 │  🔒     Private     192.168.3.147:12345   │
 │  🌍      Public    87.191.159.105:12345   │
```
alaeddine-13, line 109: can we add a hint that we can use block if we want to serve forever ?

tests/integration/deployments/test_deployment.py
```diff
@@ -395,48 +394,16 @@ def bar(self, docs, **kwargs):
...
-    yield
-
-    stop_event.set()  # set event and stop (unblock) the Deployment
-    t.join()
-
-
-@pytest.mark.parametrize('served_depl', [False, True], indirect=True)
-def test_deployment_dynamic_batching(served_depl, exposed_port):
-    docs = Client(port=exposed_port).post(on='/bar', inputs=DocumentArray.empty(5))
+    depl = Deployment(uses=MyServeExec, **kwargs)
```
alaeddine-13, line 405: it's nice to have this version of the test, but can we have both version ? (so that the usage of block is well covered in our CI)

docs/concepts/executor/serve.md
```diff
@@ -59,7 +58,7 @@ class MyExec(Executor):
 
 
 with Deployment(uses=MyExec, port=12345, replicas=2) as dep:
-    dep.block()
+    print(dep.post(inputs=DocumentArray.empty(1), on='/foo').texts)
```
alaeddine-13, line 61: can we split this ?
maybe do
```
docs = dep.post(inputs=DocumentArray.empty(1), on='/foo')
print(docs.texts)
```
same for others

alaeddine-13:
closing since https://github.com/jina-ai/jina/pull/5590 was merged

PR #5579, opened 2023-01-06
feat: run warmup on Runtimes and Executor

### LINKED ISSUE #5467: feat: run warmup requests on runtime startup to ensure that service is ready to accept connections

girishc13:
**Goals:**

- resolves #5467 
- Gateway and Head runtime now starts a warmup tasks on the `run_forever` runtime method.
- The warmup task is created by the `GatewayStreamer` or the Head `RequestHandler` that delegates the warmup per deployments to the connection pool.
  - A single task per deployment is created and await until completion.
- The `GrpcConnectionPool` fetches the available shards and or replica (target) channels and creates a `JinaInfoRPC` request task. 
  - One task is created for each channel.
  - The warmup cycle runs until:
    - all targets return a successful response or
    - a max timeout of 5 minutes is reached or
    - a stop event is set by the runtime due to early tear down of the runtime.
  - In the warmup cycle, only unsuccessful targets are attempted.
- `CompositeGateway` doesn't support the warmup due to the creation of separate `GatewayStreamer` objects for each type of Gateway. The issue will be addressed in #5592.
  - the CompositeGateway methods `shutdown`, `run_server`, `setup_server` now handle each gateway operation as a task per created gateway.


jina/serve/networking.py
```diff
@@ -1114,6 +1113,71 @@ async def task_wrapper():
...
+                    return
+
+                await asyncio.sleep(0.2)
+        except Exception as ex:
+            self._logger.error(f'error with warmup up task: {ex}')
+            return
+
+    def __extract_target_to_channel(self, deployment):
+        replica_set = set()
+        replica_set.update(self._connections.get_replicas_all_shards(deployment))
```
girishc13, line 1171: I can extract out the warmup logic into a `Warmer` class but the `self._connections` is the inner class `_ConnectionPoolMap` of the `GrpcConnectionPool`. Does the inner class still make sense?

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -147,7 +147,7 @@ async def _process_results_at_end_gateway(
                         and not self._gathering_endpoints
                     ):
                         self._gathering_endpoints = True
-                        asyncio.create_task(gather_endpoints(request_graph))
+                        await asyncio.create_task(gather_endpoints(request_graph))
```
JoanFM, line 150: no please, this is created in the backround and not awaited in purpose

JoanFM (changes requested):

jina/serve/networking.py
```diff
@@ -1,6 +1,8 @@
 import asyncio
 import ipaddress
 import os
+import threading
+import time
```
JoanFM, line 5: please remove unneeded imports

jina/serve/runtimes/asyncio.py
```diff
@@ -1,6 +1,7 @@
 import argparse
 import asyncio
 import signal
+import threading
```
JoanFM, line 4: remove unneeded import

jina/serve/runtimes/head/request_handling.py
```diff
@@ -1,6 +1,9 @@
 import asyncio
-from typing import TYPE_CHECKING, Dict, Optional, Tuple
+import threading
```
JoanFM, line 2: please cleanup imports

jina/serve/networking.py
```diff
@@ -178,6 +183,10 @@ def add_connection(self, address: str, deployment_name: str):
             stubs, channel = self._create_connection(address, deployment_name)
             self._address_to_channel[address] = channel
             self._connections.append(stubs)
+            # create a new set of stubs and channels for warmup to avoid
+            # loosing channel during remove_connection or reset_connection
+            stubs, _ = self._create_connection(address, deployment_name)
+            self._warmup_stubs.add(stubs)
```
JoanFM, line 189: I do not see where they are used

JoanFM (changes requested):

jina/serve/networking.py
girishc13: Check the usage of the property method `self.warmup_stubs`. What's the preference when exposing `read only` properties?

JoanFM: ah okey

JoanFM (approved):

PR #5586, opened 2023-01-10
fix(pod parser): use mixin_hub_pull_options_parser from hubble

nomagick:
**Goals:**

- Currently Jina core has its own `mixin_hub_pull_options_parser`
- When hubble modifies its version of `mixin_hub_pull_options_parser` jina core might break because of parameter difference.
- This is meant to offload the `mixin_hub_pull_options_parser` to hubble
- Note that the bubble SDK dependency version has been bumped to guarantee the new compatible version of `mixin_hub_pull_options_parser` from bubble sdk. 


jina/parsers/__init__.py
```diff
@@ -17,6 +17,7 @@ def set_pod_parser(parser=None):
 
     from jina.parsers.orchestrate.base import mixin_scalable_deployment_parser
     from jina.parsers.orchestrate.pod import mixin_pod_parser, mixin_hub_pull_options_parser
+    from hubble.executor.parsers.pull import mixin_hub_pull_options_parser
```
JoanFM, line 20: it is done in purpose not to add this,  so that `instalL_requirements` is redefined by us

JoanFM (changes requested):

jina/parsers/__init__.py
nomagick: In the latest SDK I've separated the `install_requirements` just for this reason.
https://github.com/jina-ai/jina-hubble-sdk/pull/150/files

nomagick: I'll also bump the version lock in this PR

JoanFM, line 20: I did on purpose to have Jina's own parser to reuse install requirements so that I can edit the description

JoanFM (changes requested):

JoanFM (approved):

PR #5591, opened 2023-01-10
feat: relax protobuf version

alaeddine-13:
try jina with protobuf 3.19 to see what is missing to support it


JoanFM (changes requested):
lets edit setup.py for python 3.11

alaeddine-13:
> lets edit setup.py for python 3.11

I don't think it's really needed. We have now a wider range from protobuf 3.19 to latest. For python 3.11, we let pip decide which version to install. The limitation will come from the grpcio package eventually and if the environment does not support the protobuf version imposed, jina will not work anyways

PR #5594, opened 2023-01-12
fix: disable timeout for liveness probe in k8s and keep only k8s timeout

alaeddine-13:
As an executor is increasingly stressed by load, the health check latency is affected.
It may eventually exceed the default 1 second timeout and therefore, an Executor might be killed by k8s.
This PR disables the ping timeout in liveness probes, keeping just the k8s timeout


jina/resources/k8s/template/deployment-uses-before-after.yml
```diff
@@ -68,6 +68,7 @@ spec:
               - ping
               - executor
               - 127.0.0.1:{port}
+              - --timeout 9
```
JoanFM, line 71: I do not think this is how is passed?

JoanFM (changes requested):
Can u add a test that can simulate what we were observing? putting time sleeps and overcharging should be good

JoanFM (approved):

PR #5598, opened 2023-01-13
feat: use single gateway streamer for CompositeGateway

### LINKED ISSUE #5592: feat: use single GatewayStreamer object for all gateways in the CompositeGateway
The `GatewayStreamer` object is instantiated in the `BaseGateway`. The `CompositeGateway` itself is a sub class of the `BaseGateway` and the respective protocol specific `Gateway` is also a sub class of the `BaseGateway`. Which means that there is at least one streamer object for the `CompositeGateway` class and one for each protocol specific gateway. This complicates the warmup logic (#5467) because by default only the `CompositeGateway` streamer object is being triggered. Using a single streamer object is more compact and reduces the overhead of creating multiple connection pools within each streamer object.

girishc13:
**Goals:**

- resolves #5592


JoanFM (changes requested):
make sure the close method is safe

alaeddine-13:
> make sure the close method is safe

I believe it's fine without a lock. The streamer.close is already called once, when the runtime is shutdown

JoanFM (approved):

Issue #5584, opened 2023-01-09
Refactor: use K8sDeploymentConfig inside Deployment to generate kubernetes yamls


alaeddine-13:
After Deployment is exposed to serve Executors, we want to use it in order to generate kubernetes yaml as well.
This means, the Deployment should use K8sDeploymentConfig to implement `to_kubernetes_yaml`.

Potentially, we also want to refactor Flow.to_kubernetes_yaml to use the new method and maybe the same for Executor (or deprecate Executor.to_kubernetes_yaml)

PR #5608, opened 2023-01-17
feat: add to_kubernetes_yaml to Deployment

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/5584


jina/orchestrate/deployments/__init__.py
```diff
@@ -134,14 +134,13 @@ def __init__(
         metrics_exporter_host: Optional[str] = None, 
         metrics_exporter_port: Optional[int] = None, 
         monitoring: Optional[bool] = False, 
-        name: Optional[str] = None, 
+        name: Optional[str] = 'executor',
```
JoanFM, line 137: weird

JoanFM (changes requested):
Check the comment, and the prefer-platform issue

alaeddine-13:
> Check the comment, and the prefer-platform issue

I don't think there is anything wrong with it. This argument shouldn't exist because prefer_platform does not exist for executors. Try: jina executor --help

JoanFM (approved):

docs/concepts/executor/serve.md
```diff
@@ -157,6 +155,33 @@ The above example deploys the `DummyHubExecutor` from Executor Hub into your Kub
 The Executor you use needs to be already containerized and stored in a registry accessible from your Kubernetes cluster. We recommend Executor Hub for this.
 ````
 
+Once the Executor is deployed, you can expose a service:
```
girishc13, line 158: What's the reason behind not exposing the service automatically?

girishc13: Nevermind. This is only for the external IP.

docs/concepts/executor/serve.md
```diff
@@ -157,6 +155,33 @@ The above example deploys the `DummyHubExecutor` from Executor Hub into your Kub
...
+```bash
+kubectl expose deployment executor --name=executor-exposed --type LoadBalancer --port 80 --target-port 8080 -n my-namespace
+sleep 60 # wait until the external ip is configured
+```
+
+Let's export the external IP address created and use it to send requests to the Executor. 
+```bash
+export EXTERNAL_IP=`kubectl get service executor-exposed -n my-namespace -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'`
+```
+To send requests using {meth}`~jina.Client`, make sure to set `stream=False`:
```
girishc13, line 168: If a Gateway is added by default then there is no need to set `stream=False`.

docs/concepts/executor/serve.md
alaeddine-13: I guess it depends on how the user wants to do so, which varies from kubernetes provider to another.
 I guess the service we expose already is mostly useful for jcloud. Not sure how a user would leverage it without portforwarding

docs/concepts/executor/serve.md
girishc13: My misunderstanding again. The `include_gateway` is used only in the overload method in the Flow.

girishc13 (dismissed):

JoanFM (dismissed):

JoanFM (approved):

PR #5617, opened 2023-01-23
fix: add proper error message to not-found errors

### LINKED ISSUE #5613: Error reporting when DNS exists, but route is not valid
When an external Executor/Flow is down, but there is no DNS error (e.g. because it is behind an API gateway), then the error reporting does not show which Executor is the failing one.

**Reproduce:**
```python
from jina import Flow

f = Flow().add(host='https://blah.wolf.jina.ai/', external=True)  # this Flow does not exist, but no DNS issue
with f:
    f.post(inputs=Document, on='/foo')
```
On the client:
```text
jina.excepts.BadServerFlow: gRPC error: StatusCode.UNKNOWN Unexpected <class 'grpc.aio._call.AioRpcError'>: <AioRpcError of RPC that terminated with:
	status = StatusCode.NOT_FOUND
	details = "no Route matched with those values"
	debug_error_string = "{"created":"@1674229068.908652727","description":"Error received from peer ipv4:35.169.210.186:443","file":"src/core/lib/surface/call.cc","file_line":952,"grpc_message":"no Route matched with those values","grpc_status":5}"
>
```
On the gateway:
```
ERROR  gateway/rep-0/GatewayRuntime@91339 Error while                           
       getting responses from deployments: <AioRpcError of                      
       RPC that terminated with:                                                
               status = StatusCode.NOT_FOUND                                    
               details = "no Route matched with those                           
       values"                                                                  
               debug_error_string =                                             
       "{"created":"@1674229068.908652727","description":"E…                    
       received from peer                                                       
       ipv4:35.169.210.186:443","file":"src/core/lib/surfac…                    
       Route matched with those values","grpc_status":5}"                       
       > 
```

Reference: https://jinaai.slack.com/archives/C018F60RBL5/p1674205594827639

JohannesMessner:
Signed-off-by: Johannes Messner <messnerjo@gmail.com>

**Goals:**

When an external Executor/Flow is behind an API Gateway (which is the case for JCloud), but is down, then DNS resolution succeeds but a "resource" (the Exec/Flow) cannot be found, resulting in a gRPC error with `NOT_FOUND` code.

This error case was not properly handled **before,** giving output like the following:

image

The user had no way of knowing what part of the Flow failed.

**After** this PR, the affected deployment and its address will be displayed:

```console
ERROR  gateway/rep-0/GatewayRuntime@123711 Error while       [01/23/23 12:35:15]
       getting responses from deployments: no Route matched                     
       with those values                                                        
       trailing_metadata=Metadata((('date', 'Mon, 23 Jan                        
       2023 11:35:15 GMT'), ('content-length', '0'),                            
       ('x-kong-response-latency', '0'), ('server',                             
       'kong/3.0.2')))                                                          
       trailing_metadata=Metadata((('date', 'Mon, 23 Jan                        
       2023 11:35:15 GMT'), ('content-length', '0'),                            
       ('x-kong-response-latency', '0'), ('server',                             
       'kong/3.0.2')))                                                          
       |Gateway: Connection error with deployment                               
       `executor0` at address(es) {'blah.wolf.jina.ai'}.                        
       Connection with {'blah.wolf.jina.ai'} succeeded, but                     
       `executor0` was not found. Possibly `executor0` is                       
       behind an API gateway but not reachable.                                 
       trailing_metadata=Metadata((('date', 'Mon, 23 Jan                        
       2023 11:35:15 GMT'), ('content-length', '0'),                            
       ('x-kong-response-latency', '0'), ('server',                             
       'kong/3.0.2')))
```

**TODO:**

- [x] add a test

- resolves #5613 

### DIFF
--- a/jina/clients/base/grpc.py
+++ b/jina/clients/base/grpc.py
@@ -277,6 +277,11 @@ class GRPCBaseClient(BaseClient):
                                     f'{msg}\nThe ongoing request is terminated as the server is not available or closed already.'
                                 )
                                 raise ConnectionError(my_details)
+                            if my_code == grpc.StatusCode.NOT_FOUND:
+                                self.logger.error(
+                                    f'{msg}\nThe ongoing request is terminated as a resource cannot be found.'
+                                )
+                                raise ConnectionError(my_details)
                             elif my_code == grpc.StatusCode.DEADLINE_EXCEEDED:
                                 self.logger.error(
                                     f'{msg}\nThe ongoing request is terminated due to a server-side timeout.'
--- a/jina/serve/networking.py
+++ b/jina/serve/networking.py
@@ -1000,23 +1000,26 @@ class GrpcConnectionPool:
         # connection failures have the code grpc.StatusCode.UNAVAILABLE
         # cancelled requests have the code grpc.StatusCode.CANCELLED
         # timed out requests have the code grpc.StatusCode.DEADLINE_EXCEEDED
+        # if an Executor is down behind an API gateway, grpc.StatusCode.NOT_FOUND is returned
         # requests usually gets cancelled when the server shuts down
         # retries for cancelled requests will hit another replica in K8s
         self._logger.debug(
             f'GRPC call to {current_deployment} errored, with error {format_grpc_error(error)} and for the {retry_i + 1}th time.'
         )
-        if (
-            error.code() != grpc.StatusCode.UNAVAILABLE
-            and error.code() != grpc.StatusCode.CANCELLED
-            and error.code() != grpc.StatusCode.DEADLINE_EXCEEDED
-            and error.code() != grpc.StatusCode.UNKNOWN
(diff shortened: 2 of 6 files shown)


JoanFM (dismissed):

alaeddine-13 (commented):
it feels unclear to the user when we say 'resource could not be found'
Can't we give more explicit hints ? maybe hint that the executor server can be down or not responding,...

JohannesMessner:
> for the head runtime, it feels like the message is not clear like the worker runtime message and we just say 'resource was not found' Can't we give more explicit hints ? maybe hint that the executor server can be down or not responding,...

The issue is that on the client side the message from the head and from the gateway will be concatenated, so it would just duplicate the same message. My thinking was that the most important piece of information to the user will be here that the head is still up and that its message agrees with the gateway message. Not sure if repeating everything is valuable.
Or do you think some users will only/preferrably check the head logs and not the gateway/client?

JoanFM (approved):

PR #5621, opened 2023-01-24
refactor: small refactor in threads for waiting Flow readiness

JoanFM:
**Goals:**
Task from rich local in the thread stack

### DIFF
--- a/jina/helper.py
+++ b/jina/helper.py
@@ -166,8 +166,8 @@ def deprecated_method(new_function_name):
 
 
 def retry(
-    num_retry: int = 3,
-    message: str = 'Calling {func_name} failed, retry attempt {attempt}/{num_retry}. Error: {error!r}',
+        num_retry: int = 3,
+        message: str = 'Calling {func_name} failed, retry attempt {attempt}/{num_retry}. Error: {error!r}',
 ):
     """
     Retry calling a function again in case of an error.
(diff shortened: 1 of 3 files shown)


jina/helper.py
```diff
@@ -1237,15 +1237,8 @@ def _get_ip(url):
...
-    threads = []
-
     for idx, ip in enumerate(ip_server_list):
-        t = threading.Thread(target=_get_ip, args=(ip,))
-        threads.append(t)
-        t.start()
-
-    for t in threads:
-        t.join(timeout)
+        results.append(_get_ip(ip))
```
samsja, line 1241:
```suggestion
        r = _get_ip(ip)
        if r:
            return r
```

Now that it is not async we can just return the first one that is not None to save time. (As we only return one IP later anyway)

jina/helper.py
```diff
@@ -1237,15 +1237,8 @@ def _get_ip(url):
...
 
     for r in results:
```
samsja, line 1243: this can be removed if we apply the change in the for loop

jina/helper.py
```diff
@@ -1237,15 +1237,8 @@ def _get_ip(url):
...
         if r:
```
samsja, line 1244: this can be removed if we apply the suggestion above

jina/helper.py
JoanFM: Good

jina/helper.py
JoanFM: Suggestion applied, good catch!

girishc13 (approved):

samsja (approved):

Issue #5552, opened 2022-12-21
Create Deployment YAML separate from Executor YAML


alaeddine-13:
create deployment yaml separate from Executor yaml

* add Deployment.load_config

A deployment can use an Executor YAML like so:
 Deployment(uses='executor.yaml', replicas=...)

But a deployment YAML is used like so:
Deployment.load_config(uses='deployment.yaml')

Possible usages of executors inside Flow YAML:
1. executor config is embedded in config
```yaml
 jtype: flow
 executors:
   - jtype: Deployment
     uses:
         jtype: MyExecutor
         py_modules: executor.py
     replicas: 2
```

2. old usage for backward compatibility:
```yaml
 jtype: flow
 executors:
    - uses: executor.yaml
      replicas: 12
```
3. ~~simply `deployment.yaml`~~
```yaml
 jtype: flow
 executors:
    - uses: deployment.yaml
      replicas: 2
      shard: 3
```
instead check if this is possible:
```yaml
 jtype: flow
 executors/deployments:
   - deployment.yaml
   - uses: executor.yaml
      replicas: 12
   - jtype: Deployment
     uses:
         jtype: MyExecutor
         py_modules: executor.py
     replicas: 2
```

Deployment args include protocol, shard, replicas, expose-gateway

Executor args include the rest

PR #5189, opened 2022-09-21
feat: allow passing custom gateway in Flow

alaeddine-13:
This PR allows users to pass a custom gateway class to Flow definition similarly to Executors.

- [x] Make custom gateway JAML compatible and loadable from config
- [x] Allow passing custom gateway as a class
- [x] Make sure `jina gateway --uses config.yml` works
- [x] Allow passing custom gateway as yaml
- [x] Allow passing custom gateway as docker container
- [x] testing
    - test jaml compatibility of gateway in tests/unit/jaml (https://github.com/jina-ai/jina/pull/5189/commits/79a1a45fb84fc81aecb1788f553298a6f07c4751 and https://github.com/jina-ai/jina/pull/5189/commits/b16446e774ae3988f0c54995ef0073a3c3c583f1)
    - test cli with custom gateway in tests/unit/test_cli.py
    - yaml config tests in tests/unit/test_yaml_parser.py (gateway uses +~~ flow uses containing gateway uses~~ TBD in https://github.com/jina-ai/jina/issues/5215) (https://github.com/jina-ai/jina/pull/5189/commits/2f1af2a04cf1edde0f346e10b9de43512673ccd6)
    - ~~test custom gateway in Flow API (python class + config + docker) in tests/unit/orchestrate/flow~~ To be done in https://github.com/jina-ai/jina/issues/5215
    - test custom gateway in gateway runtime tests/unit/serve/gateway as yaml and python class name (https://github.com/jina-ai/jina/pull/5189/commits/59cb4aef4e64ff35e119fd04ecc14f7ac61c84a0)
    - container pod tests for gateway in tests/unit/orchestrate/pods/container (we already have some in tests/unit/orchestrate/pods/container/gateway-runtime, they were removed), (https://github.com/jina-ai/jina/pull/5189/commits/a4763a2ef1e759a11205056832001b919d9329b6 and https://github.com/jina-ai/jina/pull/5189/commits/d339836702f6cdda657bea47336077bf1dceb13e)
    - custom gateway in docker-compose (https://github.com/jina-ai/jina/pull/5189/commits/2983897f1583958b24ccffcfcf81df928535d898)
    - custom gateway in k8s (https://github.com/jina-ai/jina/pull/5189/commits/36f530e19a81b33c391587062acf40d943bb7458)
    - make sure these tests are called in CI
- [x] todos in PR

Issue:
~~Gateway class ** must ** follow protocol (otherwise, GRPCRuntime will attempt to start an HTTPGateway)~~
[...]

PR #5628, opened 2023-01-26
feat: deployment yaml

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/5552

TODOs:
- [x] Separate deployment parser and executor parser
- [x] remove some arguments from executor
- [x] Add jaml parser for Deployment under jaml/parsers/deployment
- [x] Flow.add() should be able to decide whether to call Deployment.load_config(uses=deployment_uses) or Deployment(uses=executor_uses) ==> rely on jtype to infer this, by default, infer as executor yaml
- [x] Add schema for deployment
- [x] Add Deployment.load/save_config
- [ ] Documentation
- [ ] tests:
    - [x] Json schema tests
    - [x] Yaml parser tests
    - [x] Deployment from yaml (embedded executor config + nested executor yaml)
    - [x] Flow with add deployment yaml
    - [x] Flow from yaml with deployment yaml

PS: can get inspired from #5342 and #5189


jina/orchestrate/deployments/__init__.py
```diff
@@ -47,7 +49,13 @@
     from jina.serve.executors import BaseExecutor
 
 
-class Deployment(PostMixin, BaseOrchestrator):
+class DeploymentType(type(ExitStack), type(JAMLCompatible)):
```
JoanFM, line 52: we do not want `ExitStack`

JoanFM (changes requested):

jina/jaml/parsers/__init__.py
```diff
@@ -1,6 +1,7 @@
 import warnings
 from typing import List, Optional, Type
 
+from jina import Deployment
```
JoanFM, line 4: can we avoid importing directly from `jina`? Import from its own module

jina/orchestrate/deployments/__init__.py
JoanFM: Remove `ExitStack` since we are not using its features

jina/orchestrate/flow/base.py
```diff
@@ -1208,47 +1209,51 @@ def add(
...
+            deployment = Deployment.load_config(deployment)
 
-        op_flow._deployment_nodes[deployment_name] = Deployment(
-            args, needs, include_gateway=False
-        )
+        op_flow._deployment_nodes[deployment_name] = deployment
 
-        if not args.floating:
-            op_flow._last_deployment = deployment_name
+        # TODO: uncomment later
```
JoanFM, line 1254: remember about this

JoanFM (changes requested):

tests/unit/yaml/test-flow-deployment-nested-config.yml
```diff
@@ -0,0 +1,3 @@
+!Flow
+executors:
```
JoanFM, line 2: make sure this can also be called `deployments`

tests/unit/orchestrate/flow/flow-construct/test_flow_deployment.py
```diff
@@ -23,3 +29,16 @@ def test_flow_deployment(deployment_config):
...
+@pytest.mark.parametrize(
+    'flow_config',
+    [_flow_yaml_with_deployment_config, _flow_yaml_with_deployment_exec_config],
+)
+def test_flow_load_config_with_deployment(flow_config):
+
+    flow = Flow.load_config(flow_config)
+    with flow:
+        docs = flow.post(on='/', inputs=DocumentArray.empty(3))
+        assert len(docs) == 3
```
JoanFM, line 59: do assertions outside the context manager

jina/orchestrate/deployments/__init__.py
alaeddine-13: not possible, python will fail and say the metaclass of Deployment should be a subclass of its parent's metaclasses (which are ExitStack and JAMLCompatible)

JoanFM (approved):
Documenting this would be in another PR?

alaeddine-13:
> Documenting this would be in another PR?

yeah let's do this in another PR

JoanFM (approved):

PR #5632, opened 2023-01-27
refactor: no need to run summary in thread

JoanFM:
**Goals:**
Avoid problems of concurrency with `urllib`


girishc13:
So no more summary table with the ip's? Ok if it avoids the seg faults with CI.

girishc13 (approved):

JoanFM:
No, @girishc13 there will be summary, but it will not be computed in a thread

PR #5633, opened 2023-01-27
refactor: remove manual deletion of channel resources or objects

### LINKED ISSUE #5622: refactor: extract GrpcConnectionPool implementation into a package
Currently the `GrpcConnectionPool` class that handles gRPC channels and sending requests to targets in a single giant file. The goals of the refactor are:
- create a new networking package to extract inner classes to package inner classes.
- add unit tests to each inner class implementation.
- use correct synchornization (asyncio or thread) when resetting or recreating new gRPC channels and stubs which are stored in a dict.

girishc13:
**Goals:**

- resolves #5622 
- In the `reset_connection` method, the channel was being manually deleted using the `del` keyword which can cause segmentation fault if another co-routine was having a reference the deleted channel. The alternatives to avoid deleting the channel reference are:
  1.  use a Lock to detect usage or
  2. allow the GC to collect the unused channel when all references have been removed. 
- The ii choice is more straight forward since the co-routines that maybe holding a channel are usually short lived due to the timeout on the request. Also destroying the channel has very little benefit because the underlying TCP connection will be reused or recreated by the new connection.
- The removed channel doesn't need to be returned because there is no further usage.


girishc13:
I opened a [bug](https://github.com/jina-ai/jina/issues/5636) ticket related to custom logging configuration. The logging check using a `FileHandler` was one potential way to check the reconnect attempts. Due to multiprocessing using instrumentation might be the only other way.

JoanFM (approved):

PR #5635, opened 2023-01-29
feat: add argument to remove or preserve root logging handlers

### LINKED ISSUE #5629: feat: add argument to remove or preserve root logging handlers
**Describe the feature**
JinaLogger removes all root logging handlers preventing output from modules using the root logger. This prevents logging of useful debug/trace logs of the module.

**Your proposal**

Add a new argument `suppress_root_logging` which gives the users the flexibility to switch off (current behavior) or enable root logger to get more useful information.

Jake-00:
**Goals:**

- resolves #5629

**Updates:**
1. To add new argument `suppress_root_logging` for class `JinaLogger`'s constructor
2. To add if statement which determines whether to remove or preserve root logging handlers


Jake-00:
hi @JoanFM , plz take a glance at code when free

jina/logging/logger.py
```diff
@@ -117,6 +118,7 @@ def __init__(
         name: Optional[str] = None,
         log_config: Optional[str] = None,
         quiet: bool = False,
+        suppress_root_logging: bool = False,
```
girishc13, line 121: The avoid breaking current behavior, the default value is True.

girishc13 (changes requested):
This is a good start but the value can only be used by users writing new Executors. It cannot be configured by the current entry points like Flow and Client. If you're up to the task, you can make a few more changes to add the same argument at least to the Flow and the Client parser. 
You can refer to the the `parsers` package for adding the new argument and you can refer to the following lines where the argument will be used automatically:
- https://github.com/jina-ai/jina/blob/c57fdbb11e4c318e31ab1d778ded3d4183ab1e0c/jina/orchestrate/flow/base.py#L519-L524
- https://github.com/jina-ai/jina/blob/c57fdbb11e4c318e31ab1d778ded3d4183ab1e0c/jina/clients/base/__init__.py#L40

Jake-00:
Hi @girishc13 . Thanks for your detailed instruction. There are two things which need to clarify:
* As for Flow, to append argument in file `jina\parsers\orchestrate\base.py` is enough.
```python
def mixin_essential_parser(parser, default_name=None):
    # leave out original lines
    gp.add_argument(
        '--suppress_root_logging',
        action='store_true',
        default=False,
        help='If set, then no root handlers will be suppressed from logging.',
    )
```
* As for Client, adding modified `mixin_essential_parser()` in function `set_client_cli_parser()` of file `jina\parsers\__init__.py` is allowed.
```python
def set_client_cli_parser(parser=None):
    # leave out original lines
    mixin_essential_parser(parser)  # new
    mixin_client_gateway_parser(parser)
    mixin_client_features_parser(parser)
    mixin_client_protocol_parser(parser)
    mixin_prefetch_parser(parser)
```

girishc13:
@Jake-00, the `mixin_essential_parser` cannot be added to the client parser because it contains irrelevant arguments to the client. My suggestion is to create a new file `logging.py` under the `parsers` package which has the following method:

```python
def mixin_suppress_root_logging_parser(parser):
   parser.add_argument(
          '--suppress_root_logging',
          action='store_true',
          default=False,
          help='If set, then no root handlers will be suppressed from logging.',
      )
```

This method can now be individually added after the usages of `mixin_essential_parser` and also inside the `set_client_cli_parser` method independently.

Jake-00:
Hi @girishc13 , I have followed your suggestion and update code with new commit, plz check it out when free, thanks

jina/parsers/logging.py
```diff
@@ -0,0 +1,10 @@
+"""Extensive argparser module for logging"""
+
+
+def mixin_suppress_root_logging_parser(parser):
+    parser.add_argument(
+        '--suppress_root_logging',
```
girishc13, line 6: CLI args are separated by `-` instead of underscore. The python variable name will be available with underscores.

girishc13 (changes requested):
One minor change. Otherwise looks good.

jina/parsers/logging.py
Jake-00: I get your point and make the change.

Jake-00:
I just now append docstring for function `mixin_suppress_root_logging_parser()`.

girishc13 (approved):

Jake-00:
hi @girishc13, should I add more test to improve test coverage rate? But I do not why the updated code would affect `schemas` module.

JoanFM (approved):

JoanFM:
Thanks for this amazing contribution @Jake-00 !!!! Happy to have you as part of the community!!

PR #5642, opened 2023-01-31
refactor: handle rich and threads differently

JoanFM:
**Goals:**
Try to fix CI segfaults


alaeddine-13 (commented):
LGTM. I would just suggest renaming `results_access` to `results_lock` for readability

PR #5650, opened 2023-02-02
feat: refactor streamer.stream method to yield reponse and executor error

### LINKED ISSUE #5646: offer stream method in GatewayStreamer that allows users to handle exceptions from executors instead of silently returning empty docs
We should offer a public method in GatewayStreamer called `stream` that allows users to handle exceptions from executors instead of silently returning empty docs.
For instance, the interface can look like this:
```
for docs, error in streamer.stream(...):
    if error:
         # handle error
    else:
         # process documents
```
The existing `stream` method is only used internally so we can just rename it and expose a new stream method. It should be documented under https://docs.jina.ai/concepts/gateway/customization/#calling-executors-with-gatewaystreamer instead of `stream_docs`

girishc13:
**Goals:**

- resolves #5646 
- rename existing `stream` method that implements the gRPC Call stream method to `rpc_stream`.
- create a new `stream` method that accepts an request iterator and yields a tuple of response and the underlying Executor error.
  - The `ExecutorError` class is used to wrap the `jina_pb2.StatusProto.ExceptionProto` information so that the user can raise the exception if necessary.  
- The new `stream` method doesn't catch all types of exceptions. Exceptions from the`stream` method will be raised.

### DIFF
--- a/docs/concepts/gateway/customization.md
+++ b/docs/concepts/gateway/customization.md
@@ -245,7 +245,7 @@ async def get(text: str):
             errors.append(error)
         else:
             results.append(docs[0].text)
-    return {'results': results, errors=[error.name for error in errors]}
+    return {'results': results, 'errors': [error.name for error in errors]}
 ```
 
 (executor-streamer)=
--- a/docs/concepts/gateway/customization.md
+++ b/docs/concepts/gateway/customization.md
@@ -224,7 +224,7 @@ class MyGateway(FastAPIBaseGateway):
         return app
 ```
 
-### Recovering Executor error
+### Recovering Executor errors
 Exceptions raised by an `Executor` are captured in the server object which can be extracted by using the {meth}`jina.serve.streamer.stream()` method. The `stream` method
 returns an `AsyncGenerator` of a tuple of `DocumentArray` and an optional {class}`jina.excepts.ExecutorError` class that be used to check if the `Executor` has issues processing the input request.
 The error can be utilized for retries, handling partial responses or returning default responses.
--- a/docs/concepts/gateway/customization.md
+++ b/docs/concepts/gateway/customization.md
@@ -224,6 +224,30 @@ class MyGateway(FastAPIBaseGateway):
         return app
 ```
 
+### Recovering executor error
+Exceptions raised by an `Executor` are captured in the server object which can be extracted by using the {meth}`jina.serve.streamer.stream()` method. The `stream` method
+returns an `AsyncGenerator` of a tuple of `DocumentArray` and an optional {class}`jina.excepts.ExecutorError` class that be used to check if the `Executor` has issues processing the input request.
+The error can be utilized for retries, handling partial responses or returning default responses.
+
+```{code-block} python
+---
+emphasize-lines: 5, 6, 7, 8, 9, 10, 11, 12
+---
+@app.get("/endpoint")
+async def get(text: str):
+    results = []
+    errors = []
+    async for docs, error in self.streamer.stream(
+        docs=DocumentArray([Document(text=text)]),
+        exec_endpoint='/',
+    ):
+        if error:
+            errors.append(error)
+        else:
+            results.append(docs[0].text)
+    return {'results': results, errors=[error.name for error in errors]}
+```
+
 (executor-streamer)=
 ## Calling an individual Executor
 An `executor` object is injected by Jina to your gateway class which allows you to call individual Executors from the Gateway.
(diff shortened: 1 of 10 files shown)


jina/excepts.py
```diff
@@ -145,3 +145,56 @@ def details(self):
                 return self._details
 
         return self.og_exception.details()
+
+
+class ExecutorError(RuntimeError, BaseJinaException):
```
JoanFM, line 150: Let's also add some unit tests for this

docs/concepts/gateway/customization.md
```diff
@@ -224,6 +224,30 @@ class MyGateway(FastAPIBaseGateway):
         return app
 ```
 
+### Recovering executor error
```
alexcg1, line 227:
```suggestion
### Recovering Executor errors
```
Please ensure Executor capitalized throughout

jina/serve/streamer.py
```diff
@@ -143,6 +155,52 @@ def stream(self, *args, **kwargs):
...
+        parameters: Optional[Dict] = None,
+        results_in_order: bool = False,
+    ) -> AsyncIterator[Tuple[Union[DocumentArray, 'Request'], 'ExecutorError']]:
+        """
+        stream documents and yield responses and unpacked executor exception if any.
+
+        :param docs: The Documents to be sent to all the Executors
+        :param request_size: The amount of Documents to be put inside a single request.
+        :param return_results: If set to True, the generator will yield Responses and not `DocumentArrays`
+        :param exec_endpoint: The executor endpoint to which to send the Documents
```
alexcg1, line 174: Capitalization

alexcg1 (changes requested):

docs/concepts/gateway/customization.md
```diff
@@ -224,7 +224,7 @@ class MyGateway(FastAPIBaseGateway):
         return app
 ```
 
-### Recovering executor error
+### Recovering Executor error
```
alexcg1, line 227: Should this be singular or plural?

jina/serve/streamer.py
```diff
@@ -169,24 +172,30 @@ async def stream(
...
+                exception = result.status.exception
                 error = ExecutorError(
                     name=exception.name,
                     args=exception.args,
                     stacks=exception.stacks,
                     executor=exception.executor,
                 )
-            yield response, error
+            if not return_responses:
+                result.document_array_cls = return_type
```
JoanFM: why do u need this update?

docs/concepts/gateway/customization.md
girishc13: Depends on how the user batches the input docs. If there is not batching then there will be only one otherwise it will match the batch size.

jina/serve/streamer.py
girishc13: I was using the same method signature as before which has the `return_type` param. Its not valid anymore after using the same method signature as `stream_docs` method.

docs/concepts/gateway/customization.md
alexcg1: Let's make it plural. It sounds more natural

alexcg1 (changes requested):

girishc13:
@alexcg1 Can you add your last approval?

alexcg1 (approved):

jina/serve/streamer.py
```diff
@@ -143,6 +155,52 @@ def stream(self, *args, **kwargs):
...
+                error = ExecutorError(
+                    name=exception.name,
+                    args=exception.args,
+                    stacks=exception.stacks,
+                    executor=exception.executor,
+                )
+            if return_results:
+                yield result, error
+            else:
+                result.document_array_cls = DocumentArray
```
JoanFM, line 201: why is this needed?

JoanFM (changes requested):

jina/serve/streamer.py
girishc13: This block is the same as the block in in the async [post](https://github.com/jina-ai/jina/blob/master/jina/clients/mixin.py#L387) method which also returns an async generator.

JoanFM: @samsja do you know why is this needed?

samsja: > result.document_array_cls = DocumentArray
this is needed to for docarray v2 so that the DataRequest can be serialize to protobuf using the correct class (i,e, the return schema from the executor)

JoanFM: okk

JoanFM (approved):

jina/serve/streamer.py
samsja: but since it is a resulst here and not a request I am not sure it is needed

PR #5651, opened 2023-02-03
idempotent request

### LINKED ISSUE #5648: docs: explain idempotent requests
Idempotent requests are a standard of the REST POST and PUT api's wherein duplicate requests or re-tried requests for certain use cases have a single effect and outcome. For example, user account creation is unique and retries will return an error response. In a similar way asynchronous jobs that have the same identifier will only be scheduled to run once. This concept must be documented in jina so that the custom Executors are aware that retrying requests with side effects are carefully handled.

Sre-n:
**Goals:**

- resolves #5648 
- In the above issue it was asked to document idempotent request which is an important concept tat was to be adhered to
- Briefly understanding the concepts need to commit more

- docs
- ...


JoanFM:
Hello @Sre-n , thanks for the contribution. However, we believe that this will need to be taken care by someone in our internal team because documenting this requires an in-depth understanding of some of our internals, and they are hard to explain to an external person.

PR #5656, opened 2023-02-03
docs: refactor example code for experimenting with Otel

girishc13:
**Goals:**

- provide a more concrete local example for the confusion in https://github.com/jina-ai/example-grafana-prometheus/issues/4

### DIFF
--- a/docs/cloud-nativeness/opentelemetry.md
+++ b/docs/cloud-nativeness/opentelemetry.md
@@ -140,6 +140,7 @@ Run the Flow and a sample request that we want to instrument locally. If the bac
 
 ```python
 from jina import Flow, Document, DocumentArray
+import time
 
 with Flow(
     tracing=True,
@@ -150,6 +151,8 @@ with Flow(
     metrics_exporter_port=4317,
 ).add(uses='jinaai://jina-ai/SimpleIndexer') as f:
     f.post('/', DocumentArray([Document(text='hello')]))
+    # allow some time for the metrics data export
+    time.sleep(3)
 ```
 
 ## Viewing Traces in Jaeger UI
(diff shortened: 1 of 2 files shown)


JoanFM (changes requested):
Lets change the docs to use Flow with block and use an external client

docs/cloud-nativeness/opentelemetry.md
```diff
@@ -138,21 +138,43 @@ The Prometheus configuration now only needs to scrape from the OpenTelemetry Col
...
+
+client = Client(
+    host='grpc://localhost:54321',
+    tracing=True,
+    traces_exporter_host='http://localhost',
+    traces_exporter_port=4317,
+)
+client.post('/', DocumentArray([Document(text='hello')]))
+# allow some time for the tracing data export
+time.sleep(3)
```
JoanFM, line 173: no need to sleep here

JoanFM: Just comment that there may be a delay

JoanFM (changes requested):

docs/cloud-nativeness/opentelemetry.md
girishc13: Sleep is needed otherwise the Client process cannot export the data. You won't see the client span.

JoanFM: Why would this be true? The Python process should not exit until the client has finished. This is how it should be.

If you have a client and u expect to see the logs, you should not aim to have an sleep, otherwise we have a huge bug here

JoanFM (changes requested):

docs/cloud-nativeness/opentelemetry.md
girishc13: Its not about the client finishing but the usage of the batch exporter which is running in another thread. When the process exits it doesn't wait to flush the last batch. The other option is the expose the use the `client.teardown_instrumentation` method that I added that will ensure `flush` and `shutdown` calls.

JoanFM: yes, I think this is what should happen. But I do not agree that the process does not wait for the threads to finish. Python will wait for them to finish AFAIK

girishc13: The thread will finish but the thread won't ensure that the last batch has been exported.

JoanFM (approved):

Issue #5661, opened 2023-02-07
On master, Exec methods decorated with `Optional[DocumentArray]` do not work


JoanFM:
**Describe the bug**

```python

from typing import Optional
from jina import Flow, Executor, requests, DocumentArray
class A(Executor):

    @requests(on='/index')
    def index(
            self, docs: Optional[DocumentArray], parameters: dict = {}, **kwargs
    ):
        pass


f = Flow().add(uses=A)

with f:
    f.post(on='/index', inputs=DocumentArray([Document(text='a') for _ in range(2)]))
```

Most likely related to #5603

PR #5663, opened 2023-02-07
Fix v2 docarray optional

samsja:
# Context

This pr adds a default fallback if the type hint of the Executor is not correct.

this fix : https://github.com/jina-ai/jina/issues/5661

### DIFF
--- a/jina/serve/executors/__init__.py
+++ b/jina/serve/executors/__init__.py
@@ -112,6 +112,7 @@ class _FunctionWithSchema(NamedTuple):
     def get_function_with_schema(fn: Callable) -> T:
 
         docs_annotation = fn.__annotations__.get('docs', None)
+
         if type(docs_annotation) is str:
             warnings.warn(
                 f'`docs` annotation must be a type hint, got {docs_annotation}'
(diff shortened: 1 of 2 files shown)


jina/serve/executors/__init__.py
```diff
@@ -112,22 +112,38 @@ class _FunctionWithSchema(NamedTuple):
...
         if type(docs_annotation) is str:
             warnings.warn(
                 f'`docs` annotation must be a type hint, got {docs_annotation}'
                 ' instead, you should maybe remove the string annotation. Default value'
                 'DocumentArray will be used instead.'
             )
             docs_annotation = None
+        elif not isinstance(docs_annotation, type):
+            warnings.warn(
+                f'`docs` annotation must be a class if you want to use it'
```
JoanFM, line 125: Maybe u can add something specific to `Optional` that would be the most common usecase

JoanFM (changes requested):

jina/serve/executors/__init__.py
samsja: but do we even expect ppl using `Optional` here ? I think this is a relic from the past when we were sending the da to the Executor even if it was empty ?

JoanFM: but many old hub Executors are using this, so we do expect people to use old Executors yes

JoanFM: ANNLiteIndexer for instance is used and has this behavior

samsja: I see I see

JoanFM (approved):

PR #5667, opened 2023-02-08
feat: add survey placeholder

samsja:
# context

add a message for survey

### DIFF
--- a/jina/orchestrate/flow/base.py
+++ b/jina/orchestrate/flow/base.py
@@ -1975,7 +1975,7 @@ class Flow(
         print(
             'Do you love Open Source? Help us get better and be heard in just 1 minute and 30 seconds :sparkling_heart: '
             ' '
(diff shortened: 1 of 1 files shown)


JoanFM (approved):

samsja:
Screenshot from 2023-02-13 09-51-11

FYI : how it looks like

Personally I think that it takes too much place on the screen :sweat_smile: we are expecting to show this message every time someone start a flow :thinking: 

Maybe we could display it only the first X times the user start a flow ?

JoanFM:
> Screenshot from 2023-02-13 09-51-11
> 
> FYI : how it looks like
> 
> Personally I think that it takes too much place on the screen sweat_smile we are expecting to show this message every time someone start a flow thinking
> 
> Maybe we could display it only the first X times the user start a flow ?

We cannot know that

PR #5679, opened 2023-02-09
test: trying to fix deployment integration tests

alaeddine-13:
test: trying to fix deployment integration tests

### DIFF
--- a/jina/jaml/parsers/flow/v1.py
+++ b/jina/jaml/parsers/flow/v1.py
@@ -75,6 +75,7 @@ class V1Parser(VersionedYAMLParser):
                     deployments,
                     extra_search_paths=data.get('with', {}).get('extra_search_paths'),
                     include_gateway=False,
+                    noblock_on_start=True,
                 )
                 getattr(obj, 'add')(dep, copy_flow=False)
             elif (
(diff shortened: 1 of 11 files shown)


jina/jaml/parsers/flow/v1.py
```diff
@@ -75,6 +75,7 @@ def parse(
                     deployments,
                     extra_search_paths=data.get('with', {}).get('extra_search_paths'),
                     include_gateway=False,
+                    noblock_on_start=True,
```
JoanFM, line 78: I do not understand, why would this fix the issue, if at the end you are passing `True`

JoanFM, line 78: I do not understand, why would this fix the issue, if at the end you are passing `True`

JoanFM (changes requested):

jina/jaml/parsers/flow/v1.py
alaeddine-13: The issue is not with Fload().add('deployment.yml'). It happens when the user attempts to do `Deployment.load_config('deployment.yml')` and in this case, noblock_on_start will be set to false

JoanFM (approved):

tests/integration/deployments/test_deployment.py
```diff
@@ -343,7 +343,8 @@ def _create_regular_deployment(
         args.uses_after = executor if executor else 'NameChangeExecutor'
     if uses_before:
         args.uses_before = executor if executor else 'NameChangeExecutor'
-    return Deployment(args, include_gateway=False)
+    # is this needed ?
+    return Deployment(args, include_gateway=False, noblock_on_start=False)
```
alaeddine-13, line 347:
```suggestion
    return Deployment(args, include_gateway=False)
```

PR #5682, opened 2023-02-09
refactor: cancel warmup tasks

JoanFM:
Signed-off-by: Joan Fontanals Martinez <joan.martinez@jina.ai>

**Goals:**

- resolves #ISSUE-NUMBER
- ...
- ...

- ...
- ...

### DIFF
--- a/jina/serve/runtimes/gateway/__init__.py
+++ b/jina/serve/runtimes/gateway/__init__.py
@@ -125,14 +125,14 @@ class GatewayRuntime(AsyncNewLoopRuntime):
 
     async def async_teardown(self):
         """Shutdown the server."""
-        await self.cancel_warmup_task()
+        self.cancel_warmup_task()
         await self.gateway.streamer.close()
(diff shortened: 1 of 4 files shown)


jina/serve/runtimes/gateway/__init__.py
```diff
@@ -125,14 +125,14 @@ async def _wait_for_cancel(self):
 
     async def async_teardown(self):
         """Shutdown the server."""
-        await self.cancel_warmup_task()
+        self.cancel_warmup_task()
```
girishc13, line 128: I think this will cause some other issues because the task is relying on the gRPC connection pools. If the `gateway.streamer.close()` will execute before the shutdown then it will again lead to a seg fault on the warmup task.

JoanFM: but those task are safe to cancel

girishc13 (approved):

PR #5691, opened 2023-02-14
feat: interoperability docarray v2 

samsja:
# Context

this PR add schema field mapping feature from docarray v2 to ensure interoperability.

Relying on this pr : https://github.com/docarray/docarray/pull/1133

```python
    class InputDoc(BaseDocument):
        link: AnyUrl
        array: NdArray

    class OutputDoc(BaseDocument):
        embedding: NdArray

    class MyExec(Executor):
        @requests(on='/bar')
        def bar(
            self, docs: DocumentArray[InputDoc], **kwargs
        ) -> DocumentArray[OutputDoc]:

            for doc in docs:
                assert doc.link == 'hello.png'
                assert (doc.array == np.zeros(3)).all()

            docs_return = DocumentArray[OutputDoc](
                [OutputDoc(embedding=np.zeros(5)) for _ in range(len(docs))]
            )
            return docs_return

    class A(BaseDocument):
        url: AnyUrl
        tensor: NdArray

    with Deployment(uses=MyExec, field_map={'url': 'link', 'tensor': 'array'}) as dep:
        docs = dep.post(
            on='/bar',
            inputs=A(url='hello.png', tensor=np.zeros(3)),
            return_type=DocumentArray[OutputDoc],
        )
        assert docs[0].embedding.shape == (5,)
        assert docs.__class__.document_type == OutputDoc
```

### DIFF
--- a/docs/concepts/flow/deployment-args.md
+++ b/docs/concepts/flow/deployment-args.md
@@ -56,3 +56,4 @@
 | `external` | The Deployment will be considered an external Deployment that has been started independently from the Flow.This Deployment will not be context managed by the Flow. | `boolean` | `False` |
 | `grpc_metadata` | The metadata to be passed to the gRPC request. | `object` | `None` |
 | `tls` | If set, connect to deployment using tls encryption | `boolean` | `False` |
+| `field_map` |  | `object` | `{}` |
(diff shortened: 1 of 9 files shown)


JohannesMessner:
Wouldn't it make more sense to be able to pass the `field_map` at post-time instead of deployment init-time?

The way it is currently done, a user still needs to create both the deployment creation and the sending, so **they need to control client side and server side**. And every client still needs to follow this schema.
I can see this being useful only for one scenario:
- A user pulls an Executor from the Hub, but controls everything else

In contrast, if the map can be passed at post-time, the user **only needs to control the client side**.
This opens up new use cases that are not covered by the current approach:
- chaining executors in a complex Flow topology
- **serving a model publicly and letting different people connect to it**

PR #5698, opened 2023-02-19
docs: replace the legacy term pod with microservice

nan-wang:
**Goals:**

- Replace the legacy term `pod` with `microservice` on this page. https://docs.jina.ai/concepts/flow/instrumentation/


JoanFM (changes requested):
each microservice is its own microservice does not make sense

JoanFM:
I do not think this is needed

Issue #5701, opened 2023-02-20
Jina does not support docarray v2 with protobuf 4


alaeddine-13:
Jina does not support docarray v2 with protobuf 4
steps to reproduce:
pip install jina "git+https://github.com/docarray/docarray@feat-rewrite-v2#egg=docarray[common,torch,image]"
pip install protobuf==4.22.0
python -c "from jina import Executor"
output:
```
  File "/Users/alaeddine/PycharmProjects/jina/jina/orchestrate/pods/__init__.py", line 16, in <module>
    from jina.serve.runtimes.asyncio import AsyncNewLoopRuntime
  File "/Users/alaeddine/PycharmProjects/jina/jina/serve/runtimes/asyncio.py", line 14, in <module>
    from jina.serve.networking.utils import send_health_check_sync
  File "/Users/alaeddine/PycharmProjects/jina/jina/serve/networking/__init__.py", line 13, in <module>
    from jina.proto import jina_pb2
  File "/Users/alaeddine/PycharmProjects/jina/jina/proto/jina_pb2.py", line 4, in <module>
    from .pb.jina_pb2 import *
  File "/Users/alaeddine/PycharmProjects/jina/jina/proto/pb/jina_pb2.py", line 15, in <module>
    import docarray.proto.docarray_pb2 as docarray__pb2
ModuleNotFoundError: No module named 'docarray.proto.docarray_pb2'

```

PR #5702, opened 2023-02-21
fix: docarray v2 support with protobuf 4

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/5701


JoanFM (changes requested):
make sure to change the protobuf script that does this change

JoanFM (approved):

PR #5714, opened 2023-02-24
chore: unpin grpcio version

alaeddine-13:
Trying to unpin grpcio version and see compatibility with jina


Tanguyabel (dismissed):
Tested locally on my M2 machine and it works like a charm. I get not more crash when running an executor.

PR #5731, opened 2023-03-01
refactor: grpc logic to handler

JoanFM:
**Goals:**
First step towards refactoring runtime

### DIFF
--- a/jina/serve/gateway.py
+++ b/jina/serve/gateway.py
@@ -76,7 +76,7 @@ class BaseGateway(JAMLCompatible, metaclass=GatewayType):
 
         import json
 
-        from jina.serve.streamer import GatewayStreamer, _ExecutorStreamer
+        from jina.serve.runtimes.gateway.streamer import GatewayStreamer, _ExecutorStreamer
 
         graph_description = json.loads(self.runtime_args.graph_description)
         graph_conditions = json.loads(self.runtime_args.graph_conditions)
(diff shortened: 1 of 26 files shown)


alaeddine-13 (commented):
what is the plan for wait_for_ready_or_shutdown and is_ready?
I guess with the runtime becoming separate from serving logic, this logic should be offloaded to the handler as well ?

JoanFM:
> runtime

I do not have a specific strategy, but anyway it is not relevant

jina/serve/gateway.py
```diff
@@ -73,53 +74,13 @@ def __init__(
...
-            graph_conditions=graph_conditions,
-            deployments_metadata=deployments_metadata,
-            deployments_no_reduce=deployments_no_reduce,
-            timeout_send=self.runtime_args.timeout_send,
-            retries=self.runtime_args.retries,
-            compression=self.runtime_args.compression,
-            runtime_name=self.runtime_args.runtime_name,
-            prefetch=self.runtime_args.prefetch,
+        self.logger = JinaLogger(self.name, **vars(self.runtime_args))
+        self._request_handler = GatewayRequestHandler(
```
alaeddine-13, line 78: not sure I understand the motivation behind blending together streamer and executor into `request_handler` object.
I mostly see the Gateway unaffected by the upcoming refactoring in the sense that users who implement a custom gateway will own the serve and application/request handling layer.

I think since this class is exposed to the user, we should keep it intact and in case we need to adapt it to our case, better to have a jina specific `BaseGateway` to be used as the Base server

alaeddine-13 (approved):

PR #5733, opened 2023-03-02
feat: retry client grpc stream and unary rpc methods

### LINKED ISSUE #5730: bug: retries are not handled correctly
**Describe the bug**

The below snippet doesn't respect the `max_attempts` parameter and immediately raises a `BadClient` error.
```python
from jina import Client, Document, DocumentArray

Client(host='https://dear-bedbug-9e545b601d.wolf.jina.ai/').post(
    on='/',
    inputs=DocumentArray.empty(),
    max_attempts=100,
)
```

**Describe how you solve it**

The `send_message` method in `jina/clients/base/helper.py` must be modified as below:
```python
try:
    response = await self.session.post(url=self.url, json=req_dict).__aenter__()
    response.raise_for_status()
    return response
except:
         ....
```

---

**Environment**

**Screenshots**

girishc13:
**Goals:**

- resolves #5730 
- extract out grpc streming and unary logic to classes.
- create methods for retries.
- return retry attempts as rpc metadata.
- gRPC `client.post` method with the default `stream=True` parameter only supports the default gRPC retry mechanism which might not succeed or attempt a retry due to design restrictions. The user needs to implement their own retry logic as shown in the documentation snippet.
- [x] update and add tests.


tests/unit/clients/python/test_client.py
```diff
@@ -216,3 +287,183 @@ def test_deployment_sync_client(mocker, use_stream):
...
+        stop_event.set()
+        t.join()
+        t.terminate()
+
+
+@pytest.mark.timeout(90)
+@pytest.mark.asyncio
+@pytest.mark.parametrize('flow_or_deployment', ['deployment', 'flow'])
+@pytest.mark.ignore
+async def test_async_grpc_stream_transient_error(
```
girishc13, line 418: @alaeddine-13 @JoanFM This test is ignored because it creates a segmentation fault. I need some help to find the root cause.

girishc13: Looking at the logs, the warmup tasks plus the client retires might be causing an issue with the grpc asyncio implementation.

docs/concepts/client/transient-errors.md
```diff
@@ -0,0 +1,102 @@
...
+
+The {meth}`~jina.clients.mixin.PostMixin.post` method supports the `stream` boolean parameter. If set to `True`,
+the **gRPC** server side streaming RPC method will be invoked. If set to `False`, the server side unary RPC method will
+be invoked. Some important implication of
+using retires with **gRPC** are as follows:
+
+1. The built-in **gRPC** retries are limited in scope and are implemented to work under certain circumstances. More
+   details are specified in the [design document](https://github.com/grpc/proposal/blob/master/A6-client-retries.md).
+2. If `stream` parameter is set to True (defaults to True), the retry must
+   be handled as below because the result must be consumed to check for errors in the stream of responses.
```
alexcg1, line 41: Just use bullets unless we're calling back to specific list items later in the doc

docs/concepts/client/transient-errors.md
```diff
@@ -0,0 +1,102 @@
...
+to Exceptions caused by an Executor, but in case of network connectivity issues, an Exception will be raised.
+
+## Retries with a large inputs or long-running operations
+
+When using the gRPC client, it is recommended to set the `stream` parameter to False so that the unary RPC is invoked by
+the {class}`~jina.Client`
+which performs the retry internally with the request from the `inputs` iterator or generator. The `request_size`
+parameter must also be set to perform smaller operations which can be retried without much overhead on the server.
+
+The **HTTP** and **Websocket**
```
alexcg1, line 98:
```suggestion
The **HTTP** and **WebSocket**
```

alexcg1 (changes requested):

jina/clients/base/retry.py
```diff
@@ -0,0 +1,53 @@
+import asyncio
+import random
+
+import grpc
+
+
+async def wait_or_raise_err(
+    attempt, err, max_attempts, backoff_multiplier, initial_backoff, max_backoff
```
JoanFM, line 8: can we have some basic type hinting?

JoanFM (changes requested):

JoanFM:
[](url)

jina/clients/base/stream_rpc.py
```diff
@@ -0,0 +1,68 @@
...
+        self.p_bar = p_bar
+        self.on_error = on_error
+        self.on_done = on_done
+        self.on_always = on_always
+        self.metadata = metadata
+        self.continue_on_error = continue_on_error
+        self.channel = channel
+        self.kwargs = kwargs
+
+    async def stream_rpc_with_retry(self):
```
numb3r3, line 46: `stream_rpc_with_retry` is misleading. As you described, the stream RPC does not support retry.

jina/clients/base/stream_rpc.py
```diff
@@ -0,0 +1,68 @@
...
+        compression,
+        **kwargs
+    ):
+        self.compression = compression
+        self.show_progress = show_progress
+        self.logger = logger
+        self.max_backoff = max_backoff
+        self.initial_backoff = initial_backoff
+        self.backoff_multiplier = backoff_multiplier
+        self.max_attempts = max_attempts
```
numb3r3, line 35: The `max_attempts` would not take effect, right? I'm not so sure that is it reasonable to remind the user `max_attemps` is not considered in stream rpc.

jina/clients/base/unary_rpc.py
```diff
@@ -0,0 +1,126 @@
...
+    async def unary_rpc_with_retry(self):
+        """Wraps the unary rpc call with retry loop based on the retry params.
+        :yields: Responses received from the target.
+        """
+        stub = jina_pb2_grpc.JinaSingleDataRequestRPCStub(self.channel)
+
+        def _request_handler(
+            request: 'Request',
+        ) -> 'Tuple[asyncio.Future, Optional[asyncio.Future]]':
+            async def _with_retry(req: 'Request'):
```
numb3r3, line 72: where `_with_retry` is imported? Is it an error here?

jina/clients/base/unary_rpc.py
```diff
@@ -0,0 +1,126 @@
...
+                for attempt in range(1, self.max_attempts + 1):
```
numb3r3, line 73: Shall we add an assert `self.max_attempts >= 1` to valide the parameter? If the `self.max_attempts` is set to 0, then, the request would not be processed.

jina/clients/base/unary_rpc.py
numb3r3: ok, i see, the `_with_retry` is implemented in a mixin.

numb3r3: hold on, I'm a bit confusing. It seems `_with_retry` implementation (`jina/clients/mixin.py`) supports multiple attempts. Why do we use a for-loop below to retry `max_attempts` .

numb3r3 (changes requested):

jina/clients/base/stream_rpc.py
girishc13: It's still valid because the grpc.serviceoptions are still enabled and the retry might work if certain conditions are met. If the documentation is not clear, I will update it.

jina/clients/base/unary_rpc.py
girishc13: I will just log a warning and reset the param to min 1 instead of an assert.

jina/clients/base/unary_rpc.py
girishc13: I need to update the code but good catch. The `_with_retry` in the (jina/clients/minxin.py) must only be used when sync client and when `stream=True`. If `stream=False` then the retry here must take effect. There are a lot of considerations and optimizations to make it easy for the user.

docs/concepts/client/transient-errors.md
```diff
@@ -0,0 +1,108 @@
...
+
+```{hint}
+The retry parameters `max_attempts`, `initial_backoff`, `backoff_multiplier` and `max_backoff` of the {meth}`~jina.clients.mixin.PostMixin.post` method will be used to set the **gRPC** retry service options. This improves the chances of success if the gRPC retry conditions are met.
+```
+
+## Continue streaming when an Executor error occurs
+
+The {meth}`~jina.clients.mixin.PostMixin.post` accepts a `continue_on_error` parameter. When set to `True`, the Client
+will keep trying to send the remaining requests. The `continue_on_error` parameter will only apply
+to Exceptions caused by an Executor, but in case of network connectivity issues, an Exception will be raised.
```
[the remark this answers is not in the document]
girishc13: I added more clarification below. Please review.

girishc13:
@alexcg1 Please do re-review of the docs.

JoanFM (approved):

girishc13:
@numb3r3 @ZiniuYu Please check again.

JoanFM (dismissed):

tests/unit/clients/python/test_client.py
JoanFM: I think we should remove `warmup tasks`

JoanFM (approved):

PR #5735, opened 2023-03-03
refactor: first step refactor, one Runtime class

JoanFM:
**Goals:**
Refactor the `Runtime` module so that we can add different protocols easily to different entities like Worker and Head.

- [x] Have a single Runtime class
- [x] Have a BaseServer and 4 server subclasses (GRPC, HTTP, WebScoket, Composite)
- [x] Have BaseGateway just inherit from BaseServer with a thin layer just providing JAMLCompatibiolity and make sure it is backwards compatible.
- [x] Each server will have its RequestHandler injected depending on the `runtime_cls` (Rely on `runtime_cls` for backwards compatibility)
- [x] Review logging
- [x] Make sure `send_telemetry_event` is backwards compatible
- [x] Make sure all tests pass as used to be


girishc13 (approved):
We need to do some refactoring to make the reduce the long methods. I don't any major issues for now. Let's merge and do some follow up work.

PR #5736, opened 2023-03-03
fix: enable root logging always

girishc13:
**Goals:**

- enable root logging always to allow DEBUG logging of other dependencies.

The current JINA_LOG_LEVEL controls the log level of the `JinaLogger`. Debug logging of other dependencies like urllib3 can be enabled by adding `logging.get_logger('urllib3').setLevel(logging.DEBUG)` at start of the Flow when running locally. This of course has limitations when deploying Gateway & Executors to the cloud.

### DIFF
--- a/jina/logging/logger.py
+++ b/jina/logging/logger.py
@@ -118,7 +118,6 @@ class JinaLogger:
         name: Optional[str] = None,
         log_config: Optional[str] = None,
         quiet: bool = False,
-        suppress_root_logging: bool = True,
         **kwargs,
     ):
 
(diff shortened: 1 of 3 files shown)


JoanFM (approved):

jina/logging/logger.py
```diff
@@ -105,7 +105,7 @@ class JinaLogger:
 
     :param context: The context identifier of the class, module or method.
     :param log_config: The configuration file for the logger.
-    :param suppress_root_logging: The flag determines whether to suppress root logging.
+    :param suppress_root_logging: Deprecated flag that determines whether to suppress root logging. Root logging is enabled by default.
```
JoanFM, line 108: we can remove this since, it will be swallowed by `kwargs`

JoanFM (changes requested):

JoanFM (approved):

Issue #5737, opened 2023-03-04
[bug] [feature request] Starter documentation links do not jump and become unnecessary


zhimin-z:
**Describe the bug**
There exist two unnecessary links in the starter tutorial.
Here is the starter documentation: https://docs.jina.ai
Here are the unnecessary links in the starter:
![image](https://user-images.githubusercontent.com/8592144/222920954-b0faedbf-68e8-4c48-bbe5-f773472ba29b.png)

**Screenshots**
After clicking, you land on exactly the same page as the current page...
Why shall we create two links that do not jump at all? 
Maybe we can make a fancy way to stress Jina to have the two functions rather than staying on the original documentation page.

zhimin-z:
@JoanFM @alaeddine-13 I found these two links jump to the same place, is that designed as so. Who do we need two links to jump to the exact same place? why not one?

PR #5740, opened 2023-03-06
docs: add readme link

alaeddine-13:
fix readme link
closes: https://github.com/jina-ai/jina/issues/5737


JoanFM (approved):

zhimin-z:
@JoanFM @alaeddine-13 I found these two links jump to the same place, is that designed as so. Who do we need two links to jump to the exact same place? why not one?

PR #5741, opened 2023-03-06
fix: composite gateway tracing support

alaeddine-13:
fix: composite gateway tracing support

before the PR, doing `jina gateway --protocol HTTP GRPC --port 8081 8085 --tracing` should result in an error
```grpc._cython.cygrpc.Channel.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__```

### DIFF
--- a/jina/serve/runtimes/gateway/request_handling.py
+++ b/jina/serve/runtimes/gateway/request_handling.py
@@ -38,27 +38,24 @@ class GatewayRequestHandler:
         deployments_metadata = json.loads(self.runtime_args.deployments_metadata)
         deployments_no_reduce = json.loads(self.runtime_args.deployments_no_reduce)
 
-        print('init streamer')
-        if streamer:
-            self.streamer = streamer
-        else:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: jina/serve/runtimes/gateway/request_handling.py [hunk 1]
```diff
@@ -30,23 +38,27 @@ def __init__(
...
-            runtime_name=self.runtime_args.runtime_name,
-            prefetch=self.runtime_args.prefetch,
-            logger=self.logger,
-            metrics_registry=self.runtime_args.metrics_registry,
-            meter=self.runtime_args.meter,
-            aio_tracing_client_interceptors=self.runtime_args.aio_tracing_client_interceptors,
-            tracing_client_interceptor=self.runtime_args.tracing_client_interceptor,
-        )
+        print('init streamer')
+        if streamer:
```
JoanFM, line 42: use the `self.streamer = streamer or ` syntax

PR #5764, opened 2023-03-16
feat: enable HTTP and Composite protocols for Deployment

JoanFM:
**Goals:**
This PR would enable exposing a single-Executor Deployment with an HTTP protocol and a Composition of `HTTP` and `GRPC` protocols.

- [x] Expose input and output models for each `endpoint` in `endpoint_discovery` method so that it can be used by Head or Gateway to build the FastAPI app for Executor locally.
- [x] Forbid feature when `shards` > 1
- [x] Check TLS support
- [x] Given this information, build a FastAPI app that has each `endpoint` mapped to a `app.post` method with the endpoint in `path` and the corresponding input and output models. (DocArray with Parameters).
- [x] Make sure to wrap the models with `parameters` into an extended model
- [x] Expose this app for Worker, Head, and Gateway.
- [x] Map Exceptions to HTTP Exceptions inside these FastAPI apps.
- [x] Adapt the client not to send full Requests in this case and handle Exceptions properly. For this, HTTPClient should be able to check the RequestModel expected
- [x] Check resulting Swagger UIs.
- [x] Add tests.
- [ ] Test with docarray v1 and v2
- [x] Test with/without shards and with/without include_gateway
- [x] Test fast and slow-to-load Executors (test endpoint_discovery)
- [x] Test with and without accepting parameters, and Executors returning parameters
- [ ] Test docker compose and K8s
- [x] Test all the rest with composition of `grpc` and `http`


JoanFM:
-

tests/integration/deployment_http_composite/test_deployment_http_composite.py
```diff
@@ -0,0 +1,139 @@
+import os
+import time
+import pytest
+
+from jina import Deployment, Executor, requests, Client, DocumentArray
+from jina.helper import random_port
+
+
+class SingleExecutorDeployment(Executor):
```
alaeddine-13, line 9: what about `d.post(on='/bar', inputs=DocumentArray.empty(10), parameters={...})` how does it behave ?

jina/clients/base/http.py
```diff
@@ -18,6 +18,34 @@
...
+        proto = 'https' if self.args.tls else 'http'
+        target_url = f'{proto}://{self.args.host}:{self.args.port}/openapi.json'
+        try:
+            async with aiohttp.ClientSession() as session:
+                async with session.get(target_url) as response:
+                    content = await response.read()
+                    openapi_response = json.loads(content.decode())
+                    self._endpoints = extract_paths_by_method(openapi_response).get('post', [])
+        except:
+            pass
```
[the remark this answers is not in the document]
JoanFM: in most cases it would.

tests/integration/deployment_http_composite/test_deployment_http_composite.py
JoanFM: oh good point, will test

PR #5767, opened 2023-03-20
chore: update README.md to handle problems during step by step learning

learningpro:
add transformers and sentencepiece in pip install commands

add import line in example python file:
from translate_executor import Translator


alaeddine-13 (approved):
LGTM

JoanFM (approved):

PR #5770, opened 2023-03-22
test: add integration test for stateful executor

alaeddine-13:
stateful executor tests


jina/serve/consensus/run.go
```diff
@@ -201,6 +203,7 @@ func Run(myAddr string,
         sig := <-sigchnl
         log.Printf("Signal %v received", sig)
         grpcServer.Stop()
+        sock.Close()
```
JoanFM, line 206: why is it needed if it is already defered?

JoanFM: is it because the `os.Exit`?

alaeddine-13: according to my research it should be called in sig term handling

PR #5777, opened 2023-03-29
ci: build wheels for stateful executor jraft golang package

alaeddine-13:
ci: build wheels for stateful executor jraft golang package


.github/workflows/stateful-build-wheels.yml
```diff
@@ -0,0 +1,149 @@
+name: Jina v-raft Workflow
```
JoanFM, line 1: this is supposed to be a temporary file?

JoanFM (changes requested):

.github/workflows/stateful-build-wheels.yml
alaeddine-13: yes, before merging jina-v-raft to master, we don't rely on `stateful` tag anymore and incorporate the wheels building step in the regular release workflow

PR #5821, opened 2023-04-23
fix: encoding

florian-hoenicke:
Currently there are encoding problems when using jina under windows as we can see in this issue:
https://github.com/jina-ai/gptdeploy/issues/49

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ try:
     if 'JINA_VERSION' not in os.environ:
         pkg_name = 'jina'
         libinfo_py = path.join(repo_dir, pkg_name, '__init__.py')
-        libinfo_content = open(libinfo_py, 'r').readlines()
+        libinfo_content = open(libinfo_py, 'r', encoding='utf-8').readlines()
         version_line = [
             l.strip() for l in libinfo_content if l.startswith('__version__')
         ][0]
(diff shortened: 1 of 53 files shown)


JoanFM:
Have you tried doing this?

florian-hoenicke:
@JoanFM Yes, I just tried PYTHONUTF8=1, , but it is also not working

JoanFM:
I am a little concerned if this may cause any trouble for Windows user who write their files on Windows and for whom the encoding would not be this one.

florian-hoenicke:
ASCII is a subset of UTF-8. So those files would be able to open.
With this, we have covered almost all possible cases
What you can do at some point is to detect the encoding
```
import chardet
with open(file_path, 'rb') as file:
    raw_data = file.read()
    encoding = chardet.detect(raw_data)['encoding']
with open(file_path, 'r', encoding=encoding) as file:
    content = file.read()
```

florian-hoenicke:
utf-8 is standard
VS code uses it:
image
Pycharm uses it:
image
Atom uses it:
image
Even python itself promotes utf-8
image

There is no reason we should support something else.

JoanFM (approved):

florian-hoenicke:
Is it normal that the graceful test takes that long and the ecosystem test is failing? @JoanFM

PR #5825, opened 2023-04-24
feat: catch ContentTypeError for retry in POST requests

Tanguyabel:
When querying a POST route we always try to retrive a json format from the response. It happens that we sometimes receive a plain html format, in case the json cannot be decoded, ended up in a raise of the following error: `Attempt to decode JSON with unexpected mimetype: text/html; charset=utf-8`. Adding this type of Exception in the catch allow to perform a retry.

### DIFF
--- a/jina/clients/base/helper.py
+++ b/jina/clients/base/helper.py
@@ -150,7 +150,7 @@ class HTTPClientlet(AioHttpClientlet):
                     r_str = response.text
                 handle_response_status(response.status, r_str, self.url)
                 return response
-            except (ValueError, ConnectionError, BadClient, aiohttp.ClientError, aiohttp.ContentTypeError) as err:
+            except (ValueError, ConnectionError, BadClient, aiohttp.ClientError) as err:
                 await retry.wait_or_raise_err(
                     attempt=attempt,
                     err=err,
--- a/jina/clients/base/helper.py
+++ b/jina/clients/base/helper.py
@@ -147,7 +147,7 @@ class HTTPClientlet(AioHttpClientlet):
                 r_str = await response.json()
                 handle_response_status(response.status, r_str, self.url)
                 return response
-            except (ValueError, ConnectionError, BadClient, aiohttp.ClientError) as err:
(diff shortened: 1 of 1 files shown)


jina/clients/base/helper.py
```diff
@@ -147,7 +147,7 @@ async def send_message(self, request: 'Request'):
                 r_str = await response.json()
                 handle_response_status(response.status, r_str, self.url)
                 return response
-            except (ValueError, ConnectionError, BadClient, aiohttp.ClientError) as err:
+            except (ValueError, ConnectionError, BadClient, aiohttp.ClientError, aiohttp.ContentTypeError) as err:
```
girishc13, line 150: Why don't we check the content-type to extract the `r_str`. Here the status code is more important or should be checked first before extracting the payload. The underlying issue is a `504` status code with some default response from the Ingress component which is used raise a new error. You can check the `handle_response_status` method also which already handles the `504` http error code.

Tanguyabel: You are right, I am going to check the status code prior to decoding.

JoanFM (approved):

jina/clients/base/helper.py
```diff
@@ -171,7 +171,10 @@ async def send_message(self, request: 'Request'):
                 response = await self.session.post(
                     **request_kwargs
                 ).__aenter__()
-                r_str = await response.json()
+                try:
+                    r_str = await response.json()
+                except aiohttp.ContentTypeError:
+                    r_str = response.text
```
girishc13, line 177: This should be `response.text()` since we are using aiohttp package.
https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request

Tanguyabel: Weird, I tried, and `response.text` was the one that worked here. If this comes up as an error I will create a quick fix.

JoanFM: @girishc13 could it be that you have already `awaited` the `response`?

girishc13: Ah yes, could be. Since we have already have the response the `.text` is readily accessible.

PR #5845, opened 2023-05-04
ci integrate wheels building

alaeddine-13:
After this PR, each release will:
- build wheels for each platform in a matrix job
- upload wheels to github artifacts
- in the release job, will download wheels all together
- will publish them to pypi


scripts/release.sh
```diff
@@ -36,7 +36,6 @@ function pub_pypi {
     # publish to pypi
     clean_build
     cp extra-requirements.txt jina/resources/
-    python setup.py sdist
```
alaeddine-13, line 39: Actually I think this should not be removed, need to try out

JoanFM: how can we try this out?

JoanFM (approved):

PR #5899, opened 2023-05-30
feat: streaming with HTTP SSE for deployment

alaeddine-13:
feat: streaming with HTTP SSE for deployment
closes: https://github.com/jina-ai/jina/issues/5898


jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -44,18 +54,36 @@ class InnerConfig(BaseConfig):
 
     app = FastAPI()
 
-    def add_route(endpoint_path, input_model, output_model, input_doc_list_model=None, output_doc_list_model=None):
-        app_kwargs = dict(path=f'/{endpoint_path.strip("/")}',
-                          methods=['POST'],
-                          summary=f'Endpoint {endpoint_path}',
-                          response_model=output_model, )
+    if cors:
```
JoanFM, line 57: please add tests for this feature independently of `SSE`

jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -70,6 +98,8 @@ async def post(body: input_model, response: Response):
             req.parameters = body.parameters
             req.header.exec_endpoint = endpoint_path
             resp = await caller(req)
+            if request_models_map[endpoint_path]['is_generator']:
```
JoanFM, line 101: this `is_generator`  flag should be computed outside the `post` so that u do not do this access here

jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -112,6 +109,36 @@ async def post(body: input_model, response: Response):
...
+            else:
+                req.data.docs = DocList[input_doc_list_model](body.data)
+
+            if body.header is not None:
+                req.header.request_id = body.header.request_id
+
+            req.parameters = body.parameters
+            req.header.exec_endpoint = endpoint_path
+            resp = await caller(req)
+            return EventSourceResponse(resp)
```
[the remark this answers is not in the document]
alaeddine-13: `resp` is a generator yes
`EventSourceResponse(resp)` will stream events to the connection based on the generator

docs/concepts/serving/executor/add-endpoints.md
```diff
@@ -187,6 +187,70 @@ with Deployment(uses=MyExec) as dep:
 {"__results__": {"my_executor/rep-0": {"internal_parameter": 20.0}}}
 ```
   
+
+## Streaming endpoints
```
JoanFM, line 191: You have to say that this is currently only available in HTTP protoocl and Deployment

JoanFM (changes requested):

docs/concepts/serving/executor/add-endpoints.md
```diff
@@ -194,6 +194,12 @@ This is useful when you want to return Documents one by one and you want the cli
 they arrive. This can be helpful for Generative AI use cases, where a Large Language Model is used to generate text 
 token by token and the client displays tokens as they arrive.
 Streaming endpoints receive one Document as input and yields one Document at a time.
+```{admonition} Note
+:class: note
+
+Streaming endpoints are only supported for HTTP protocol and for Deployment.
```
JoanFM, line 200: Streaming endpoints are currently only available when Executors are served inside a single `Deployment` with `HTTP` protocool

JoanFM (changes requested):

docs/concepts/serving/executor/add-endpoints.md
alaeddine-13: check here: https://feat-deployment-http-sse--jina-docs.netlify.app/concepts/serving/executor/add-endpoints/#streaming-endpoints

jina/serve/runtimes/worker/http_fastapi_app.py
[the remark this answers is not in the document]
JoanFM: can u then call it resp_generator?

jina/clients/base/http.py
```diff
@@ -212,3 +223,30 @@ def _result_handler(result):
...
+        self,
+        on: str,
+        inputs: 'Document',
+        parameters: Optional[Dict] = None,
+        return_type: Type[Document] = Document,
+        timeout: Optional[int] = None,
+        **kwargs,
+    ):
+        proto = 'https' if self.args.tls else 'http'
+        url = f'{proto}://{self.args.host}:{self.args.port}/{on.strip("/")}'
```
JoanFM, line 237: this `on` is recomputed based on the `logic` used in `post` where it checks the OpenAPI.json to detect if the endpoint is there and otherwise it goes to default or something like that?

jina/clients/mixin.py
```diff
@@ -521,6 +521,39 @@ async def post(
...
+    async def stream_doc(
+        self,
+        on: str,
+        inputs: Document,
+        parameters: Optional[Dict] = None,
+        return_type: Type[Document] = Document,
+        timeout: Optional[int] = None,
+        **kwargs,
+    ) -> AsyncGenerator[None, 'Document']:
+        """Send one document to a streaming endpoint and receive results asynchronisly.
```
JoanFM, line 533: not only results asyncronously, it receives also a single Document

jina/helper.py
```diff
@@ -1666,3 +1664,14 @@ def _telemetry():
             pass
 
     threading.Thread(target=_telemetry, daemon=True).start()
+
+
+def is_generator(func):
```
JoanFM, line 1669: is this needed?

jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -82,33 +109,105 @@ async def post(body: input_model, response: Response):
...
+                __config__=output_doc_model.__config__,
             )
 
-            add_route(endpoint,
-                      input_model=endpoint_input_model,
-                      output_model=endpoint_output_model,
-                      input_doc_list_model=input_doc_model,
-                      output_doc_list_model=output_doc_model)
+            if is_generator:
+                add_streaming_post_route(
```
JoanFM, line 189: I do not understand here, why 2 routes are added?

JoanFM (changes requested):
You need to make sure these `streaming tests` are passing in CI with the `docarray-v-two-tests`

JoanFM (commented):
Make sure this also works for `docarray_v1`

jina/helper.py
alaeddine-13: yes

jina/serve/runtimes/worker/http_fastapi_app.py
alaeddine-13: yeah actually the post one is not needed, removed it

jina/clients/mixin.py
alaeddine-13: I improved the wording to specify that it receives results document by document

jina/clients/base/http.py
alaeddine-13: it's actually inspired from this section: https://github.com/jina-ai/jina/blob/463eff659eeb5ac27be7fa1c569550ee546e3796/jina/clients/base/http.py#L148

JoanFM: but what if the `on` is not a deployed endpoint? as it can happen for docarray v1

JoanFM: Check the 3 if elses here:

```python
            if endpoint != '' and endpoint in self._endpoints:
                url = f'{proto}://{self.args.host}:{self.args.port}/{on.strip("/")}'
            elif has_default_endpoint:
                url = f'{proto}://{self.args.host}:{self.args.port}/default'
            else:
                url = f'{proto}://{self.args.host}:{self.args.port}/post'
```

jina/serve/runtimes/worker/http_fastapi_app.py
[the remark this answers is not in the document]
alaeddine-13: not that the snippet is outdated, this is the actual one:
```python
            event_generator = _gen_dict_documents(await caller(req))
            return EventSourceResponse(event_generator)
```

I think that event_generator is more accurate, but will change it if you believe `resp_generator` is more accurate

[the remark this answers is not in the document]
alaeddine-13: not that the snippet is outdated, this is the actual one:
```python
            event_generator = _gen_dict_documents(await caller(req))
            return EventSourceResponse(event_generator)
```

I think that event_generator is more accurate, but will change it if you believe `resp_generator` is more accurate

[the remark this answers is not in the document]
JoanFM: okey event generator is good

docs/concepts/serving/executor/add-endpoints.md
```diff
@@ -187,6 +187,76 @@ with Deployment(uses=MyExec) as dep:
...
+```
+
+A streaming endpoint has the following signature:
+
+```python
+from jina import Executor, requests, Document, Deployment
```
JoanFM, line 206: can u add this as supported part in `docarray v2` support, where u show an example with the new way?

JoanFM (changes requested):

jina/clients/base/http.py
alaeddine-13: IMO, it's best to not mutate the endpoint that the user specifies. If it's not a deployed endpoint, it will fail.
I changed to the following logic: if the endpoint is in self._endpoints or there are no default endpoints, then use the endpoint as it is, otherwise it will use the default endpoint.
It does not make sense to use `/post` in this case

JoanFM (approved):

Issue #5690, opened 2023-02-14
Introduce Job concepts to make Executors suitable for 1-time Request


JoanFM:
**Describe the feature**
Provide this:

```python
job = Job(uses=Executor, retries=3, ...)
job.run(on='/foo', inputs='s3....', output='s3....')
job.to_kubernetes_yaml() # should map to a K8s BatchJob
```

asuzukosi:
@JoanFM Hey I would like to work on this problem, please can you provide some additional information on how this feature would work in practice

PR #5948, opened 2023-07-04
docs: add docs for secrets & jobs

npitsillos:
**Goals:**
- Adds documentation for the `jcloud` for the `secrets` and `jobs` support


docs/concepts/jcloud/index.md
```diff
@@ -164,12 +196,14 @@ To access the [Grafana](https://grafana.com/)-powered dashboard, first get {ref}
...
+### List Flows, Secrets or Jobs
+
+#### List Flows
 
-To list all of your "Serving" Flows:
+To list all of your "Starting", "Serving", "Failed", "Updating", and "Paused" Flows:
 
 ```bash
-jc list
+jc list flows
```
deepankarm, line 206: Is this a breaking change? Does `jc list/update/remove` etc need params `flow/job/secret` etc now?

docs/concepts/jcloud/index.md
```diff
@@ -82,6 +84,36 @@ jina flow --uses flow.yml
...
+#### Create a Secret
+
+To create a Secret for a Flow:
+
+```bash
+jc create secret mysecret -f rich-husky-af14064067 --from-literal "{'env-name': 'secret-value'}"
+```
+
+```{tip}
+You can optionally pass the `--update` flag to automatically update the Flow spec with the updated secret information. This flag will update the Flow which is hosted on the cloud. Finally, you can also optionally pass a Flow's yaml file path with `--path` to update the yaml file locally.
```
JoanFM, line 98: Link to the `env-from-secrets` parameter in Jina Docs. https://docs.jina.ai/cloud-nativeness/kubernetes/#deploy-flow-with-custom-environment-variables-and-secrets

docs/concepts/jcloud/index.md
```diff
@@ -82,6 +84,36 @@ jina flow --uses flow.yml
...
+#### Create a Job
+
+To create a Job for a Flow:
+
+```bash
+jc create job job-name rich-husky-af14064067 image 'job entrypoint' --timeout 600 --backofflimit 2
+```
+
+```{tip}
+`image` can be any Executor image passed to a Flow's Executor `uses` or any normal docker image prefixed with `docker://`
```
JoanFM, line 114: in this case it is not true, unfortunately, an Executor image cannot be converted to a Flow. We have a ticket about this, but no time for it. https://github.com/jina-ai/jina/issues/5690

JoanFM (changes requested):

docs/concepts/jcloud/index.md
npitsillos: Yes we need this to distinguish between the resource to create.

docs/concepts/jcloud/index.md
npitsillos: You mean it cannot be converted to a job? I added tests for the images here https://github.com/jina-ai/jina-operator/blob/main/api/test/e2e/jobs_test.go and it is running fine adding the executor image.

JoanFM: but what is the entrypoint?

npitsillos: Keep in mind I am only running `jina` as the entrypoint and checking that the logs are not empty in this case.

JoanFM (approved):

docs/concepts/jcloud/index.md
JoanFM: but can we have `flow` as default? to avoid breaking changes?

npitsillos:
Let's not merge this until I update `jcloud` to support the resource as a sub-command. Currently the user should run `jc list jobs` however we will change it to `jc jobs list` instead. Same for the other resources

deepankarm (changes requested):
image

The sections are weirdly structured. Please change it to resource-wise - one section each for `Flows`, `Jobs` & `Secrets`

PR #6013, opened 2023-08-03
docs: do not use tokenizer in client for llm streaming example

alaeddine-13:
docs: do not use tokenizer in client for llm streaming example

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,7 @@ class PromptDocument(BaseDoc):
 
 class TokenDocument(BaseDoc):
     token_id: int
+    generated_text: str
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -368,6 +368,7 @@ class PromptDocument(BaseDoc):
 
 class TokenDocument(BaseDoc):
     token_id: int
+    generated_text: str
```
JoanFM, line 371: Lets do not call it TokenDocument anymore

JoanFM (changes requested):
I would change the naming

JoanFM (approved):

Issue #6019, opened 2023-08-07
Getting endpoints failed: 'definitions'. Waiting for another trial


Song-Gy:
**Describe your proposal/problem**

When the docs structure of the set access node involves nesting, the service will fail to start normally;
---

**Environment**

jina- 3.19.1
docarray-0.37.0
**Screenshots**

```python
class RectMeta(BaseDoc):
    top: float
    left: float
    width: float
    height: float

class ObjMeta(BaseDoc):
    classId: Optional[int]
    objectId: Optional[int]
    confidence: Optional[float]
    rectParams: Optional[RectMeta]

class FrameMeata(BaseDoc):
    taskId: str
    # Path of source
    dataPath: str
    # Current frame number of source
    frameNum: Optional[int]
    # List of objects of type VizorObjectMetaDoc in usr for give frame
    objMetaList: Optional[List[ObjMeta]]
class BLSExec(Executor):
    def __init__(self,
         
                 *args,
                 **kwargs):
        super().__init__(*args, **kwargs)

    @requests(on="/submit")
    async def submit(self,
                     docs: DocList[FrameMeata],
                     **kwargs) -> DocList[ResponseDoc]:
        print(docs)
        for doc in docs:
            doc.summary()
        rets = DocList[ResponseDoc]()
        # for  in docs:
        # doc.summary()
        # rets.append(self.controller.process_submit(doc=doc))
        return rets

bls_flow = (Flow(log_config=str(log_cfg.absolute()))
                        .add(name='inference',
                             needs='gateway',
                             uses=BLSExec)
                        )
bls_flow = bls_flow.config_gateway(protocol='http', ports=11777)
bls_flow.start()
```
After starting the service according to the above code, the console keeps outputting the following information, and the client cannot access the service normally
```shell
2023-08-07 02:49:31,484 topology_graph.py[161] DEBUG Getting Endpoints data from viznrn_bls
2023-08-07 02:49:31,485 request_handling.py[881] DEBUG got an endpoint discovery request
2023-08-07 02:49:31,510 topology_graph.py[608] WARNING Getting endpoints failed: 'definitions'. Waiting for another trial
2023-08-07 02:49:32,512 topology_graph.py[161] DEBUG Getting Endpoints data from viznrn_bls
2023-08-07 02:49:32,513 request_handling.py[881] DEBUG got an endpoint discovery request
[...]

PR #6021, opened 2023-08-07
fix: refer to global schema definitions when creating models

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/6019


JoanFM (changes requested):
Can we have a test for the specific issue, maube create a new file under docarray_v2 with test_issues.py where we add tests for specific issues.

tests/unit/serve/runtimes/test_helper.py
```diff
@@ -108,26 +116,43 @@ class CustomDoc(BaseDoc):
         u: Union[str, int]
         lu: List[Union[str, int]] = [0, 1, 2]
         tags: Optional[Dict[str, Any]] = None
+        nested: NestedDoc
```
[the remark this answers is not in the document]
alaeddine-13: @JoanFM  my bad the issue appears only at 2 or more levels of nesting. I updated the tests

tests/integration/docarray_v2/test_issues.py
```diff
@@ -0,0 +1,41 @@
+from typing import List, Optional
+
+from docarray import BaseDoc, DocList
+
+from jina import Executor, Flow, requests
+
+
+class Nested2Doc(BaseDoc):
```
JoanFM, line 8: can we copy the exact same structure as the issue?

JoanFM (changes requested):

tests/integration/docarray_v2/test_issues.py
alaeddine-13: I think the only difference is the use of `Optional` and `List`. I added another case for this

Issue #6020, opened 2023-08-07
Streaming Endpoint Issues with Gateways


chenxiaolong-coder:
**Describe the bug**

I found that the gateway caches the data to a certain size before sending it to the client, which can cause the typewriter effect to fail
**Describe how you solve it**

    set param include_gateway to False like this:

    dep = Deployment(uses=XunFeiLLM, timeout_ready=-1, protocol='http', port=80, cors=True,
                     reload=True, timeout_send=-1, include_gateway=False)

**Environment**

- jina 3.20.0
- docarray 0.36.0
- jcloud 0.2.12
- jina-hubble-sdk 0.39.0
- jina-proto 0.1.27
- protobuf 4.23.4
- proto-backend upb
- grpcio 1.47.5
- pyyaml 6.0.1
- python 3.10.0
- platform Windows
- platform-release 10
- platform-version 10.0.22621
- architecture AMD64
- processor Intel64 Family 6 Model 151 Stepping 2, GenuineIntel
- uid 123870401841410
- session-id 0e41f950-3500-11ee-b0dc-70a8d34cc902
- uptime 2023-08-07T16:54:44.628100
- ci-vendor (unset)
- internal False
* JINA_DEFAULT_HOST (unset)
* JINA_DEFAULT_TIMEOUT_CTRL (unset)
* JINA_DEPLOYMENT_NAME (unset)
* JINA_DISABLE_UVLOOP (unset)
* JINA_EARLY_STOP (unset)
* JINA_FULL_CLI (unset)
* JINA_GATEWAY_IMAGE (unset)
* JINA_GRPC_RECV_BYTES (unset)
* JINA_GRPC_SEND_BYTES (unset)
* JINA_HUB_NO_IMAGE_REBUILD (unset)
* JINA_LOG_CONFIG (unset)
* JINA_LOG_LEVEL (unset)
* JINA_LOG_NO_COLOR (unset)
* JINA_MP_START_METHOD (unset)
* JINA_OPTOUT_TELEMETRY (unset)
* JINA_RANDOM_PORT_MAX (unset)
* JINA_RANDOM_PORT_MIN (unset)
* JINA_LOCKS_ROOT (unset)
* JINA_K8S_ACCESS_MODES (unset)
* JINA_K8S_STORAGE_CLASS_NAME (unset)
* JINA_K8S_STORAGE_CAPACITY (unset)
* JINA_STREAMER_ARGS (unset)
**Screenshots**

[...]

PR #6024, opened 2023-08-08
fix: make gateway load balancer should stream results

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/6020

Prior to this request, a Gateway Load balancing server will send an incoming request to the executor and wait until it returns all chunks before sending them back to the client.
After this fix, the gateway will rather stream chunks as soon as they come to the gateway


jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -185,12 +186,23 @@ async def _load_balance(self, request):
             async with aiohttp.ClientSession() as session:
                 if request.method == 'GET':
                     async with session.get(target_url) as response:
-                        content = await response.read()
-                        return web.Response(
-                            body=content,
+                        # Create a StreamResponse with the same headers and status as the target response
```
JoanFM, line 189: this also works for getting the OpenAPI.json or forwarding any other GET command?

alaeddine-13: `/openapi.json`, `/docs` and `/redoc` endpoints work as expected

alaeddine-13:
@JoanFM I added test cases that assert the type write effect is preserved (docs are streamed as soon as they come in to the gateway). The one with protocol=http and include_gateway=True fails if the fix commit is reverted

JoanFM (approved):

Issue #6012, opened 2023-08-03
Docarray v2 ser-de issue when using streamer.stream


girishc13:
**Describe the bug**

1. Create a common module to define the document model. 
2. Import the common document model in the gateway and the executor.
3. Use `self.streamer.stream` in the gateway to send request to a specific executor.

The documents (shared document model) returned by the executor cannot be added to a DocList of the same shared document model defined in the executor. The gateway throws the following exception.
```
DEBUG  text_encoder/rep-0@132549 recv DataRequest at /encode [08/03/23 11:00:53]
       with id: d624f1fa2e3246499d678973990772e0                                
INFO:     127.0.0.1:53964 - "POST /index HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/fastapi/applications.py", line 289, in __call__
    await super().__call__(scope, receive, send)
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/girishc13/Documents/Code/jina-ai/eatsmarter-prototype/venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
[...]

PR #6027, opened 2023-08-11
feat: add return_type parameter to gateway streamer methods to allow …

alaeddine-13:
closes: https://github.com/jina-ai/jina/issues/6012

GatewayStreamer by default will fetch executor input and output schemas and reconstruct docarray models dynamically. The output schemas will be used to cast the responses at the gateway level.
Although the casted responses are nearly identical to the original schemas defined at executor level, they might fail some checks, for instance, if the user defines a doc schema `MyDoc`, the following check will fail:
`assert isinstance(doc, MyDoc)`
similarly, adding `doc` to a DocList[MyDoc] will fail the type checking.

This PR allows the user of a GatewayStreamer to specify the return_type that will be used for casting and the output responses received from methods of the gateway streamer will match exactly the specified `return_type`.

- [x] tests
- [x] update the documentation


alaeddine-13:
Currently this PR depends on an issue with @docarray:
@docarray's UpdateMixin, only updates a document if the other document is exactly the same type.
However, there are cases where 2 documents are generated from dynamic schemas that are exactly the same. Even though the fields are exactly the same, the update operation will fail because in docarray checks that the schemas belong to the same reference

alaeddine-13:
depends on https://github.com/docarray/docarray/pull/1755

.github/workflows/ci.yml
```diff
@@ -459,6 +459,8 @@ jobs:
           python -m pip install wheel
           WHEEL_FILE=$(ls dist/*whl)
           pip install "$WHEEL_FILE[common,devel,test]" --no-cache-dir
+          pip uninstall docarray -y
```
JoanFM, line 462: is this still needed? this should not be needed right?

JoanFM (changes requested):

PR #6038, opened 2023-09-05
refactor: refactor the setup.py file

ai-naymul:
In the setup.py file there are some constant text that are hardcoded into the file, in this PR those hardcode text are converted into constants for better code readability


JoanFM (approved):
Thanks for the contribution!

PR #6043, opened 2023-09-07
refactor: refactor env var reading in docs/conf.py

ai-naymul:
Centralized the environment variable in the conf.py, refactor the code in constants.py


docs/conf.py
```diff
@@ -3,6 +3,11 @@
 import sys
 from os import path
 
+version = ''
+# Centralize environment variables
+JINA_VERSION = os.environ.get('JINA_VERSION')
+SPHINX_MULTIVERSION_VERSION = os.environ.get('SPHINX_MULTIVERSION_VERSION', version)
```
JoanFM, line 9: this is not what is intended. See that `version` is used later, and thus this reading should be placed after the `version` var initialization

docs/conf.py
```diff
@@ -3,6 +3,11 @@
 import sys
 from os import path
 
+version = ''
+# Centralize environment variables
+JINA_VERSION = os.environ.get('JINA_VERSION')
```
JoanFM, line 8:
```suggestion
JINA_VERSION = os.environ.get('JINA_VERSION', None)
```

JoanFM (changes requested):

docs/conf.py
ai-naymul: Okay I let me remove that...

docs/conf.py
```diff
@@ -3,10 +3,10 @@
 import sys
 from os import path
 
-version = ''
+
 # Centralize environment variables
 JINA_VERSION = os.environ.get('JINA_VERSION')
-SPHINX_MULTIVERSION_VERSION = os.environ.get('SPHINX_MULTIVERSION_VERSION', version)
+SPHINX_MULTIVERSION_VERSION = os.environ.get('SPHINX_MULTIVERSION_VERSION', None)
```
JoanFM, line 9: no, this is not correct, this assignment needs to go much lowrr, aftrr the version variable is properly set

JoanFM (changes requested):

docs/conf.py
ai-naymul: > no, this is not correct, this assignment needs to go much lowrr, aftrr the version variable is properly set

Oh. I am extremely sorry I didn't notice that...

ai-naymul: Is that fine?
kolla

JoanFM: no. Follow the flow, in master code we have:

```python
os.environ.get("SPHINX_MULTIVERSION_VERSION", version) on line 170
```
and on line 31 we have:

```python
version = __version__ 
```

With your changes, the default would be empty, in master the default would be the value of `__version__`. So ur code does not refactor, but changes the flow of the code

ai-naymul: > no. Follow the flow, in master code we have:
> 
> ```python
> os.environ.get("SPHINX_MULTIVERSION_VERSION", version) on line 170
> ```
> 
> and on line 31 we have:
> 
> ```python
> version = __version__ 
> ```
> 
> With your changes, the default would be empty, in master the default would be the value of `__version__`. So ur code does not refactor, but changes the flow of the code


Should I add a empty `version` string at the begining?
Screenshot_6

JoanFM: no, there is no need. But you cannot do this 

`os.environ.get("SPHINX_MULTIVERSION_VERSION", version)`

So early in the code

JoanFM: the code 

```python
SPHINX_MULTIVERSION_VERSION  = os.environ.get('SPHINX_MULTIVERSION_VERSION', version)
```
needs to be after line 31

JoanFM: yes

ai-naymul: > 

Ok

JoanFM (approved):

JoanFM:
Thanks for the contribution!!

Issue #6090, opened 2023-10-18
Streaming breaks for HTTP client when fields are not str, int, or float 


NarekA:
**Describe the bug**

Http streaming breaks when documents have fields which are not `str`, `int`, or `float`. This includes dicts, bools, and nested objects. The example below works for grpc but not http.

```python
import json
import time
from typing import Dict

from docarray import BaseDoc
from jina import Executor, Deployment, Client, requests
from pydantic import Field


# first define schemas
class MyDocument(BaseDoc):
    text: str
    other_doc: Dict = Field(default_factory=dict)

    def to_request_data(self):
        return json.dumps({"params": self.dict()}, indent=2)

# then define the Executor
class MyExecutor(Executor):
    @requests(on="/stream")
    async def task(self, doc: MyDocument, **kwargs) -> MyDocument:
        for i in range(10):
            yield MyDocument(text=f"hello world {doc.text} {i}")
            time.sleep(1)


example_doc = MyDocument(text="my input text")


async def stream_docs():
    protocol = "http"
    with Deployment(
        uses=MyExecutor, port=[12346], protocol=[protocol], cors=True
    ) as dep:
        client = Client(port=12346, protocol=protocol, asyncio=True)
        async for doc in client.stream_doc(
            on="/stream",
            inputs=example_doc,
            input_type=MyDocument,
            return_type=MyDocument,
        ):
            print(doc)



if __name__ == "__main__":
    import asyncio

    asyncio.run(stream_docs())

```

**Describe how you solve it**

The source of the problem seems to be [here](https://github.com/jina-ai/jina/blob/master/jina/clients/base/helper.py#L210). The input is being passed into streaming executor as params instead of a request body. It should either

1. Use http get with a body. See [this link](https://fastapi.tiangolo.com/tutorial/body/)

    > Sending a body with a GET request has an undefined behavior in the specifications, nevertheless, it is supported by FastAPI, only for very complex/extreme use cases.

2. Use a different http method
3. Serialize/deserialize the fields before IO.

---

**Environment**

**Screenshots**

 
Here is a stacktrace:
```
Traceback (most recent call last):
  File ".../app.py", line 71, in <module>
    asyncio.run(stream_docs())
  File "...python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "...python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File ".../app.py", line 53, in stream_docs
[...]

PR #6091, opened 2023-10-19
Use Post for Streaming Endpoints

NarekA:

This PR changes the get endpoints to post endpoints. Not sure if we want to keep the get endpoints for backwards compatibility.

**Goals:**


JoanFM (changes requested):
Can't we keep a GET endpoint together with a POST endpoint, and have the Jina client use the POST? And then we would need to document the limitations of the GET endpoint

jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -130,20 +130,19 @@ def add_streaming_get_route(
 
         @app.api_route(
             path=f'/{endpoint_path.strip("/")}',
-            methods=['GET'],
+            methods=['POST'],
```
alaeddine-13, line 133: I would rather duplicate the streaming endpoint in order to have 2 versions, post and get. The jina client should use the post method

NarekA: Sounds good to me, will add the get endpoint back.

NarekA:
Added the get endpoints back, and I will add some tests for the post endpoint. This feature will also align us with openai since they are using [streaming on post endpoints](https://platform.openai.com/docs/api-reference/chat/create) too.

JoanFM (changes requested):
Hey @NarekA,

This looks quite cool, I woule be missing a test proving that the original issue is solved. You can add it under the '/tests/integration/docarray_v2' folder.

NarekA:
@JoanFM @alaeddine-13 created a new [PR](https://github.com/jina-ai/jina/pull/6093) that follows conventions, working on adding tests too.

PR #6093, opened 2023-10-20
fix: add and use post endpoint for streaming

NarekA:
Making a new PR that follows conventions to replace https://github.com/jina-ai/jina/pull/6091

This PR changes the get endpoints to post endpoints. Not sure if we want to keep the get endpoints for backwards compatibility.

**Goals:**

### DIFF
--- a/docs/concepts/serving/executor/add-endpoints.md
+++ b/docs/concepts/serving/executor/add-endpoints.md
@@ -382,7 +382,7 @@ Streaming endpoints receive one Document as input and yields one Document at a t
 
 Streaming endpoints are only supported for HTTP and gRPC protocols and for Deployment and Flow with one single Executor.
 
-For HTTP deploymnent streaming executors generate both a GET and POST endpoint.
+For HTTP deployment streaming executors generate both a GET and POST endpoint.
 The GET endpoint support documents with string, integer, or float fields only,
 whereas, POST requests support all docarrays.
 The Jina client uses the POST endpoints.
--- a/docs/concepts/serving/executor/add-endpoints.md
+++ b/docs/concepts/serving/executor/add-endpoints.md
@@ -381,6 +381,11 @@ Streaming endpoints receive one Document as input and yields one Document at a t
 :class: note
 
 Streaming endpoints are only supported for HTTP and gRPC protocols and for Deployment and Flow with one single Executor.
+
+For HTTP deploymnent streaming executors generate both a GET and POST endpoint.
+The GET endpoint support documents with string, integer, or float fields only,
+whereas, POST requests support all docarrays.
+The Jina client uses the POST endpoints.
 ```
 
 A streaming endpoint has the following signature:
--- a/jina/clients/base/helper.py
+++ b/jina/clients/base/helper.py
@@ -200,7 +200,7 @@ class HTTPClientlet(AioHttpClientlet):
         request_kwargs = {
             'url': self.url,
             'headers': {'Accept': 'text/event-stream'},
-            'json': doc.dict(),
+            'json': doc.dict() if docarray_v2 else doc.to_dict(),
         }
 
         async with self.session.post(**request_kwargs) as response:
--- a/jina/clients/base/helper.py
+++ b/jina/clients/base/helper.py
@@ -197,19 +197,15 @@ class HTTPClientlet(AioHttpClientlet):
         :param on: Request endpoint
         :yields: responses
         """
-        if docarray_v2:
-            req_dict = doc.dict()
-        else:
-            req_dict = doc.to_dict()
+        req_dict = doc.json()
 
         request_kwargs = {
             'url': self.url,
             'headers': {'Accept': 'text/event-stream'},
         }
-        req_dict = {key: value for key, value in req_dict.items() if value is not None}
-        request_kwargs['params'] = req_dict
+        request_kwargs['data'] = req_dict
 
-        async with self.session.get(**request_kwargs) as response:
+        async with self.session.post(**request_kwargs) as response:
             async for chunk in response.content.iter_any():
                 events = chunk.split(b'event: ')[1:]
                 for event in events:
--- a/jina/clients/base/helper.py
+++ b/jina/clients/base/helper.py
@@ -197,13 +197,11 @@ class HTTPClientlet(AioHttpClientlet):
         :param on: Request endpoint
         :yields: responses
         """
-        req_dict = doc.json()
-
         request_kwargs = {
             'url': self.url,
             'headers': {'Accept': 'text/event-stream'},
+            'data': doc.json(),
         }
-        request_kwargs['data'] = req_dict
 
         async with self.session.post(**request_kwargs) as response:
             async for chunk in response.content.iter_any():
--- a/jina/clients/base/helper.py
+++ b/jina/clients/base/helper.py
@@ -200,7 +200,7 @@ class HTTPClientlet(AioHttpClientlet):
         request_kwargs = {
             'url': self.url,
             'headers': {'Accept': 'text/event-stream'},
-            'data': doc.json(),
+            'json': doc.dict(),
         }
 
         async with self.session.post(**request_kwargs) as response:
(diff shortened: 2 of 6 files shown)


tests/integration/docarray_v2/test_issues.py
```diff
@@ -92,3 +94,50 @@ def foo(self, docs: DocList[A], **kwargs) -> DocList[A]:
...
+        ):
+            docs.append(doc)
+
+    assert [d.text for d in docs] == [
+        "hello world my input text 0",
+        "hello world my input text 1",
+        "hello world my input text 2",
+        "hello world my input text 3",
+    ]
+    assert docs[0].nested_field.name == "test_name"
```
NarekA, line 142: @JoanFM I added this test to make sure complex Docs can be streamed

jina/serve/runtimes/gateway/http_fastapi_app_docarrayv2.py
```diff
@@ -258,6 +258,29 @@ async def streaming_get(request: Request):
             async def event_generator():
                 async for doc, error in streamer.stream_doc(
                     doc=input_doc_model(**query_params), exec_endpoint=endpoint_path
+                ):
+                    if error:
+                        raise HTTPException(status_code=499, detail=str(error))
```
JoanFM, line 263: can we get this 499 from the proper fastapi status constants?

tests/integration/docarray_v2/test_issues.py
```diff
@@ -92,3 +94,50 @@ def foo(self, docs: DocList[A], **kwargs) -> DocList[A]:
...
+        @requests(on="/stream")
+        async def stream(
+            self, doc: InputWithComplexFields, parameters: Optional[Dict] = None, **kwargs
+        ) -> InputWithComplexFields:
+            for i in range(4):
+                yield InputWithComplexFields(text=f"hello world {doc.text} {i}")
+
+    docs = []
+    protocol = "http"
+    with Deployment(uses=MyExecutor, protocol=protocol, port=11112) as dep:
```
JoanFM, line 125: for the tests we have a helper method that gives us a random port, please use it instead of giving a fixed one. That method makes sure there are no collisions

JoanFM (changes requested):

jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -143,7 +143,27 @@ async def streaming_get(request: Request):
...
+        @app.api_route(
+            path=f'/{endpoint_path.strip("/")}',
+            methods=['POST'],
+            summary=f'Streaming Endpoint {endpoint_path}',
+        )
+        async def streaming_post(body: input_doc_model, request: Request):
+            req = DataRequest()
+            req.header.exec_endpoint = endpoint_path
+            if not docarray_v2:
+                from docarray import Document
```
JoanFM, line 161: why is this import needed?

JoanFM (changes requested):
I like how the solution looks, I just have some small comments.

jina/clients/base/helper.py
```diff
@@ -197,19 +197,15 @@ async def send_streaming_message(self, doc: 'Document', on: str):
         :param on: Request endpoint
         :yields: responses
         """
-        if docarray_v2:
-            req_dict = doc.dict()
-        else:
-            req_dict = doc.to_dict()
+        req_dict = doc.json()
```
JoanFM, line 200: I would not do this change, I think the signature of `json` may return a `str`. Why is there a need to change it?

JoanFM (changes requested):

jina/clients/base/helper.py
NarekA: In order to be passed as `request_kwargs['data']` rather than `request_kwargs['params']`, it needs to be serialized as a string.

jina/serve/runtimes/worker/http_fastapi_app.py
NarekA: Removed!

jina/serve/runtimes/gateway/http_fastapi_app_docarrayv2.py
NarekA: I can't seem to find a fastapi constant for this, can you find me an example? This code is copied from the existing get endpoint.

NarekA:
@JoanFM I addressed the your comments. Please let me know If there's anything else I can do.

JoanFM:
Hey @NarekA ,

Thanks for the contrubution, could yiu add a note in the documentation https://docs.jina.ai/concepts/serving/executor/add-endpoints/#streaming-endpoints making explicit note that SSE can only be used with flat schemas?

jina/clients/base/helper.py
NarekA: Was able to keep this by using `json` instead of `data`

docs/concepts/serving/executor/add-endpoints.md
```diff
@@ -381,6 +381,11 @@ Streaming endpoints receive one Document as input and yields one Document at a t
 :class: note
 
 Streaming endpoints are only supported for HTTP and gRPC protocols and for Deployment and Flow with one single Executor.
+
+For HTTP deploymnent streaming executors generate both a GET and POST endpoint.
+The GET endpoint support documents with string, integer, or float fields only,
+whereas, POST requests support all docarrays.
+The Jina client uses the POST endpoints.
```
NarekA, line 388: @JoanFM Created the following docs

docs/concepts/serving/executor/add-endpoints.md
```diff
@@ -381,6 +381,11 @@ Streaming endpoints receive one Document as input and yields one Document at a t
 :class: note
 
 Streaming endpoints are only supported for HTTP and gRPC protocols and for Deployment and Flow with one single Executor.
+
+For HTTP deploymnent streaming executors generate both a GET and POST endpoint.
```
JoanFM, line 385: typo in deployment

JoanFM (changes requested):

docs/concepts/serving/executor/add-endpoints.md
NarekA: Fixed

jina/clients/base/helper.py
JoanFM: but where does the need for passing as 'str' come from?

NarekA: If we were using `data`, data needs to be a string, but since I swithced to json, I am able to use dict again. Here is the current code:
```python
 'json': doc.dict() if docarray_v2 else doc.to_dict(),
```

JoanFM:
> @JoanFM I addressed the your comments. Please let me know If there's anything else I can do.

It looks good to me, but there seems to be a failing test

NarekA:
@JoanFM I am trying to fix the tests, but I'm having issues recreating CI tests locally, or even seeing what the failure is in CI.

tests/integration/docarray_v2/test_streaming.py
```diff
@@ -143,17 +143,19 @@ async def test_streaming_delay(protocol, include_gateway):
...
-        async for doc in client.stream_doc(
+        stream = client.stream_doc(
             on='/hello',
             inputs=MyDocument(text='hello world', number=i),
             return_type=MyDocument,
-        ):
+        )
+        start_time = None
+        async for doc in stream:
+            start_time = start_time or time.time()
```
NarekA, line 153: Looks like streaming takes longer to initialize, but has all of the data emitted at once:
```
Expected delay to be less than 0.5, got 7.867813110351562e-06 on iteration 1
Expected delay to be less than 1.0, got 0.00012993812561035156 on iteration 2
Expected delay to be less than 1.5, got 0.0001659393310546875 on iteration 3
Expected delay to be less than 2.0, got 0.00019884109497070312 on iteration 4
Expected delay to be less than 2.5, got 0.00022912025451660156 on iteration 5
``

NarekA: All these times do not include a 2.5 second initialization time.

NarekA: Just realizes the gateway seems to be breaking the streaming functionality here. Will look into why this is happening.

NarekA: Looks like the gateway might be breaking streaming.

JoanFM:
Thanks @NarekA for the great contribution

PR #6097, opened 2023-10-25
Revert "fix: add and use post endpoint for streaming"

NarekA:
Reverts jina-ai/jina#6093

@JoanFM, This PR had an issue, the change I made to the streaming delay test was a patch. It looks like streaming does not work when I use post with a gateway, however, I have been able to update the endpoint so `get` works with either query params or a json body. Let's revert this and I will push my new branch.

### DIFF
--- a/tests/integration/docarray_v2/test_streaming.py
+++ b/tests/integration/docarray_v2/test_streaming.py
@@ -143,19 +143,17 @@ async def test_streaming_delay(protocol, include_gateway):
     ):
         client = Client(port=port, protocol=protocol, asyncio=True)
         i = 0
-        stream = client.stream_doc(
+        start_time = time.time()
+        async for doc in client.stream_doc(
             on='/hello',
             inputs=MyDocument(text='hello world', number=i),
             return_type=MyDocument,
-        )
-        start_time = None
-        async for doc in stream:
-            start_time = start_time or time.time()
+        ):
             assert doc.text == f'hello world {i}'
             i += 1
-            delay = time.time() - start_time
+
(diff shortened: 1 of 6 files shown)


JoanFM:
Can you open in parallel a PR with your fix to the GET parameter, before we actually revert this?

JoanFM:
Can you tell me how you reproduce the issue with a Gateway? I may be able to have it working?

NarekA:
@JoanFM I was able to do this without adding a post endpoint in the [following PR](https://github.com/jina-ai/jina/pull/6098). That PR just uses the existing get endpoint but adds an optional body param.

tests/integration/docarray_v2/test_streaming.py
```diff
@@ -143,19 +143,17 @@ async def test_streaming_delay(protocol, include_gateway):
...
-        async for doc in stream:
-            start_time = start_time or time.time()
+        ):
             assert doc.text == f'hello world {i}'
             i += 1
-            delay = time.time() - start_time
+
             # 0.5 seconds between each request + 0.5 seconds tolerance interval
-            assert delay < (0.5 * i), f'Expected delay to be less than {0.5 * i}, got {delay} on iteration {i}'
+            assert time.time() - start_time < (0.5 * i) + 0.5
```
NarekA, line 156: @JoanFM If you want to reproduce the gateway issue, just revert this test, but you might not need to because of https://github.com/jina-ai/jina/pull/6098

NarekA:
@JoanFM do you prefer to use the current endpoints with the gateway fixed or this?

JoanFM:
I think the idea of keeping only the `GET` endpoint is good.

PR #6098, opened 2023-10-25
fix: use body for streaming instead of params

NarekA:
Making a new PR that follows conventions to replace https://github.com/jina-ai/jina/pull/6091 and https://github.com/jina-ai/jina/pull/6093

This PR changes the get endpoints to have the data passed in the request body rather than the params.

**Goals:**

Also related: https://github.com/jina-ai/jina/pull/6097


JoanFM:
I am not sure if I would prefer to keep the 2 endpoints separate and fix the underlying issue. @alaeddine-13 is the integration with SSE as in documentation possible still?

NarekA:
@JoanFM I'm just taking the existing endpoint and allowing it to accept either parameters or a request body. I'm happy to add a post endpoint as well, but I need your help with the gateway.

NarekA:
Here is a [branch on my fork](https://github.com/NarekA/jina/pull/1) that will reproduce the getway error

tests/integration/docarray_v2/test_streaming.py
```diff
@@ -143,19 +143,17 @@ async def test_streaming_delay(protocol, include_gateway):
...
-        async for doc in stream:
-            start_time = start_time or time.time()
+        ):
             assert doc.text == f'hello world {i}'
             i += 1
-            delay = time.time() - start_time
+
             # 0.5 seconds between each request + 0.5 seconds tolerance interval
-            assert delay < (0.5 * i), f'Expected delay to be less than {0.5 * i}, got {delay} on iteration {i}'
+            assert time.time() - start_time < (0.5 * i) + 0.5
```
NarekA, line 156: At the very least, we should merge this and fix the gateway.

alaeddine-13:
@NarekA why do you want to change the GET endpoint method ? 
I thought the agreement was that, if you have basic requirements and would like to use client libraries, you use GET endpoint and if you need customized schemas and long input, you use the POST method.
What is wrong with this approach ?

alaeddine-13:
> @NarekA why do you want to change the GET endpoint method ? I thought the agreement was that, if you have basic requirements and would like to use client libraries, you use GET endpoint and if you need customized schemas and long input, you use the POST method. What is wrong with this approach ?

Actually now I realize you want to allow passing either of body or query parameters.
I'm not sure if that's a good practice when building HTTP services but I think it's a nice idea actually. Yes @JoanFM  this should keep compatibility with the documentation section where we use a standard js client.
A test would be great to have though or at least we need to try out the documentation on this PR before merging

NarekA:
@alaeddine-13 The current implementation with the POST method is fine, but it seems to break when using a deployment with a gateway. The stream all comes out at once. I modified [this test (and am changing it back)](https://github.com/jina-ai/jina/pull/6098/files/3d1b07fe3cdfdf2c602cf3045d2c01d3ece982dd#diff-4a9008012b5ba14919be3fd22f1e0da267ca041c67411dd0343fabb0065c203b) not realizing that I was fixing the very thing it was testing for. The parameterized test passes for all the scenarios except for http+gateway. Using get instead of post seems to fix the gateway and we have only one endpoint so it's less confusing.

JoanFM (changes requested):
Is there or can we have a test where the API is called with the query params?

NarekA:
If we do allow post, I like the idea of just using the same route with 2 methods: `methods=['GET', 'POST'],`. That way we have consistency between the endpoints, and don't have to maintain 2 routes.

jina/serve/runtimes/gateway/http_fastapi_app_docarrayv2.py
```diff
@@ -252,35 +252,14 @@ def add_streaming_routes(
             methods=['GET'],
             summary=f'Streaming Endpoint {endpoint_path}',
         )
-        async def streaming_get(request: Request):
-            query_params = dict(request.query_params)
+        async def streaming_get(request: Request, body: input_doc_model = None):
+            if not body:
```
NarekA, line 256: We can add a conditionals like `and request.method == 'GET'` here.

JoanFM: I think we may not need `POST` at all then.

NarekA: That's the current state of this PR, and it fixes everything. Can't tell if tests need to be re-run or if they're actually failing.

JoanFM: you can see a test failing here https://github.com/jina-ai/jina/actions/runs/6647328275/job/18084538375?pr=6098

NarekA: I fixed that issue, I think the current failures just need re-runs, can you take a look?

NarekA: Nevermind, found one that's not flakyness

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -184,8 +184,17 @@ async def _load_balance(self, request):
...
-                    async with session.get(target_url) as response:
+                    request_kwargs = {}
+                    if request.body_exists():
+                        payload = await request.json()
+                        if payload:
+                            request_kwargs['json'] = payload
+
+                    async with session.get(
+                        url=target_url, **request_kwargs
+                    ) as response:
```
NarekA, line 197: Found out where the gateway forwarding is happening.

NarekA:
Only one test job failing, and I can't find a clear stack trace in it so it might be flakiness.

JoanFM (approved):

PR #6122, opened 2023-12-04
fix: loadbalance stream based on response

NarekA:
The current load balancer assumes all `GET` requests are streaming and all `POST` requests are not. This may not be true for user-added fast-api endpoints and in the past we have talked about using `POST` for streaming. (One of the benefits of this is the Swagger UI better documents the payload for `POST` endpoints).

It makes more sense to use the response content-type to determine when to stream.

**Goals:**

- Allow Post streaming POST endpoints and non-streaming GET endpoints


JoanFM:
Hey @NarekA,

Thanks for the contribution, can you check the failed tests and add a test showing what u are intending to fix?

NarekA:
@JoanFM I think the issue came about because I was using `request.content_type` instead of `response.content_type`. I think the latest commit should fix it.

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -157,18 +159,23 @@ async def _load_balance(self, request):
...
+                try:
+                    payload = await request.content.read()
+                    if payload:
+                        request_kwargs['data'] = payload
+                except Exception:
+                    self.logger.debug('No JSON payload found in request')
+
+                async with _RequestContextManager(
+                    session._request(request.method, target_url, **request_kwargs)
+                ) as response:
```
cenhao, line 175: @NarekA why not directly use the `request` method from the session?, it's already wrapping the context manager for you.

```
    def request(
        self, method: str, url: StrOrURL, **kwargs: Any
    ) -> "_RequestContextManager":
        """Perform HTTP request."""
        return _RequestContextManager(self._request(method, url, **kwargs))
        ```

File: jina/serve/runtimes/gateway/request_handling.py [hunk 2]
```diff
@@ -157,18 +159,23 @@ async def _load_balance(self, request):
...
-                    except Exception:
-                        self.logger.debug('No JSON payload found in request')
-
-                    async with session.get(
-                        url=target_url, **request_kwargs
-                    ) as response:
+                request_kwargs = {
+                    'headers': request.headers,
+                    'params': request.query,
+                }
```
cenhao, line 165: The original code did not forward the headers & query, which makes me wonder if it's intentional?

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -185,20 +192,12 @@ async def _load_balance(self, request):
...
-                        return web.Response(
-                            body=content,
-                            status=response.status,
-                            content_type=response.content_type,
-                        )
+                    content = await response.read()
+                    return web.Response(
+                        body=content,
+                        status=response.status,
+                        content_type=response.content_type,
```
cenhao, line 199: should the header be added here as well?

JoanFM:
Most of the errors on tests will be fixed by #6124

jina/serve/runtimes/gateway/request_handling.py
JoanFM: it was not intentional

jina/serve/runtimes/gateway/request_handling.py
NarekA: Somehow I missed this, will fix.

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -157,18 +160,23 @@ async def _load_balance(self, request):
...
             async with aiohttp.ClientSession() as session:
 
-                if request.method == 'GET':
-                    request_kwargs = {}
-                    try:
-                        payload = await request.json()
-                        if payload:
-                            request_kwargs['json'] = payload
-                    except Exception:
-                        self.logger.debug('No JSON payload found in request')
```
cenhao, line 167: Looking at the original implementation, i have this idea:
it looks like the original logic was only to write a debug log which is not useful at all for production application. Can we just act as a pure proxy here for performance consideration? Something like:
```
async with session.request(request.method, data=request.iter_any(), **request_kwargs) as response:
    ....
```

@NarekA @JoanFM

jina/serve/runtimes/gateway/request_handling.py
JoanFM: yes this would be a good idea

jina/serve/runtimes/gateway/request_handling.py
NarekA: For some reason, if I try to pass the content in any other way besides the json field, I get an error here. I've tried everything at this point, if you can get this to work, I am interested.

NarekA:
@JoanFM I'm finding these tests to be prohibitively difficult to debug. Any idea what is going wrong? I need a better way of replicating them locally.

jina/serve/runtimes/gateway/request_handling.py
cenhao: what error do you see?

jina/serve/runtimes/gateway/request_handling.py
```diff
@@ -157,48 +157,34 @@ async def _load_balance(self, request):
...
+                request_kwargs = {}
+                try:
+                    payload = await request.json()
+                    if payload:
+                        request_kwargs['json'] = payload
+                except Exception:
+                    self.logger.debug('No JSON payload found in request')
+
+                async with session.request(
+                    request.method, url=target_url, **request_kwargs
```
cenhao, line 169: You might consider adding `auto_decompress=False` here. I've seen issue when upstream compresses the response, without this flag, together with the upstream header, there will be an error.

NarekA: Thanks, just fixed this.

JoanFM (approved):

JoanFM (approved):

Issue #6049, opened 2023-09-20
Best Way to Pass Request Headers to Executor


NarekA:
**Describe your proposal/problem**

I've been looking for a way to pass http request headers to my Executor, and this is the only way I've been able to do it.

```python

class MyGateway(FastAPIBaseGateway):
    @property
    def app(self):
        from fastapi import FastAPI

        app = FastAPI(title="Test Gateway")

        @app.get(path="/config")
        async def custom_endpoint(my_input: MyInput, request: Request):
            docs1: DocList[RequestOutput] = await self.executor["executor0"].post(
                on="/api/v0/test",
                inputs=DocList[Input]([my_input]),
                parameters=request.headers or {"no": "headers"},
                return_type=DocList[RequestOutput],
            )
            return docs1.to_json()

        return app

```

Is there a simpler way to do this? The approach above only works for flows, is there a way that works for deployments?

---

**Environment**

- jina 3.20.3
- docarray 0.33.0
- jcloud 0.2.16
- jina-hubble-sdk 0.39.0
- jina-proto 0.1.27
- protobuf 3.20.0
- proto-backend python
- grpcio 1.48.0rc1
- pyyaml 6.0.1
- python 3.9.17
- platform Darwin
- platform-release 22.6.0
- platform-version Darwin Kernel Version 22.6.0: Wed Jul 5 22:21:53
  PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6020
- architecture arm64
- processor arm
- uid 11592019587001
- session-id 0062ff1e-574c-11ee-9138-0a8afa35afb9
- uptime 2023-09-19T17:24:02.674004
- ci-vendor (unset)
- internal False
* JINA_DEFAULT_HOST (unset)
* JINA_DEFAULT_TIMEOUT_CTRL (unset)
* JINA_DEPLOYMENT_NAME (unset)
* JINA_DISABLE_UVLOOP (unset)
* JINA_EARLY_STOP (unset)
* JINA_FULL_CLI (unset)
* JINA_GATEWAY_IMAGE (unset)
* JINA_GRPC_RECV_BYTES (unset)
* JINA_GRPC_SEND_BYTES (unset)
* JINA_HUB_NO_IMAGE_REBUILD (unset)
* JINA_LOG_CONFIG (unset)
* JINA_LOG_LEVEL (unset)
* JINA_LOG_NO_COLOR (unset)
* JINA_MP_START_METHOD (unset)
[...]

PR #6126, opened 2023-12-07
feat: pass request-headers as metadata

NarekA:
This PR is attempts to implement the solution in the discussion of https://github.com/jina-ai/jina/issues/6049. It adds a new parameter `metadata` for jina endpoints which can be used to pass in http headers or GRPC metadata.

This PR only implements the http headers, GRPC metadata will have to be done in a separate PR

**Goals:**

- resolves 6049

TODO:
- [ ] Add tests
- [ ] Add documentation


jina/proto/docarray_v1/jina.proto
```diff
@@ -122,6 +122,8 @@ message DataRequestProto {
     }
 
     DataContentProto data = 4; // container for docs and groundtruths
+
+    google.protobuf.Struct metadata = 5; // extra kwargs that will be used in executor
```
NarekA, line 126: Will update this comment

jina/serve/runtimes/worker/http_fastapi_app.py
```diff
@@ -86,15 +86,18 @@ def add_post_route(
...
-        async def post(body: input_model, response: Response):
+        async def post(body: input_model, response: Response, request: Request):
 
             req = DataRequest()
             if body.header is not None:
                 req.header.request_id = body.header.request_id
 
             if body.parameters is not None:
                 req.parameters = body.parameters
+            req.metadata = dict(request.headers or {"no_headers": "true"})
```
NarekA, line 100: Will remove this default value

jina/serve/runtimes/gateway/http_fastapi_app_docarrayv2.py
```diff
@@ -187,8 +187,10 @@ def add_post_route(
         )
         app_kwargs['response_class'] = DocArrayResponse
 
+        from fastapi import Request
+
         @app.api_route(**app_kwargs)
-        async def post(body: input_model, response: Response):
+        async def post(body: input_model, response: Response, request: Request):
```
cenhao, line 193: Does this means the header will only be available when gateway is enabled?

cenhao: Anyway - having access to HTTP header is very important for our production application - we are an infra team supporting different businesses, and we use headers to communicate extra information like auth, tracing id, etc.

@JoanFM

NarekA: It should work for standalone-fast-api deployments too see [this change](https://github.com/jina-ai/jina/pull/6126/files#diff-76f86adabfbe802643bffc1fcd78a47e594e330aa0c0d3e787c166d21abb46caR92).

jina/constants.py
```diff
@@ -53,6 +53,7 @@
 __args_executor_func__ = {
     'docs',
     'parameters',
+    'metadata',
```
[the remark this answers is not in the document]
NarekA: cc @JoanFM

jina/serve/runtimes/gateway/http_fastapi_app_docarrayv2.py
JoanFM: hey @cenhao ,

The feature should be available with and without gateway.

Btw, would be cool to know about ur usecase and how u use/plan to use Jina

jina/proto/docarray_v1/jina.proto
NarekA: Will remove this and the proto changes that resulted from it.

NarekA:
@JoanFM I'm happy to change the parameter to headers, but I'm afraid it's too close to `header` which I've noticed is used for passing `request_id` during orchestration. How about `request_headers`?

JoanFM:
> @JoanFM I'm happy to change the parameter to headers, but I'm afraid it's too close to `header` which I've noticed is used for passing `request_id` during orchestration. How about `request_headers`?

it does not matter, header is something that is quite hidden from the Executor developer. I would still use `headers`

PR #6055, opened 2023-09-25
feat(sagemaker): support batch-transform

deepankarm:
## Goals

PR in continuation to https://github.com/jina-ai/jina/pull/6046. 

Sagemaker still calls the `POST /invocations` route during batch-transform but with different content-type & body. To enable both `application/json` (for inference) and `text/csv` (for batch-transform, I removed the pydantic model from the path signature, rather use fastapi's `Request` class and evaluate the body according to the content-type. 

For batch-transform, 

- Expect `text/csv` or `application/csv` content type.
- Validate if the format is correct.
- Convert each row in the csv to `input_model` (user-defined model in the Executor route).
- Invoke the route for all inputs together and return the response. This makes sure the values appended to the output csv by sagemaker has a similar format as inference. 

#### Gotchas

- Sagemaker doesn't support CSV with headers. We follow the same restriction. 
- Rows in the input csv should have all fields defined in the input type (including optional fields like `id`)

  ```python
  from docarray import BaseDoc
  
  class TextDoc(BaseDoc):
      text: str
  ```

  Following inputs are valid

  ```csv
  1,abcd
  2,efgh
  3,ijkl
  4,mnop
  5,qrst
  6,uvwx
  7,yzab
  8,cdef
  9,ghij
  10,klmn
  ```

  Following CSV is invalid.

  ```csv
  abcd
  efgh
  ijkl
  mnop
  qrst
  uvwx
  yzab
  cdef
  ghij
  klmn
  ```

#### End-to-end tests

I've manually tested the Executor after pushing a model to S3, the Executor image to ECR & running batch-transform jobs for both `SingleLine` and `MultiLine` records.

---

- [x] check and update documentation. See [guide](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md#-contributing-documentation) and ask the team.

PR #6145, opened 2024-02-29
fix: batch transform update for sagemaker reranker integration

zac-li:
**Goals:**
This PR made a required improvement (based off of https://github.com/jina-ai/jina/pull/6055/) on the SageMaker Batch Transform routine so that Reranking SageMaker model can perform Batch Transform as expected.

Previously, Jina Batch Transform endpoint has no problem reading simple data in each row from CSV and construct a input model and perform inference, for example, for 

`0f37fc919deb46e8bcafd92f1feba65f, simple text`, it will be converted to:

```python
TextDoc(
    id='0f37fc919deb46e8bcafd92f1feba65f',
    text='simple text'
)
```

However, for more complex model (e.g., includes `List` of models), such as:

```python
class RankInput(OriginalRankInput):
    query: Union[str, TextDoc]
    documents: List[TextDoc]
    top_n: Optional[int]
```
which is represented as such in CSV:

```
07937cd63aed436b9e6c34a86783f55f,where's the dog,[{\"text\": \"the dog is in my house\"}\, {\"text\": \"the cat looks good\"}\, {\"text\": \"fish chips\"}],2
```
 The original implementation would fail when parsing.

So this PR proposed a `construct_model_from_line` that can deal with more complex cases like above.

Test cases have been added to verify the fix.


tests/integration/docarray_v2/sagemaker/test_reranking.py
```diff
@@ -0,0 +1,112 @@
...
+    def __enter__(self):
+        self._old_cwd.append(os.getcwd())
+        os.chdir(self.path)
+
+    def __exit__(self, *excinfo):
+        os.chdir(self._old_cwd.pop())
+
+
+def test_provider_sagemaker_pod_inference():
+    with chdir(os.path.join(os.path.dirname(__file__), "SampleRerankerExecutor")):
```
JoanFM, line 29: u do not need this, u can pass the path to the `config.yml` without this

tests/integration/docarray_v2/sagemaker/test_reranking.py
```diff
@@ -0,0 +1,112 @@
...
+            assert resp.status_code == 200
+            resp_json = resp.json()
+            assert len(resp_json["data"]) == 1
+            assert (
+                resp_json["data"][0]["results"][0]["document"]["text"] == "first result"
+            )
+
+
+def test_provider_sagemaker_pod_batch_transform_for_reranker_valid():
+    with chdir(os.path.join(os.path.dirname(__file__), "SampleRerankerExecutor")):
```
JoanFM, line 69: same here

JoanFM (changes requested):

jina/serve/runtimes/worker/http_sagemaker_app.py
```diff
@@ -145,6 +154,47 @@ async def post(request: Request):
...
                     )
 
+                def construct_model_from_line(
+                    model: Type[BaseModel], line: List[str]
+                ) -> BaseModel:
+                    parsed_fields = {}
+                    model_fields = model.__fields__
+
+                    for field_name, field_str, field_info in zip(
+                        model_fields.keys(), line, model_fields.values()
```
JoanFM, line 164: this looks like a very weird way to iterate.

```suggestion
                        for field_str, (field_name, field_info) in zip (line, model_fields.items())
```

JoanFM (changes requested):

jina/serve/runtimes/worker/http_sagemaker_app.py
zac-li: ah right

tests/integration/docarray_v2/sagemaker/test_reranking.py
zac-li: fixed

tests/integration/docarray_v2/sagemaker/test_reranking.py
zac-li: fixed

PR #6149, opened 2024-03-07
feat: support provider endpoint in jina executor

zac-li:
## Goals ##

Added `--provider-endpoint` in `jina executor serve` so that we can specify the endpoint used in SageMaker integration for `/invocations`. 

This is needed for Colbert use case as in the Executor we have multiple endpoints, and each SageMaker model package needs to use different endpoint for `/invocations` from the same Executor container.


jina/enums.py
```diff
@@ -271,6 +271,14 @@ class ProviderType(BetterEnum):
     SAGEMAKER = 1  #: AWS SageMaker
 
 
+class ProviderEndpointType(BetterEnum):
+    """Provider endpoint type."""
+
+    NONE = 0
+    RANK = 1
+    ENCODE = 2
```
deepankarm, line 279: Any reason we have rank & encode hard-coded here?

JoanFM: oh yes, this should not be an enum. This should be free

zac-li: ok I can remove that

JoanFM: pls remove yes

jina/orchestrate/deployments/__init__.py
```diff
@@ -35,6 +35,7 @@
     PollingType,
     ProtocolType,
     ProviderType,
+    ProviderEndpointType,
```
JoanFM, line 38: remove

jina/parsers/orchestrate/pod.py
```diff
@@ -4,7 +4,7 @@
 from dataclasses import dataclass
 from typing import Dict
 
-from jina.enums import PodRoleType, ProtocolType, ProviderType
+from jina.enums import PodRoleType, ProtocolType, ProviderEndpointType, ProviderType
```
JoanFM, line 7: remove import

JoanFM (changes requested):

JoanFM (approved):

PR #6159, opened 2024-03-29
feat: executor to azure custom container

zac-li:
**Notes**

- This PR enables executor to run as a Fastapi app in Azure container.
- Unlike Sagemaker, multiple endpoints can be exposed in one container (no validation like SageMaker would be called).
- Tests structure is also restructured; now both sagemaker and azure tests are under `csp` folder and sharing the same executor fixtures.


jina/serve/runtimes/worker/request_handling.py
```diff
@@ -192,7 +192,29 @@ async def _shutdown():
         return extend_rest_interface(app)
 
     def _http_fastapi_sagemaker_app(self, **kwargs):
```
JoanFM, line 194: why don't we have only one `_http_fastapi_csp_app`?

JoanFM (changes requested):

JoanFM (approved):

PR #6198, opened 2024-09-15
ci: unblock grpcio version

JoanFM:
**Goals:**

- resolves #ISSUE-NUMBER
- ...
- ...

- ...
- ...


abeppu:
Hey there, 
Sorry to reach out without context, but I'm wondering if there's any likelihood of these updates to dependencies being merged in the near term?
I was working on using vectordb in an existing, mature codebase. The fact that jina has explicit upper-bounds on the versions of these libraries, and that those upper bounds are rather old (and older than the *lower* bounds of some other dependencies in our codebase), makes it actually rather messy and challenging for us to use these libraries. Basically, any component that already has a dependency on a newer version of grpcio, opentelemetry-sdk or indeed pydantic or uvicorn cannot cleanly add jina (or vectordb) to its environment or built artifacts.

JoanFM:
I will try to get some time to look into it. If u can check what needs to be done to pass these failing tests, it would be of great help

andrev24 (commented):
How do I check my people's instagram

PR #6229, opened 2025-02-13
feat: support param in sm batch

zac-li:
**Goals:**

Currently, the FastAPI provided by `jina` can’t handle SageMaker Batch Transform requests when both the doc list and parameters (e.g., late_trunking set to True/False) need to be considered. In the CSV, there’s no way to specify parameters.

This PR introduces a way for users to set both the document list and parameters. While it’s not the cleanest solution, it works.

Essentially, the first row will be used for parameters, and it must include `params_row` as a signature to identify it as the parameters row. The remaining rows will still be used as doc rows.

Example:

```
0,params_row,retrieval.query,false,2
1,abcd
2,efgh
3,ijkl
4,mnop
5,qrst
6,uvwx
7,yzab
8,cdef
9,ghij
10,klmn

```


jina/serve/runtimes/worker/http_csp_app.py
```diff
@@ -154,50 +155,81 @@ async def post(request: Request):
...
+                        field_str = line[idx]  # Corresponding value from the row
+
+                        try:
+                            # Handle Literal types (e.g., Optional[Literal["value1", "value2"]])
+                            origin = get_origin(field_type)
+                            if origin is Literal:
+                                literal_values = get_args(field_type)
+                                if field_str not in literal_values:
+                                    raise HTTPException(
+                                        status_code=400,
```
JoanFM, line 182: use thr fastapi status codes and not plain numbers

jina/serve/runtimes/worker/http_csp_app.py
```diff
@@ -154,50 +155,81 @@ async def post(request: Request):
...
+                                        # General case: try converting to the target type
+                                        try:
+                                            parsed_fields[field_name] = field_type(field_str)
+                                        except (ValueError, TypeError):
+                                            # Fallback to parse_obj_as when type is more complex, e., AnyUrl or ImageBytes
+                                            parsed_fields[field_name] = parse_obj_as(field_type, field_str)
+
+                        except Exception as e:
+                            raise HTTPException(
+                                status_code=400,
```
JoanFM, line 230: same

jina/serve/runtimes/worker/http_csp_app.py
```diff
@@ -74,7 +75,7 @@ def add_post_route(
             input_model,
             output_model,
             input_doc_list_model=None,
-            output_doc_list_model=None,
+            parameters_model=None,
```
JoanFM, line 78: reviewing on phone, so cannot dig much deeper. Why is this removed?

JoanFM (changes requested):
please define what is being done with more detail.

jina/serve/runtimes/worker/http_csp_app.py
zac-li: done

jina/serve/runtimes/worker/http_csp_app.py
zac-li: done

jina/serve/runtimes/worker/http_csp_app.py
zac-li: oh this `output_doc_list_model ` was never used

JoanFM (dismissed):

